PR #6, opened 2016-02-03
explain data array in Vertex struct

shlomizadok:


purpleidea:
@shlomizadok As discussed, I'm not letting your first patch be something silly like this. I'll try to put together an "areas to patch guide" in the near future.

PR #5, opened 2016-02-02
Build docs

ffrank:
Issues found while following README to compile `mgmt`:

* the `go get` commands are inlined in the rendering on GitHub
* the `golang 1.3` package from Debian stable errors out of `make build` because `go generate` is not a thing yet

PR #7, opened 2016-02-03
README: also list stringer as a dependency

ffrank:
Your advice in #5 worked nicely. Debian apparently does not (yet) package `stringer`, so let's mention it and give help.


purpleidea:
LGTM, merged, thanks!

PR #9, opened 2016-02-06
Made the `make deps` command work on OSX.

walterheck:
In order to make this work we check if we're on OSX and if we are, we use brew to install etcd (on linux it's installed from source for some reason)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ go:
 dist: trusty
 sudo: required
 install: 'make deps'
+os:
+  - linux
+  - osx
 script: 'make test'
 matrix:
   allow_failures:
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -1,29 +1,52 @@
 #!/bin/bash
 # setup a simple go environment
 
+# see if we're running a testrun in travis
 travis=0
 if env | grep -q '^TRAVIS=true$'; then
 	travis=1
 fi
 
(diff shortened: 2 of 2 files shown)


purpleidea:
@walterheck Thanks for the patch, but as you should see this failed to pass the CI. This was probably because of extra whitespace at the end of lines, and in addition, I think you used spaces for indenting code instead of tabs.

If you could resubmit so that it passes, that would be great, and I'll have a look.

Cheers!

walterheck:
@purpleidea fixed the spaces thing (first project I've contributed to in a loooong time that uses tabs instead of spaces ;) ). However now it's failing and I'm not sure why. It seems like due to the change of flow of logic now etcd doesn't get installed where I thought it was getting installed to. Can't we just switch to yum install etcd or does that mess with travis too much?

Need a bit of help here, not much familiar with how travis works internally.

purpleidea:
@walterheck If you look in the logs, you'll see etcd binary is no longer found due to your patch. If you still want to do this, you'll have to fix it up. I'd recommend leaving MacOSX support out for now, until a lot of the other pieces required for it are ready. If you'd like to maintain a feat/macosx branch, I think that would make more sense.

purpleidea:
@walterheck This still doesn't build, so I'm going to close this for now, but feel free to maintain a feat/osx branch in the meantime. Cheers!

PR #17, opened 2016-02-22
Decided to start with tests for FileRes

witlessbird:
- need to figure out how to test FileRes.Watch
 - FileApply and CheckApply need tests
 - Compare method is broken?

will add more tests as I go.

### DIFF
--- a/examples/content_example_dir/test_dir_2/test_file_2
+++ b/examples/content_example_dir/test_dir_2/test_file_2
@@ -1,0 +1,1 @@
+Hello world once more!
--- a/examples/content_example_dir/test_file_1
+++ b/examples/content_example_dir/test_file_1
@@ -1,0 +1,1 @@
+Hello, World.
--- a/examples/graph0a.yaml
+++ b/examples/graph0a.yaml
@@ -15,4 +15,4 @@ edges:
     name: noop1
   to:
     res: file
-    name: file1
+    name: dir1
--- a/examples/graph0a.yaml
+++ b/examples/graph0a.yaml
@@ -11,8 +11,8 @@ resources:
 edges:
 - name: e1
   from:
-    res: noop
+    kind: noop
     name: noop1
   to:
-    res: file
+    kind: file
     name: dir1
--- a/examples/graph0a.yaml
+++ b/examples/graph0a.yaml
@@ -1,0 +1,18 @@
+---
+graph: mygraph
+comment: hello world example
+resources:
+  noop:
+  - name: noop1
+  file:
+  - name: dir1
+    path: "/tmp/mgmt-hello-world-dir"
+    state: exists
+edges:
+- name: e1
+  from:
+    res: noop
+    name: noop1
+  to:
+    res: file
+    name: file1
--- a/examples/graph0b.yaml
+++ b/examples/graph0b.yaml
@@ -6,7 +6,8 @@ resources:
   - name: noop1
   file:
   - name: dir1
-    path: "/tmp/mgmt-hello-world-dir/"
+    path: "/tmp/mgmt-hello-world-dir-2/"
+    content: "./examples/content_example_dir"
     state: exists
 edges:
 - name: e1
--- a/examples/graph0b.yaml
+++ b/examples/graph0b.yaml
@@ -12,8 +12,8 @@ resources:
 edges:
 - name: e1
   from:
-    res: noop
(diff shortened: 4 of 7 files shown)


purpleidea:
@witlessbird Cool... I guess I'll let you keep hacking at this until it's ready, but tests are a great thing! Cheers!
Note there are a few related tests in misc_test.go

Cheers!

witlessbird:
@purpleidea: do you know what "The following file headers are not properly formatted:" build failure for file_test.go is about?

purpleidea:
On Mon, Feb 22, 2016 at 5:27 PM, Dmitri Dolguikh <notifications@github.com>
wrote:

> @purpleidea <https://github.com/purpleidea>: do you know what "The
> following file headers are not properly formatted:" build failure for
> file_test.go is about?


Yeah, just put the exact comment boiler plate at the top in your file as is
in main.go or misc_test.go or whatever. they're all the same

witlessbird:
@ffrank: fixed formatting

ffrank:
Hmm, according to Travis, something is still amiss. Also, fixing it in a new commit seems messy to me. Since this is not yet merged, it would be cleaner to `rebase -i` and amend the commits that broke formatting in the first place.

purpleidea:
On Sun, Apr 10, 2016 at 3:06 PM, Felix Frank <notifications@github.com>
wrote:

> Hmm, according to Travis, something is still amiss. Also, fixing it in a
> new commit seems messy to me. Since this is not yet merged, it would be
> cleaner to rebase -i and amend the commits that broke formatting in the
> first place.
>

I would work on just getting everything working and passing in however many
commits, and then when you're (both?) happy I'll do a review, and when
that's done and fixes are in, we can squash it all into one or two commits.

mohae:
gofmt your code for consistent and idiomatic formatting

witlessbird:
This should make Travis happy.

purpleidea:
I just noticed that this is passing... Here's a quick initial review if anyone is interested:

1) Can you remove the use of the "guardian" fileutils lib? I don't think we need an extra dep at this point.
2) Can you rebase to master and squash this all into a single patch please?

Ping me when ready and I'll do a more thorough review.
Thanks!

ffrank:
Iirc we need to rewrite the directory syncing logic in order to shed the dependency.

I have that on my plate, but was blocked for a couple of weeks. Hopeful to find some time soon.

witlessbird:
@ffrank: I can do the directory synching/copying logic if you don't have time atm... Let me know!

ffrank:
I've started this weekend. It's quite a mouthful, but I'll keep at it for a few more days. Pretty sure I'm half way there already.

ffrank:
The code works now, in principle. However, through our in-person discussions, we found two remaining issues:
 1. the inotify watches must be removed when no longer used
 2. the CheckApply function must indicate whether change was necessary (the `checkok` flag)

I'm working on adding the second. However, I now believe that the first is a non-issue, because
 1. files that are removed (by `mgmt` or foreign processes) implicitly remove their watch
 2. calling `watcher.Add()` with a path that is already watched does not use an additional watch

As a result, there is no way for us to hold on to watches that are no longer needed (except those on parent directories, which are not the issue here).

I propose we just keep adding watches as needed (recursively in managed directories), and rely on their automatic removal (except for parent directories of course). Cool?

purpleidea:
> calling watcher.Add() with a path that is already watched does not use an additional watch

But does it replace the old watch pointing to inode X, with a new one that now points to inode Y ?

Does the fsnotify lib free memory on stale watches that aren't used anymore or do we need to explicitly remove them?

Anyways, the idea is that the code looks logical and clean. Let's look at that and go from there.

purpleidea:
Since this has been stale for a while, and a friend was willing to help me through the tunnel vision of the recursive watcher, after I got that first part done, I was "bored" and hacked through an implementation of the CheckApply stuff too. Because of the inactivity here, I figured nobody would mind.

I think the code is fairly clean, and while there is still work to do to add more features, I think it's easy to see where they fit in at this point. It is:

https://github.com/purpleidea/mgmt/commit/598c74657c1488a79795daa48def4003ffa58f82

If there are tests or any other useful code left in this PR, please rebase and send it in as a new PR. Also thank you so much for getting involved in the project.

Cheers!

PS: Please test and report any issues with the updated file resource.

PR #18, opened 2016-03-10
Fix broken link to graph.yaml section

xsb:


purpleidea:
@xsb Thanks! Strangely, this fixes the issue in github markdown, but breaks the link when running pandoc... Any idea of how to fix both? The `make docs` target does the pandoc file...

xsb:
@purpleidea What about changing the section title from *graph.yaml* to *Graph definition file*?

purpleidea:
WFM :) Want to update the patch? Thanks

xsb:
Done. I also tested the PDF output after running `make docs` and the link works fine.

purpleidea:
LGTM, thanks!

purpleidea:
@xsb Also, congratulations on your first commit, let's hope it's not the last!

PR #24, opened 2016-04-06
don't abort test suite on first failure

ffrank:
The test.sh script aborts as soon as a test fails. This can save time
on the local machine, but is inconvenient in CI where an early minor
failure can mask more severe errors that will be found later.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -4,31 +4,43 @@ echo running test.sh
 echo "ENV:"
 env
 
+run-test()
+{
+	$@ || FAILURES=$( [ "$FAILURES" ] && echo "$FAILURES\\n$@" || echo "$@" )
+}
+
 # ensure there is no trailing whitespace or other whitespace errors
-git diff-tree --check $(git hash-object -t tree /dev/null) HEAD
+run-test git diff-tree --check $(git hash-object -t tree /dev/null) HEAD
 
 # ensure entries to authors file are sorted
 start=$(($(grep -n '^[[:space:]]*$' AUTHORS | awk -F ':' '{print $1}' | head -1) + 1))
-diff <(tail -n +$start AUTHORS | sort) <(tail -n +$start AUTHORS)
+run-test diff <(tail -n +$start AUTHORS | sort) <(tail -n +$start AUTHORS)
 
-./test/test-gofmt.sh
-./test/test-yamlfmt.sh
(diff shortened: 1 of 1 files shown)


purpleidea:
@ffrank I'm okay with this. Although can you change your bash \n code to use the easier pattern that doesn't require # replacement? An example of how to do this is here:

https://github.com/purpleidea/mgmt/blob/master/test/test-shell.sh#L68

You also don't need `exit 0` at the end. This is implicit in bash.

Cheers

ffrank:
Oh neat! Guess I was confused by Ruby semantics and assumed the value of the `if` condition would stick.

ffrank:
Uhm...I'm not sure this [contraption](https://github.com/purpleidea/mgmt/blob/2ab72bdf94cdc5aa3f272e69464d2aff65bba0f0/test/test-shell.sh#L8) is preferable to a `tr` based approach...?

purpleidea:
@ffrank that's not required, that's unrelated. You just need the -n part I mentioned, and the concatenation, eg: https://github.com/purpleidea/mgmt/blob/2ab72bdf94cdc5aa3f272e69464d2aff65bba0f0/test/test-shell.sh#L39

ffrank:
Yes, but I think the concatenation of `$LINE` is crucial in that script. If you just
```
FOO=$( echo one ; echo two )
```
then there is *no* newline in `$FOO`. You intersperse your values with `$LINE`, which seems to act as a "hard" line feed.

I'll just push a compromise, let me know if that's palatable :)

test.sh
```diff
@@ -4,31 +4,43 @@ echo running test.sh
...
 if env | grep -q -e '^JENKINS_URL=' -e '^BUILD_TAG=jenkins'; then
-	./test/test-omv.sh
+	run-test ./test/test-omv.sh
+fi
+run-test ./test/test-golint.sh	# test last, because this test is somewhat arbitrary
+
+if [ "$FAILURES" ] ; then
+	echo
+	echo "FAILED TESTS:"
+	echo -e $FAILURES
```
ffrank: The `-e` is to expand the the `\n` escape sequences into actual newlines.

purpleidea:
Great! I made a few nitpick fixes on a new commit after this and merged it all.

Thanks!

PR #35, opened 2016-05-14
packagekit: support Debian's 'all' architecture

ffrank:


purpleidea:
w00t, merged! First golang patch !

PR #37, opened 2016-05-14
Pkg: cache state when it is correct

purpleidea:
We forgot to cache the state when we are converged. This avoids
repetitive checking when we hit repeated backpoke()'s for example.

### DIFF
--- a/pkg.go
+++ b/pkg.go
@@ -157,7 +157,7 @@ func (obj *PkgRes) Watch(processChan chan Event) {
 			if exit, send = obj.ReadEvent(&event); exit {
 				return // exit
 			}
-			//dirty = false // these events don't invalidate state
+			dirty = false // these events don't invalidate state
 
 		case _ = <-cuuid.ConvergedTimer():
 			cuuid.SetConverged(true) // converged!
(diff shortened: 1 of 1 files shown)


ffrank:
Tested this with a graph that frequently runs an exec that depends on a package. The package uses its cached state fine.

```yaml
resources:
  pkg:
  - name: cowsay
    state: installed
  exec:
  - name: "/usr/games/cowsay boo"
    cmd: "/usr/games/cowsay boo"
    timeout: 300.0
    shell: "/bin/bash"
    watchshell: "/bin/bash"
    ifshell: ''
    watchcmd: '/bin/bash -c ''while : ; do echo "puppet run interval passed" ; /bin/sleep
      3 ; done'''
    ifcmd: 
    state: present
    pollint: 0
edges:
- name: e1
  from:
    kind: pkg
    name: cowsay
  to:
    kind: exec
    name: "/usr/games/cowsay boo"
```

(Generated from the following Puppet manifest using current `puppet-mgmtgraph` master)

```
bundle exec puppet mgmtgraph --runinterval 3 --code 'package { "cowsay": ensure => present } -> exec { "/usr/games/cowsay boo": }' >/tmp/cowsay.yaml
```

purpleidea:
w00t

PR #38, opened 2016-05-14
remove vet

tuxmea:
go 1.6 has vet included
https://github.com/hashicorp/vault/issues/1310#issuecomment-207922338


purpleidea:
@tuxmea This was useful information to know, thanks! I don't want to remove this just yet because we still build in 1.5.x and I actually still run that version too :) If it generates an error on 1.6, then a patch to disable it conditionally on > 1.6 would be welcome.

purpleidea:
@tuxmea ugh, GH doesn't notify when someone pushes new patches, so I just saw this now. It LGTM, but can you rebase to master and ping me please? Alternatively I can rebase for you if you prefer.

Thanks!

tuxmea:
@purpleidea Ping ;-)

purpleidea:
Sweet, thanks, merging after `git tpush` passes.

Issue #21, opened 2016-03-28
implement noop functionality


purpleidea:
We should implement a no-operation (noop) flag which causes a resource to not run apply.

This should be available as a metaparam (and default to false) for per resource setting, and as a --noop flag, which would set all noop values to true.

Since this should be fairly straight forward, I will tag this is an [mgmtlove](https://github.com/purpleidea/mgmt/labels/mgmtlove) bug for new contributors.

purpleidea:
Talking with @tuxmea we've agreed that:

1) There's a metaparam for noop (per resource)
2) There's a global --noop
3) Setting the global --noop, works by setting the individual noop metaparam per resource
4) Setting the global --noop disables the "export" (to etcd) operation
5) Setting the global --noop does NOT disable the "collect" (from etcd) but *does* set the noop metaparam on everything collected.

purpleidea:
Additionally, when we add the --noop flag, we want resources to Compare() as identical to existing resources that don't have it, because their state is fine, and we don't need to regenerate them in the differential graph compile.

If we have a resource that IS noop, and the new version doesn't have it, then we DO want to re-build it, because it won't necessarily be in the correct state.

purpleidea:
I think this is now fixed in https://github.com/purpleidea/mgmt/commit/9f56e4a582473840418794c2f916f4e22cbf4213

w00t.

PR #39, opened 2016-05-14
Noop meta parameter

tuxmea:
First set of noop implementation.
CLI only has the parameter, but the parameter is not yet used.
Also see https://github.com/purpleidea/mgmt/issues/21#issuecomment-219081096

### DIFF
--- a/examples/graph0noop.yaml
+++ b/examples/graph0noop.yaml
@@ -1,0 +1,24 @@
+---
+graph: mygraph
+comment: hello world example
+resources:
+  noop:
+  - name: noop1
+    meta:
+      noop: true
+  file:
+  - name: file1
+    path: "/tmp/mgmt-hello-world"
+    content: |
+      hello world from @purpleidea
+    state: exists
+    neta:
+      noop: true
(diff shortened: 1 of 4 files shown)


purpleidea:
@tuxmea Thanks for the patch... I think the LGTM, but can you rebase into a single commit? I think we should include the --noop flag with the patch that implements it. LMK if you want to claim that patch or you're abandoning it.

tuxmea:
I will abandon the whole patch as it is not working yet. Just tested on Linux. Seems as if this still needs some work.
I will continue working on the noop metparameter but not on CLI noop option.

tuxmea:
I always screw squash whith a merge. Closing this one, reopening a new PR.

purpleidea:
On Sat, May 14, 2016 at 3:29 PM, Martin Alfke <notifications@github.com>
wrote:

> Closed #39 <https://github.com/purpleidea/mgmt/pull/39>.
>

You can also just fix the branch, and `git push <yourremote> branchname
--force` if you prefer.

PR #41, opened 2016-05-14
Add "Why is etcd packaged in?" FAQ section

czerasz:


purpleidea:
@czerasz Thanks for the patch, although a few grammar issues and the lack of this feature existing in git master at the moment means i'm going to nack this patch. I've written up an answer in a dev branch which will hopefully hit git master soon. Here's that text for your early consumption and review:
```
###Can I use an existing etcd cluster instead of the automatic embedded servers?

Yes, it's possible to use an existing etcd cluster instead of the automatic,
elastic embedded etcd servers. To do so, simply point to the cluster with the
`--seeds` variable, the same way you would if you were seeding a new member to
an existing mgmt cluster.

The downside to this approach is that you won't benefit from the automatic
elastic nature of the embedded etcd servers, and that you're responsible if you
accidentally break your etcd cluster, or if you use an unsupported version.
```

PR #44, opened 2016-05-19
Simplify file watch

ffrank:
So here's my proposal for an easier logic around watching files and their parent directories if necessary. It's quite a mouthful, it is likely easier to review the initial commit first. It's the biggest, but it avoids unnecessary shortcuts and should be more comprehensible than the full diff.

The followups clear up some problems I found during testing. Here's how to reproduce. Graph:
```yaml
---
graph: mygraph
resources:
  file:
  - name: file1
    path: "/tmp/mgmt-test/a/b/c/file"
    content: |
      i am file
    state: exists
edges: []
```

Run mgmt without a/b/c/ existing. It will watch /tmp/mgmt-test. Now `mkdir -p /tmp/mgmt-test/a/b/c`. Current master will start watching `/tmp/mgmt-test/a`, and miss both a/b and a/b/c.

The new logic just finds the longest path prefix it can watch on every loop iteration, and *will* start watching `/tmp/mgmt-test/a/b/c`.

This still has a major problem: Even though mgmt is now *watching* the file's parent, it has missed its creation event, and hence not `CheckApply` is triggered. This is a tough one, unless we pragmatically decide to send resource events from *any* directory creation notification.

All that said, I hope this proposal makes it easier to reason about this code's behavior. It will definitely help me add directory handling. Note that I did away with carrying the deltaDepth in favor of doing the math whenever required.

### DIFF
--- a/file.go
+++ b/file.go
@@ -149,16 +149,16 @@ func (obj *FileRes) Watch(processChan chan Event) {
 					if watchDepth < 0 {
 						log.Fatal("somehow trying to watch file above the fs root")
 					}
-				} else if err == syscall.ENOSPC {
+					continue
+				}
+				if err == syscall.ENOSPC {
 					// XXX: occasionally: no space left on device,
 					// XXX: probably due to lack of inotify watches
 					log.Printf("%v[%v]: Out of inotify watches!", obj.Kind(), obj.GetName())
-					log.Fatal(err)
 				} else {
 					log.Printf("Unknown file[%v] error:", obj.Name)
-					log.Fatal(err)
 				}
-				continue
+				log.Fatal(err)
 			}
 			watching = true
 		}
--- a/file.go
+++ b/file.go
@@ -25,7 +25,6 @@ import (
 	"encoding/gob"
 	"io"
 	"log"
-	"math"
 	"os"
 	"path"
 	"strings"
@@ -114,7 +113,7 @@ func (obj *FileRes) Watch(processChan chan Event) {
 	//var recursive bool = false
 	//var isdir = (obj.GetPath()[len(obj.GetPath())-1:] == "/") // dirs have trailing slashes
 	//log.Printf("IsDirectory: %v", isdir)
-	var safename = path.Clean(obj.GetPath()) // no trailing slash
+	var cleanObjPath = path.Clean(obj.GetPath()) // no trailing slash
 
 	watcher, err := fsnotify.NewWatcher()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


file.go
```diff
@@ -114,121 +113,105 @@ func (obj *FileRes) Watch(processChan chan Event) {
...
-
-			} else if HasPathPrefix(event.Name, current) {
-				deltaDepth = len(PathSplit(current)) - len(PathSplit(event.Name)) // -1 or less
-
-			} else if HasPathPrefix(current, event.Name) {
-				deltaDepth = len(PathSplit(event.Name)) - len(PathSplit(current)) // +1 or more
 
-			} else {
+			if !HasPathPrefix(event.Name, cleanObjPath) && !HasPathPrefix(cleanObjPath, event.Name) {
 				// TODO different watchers get each others events!
```
ffrank: I've been thinking about this one. The behavior you describe in that bug report might be exhibited my mgmt because of the whole watching logic. Here's what may have happened, I figure:

1. You run mgmt with the `file1.yaml` example, but no `/tmp/mgmt`. `/tmp` is watched by all four file resources.
2. You create `/tmp/mgmt`. All four resources race to start watching it and run a `CheckApply`.
3. At the time `f3` is created, the `f2` resource may still be watching `/tmp/mgmt`, so the `f3` creation event ends up in its notify queue.
4. After the `f2` resource has consumed the `/tmp/mgmt` event and descended in to watch `/tmp/mgmt/f2`, it takes the `f3` event out of its queue.
5. Boom! :)

Something along those lines. It seems more plausible than a library or kernel bug.

purpleidea:
@ffrank This patch looks like it mostly renames a number of things and as you said tries to simplify stuff. While this is a fine goal, I don't really want to patch this code unless we're also adding the recursive watching in the same patch set.

PS: line 179, obj.GetName() not obj.GetName (as seen in go vet failure)

purpleidea:
If I've misunderstood, LMK.

ffrank:
I can pull it into that branch, sure thing. Have you spotted anything that you're opposed to already? If you want to postpone thorough reviews until that feature is complete, that's fine too. This can be closed in that case.

purpleidea:
@ffrank My ack/nack based on the future patch will be dependent on how correct and complete the future code is... I thought the current implementation is good and extendable, but a different implementation is fine too if it's correct/works/etc...

Closing as requested.

file.go
ffrank: Have you seen this as well? Thoughts?

purpleidea: I think the correct behaviour of the library is to only give you events you've asked for, but in the scenario that it gives us events we don't want (as seen here) we just ignore them.

PR #45, opened 2016-05-20
README: add the capnslog package to the build dependencies

ffrank:


purpleidea:
indent characters don't match, so NACK, but I kind of want to NACK all of your patches until you get a _real good_ one in like --puppet or dir first :)

ffrank:
Fixed, fwiw

PR #46, opened 2016-05-24
move doc files to /usr/share/doc/mgmt and tag as docs

jumanjiman:
After this commit:

* doc files are in `/usr/share/doc/mgmt-*/`
* `rpm -qd mgmt` lists the files
* docs are listed on single lines in spec file
  to minimize future diff churn


jumanjiman:
ps. docs are also in alpha order in the spec file

purpleidea:
@jumanjiman Sweet, someone who knows things about packaging... This LGTM, as long as you can ACK that you tested this... Thanks!

jumanjiman:
i just force-pushed an updated commit (for the `examples` line to guarantee files in subdir are included regardless of rpm version) that i'm confident is correct.

😢 unfortunately i cannot attest to fully testing this commit.
The rpm on master is incredibly difficult to build atm.

I empathize with what you're currently doing with `spec.in`.

What I'd *really* love to do is to convert the `spec.in` to use https://github.com/dgoodwin/tito from the awesome @dgoodwin which would allow to tie the git history to an rpm spec file, build with copr, submit to koji, and so forth.

If you're up for it, i can submit a series of commits (atomic, of course), that allow to build via tito.

purpleidea:
On Mon, May 23, 2016 at 9:49 PM, Paul Morgan <notifications@github.com> wrote:
>
> The rpm on master is incredibly difficult to build atm.



Why is this?? (I have automatic builds on copr:
https://copr.fedorainfracloud.org/coprs/purpleidea/mgmt/

They take your .config/copr file and build on a ```make copr```

purpleidea:
@jumanjiman I figured I'd merge this. If future RPM builds get broken, you get to keep both pieces ;)

Thanks for the patch!

purpleidea:
(It's also worth noting, that I put the RPM mechanics in place so that future users would have this as an easy to consume option, but it's not really targeted at users just yet, since it's really a dev project at the moment!) Hopefully we'll get it all shiny soon! I've got the etcd integration coming soon. Cheers!

PR #47, opened 2016-05-24
Puppet integration

### LINKED ISSUE #8: Allow mgmt to run from Puppet manifests
Still on that, but no usable results yet. I had used my first PoC Ruby script to reproduce issue #4.

My initial approach of using the output of `puppet master --compile` cannot work. Not only does it not include actual dependency edges, it's also much too raw. I will blog about this with details.

What I will try to do instead is to rely on the agent side graph generating code, to let Puppet help with validation, munging and the computation of actual edges.

ffrank:
Puppet can be used on the basis of the `ffrank-mgmtgraph` module.
There are three modes available:
* fetching catalogs from the master (--puppet agent)
* compiling a manifest from a local file (--puppet /path/to/file.pp)
* compiling a manifest from the cli (--puppet "<manifest>")
                        
Catalogs from the master are currently never refreshed. We should
add some more code to re-run the parsing function at an interval
equal to Puppet's local 'runinterval' setting.
                                        
There is also still a distinct lack of tests.
                                                
Still, this fixes #8

### DIFF
--- a/main.go
+++ b/main.go
@@ -65,6 +65,11 @@ func run(c *cli.Context) error {
 	log.Printf("Main: Start: %v", start)
 	var G, fullGraph *Graph
 
+	if c.IsSet("file") && c.IsSet("puppet") {
+		log.Println("the --file and --puppet parameters cannot be used together")
+		return cli.NewExitError("", 1)
+	}
+
 	// exit after `max-runtime` seconds for no reason at all...
 	if i := c.Int("max-runtime"); i > 0 {
 		go func() {
@@ -110,9 +115,13 @@ func run(c *cli.Context) error {
 		startchan := make(chan struct{}) // start signal
 		go func() { startchan <- struct{}{} }()
 		file := c.String("file")
-		configchan := make(chan bool)
-		if !c.Bool("no-watch") {
+		var configchan chan bool
+		var puppetchan <-chan time.Time
+		if !c.Bool("no-watch") && c.IsSet("file") {
 			configchan = ConfigWatch(file)
+		} else if c.IsSet("puppet") {
+			interval := PuppetInterval(c.String("puppet-conf"))
+			puppetchan = time.Tick(time.Duration(interval) * time.Second)
 		}
 		log.Println("Etcd: Starting...")
 		etcdchan := etcdO.EtcdWatch()
@@ -133,6 +142,8 @@ func run(c *cli.Context) error {
 				default:
 					log.Fatal("Etcd: Unhandled message: ", msg)
 				}
+			case _ = <-puppetchan:
+				// nothing, just go on
 			case msg := <-configchan:
 				if c.Bool("no-watch") || !msg {
 					continue // not ready to read config
@@ -144,7 +155,12 @@ func run(c *cli.Context) error {
 				return
 			}
 
-			config := ParseConfigFromFile(file)
+			var config *GraphConfig
+			if c.IsSet("file") {
+				config = ParseConfigFromFile(file)
+			} else if c.IsSet("puppet") {
+				config = ParseConfigFromPuppet(c.String("puppet"), c.String("puppet-conf"))
+			}
 			if config == nil {
 				log.Printf("Config parse failure")
 				continue
@@ -299,6 +315,16 @@ func main() {
 				cli.BoolFlag{
 					Name:  "noop",
 					Usage: "globally force all resources into no-op mode",
+				},
+				cli.StringFlag{
+					Name:  "puppet, p",
+					Value: "",
+					Usage: "load graph from puppet, optionally takes a manifest or path to manifest file",
+				},
+				cli.StringFlag{
+					Name:  "puppet-conf",
+					Value: "",
+					Usage: "supply the path to an alternate puppet.conf file to use",
 				},
 			},
 		},
--- a/puppet.go
+++ b/puppet.go
@@ -1,0 +1,137 @@
+// Mgmt
+// Copyright (C) 2013-2016+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"bufio"
+	"io"
+	"log"
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+const (
+	PuppetYAMLBufferSize = 65535
+)
+
+func runPuppetCommand(cmd *exec.Cmd) ([]byte, error) {
+	if DEBUG {
+		log.Printf("Puppet: running command: %v", cmd)
+	}
+
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		log.Printf("Puppet: Error opening pipe to puppet command: %v", err)
+		return nil, err
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		log.Printf("Puppet: Error opening error pipe to puppet command: %v", err)
+		return nil, err
+	}
+
+	if err := cmd.Start(); err != nil {
+		log.Printf("Puppet: Error starting puppet command: %v", err)
+		return nil, err
+	}
+
+	// XXX: the current implementation is likely prone to fail
+	// as soon as the YAML data overflows the buffer.
+	data := make([]byte, PuppetYAMLBufferSize)
+	var result []byte
+	for err == nil {
+		var count int
+		count, err = stdout.Read(data)
+		if err != nil && err != io.EOF {
+			log.Printf("Puppet: Error reading YAML data from puppet: %v", err)
+			return nil, err
+		}
+		// Slicing down to the number of actual bytes is important, the YAML parser
+		// will choke on an oversized slice. http://stackoverflow.com/a/33726617/3356612
+		result = append(result, data[0:count]...)
+	}
+	if DEBUG {
+		log.Printf("Puppet: read %v bytes of data from puppet", len(result))
+	}
+	for scanner := bufio.NewScanner(stderr); scanner.Scan(); {
+		log.Printf("Puppet: (output) %v", scanner.Text())
+	}
+	if err := cmd.Wait(); err != nil {
+		log.Printf("Puppet: Error: puppet command did not complete: %v", err)
+		return nil, err
+	}
+
+	return result, nil
+}
+
+func ParseConfigFromPuppet(puppetParam, puppetConf string) *GraphConfig {
+	var puppetConfArg string
+	if puppetConf != "" {
+		puppetConfArg = "--config=" + puppetConf
+	}
+
+	var cmd *exec.Cmd
+	if puppetParam == "agent" {
+		cmd = exec.Command("puppet", "mgmtgraph", "print", puppetConfArg)
+	} else if strings.HasSuffix(puppetParam, ".pp") {
+		cmd = exec.Command("puppet", "mgmtgraph", "print", puppetConfArg, "--manifest", puppetParam)
+	} else {
+		cmd = exec.Command("puppet", "mgmtgraph", "print", puppetConfArg, "--code", puppetParam)
+	}
+
+	log.Println("Puppet: launching translator")
+
+	var config GraphConfig
+	if data, err := runPuppetCommand(cmd); err != nil {
+		return nil
+	} else if err := config.Parse(data); err != nil {
+		log.Printf("Puppet: Error: Could not parse YAML output with Parse: %v", err)
+		return nil
+	}
+
+	return &config
+}
+
+func PuppetInterval(puppetConf string) int {
+	if DEBUG {
+		log.Printf("Puppet: determining graph refresh interval")
+	}
+	var cmd *exec.Cmd
+	if puppetConf != "" {
+		cmd = exec.Command("puppet", "config", "print", "runinterval", "--config", puppetConf)
+	} else {
+		cmd = exec.Command("puppet", "config", "print", "runinterval")
+	}
+
+	log.Println("Puppet: inspecting runinterval configuration")
+
+	interval := 1800
+	data, err := runPuppetCommand(cmd)
+	if err != nil {
+		log.Printf("Puppet: could not determine configured run interval (%v), using default of %v", err, interval)
+		return interval
+	}
+	result, err := strconv.ParseInt(strings.TrimSpace(string(data)), 10, 0)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


purpleidea:
@ffrank Cool, excited to read this patch!

I noticed you put a lot of stuff in main.go but ideally we'd like to put less there. I think the bulk of this should go in a new `puppet.go` file.

Changing the ParseConfigFromFile signature, is a bit of a dirty hack :P Instead, I think it's better to have one function that returns a graphconfig, and it takes args like c.String("puppet") and c.String("puppet-conf") and does that work there.

ffrank:
I just pushed a refactored version. Much nicer, thanks for the feedback.

main.go
```diff
@@ -65,6 +66,10 @@ func run(c *cli.Context) error {
 	log.Printf("Main: Start: %v", start)
 	var G, fullGraph *Graph
 
+	if c.IsSet("file") && c.IsSet("puppet") {
+		log.Fatal("the --file and --puppet parameters cannot be used together")
```
purpleidea: Good, but this should be a parse error instead...
Try using:

		log.Printf("blah blah...")
		return cli.NewExitError("", 1)

higher up, maybe right after the noop := part

puppet.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"bufio"
+	"io"
+	"log"
+	"os/exec"
+	"strings"
+)
+
+func PuppetConfig(puppetParameter, puppetConf string) *GraphConfig {
```
purpleidea: I like it but how about ParseConfigFromPuppet instead to be consistent with the config.go equivalent. If you'd prefer a different naming pattern that's cool, but then rename ParseConfigFromFile to match the new pattern.

purpleidea: > puppetParameter

How about puppet or puppetParam instead? Keeping things short is very golang and improves readability.

puppet.go
```diff
@@ -0,0 +1,89 @@
...
+		cmd = exec.Command("puppet", "mgmtgraph", "print", puppetConfArg, "--code", puppetParameter)
+	}
+
+	if DEBUG {
+		log.Printf("running command: %v", cmd)
+	}
+
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		log.Printf("Error opening pipe to puppet command: %v", err)
```
purpleidea: Please prefix all of your error messages with "Puppet: "
When I finish the error handling stuff we can look at having nicer structured logging (medium/long term)

puppet.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		log.Printf("Error opening error pipe to puppet command: %v", err)
+		return nil
+	}
+
+	log.Println("launching puppet translator")
```
purpleidea: Prefix these log messages too please. It ends up being shorter and more obvious, eg: "Puppet: Launching translator"

puppet.go
```diff
@@ -0,0 +1,89 @@
...
+	if err := cmd.Start(); err != nil {
+		log.Printf("Error starting puppet command: %v", err)
+		return nil
+	}
+
+	data := make([]byte, 65535)
```
purpleidea: Is this value arbitrary? If so please add it as a const in this file.

main.go
```diff
@@ -111,8 +116,15 @@ func run(c *cli.Context) error {
...
 		configchan := make(chan bool)
+		// XXX Unwatchable Puppet sources should get events from a timer,
+		// with the same interval as the Puppet agent
+		// (hint: puppet config print runinterval)
 		if !c.Bool("no-watch") {
-			configchan = ConfigWatch(file)
+			if c.IsSet("file") {
+				configchan = ConfigWatch(file)
+			} else if strings.HasSuffix(c.String("puppet"), ".pp") {
+				configchan = ConfigWatch(c.String("puppet"))
```
purpleidea: While this is clever, and you get style points, I'd actually like to argue against this, because it's not consistent with well-known puppet behaviour. As a bonus for removing it, we also don't expose the parsing logic in more than one place too.

Instead, as you XXX mentions (and I whole heartedly agree) we should just implement this :) One catch would be to check that the converge timeout isn't longer than the puppet timeout or it will never end. Maybe add a warning in this case.

purpleidea:
@ffrank Initial review done, I gotta stop now and go eat, but I'll review again after changes. Looks pretty close over all! Thanks!

puppet.go
ffrank: I considered plain `puppet`, but thought it would read quite ambiguous (what "puppet"? Is this an interface to an actual Puppet (REST?), is it a boolean flag...?)

Will think a little harder. Perhaps is fine and I'm just freaking out ;)

main.go
ffrank: My reasoning is that `mgmt run --puppet /path/to/my.pp` is most akin to `puppet apply` a.k.a. masterless. This is quite different from `puppet agent` in that it *never* uses an interval to re-converge. It has to be triggered by hand (or cron).

Since `mgmt` works differently and will likely *never* launch from cron, I figure watching the manifest is a sound equivalent under our paradigm.

purpleidea: Good points however:

1) When we're running with an agent (normal mode) we'll automatically re-run puppet every 30 min. This should be part of the patch.

2) mgmt itself _could_ more generally be run in "puppet mode" where puppet mode in this case means, run every 30 minutes from cron. To do so we use the --converged-timeout flag and we don't most of the of the instant events benefits.

puppet.go
purpleidea: Yeah, puppetParam or puppetArg or similar is fine.

main.go
ffrank: OK, that also makes sense.

One thing I'm worried about. If we choose to *not* watch the manifest file, `mgmt` will ignore changes and pick them up eventually, once the 30 minute mark passed. I suspect that this behavior will surprise users. It is also inconvenient (assuming that the user does not want to restart the `mgmt` process to force a retranslation. Please note that, on the other hand, this compromise would likely be quite acceptable to most...we kick services all the time, so why not this one.)

ffrank:
Here's a new commit. Will squash it once review is finished.

Note that the stdlib's `time` package also has a `Ticker` for dispatching regular events, but that seemed overkill to me. Hard to beat a four line go routine that does all that's needed (i.e. sleeping and signaling).

main.go
purpleidea: Here are some other points that might convince you. Let me know what you think...

1) If we're doing the .pp file watching, then this is only going to work when we have local puppet files we're running. For the other two cases (using the puppet server, or inline code) this doesn't happen, so the behaviour is unique to one use case, and not consistent elsewhere.

2) Reminder, even though we'll generate a (possibly) new graph every 30 min, it will only be different if we've updated the code, or the facts. Once it's running, you still get the dynamic event based and parallel goodness.

puppet.go
```diff
@@ -22,68 +22,119 @@ import (
...
-		log.Printf("Error starting puppet command: %v", err)
-		return nil
+		log.Printf("Puppet: Error starting puppet command: %v", err)
+		return nil, err
 	}
 
-	data := make([]byte, 65535)
-	var config GraphConfig
+	// XXX: the current implementation is likely prone to fail
+	// as soon as the YAML data overflows the buffer.
```
purpleidea: What would we need to do to fix this so that it doesn't ever happen?

puppet.go
```diff
@@ -22,68 +22,119 @@ import (
...
+	} else {
+		result, err := strconv.ParseInt(strings.TrimSpace(string(data)), 10, 0)
+		if err != nil {
+			log.Printf("Puppet: error reading numeric runinterval value (%v), using default of %v", err, interval)
+		} else {
+			interval = int(result)
+		}
+	}
+
+	ch := make(chan bool)
```
purpleidea: There's a built-in that should make this easier... https://golang.org/pkg/time/#After

main.go
```diff
@@ -67,7 +66,8 @@ func run(c *cli.Context) error {
 	var G, fullGraph *Graph
 
 	if c.IsSet("file") && c.IsSet("puppet") {
-		log.Fatal("the --file and --puppet parameters cannot be used together")
+		log.Println("the --file and --puppet parameters cannot be used together")
+		return cli.NewExitError("invalid parameters", 1)
```
purpleidea: Remove the "invalid parameters" text and leave an empty string, the log function spits out the nice message with a timestamp, where as the NewExitError function doesn't log it. This way it will be consistent throughout. (Run it to see what I mean.)

puppet.go
```diff
@@ -22,68 +22,119 @@ import (
...
-	var cmd *exec.Cmd
-	if puppetParameter == "agent" {
-		cmd = exec.Command("puppet", "mgmtgraph", "print", puppetConfArg)
-	} else if strings.HasSuffix(puppetParameter, ".pp") {
-		cmd = exec.Command("puppet", "mgmtgraph", "print", puppetConfArg, "--manifest", puppetParameter)
-	} else {
-		cmd = exec.Command("puppet", "mgmtgraph", "print", puppetConfArg, "--code", puppetParameter)
-	}
+const (
+	PUPPET_YAML_BUFFER_SIZE = 65535
```
purpleidea: Use golang standard formatting, eg camelCase or CamelCase. So PuppetYAMLBufferSize for example.

main.go
```diff
@@ -111,8 +116,15 @@ func run(c *cli.Context) error {
 		go func() { startchan <- struct{}{} }()
 		file := c.String("file")
 		configchan := make(chan bool)
+		// XXX Unwatchable Puppet sources should get events from a timer,
```
purpleidea: This message is no longer needed...

main.go
```diff
@@ -111,8 +116,15 @@ func run(c *cli.Context) error {
...
+		// with the same interval as the Puppet agent
+		// (hint: puppet config print runinterval)
 		if !c.Bool("no-watch") {
```
purpleidea: This code isn't correct, because it's wrapped inside the no-watch flag. Instead, I think it's more appropriate for the puppet thing to have it's own first class entry in the select loop just below this... Reminder, a nil channel is always ignored, so if we're not using this, then a nil value means we never run that case statement.

puppet.go
```diff
@@ -0,0 +1,140 @@
...
+	if puppetConf != "" {
+		cmd = exec.Command("puppet", "config", "print", "runinterval", "--config", puppetConf)
+	} else {
+		cmd = exec.Command("puppet", "config", "print", "runinterval")
+	}
+
+	log.Println("Puppet: inspecting runinterval configuration")
+
+	interval := 1800
+	if data, err := runPuppetCommand(cmd); err != nil {
```
purpleidea: Lastly, one added benefit, is that if we move the timer stuff into the main select loop, then this code can be inlined (eg: not nested anymore) which I think makes it more elegant too!

purpleidea:
@ffrank Okay review done. Thanks again for working on this. I hope that you find my reviews fair and that you are benefiting from them. It is good for me too, since I am (I hope) getting better a reading code!

Cheers!

puppet.go
ffrank: Ah, but don't we really need a `bool` in our channel, as opposed to a `Time` object?

main.go
ffrank: I figured that it makes sense for `--no-watch` to suppress recompiling / querying the master.

puppet.go
ffrank: What exactly would you like inlined where?

puppet.go
ffrank: We will want some loop construct that keeps reading and catenating data until we have it all.

ffrank:
Hey, thanks. Yep, good reviewing going on here :) keep it up.

puppet.go
ffrank: Actually took the cue from const DEBUG ;)

ffrank:
Oh, one thing: I noticed that some of your comments were made in the commit view of 3ebaaf6. These don't show up in the diff view, unfortunately, so it would be better to do all commenting in the diff I think.

puppet.go
purpleidea: On Thu, May 26, 2016 at 8:21 AM, Felix Frank <notifications@github.com>
wrote:

> We will want some loop construct that keeps reading and catenating data
> until we have it all.



Can you add this?

main.go
purpleidea: On Thu, May 26, 2016 at 8:19 AM, Felix Frank <notifications@github.com>
wrote:

> I figured that it makes sense for --no-watch to suppress recompiling /
> querying the master.
>

Except we won't have puppet watches on the catalog...

purpleidea:
@ffrank Good point. I apparently fail at GH UI :P LMK when ready for next review :) I'm lame and hacking this nice sunny weekend.

puppet.go
purpleidea: Bool or no bool doesn't matter. We just care about something happening in the select.

ffrank: It won't compile though. I tried to go for `Tick` rather than `After` (we do want regular events), but it's a type mismatch.

```
./puppet.go:137: cannot use time.Tick(time.Duration(interval) * time.Second) (type <-chan time.Time) as type chan bool in return argument
```

puppet.go
ffrank: Are you saying that this shouldn't be a function at all? I'm not actually inclined to move any of this into the main loop. It's nice that the respective channel is created either in the configwatch or puppet module (is that the right term?) What am I missing?

ffrank:
New commit is up. Two inline threads remain unresolved for now, see new comments above.

purpleidea:
17:03 <@purpleidea> ffrank: i think my main issue is you're trying to re-use 
                    the config chan! why? you can have your OWN select cast 
                    statement in the big select! then it doesn't have to match 
                    the type!
17:04 <@purpleidea> and then you can use the time.After thing to make your life 
                    easier.
17:05 <@purpleidea> IOW if we're using puppet configchan = nil, if we're using 
                    file puppetchan (or whatever) is nil

purpleidea:
@ffrank In any case, this is so close it's painful. Perhaps it's ready for a squash and rebase to master, and perhaps you could open it in a new pull request even. If you'd prefer not to rework the select part, I'm happy to do that as a new patch on top.

puppet.go
```diff
@@ -109,7 +108,7 @@ func ParseConfigFromPuppet(puppetParam, puppetConf string) *GraphConfig {
 	return &config
 }
 
-func PuppetInterval(puppetConf string) chan bool {
+func GetPuppetInterval(puppetConf string) int {
```
purpleidea: You don't need to change the name. Golang avoids Get prefixes when possible. You had it right before :)

puppet.go
```diff
@@ -134,12 +133,5 @@ func PuppetInterval(puppetConf string) chan bool {
 		}
 	}
 
-	ch := make(chan bool)
```
purpleidea: The code right above this can now be inlined! Right?

purpleidea:
@ffrank Can you confirm that you tested _this_ version?

purpleidea:
@mohae any comments?

ffrank:
Yes, it passed my smoke tests.

purpleidea:
On Tue, May 31, 2016 at 7:37 PM, Felix Frank <notifications@github.com>
wrote:

> Yes, it passed my smoke tests.



Woooo! Merged <fireworks emoji>

I think this counts as your first real golang patch, so welcome to the club!

If you're bored and want to do the mgmtgraph class stuff and write a blog
post, I'm excited for it. What else is there to do to make this really
usable? What else is missing?

ffrank:
Yeah no that'll about cover it. :tada:

purpleidea:
@ffrank Sweet! Thanks again for the patch and your patience during the review. I am glad to get it in git master. Check out my IRC comment.

PR #49, opened 2016-06-27
Adding Timer resource

sharadgana:
@purpleidea


purpleidea:
@sharadgana Okay, firstly, thanks for your patch! May it not be the last. Here's some initial review...

* You'll need to squash everything into a single commit. If you're not sure how to do this, LMK and I'll help.
* You have trailing whitespace in the yaml file. Look at the travis error log to see the mistake.
* timer.yaml file should be named timer1.yaml and instead of the file resource in there, how about you put an exec instead that echo's hello?
* Need one newline before the `package main` text.
* You'll need to run gofmt -w on your code. You can type make gofmt to do this for you.
* You need the relevant changes in config.go -- these should be pretty straight forward to figure out by looking at the file, but ask if you're not sure.
* You need to actually test this! Since you don't have the changes made in config.go this won't work, so I'll assume you didn't test the code before submitting :(
* Other comments in the code...

timer.go
```diff
@@ -0,0 +1,137 @@
...
+		Interval: interval,
+		Ticker:   time.NewTicker(time.Duration(interval) * time.Second),
+	}
+	obj.Init()
+	return obj
+}
+
+func (obj *TimerRes) Init() {
+	obj.BaseRes.kind = "Timer"
+	obj.BaseRes.Init() //call base init, b/c we're overrriding
```
purpleidea: You clearly copied and pasted from:
```
	obj.BaseRes.Init() // call base init, b/c we're overriding
```
which is fine, but somehow you managed to get rid of the space after the //   .... why and how???

timer.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+//Validate the params that are passed to TimerRes
```
purpleidea: Lack of leading spaces... Other places too...

timer.go
```diff
@@ -0,0 +1,137 @@
...
+	defer obj.SetWatching(false)
+	cuuid := obj.converger.Register()
+	defer cuuid.Unregister()
+
+	var send = false
+
+	for {
+		obj.SetState(resStateWatching)
+		select {
+		case _ = <-obj.Ticker.C: //Recieved the timer event
```
purpleidea: Typo: received*

timer.go
```diff
@@ -0,0 +1,137 @@
...
+			resp := NewResp()
+			processChan <- Event{
+				eventNil,
+				resp,
+				"Timer Kicked",
+				true,
+			}
+			resp.ACKWait()
+		}
+
```
purpleidea: Lots of extra newlines... Why?

purpleidea:
Overall the boilerplate is correct, but there are a bunch of nitpicks which you need to be more through about, and you didn't test this, which is bad :( Please look into the details and any other small issues you may find, and resubmit when ready! Thanks!

On the feature side, this currently only supports a simple interval, but maybe it would be interesting to have another option to support a different "function" (in the math sense) or algorithm to increment?

timer.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+		case _ = <-cuuid.ConvergedTimer():
+			cuuid.SetConverged(true)
+			continue
+		}
+		if send {
+			send = false
+			obj.isStateOK = false
+			resp := NewResp()
+			processChan <- Event{
```
purpleidea: This can probably all be on one line.

timer.go
```diff
@@ -0,0 +1,137 @@
...
+				eventNil,
+				resp,
+				"Timer Kicked",
```
purpleidea: Kicked? ticked?

timer.go
purpleidea: _ = is unnecessary

timer.go
```diff
@@ -0,0 +1,137 @@
...
+	name string
+}
+
+func NewTimerRes(name string, interval int, command string) *TimerRes {
+	obj := &TimerRes{
+		BaseRes: BaseRes{
+			Name: name,
+		},
+		Interval: interval,
+		Ticker:   time.NewTicker(time.Duration(interval) * time.Second),
```
purpleidea: I'm not sure you want to do this here... You want to create this in Watch() and also defer the Close() in the same function. Think of it as the Main() of the resource.

PR #54, opened 2016-07-07
Check for old golang versions while installing dependencies

### LINKED ISSUE #51: Build doesn't check for old golang
I tried to build `mgmt` using the default golang in debian jessie, which I later realised is 1.3.3. It would be helpful for the build scripts to check the golang version and exit with a useful message if it's too old.

jerith:
This resolves #51.


purpleidea:
w00t! congrats on your first patch! may there be many more! <fireworks emoji>

Issue #55, opened 2016-07-17
Cannot build latest master


ffrank:
Using `go1.5.4`, this is what I'm getting:

```
Building: mgmt.static, version: 0.0.3-61-ga1ed034...
go generate
stringer: checking package: etcd.go:119:15: OpOption not declared by package etcd
event.go:20: running "stringer": exit status 1
Makefile:78: recipe for target 'mgmt.static' failed
make: *** [mgmt.static] Error 1
```

purpleidea:
@ffrank Does https://github.com/purpleidea/mgmt/pull/56 fix your issue? Git master is broken on 1.6.x but not on 1.5.x for me. Vendoring sucks :(

purpleidea:
@ffrank Build is officially not broken anymore on travis. If it's broken for you, probably something is wrong/old in your $GOPATH.

Having said all this, figuring out how this godep nonsense works is turning out to be a pain. This is surely related to why golang 1.6 builds aren't working. If someone wants to work on that, please LMK.

purpleidea:
Note: I'm going to try and port the code to the "embed" code, so maybe after that's working it will get easier dep wise. https://godoc.org/github.com/coreos/etcd/embed

ffrank:
Well I did also try a fresh GOPATH, but ran into the issues addressed in #56. Tiring.

purpleidea:
@ffrank I have moved to the new embed etcd code. If you still have this issue please let me know, or create another or ping me in IRC. Thanks!

ffrank:
Hmm, ran into another issue after today's `pull` that required me to `rm *_stringer.go`.

This used to be covered by `make clean`, but not anymore since 2a0a69c. What's the reasoning here?

purpleidea:
On Wed, Jul 27, 2016 at 6:21 PM, Felix Frank <notifications@github.com>
wrote:

> Hmm, ran into another issue after today's pull that required me to rm
> *_stringer.go.
>
> This used to be covered by make clean, but not anymore since 2a0a69c
> <https://github.com/purpleidea/mgmt/commit/2a0a69c917c4108632cb6b186d5042c22e6b39c4>.
> What's the reasoning here?
>

Good question! The go generate command is broken in terms of some vendoring
stuff, or at least behaved strangely, so i just stopped including the
stringing. It works without it. If you want the strings, just run it.

PR #56, opened 2016-07-17
add new build requirements to README.md

ffrank:


purpleidea:
So all of these deps are really part of etcd, and they have their deps, and so on, so I'm not sure if it makes sense to always list these. Usually when you run go get ... https://github.com/purpleidea/mgmt/blob/master/misc/make-deps.sh#L41 you get everything automatically.

ffrank:
Until I ran those `go get`s, the compiler would complain that they were missing.

Perhaps I should have gone and update `etcd` instead? Will try that later. Hopefully that *will* fix #55 then.

We need some sort of `bundler` for this mess. In the meantime, could there perhaps be an early build target that just makes sure all library versions are adequate?

ffrank:
No dice. Just started over with a clean `$GOPATH` and want to `go get` all from the README. Results:

```
Building: mgmt.static, version: 0.0.3-61-ga1ed034...
go generate
stringer: checking package: etcd.go:62:7: could not import github.com/coreos/etcd/clientv3 (can't find import: github.com/coreos/etcd/clientv3)
```

After `go get`ting that, same with `etcdserver`, now `cockroachdb cmux`.

Running `go1.5.4` from `gimme`.

ffrank:
FWIW I'm reproducing this with a fresh clone of `mgmt` from github.

I don't really understand why it's failing. This line is in `github.com/coreos/etcd/clientv3/op.go`:

    type OpOption func(*Op)

So that's a declaration, no?

purpleidea:
@ffrank Still don't know why you have this issue, but git master now points to a commit near etcd git master. LMK if you still have issues. Closing this.

ffrank:
`make` gave me this today:

```
stringer: checking package: etcd.go:63:2: could not import github.com/coreos/etcd/embed (can't find import: github.com/coreos/etcd/embed)
```

This was solved by `go get`ting that package, which put the corresponding `.a` file in `$GOPATH/pkg`.

purpleidea:
@ffrank The problem is that you're not *recursively* cloning with git. That is provided by the git repo as a submodule. Either that or you're not using GO15VENDOREXPERIMENT=1 in your environment.

HTH

PR #57, opened 2016-08-30
Change "uname -i" to "uname -m" to be portable

joejulian:
According to the documentation for uname:

       -m, --machine
              print the machine hardware name

       -i, --hardware-platform
              print the hardware platform (non-portable)

So use the portable -m version.


purpleidea:
Sweet, merged thanks!

First patch fireworks!! <fireworks>

May there be many more... You know the gluster resource is calling to
you... ;)

PR #62, opened 2016-09-07
add unit tests for UUID comparison and resource event passing

ffrank:
This is basically a showcase for the kind of tests I'd like to start adding.

By the way, I've been wondering about `IFF` - should it not consider the kind/name combination, rather than handling pkg/ntp as equal to service/ntp?

### DIFF
--- a/resources_test.go
+++ b/resources_test.go
@@ -103,3 +103,71 @@ func TestMiscEncodeDecode2(t *testing.T) {
 		t.Error("The input and output Res values do not match!")
 	}
 }
+
+func TestIFF(t *testing.T) {
+	uuid := &BaseUUID{name: "/tmp/unit-test"}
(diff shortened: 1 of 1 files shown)


purpleidea:
@ffrank zomg, a patch with just tests. <3

Two small issues: s/t.Error/t.Errorf/ but I think the rest is amazing.

Nitpick: if you feel like capitalizing the relevant sentences so they appear like normal grammar, that's a bonus :)

purpleidea:
@ffrank Right, btw while this is great, in general the Base resource isn't used directly, which is why the combination tests you mention aren't really necessary. We typically only compare red apples to green apples, never apples to flamethrowers. (Where an apple == file resource, and flamethrower == virt resource).

Good point though!

purpleidea:
w00t. Merged! Thanks again!

PR #63, opened 2016-09-12
add puppet support documentation

ffrank:


purpleidea:
A perfect patch. Thank you merged!

PS: Good idea splitting out the docs into sub files. I think it was a good idea.

ffrank:
Me too :)

Not sure how well this works for the PDF export, I haven't tried.

Thanks for merging ❤️

purpleidea:
> Not sure how well this works for the PDF export, I haven't tried.

Good point. We can leave that as a problem for future James and future Felix.

https://www.youtube.com/watch?v=3HRrm1Hg7pw

PR #67, opened 2016-09-19
Let `env` find bash in shebang.

mikefaille:
On Nixos and GNUIX-SD, bash is chroot in package store path.
In my case : /nix/store/qvccmr6fsis4kqlvlk8pb1c8c0r0cwai-system-path/bin/bash

In any case, using `/usr/bin/env bash` is the recommended way to get bash
portable across UNIX-like systems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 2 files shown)


purpleidea:
LGTM, merged thanks!
https://github.com/purpleidea/mgmt/commit/22b0b8994940da91859067e80b35c2b28873dd0d
(I hope you don't mind that I rebased it, it was trivial.)

PR #68, opened 2016-09-19
travis : use go1.7 to be future proof.

mikefaille:
And, to reflect my own config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ go:
   - 1.4.3
   - 1.5.3
   - 1.6
+  - 1.7
   - tip
 sudo: false
 before_install: 'git fetch --unshallow'
@@ -14,6 +15,7 @@ matrix:
     - go: tip
     - go: 1.4.3
     - go: 1.6
+    - go: 1.7
 notifications:
   irc:
     channels:


purpleidea:
LGTM, merged thanks!

https://github.com/purpleidea/mgmt/commit/9715146495ccf27026a573dbe56ecf535fb1d332

(I hope you don't mind that I rebased it, it was trivial.)

mikefaille:
Cool. 

> (I hope you don't mind that I rebased it, it was trivial.)

It's ok. As as see, I'm still the author and you're the committer. My own reference : http://stackoverflow.com/questions/18750808/difference-between-author-and-committer-in-git

purpleidea:
> It's ok. As as see, I'm still the author and you're the committer

Yep, I didn't change any of that, here's what happened:

1) You created three simultaneous patch (probably in feature branches lets say) that all pointed to my git master. That's all fine.

2) The moment I merged any one of those (via fast-forward) now git master is pointing to your merged patch, and the other two of your patches is not ideally pointing to git master anymore.

There are two solutions:
a) I could have added a merge commit for each patch -- something I think is messy and unnecessary for small patches like this.
b) You could have rebase and resubmitted.

Since (b) is silly for a one-liner that will easily still apply, I just did this part and merged it. GH doesn't detect these patches as being merged because technically you never sent in a patch with that identical sha1.

HTH

purpleidea:
... The point being that rebasing doesn't necessarily mean I change the author. :)

PR #69, opened 2016-09-19
Add pcap.h dependency on Debian/Ubuntu

BinaryTiger:
```
# github.com/akrennmair/gopcap
../../akrennmair/gopcap/pcap.go:12:18: fatal error: pcap.h: No such file or directory
compilation terminated.
# github.com/purpleidea/mgmt/vendor/github.com/coreos/etcd/cmd/vendor/github.com/akrennmair/gopcap
vendor/github.com/coreos/etcd/cmd/vendor/github.com/akrennmair/gopcap/pcap.go:12:18: fatal error: pcap.h: No such file or directory
compilation terminated.
```
`make deps` returns this error on Ubuntu 16.04. The added dependency solve this issue.

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -29,6 +29,7 @@ if [ $travis -eq 0 ]; then
 		# one of these two golang tools packages should work on debian
 		$sudo_command $APT install -y golang-golang-x-tools || true
 		$sudo_command $APT install -y golang-go.tools || true
+		$sudo_command $APT install -y libpcap0.8-dev || true
 	fi
 fi
 


purpleidea:
@BinaryTiger Thanks for the patch! it LGTM, but could you:

1) Add the  || true at the end, and
2) Rebase into a single commit? If you don't know how to do this, that's fine, LMK and I will explain.

Cheers!

BinaryTiger:
Fix and squashed (if it's what you meant by rebase) into a single commit.

purpleidea:
@BinaryTiger 

Hey, so the patch is now correct, but I figured I'm going to nitpick this one a little bit because it's your first. In the commit message it says:

> Fix my whitespace style
> Add true statement to libpcap dep

From the POV of someone looking at this patch, they don't know about that history, so you don't (and shouldn't) need to mention it. Rebasing made it as though it doesn't exist anymore!

So, please update and resubmit once again. You can use git commit --amend to fix it :) And the push with --force.

purpleidea:
LGTM, merged! Now let's get a more core patch in :)

Thanks!

PR #77, opened 2016-09-30
Virt Resource WIP

nseps:
This is still WIP but it already does the essentials. It can be merged on a different branch if needed.

There is an example called virt_test.yaml It's suppose to:
* Install libvirt-bin & qemu-kvm
* Create directory structure
* Download and verify ubuntu xenial cloud image
* Create VM disc
* Create cloud init iso
* Create a VM that comes with your mgmt dir mounted on /mnt :)

You can always copy-paste the commands and use the virt_disk.yaml example that does only the last step.

The other examples are for testing the running & transient state of vms. You can use virsh destroy/undefine etc to see mgmt keeping the state.

Please comment & review if you can :)

### DIFF
--- a/resources/virt.go
+++ b/resources/virt.go
@@ -1,0 +1,521 @@
+package resources
+
+import (
+	"encoding/gob"
+	"log"
+	"fmt"
+	"strconv"
+
+	"github.com/purpleidea/mgmt/event"
+	"github.com/rgbkrk/libvirt-go"
+	"math/rand"
+	"time"
+	"bytes"
+)
+
+func init() {
+	gob.Register(&VirtRes{})
+}
+
+type VirtRes struct {
+	BaseRes `yaml:",inline"`
+	Cpus int `yaml:"cpus"`
+	Ram int `yaml:"ram"`
+	State string `yaml:"state"` //running, paused, shutoff
+	Transient bool `yaml:"transient"`
+	Boot []string `yaml:"boot"` // boot order. values: fd, hd, cdrom, network
+	Disk []diskDevice `yaml:"disk"`
+	CDRom []cdRomDevice `yaml:"cdrom"`
+	Network []networkDevice `yaml:"network"`
+	Filesystem []filesystemDevice `yaml:"filesystem"`
+	conn libvirt.VirConnection
+}
+
+type VirtUUID struct {
+	BaseUUID
+}
+
+type virtDevice interface {
+	GetXML(idx int) string
+}
+
+type diskDevice struct {
+	Source string `yaml:"source"`
+	Type string `yaml:"type"`
+}
+
+type cdRomDevice struct {
+	Source string `yaml:"source"`
+	Type string `yaml:"type"`
+}
+
+type networkDevice struct {
+	Name string `yaml:"name"`
+	MAC string `yaml:"mac"`
+}
+
+type filesystemDevice struct {
+	Access string `yaml:"access"`
+	Source string `yaml:"source"`
+	Target string `yaml:"target"`
+	ReadOnly bool `yaml:"read_only"`
+}
+
+var (
+	libvirtInitialized bool = false
+)
+
+func NewVirtRes(name string, cpus, ram int, state string, transient bool) *VirtRes {
+	obj := &VirtRes{
+		BaseRes: BaseRes{
+			Name: name,
+		},
+		Cpus: cpus,
+		Ram: ram,
+		State: state,
+		Transient: transient,
+	}
+	obj.Init()
+	return obj
+}
+
+func (obj *VirtRes) Init() error {
+	obj.BaseRes.kind = "Virt"
+
+	if !libvirtInitialized {
+		libvirt.EventRegisterDefaultImpl()
+		// run libvirt event loop
+		go func() {
+			for {
+				libvirt.EventRunDefaultImpl()
+			}
+		}()
+		libvirtInitialized = true
+	}
+
+	conn, err := libvirt.NewVirConnection("qemu:///system")
+	if err != nil {
+		// should we proceed or exit?
+		fmt.Println("Connection to libvirt faild")
+		return err
+	}
+	obj.conn = conn
+
+	obj.BaseRes.Init() // call base init, b/c we're overriding
+
+	return nil
+}
+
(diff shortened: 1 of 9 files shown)


resources/virt.go
```diff
@@ -0,0 +1,521 @@
+package resources
+
+import (
```
purpleidea: Imports should be split into three chunks (with a newline between each) first chunk normal imports, 2nd internal purpleidea/mgmt/ imports, third chunk external imports like the golang virt lib. Eg: https://github.com/purpleidea/mgmt/blob/master/resources/svc.go#L22

purpleidea: Imports should be split into three chunks (with a newline between each) first chunk normal imports, 2nd internal purpleidea/mgmt/ imports, third chunk external imports like the golang virt lib. Eg: https://github.com/purpleidea/mgmt/blob/master/resources/svc.go#L22

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+)
+
+func init() {
+	gob.Register(&VirtRes{})
+}
+
+type VirtRes struct {
+	BaseRes `yaml:",inline"`
+	Cpus int `yaml:"cpus"`
+	Ram int `yaml:"ram"`
```
purpleidea: Please add a comment here at the end to specify the units. Shouldn't it also be uint?

purpleidea: Please add a comment here at the end to specify the units. Shouldn't it also be uint? Same for CPU's, shouldn't it be unit?

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+	Network []networkDevice `yaml:"network"`
+	Filesystem []filesystemDevice `yaml:"filesystem"`
+	conn libvirt.VirConnection
+}
+
+type VirtUUID struct {
+	BaseUUID
+}
+
+type virtDevice interface {
```
purpleidea: What is virtDevice used for?

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+	return obj
+}
+
+func (obj *VirtRes) Init() error {
+	obj.BaseRes.kind = "Virt"
+
+	if !libvirtInitialized {
+		libvirt.EventRegisterDefaultImpl()
+		// run libvirt event loop
+		go func() {
```
purpleidea: How does this work? Do we ever have to shut this loop down somehow? If so, and if this requires an API change to the resource API LMK how we can help. Getting your feedback on if we designed this correctly is very valuable to us!

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+				libvirt.EventRunDefaultImpl()
+			}
+		}()
+		libvirtInitialized = true
+	}
+
+	conn, err := libvirt.NewVirConnection("qemu:///system")
+	if err != nil {
+		// should we proceed or exit?
+		fmt.Println("Connection to libvirt faild")
```
purpleidea: You can just return fmt.Errorf("Libvirt connection failed: %s", err) instead.

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+		return err
+	}
+	obj.conn = conn
+
+	obj.BaseRes.Init() // call base init, b/c we're overriding
+
+	return nil
+}
+
+func (obj *VirtRes) Validate() bool {
```
purpleidea: Should be Validate() error now. (Sorry, we changed the API, but didn't add it to the interface yet.)

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+func (obj *VirtRes) Watch(processChan chan event.Event) error {
+	if obj.IsWatching() {
+		return nil
+	}
+	obj.SetWatching(true)
+	defer obj.SetWatching(false)
+	cuuid := obj.converger.Register()
+	defer cuuid.Unregister()
+
+	ch := make(chan int, 10)
```
purpleidea: Does this need to be buffered for things to work? If we aren't ready to receive the event because we're still processing the last one, and there's no buffer for it does it drop? Using an unbuffered channel is probably better if we can do it. If not, please add a comment explaining why it's needed.

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+			processChan <- event.Event{event.EventNil, resp, "", true} // trigger process
+			resp.ACKWait()                                 // wait for the ACK()
+		}
+	}
+}
+
+func (obj *VirtRes) CheckApply(apply bool) (bool, error) {
+	log.Printf("%v[%v]: CheckApply(%t)", obj.Kind(), obj.GetName(), apply)
+
+	if obj.isStateOK { // cache the state
```
purpleidea: This isn't really being used ATM, but probably should be. Basically a watch event invalidates the cached event state, but a successful CheckApply should `true` it.

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+		return true, nil
+	}
+
+	dom, err := obj.conn.LookupDomainByName(obj.BaseRes.Name)
```
purpleidea: obj.GetName() is sufficient.

purpleidea: if err != nil ... ?

resources/virt.go
```diff
@@ -0,0 +1,521 @@
...
+		// Check for valid state
+		domState := domInfo.GetState()
+		switch obj.State {
+		case "running":
+			if domState == libvirt.VIR_DOMAIN_RUNNING {
+				break
+			}
+			if domState == libvirt.VIR_DOMAIN_BLOCKED {
+				// TODO: what should happen?
+				return false, fmt.Errorf("Domain %s blocked on resources", obj.BaseRes.Name)
```
purpleidea: obj.GetName()

purpleidea:
@nseps Hey, are you planning on finishing this off, or should any interested parties have a go at it? Just LMK your intent. Thanks

purpleidea:
@nseps Okay, so as discussed, I hacked on this a bit and merged https://github.com/purpleidea/mgmt/commit/8381d8246a522b46bf297aa915dfe3889870a9e6

I had to do a good amount of changes, but I think it's a solid base now. Thanks for your initial push. Hope to have you back and hacking. (See the commit message.)

Cheers!

PR #81, opened 2016-10-13
More info for installation for mgmt

narbutas:
add a bit more detail documentation how to install mgmt for people who is new to GOlang

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -170,7 +170,8 @@ which need to exchange information that is only available at run time.
 
 ####Blog post
 
-An introductory blog post about this topic will follow soon.
+You can read the introductory blog post about this topic here:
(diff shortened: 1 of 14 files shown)


purpleidea:
@narbutas Sorry for the delay on the review. As you know I was busy with speaking and travel. Okay:

1) This needs to be rebased into a single commit. If you need help doing this, LMK and I can guide you.
2) I think for the GOPATH stuff it should be setup permanently otherwise it will stop working when they open a new shell. Perhaps mention the lines they need to add into their ~/.bashrc or be more general with what they need to do?
3) I think the go get approach is probably better for installation. It's currently not specified with this patch how to get the code.

If you aren't in the mood to fix it up, no worries, just LMK and I'll merge something.

Thanks again!

narbutas:
Hey,

Yeah I know that, no problem. How was your presentation during conference?
Lot of people got interested?

Sure, no problem I will do that, just most likely not today, most likely
tomorrow :)

PR #83, opened 2016-10-19
Use the download-only flag when fetching dependencies

joejulian:


purpleidea:
@joejulian Okay, merged, but next time could you add a "misc: " prefix to your commit message? Thanks!

PS: What happened or broke that you needed this fix?

PR #86, opened 2016-10-24
virt: don't set emulator path

### LINKED ISSUE #85: virt emulator path is distro specific and unnecessary
In the virt resource the emulator path is specified implicitly which is incorrect and unnecessary for at least Fedora, Ubuntu, CentOS, and Arch (all that I had handy to check). 

For example, if the domain type is kvm, libvirt will use its configured emulator for that type:

From capabilities (on arch):
```
      <domain type='qemu'/>
      <domain type='kvm'>
        <emulator>/usr/sbin/qemu-system-x86_64</emulator>
      </domain>
```
so as long as the `domain type` is set to `kvm`, it will already know to use `/usr/sbin/qemu-system-x86_64`.

joejulian:
Remove the implicit emulator path from the domain definition. Libvirt is
already configured to use the correct emulator for kvm or qemu and
specifying it creates distro dependence.

Fixes #85

### DIFF
--- a/resources/virt.go
+++ b/resources/virt.go
@@ -525,9 +525,6 @@ func (obj *VirtRes) getDomainXML() string {
 	b += fmt.Sprintf("</os>")
 
 	b += fmt.Sprintf("<devices>") // start devices
-	// TODO: use capabilities to determine emulator
-	//b += "<emulator>/usr/bin/kvm-spice</emulator>" // TODO: ?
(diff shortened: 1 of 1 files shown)


purpleidea:
@joejulian Is there any reason we'd ever want to specify this for other domain types?

joejulian:
None that I can find. If one shows up later, the logic could be added then to provide it.

purpleidea:
Merged thanks!

PR #90, opened 2016-11-04
Virt allow other than kvm

vinzenz:
/cc @joejulian 

Collaborative improvements on the Virt resource

### DIFF
--- a/resources/virt.go
+++ b/resources/virt.go
@@ -359,7 +359,7 @@ func (obj *VirtRes) CheckApply(apply bool) (bool, error) {
 	dom, err := obj.conn.LookupDomainByName(obj.GetName())
 	if err == nil {
 		// pass
-	} else if virErr, ok := err.(libvirt.VirError); ok && virErr.Domain == libvirt.VIR_FROM_QEMU && virErr.Code == libvirt.VIR_ERR_NO_DOMAIN {
+	} else if virErr, ok := err.(libvirt.VirError); ok && virErr.Code == libvirt.VIR_ERR_NO_DOMAIN {
 		// domain not found
 		if obj.absent {
 			obj.isStateOK = true
--- a/resources/virt.go
+++ b/resources/virt.go
@@ -47,6 +47,12 @@ const (
 	lxcURI
 )
 
+// VirtAuth is used to pass credentials to libvirt.
+type VirtAuth struct {
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+}
+
 // VirtRes is a libvirt resource. A transient virt resource, which has its state
 // set to `shutoff` is one which does not exist. The parallel equivalent is a
 // file resource which removes a particular path.
@@ -63,6 +69,7 @@ type VirtRes struct {
 	CDRom      []cdRomDevice      `yaml:"cdrom"`
 	Network    []networkDevice    `yaml:"network"`
 	Filesystem []filesystemDevice `yaml:"filesystem"`
+	Auth       *VirtAuth          `yaml:"auth"`
 
 	conn      libvirt.VirConnection
 	absent    bool // cached state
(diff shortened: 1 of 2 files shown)


resources/virt.go
```diff
@@ -82,6 +99,14 @@ func (obj *VirtRes) Init() error {
 		}
 		libvirtInitialized = true
 	}
+	if u, err := url.Parse(obj.URI); err != nil {
+		log.Printf("%s[%s]: Parsing URI failed: %s", obj.Kind(), obj.GetName(), obj.URI)
```
purpleidea: This should return fmt.Errorf, no?

resources/virt.go
```diff
@@ -82,6 +99,14 @@ func (obj *VirtRes) Init() error {
...
+	} else {
```
purpleidea: Inline after the return

resources/virt.go
```diff
@@ -49,18 +62,21 @@ type VirtRes struct {
 	Transient  bool               `yaml:"transient"` // defined (false) or undefined (true)
 	CPUs       uint16             `yaml:"cpus"`
 	Memory     uint64             `yaml:"memory"` // in KBytes
+	OSInit     string             `yaml:"osinit"` // init used by lxc
```
purpleidea: Do you think there's any way to have this be a more general property that other domain types can use too? IOW, is there some equivalent the kvm and friends might use?

This way we can try to avoid having domain specific parameters as much as possible.

purpleidea:
PS: for some reason GH didn't send me an email notification. wazzup with that?

resources/virt.go
vinzenz: Not really since I don't need u other than for this and u is only visible within the if else portion

resources/virt.go
purpleidea: Interesting point. Is there any situation though where it would fail, but still be a valid virt connection string? If so, then I agree with you, if not we're generating a failure message which might be confusing.

resources/virt.go
joejulian: The only other way that comes to mind would be a "params" map that could take arbitrary dicts.

resources/virt.go
joejulian: :+1:

resources/virt.go
purpleidea: @joejulian If @vinzenz has no ideas, we can leave this for now. API is unstable anyways.

purpleidea:
AFAICT two issues to fix:

1) golint errors
2) URI thing to fmt.Errorf

Do we have consensus?

vinzenz:
Ack will apply the fix as soon as I have some time.

We will find some improvements over time

purpleidea:
@vinzenz Nice work. If you can rebase it in, it would be appreciated.

Enjoy the weekend.

resources/virt.go
vinzenz: Well OSInit needs to be replaced anyway, since the parameters need to be passed separately (or we would have to parse them)

I think this is too extensive for this first attempt, I wouldn't call the virt resource finished though

resources/virt.go
```diff
@@ -82,6 +100,15 @@ func (obj *VirtRes) Init() error {
 		}
 		libvirtInitialized = true
 	}
+	var u *url.URL
+	var err error
+	if u, err = url.Parse(obj.URI); err != nil {
+		return errwrap.Wrapf(err, "%s[%s]: Parsing URI failed: %s", obj.Kind(), obj.GetName(), obj.URI)
+	}
```
vinzenz: @purpleidea Changed here as requested

resources/virt.go
```diff
@@ -39,6 +40,19 @@ var (
...
 )
 
+type virtURISchemeType int
+
+const (
+	defaultURI virtURISchemeType = iota
+	lxcURI
+)
+
+// VirtAuth is used to pass credentials to libvirt.
```
vinzenz: @purpleidea Made golint happy here

resources/virt.go
joejulian: Ah, right, good point. I had only used commands without additional parameters in my tests.

This will work as is. There are simple workarounds that can be done to avoid passing parameters so this will function as-is with room for future improvement.

purpleidea:
I'm super pleased to announce that this is now in git master!

Thanks to both of you: @joejulian @vinzenz ! w00t. Looking forward to more...

PR #91, opened 2016-11-09
nspawn: Add systemd-machined support for nspawn containers

joejulian:
This adds a rudimentary resource for systemd-machined's nspawn
containers, ensuring they're either started or stopped.


examples/nspawn2.yaml
```diff
@@ -0,0 +1,7 @@
+---
+graph: mygraph
+resources:
+  nspawn:
+  - name: mgmt-nspawn1
```
purpleidea: Did you mean nspawn2?

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+		break
+	default:
+		return fmt.Errorf("Invalid State: %s", obj.State)
+	}
+	return nil
+}
+
+// Watch for state changes and sends a message to the bus if there is a change.
+func (obj *NspawnRes) Watch(processChan chan event.Event) error {
+	// Boilerplate
```
purpleidea: Agreed, but please remove the boilerplate comments for now. We know.

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	Startup := func(block bool) <-chan time.Time {
+		if block {
+			return nil // blocks forever
+		}
+		return time.After(time.Duration(500) * time.Millisecond) // 1/2 the resolution of converged timeout
+	}
+	// End Boilerplate
+
+	// This resource depends on systemd. Ensure that it's running.
+	// TODO: If systemd is not running, should we start it?
```
purpleidea: Remove this TODO. The answer is no, the code is correct.
You can also turn the above comment into a lower cased non-sentence if you'd prefer.

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	if !systemdUtil.IsRunningSystemd() {
+		return fmt.Errorf("Systemd is not running.")
+	}
+
+	// Create a private message bus
```
purpleidea: Nit pick: one line comments aren't usually sentences, they start with lower case strings, elsewhere too

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// Add a match rule to match messages going through the message bus.
+	bus.BusObject().Call("org.freedesktop.DBus.AddMatch", 0,
+		"type='signal',interface='org.freedesktop.machine1.Manager',eavesdrop='true'")
+	buschan := make(chan *dbus.Signal, 10)
+	bus.Signal(buschan)
+
+	var send = false
+	var exit = false
+	// var dirty = false
```
purpleidea: If unused, remove these please

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	// var invalid = false
+	// var activeSet = false
+	// var previous bool
+
+	for {
+		obj.SetState(ResStateWatching)
+		select {
+		// If this resource has been idle for long enough, set converged to allow
```
purpleidea: If you wish to keep this comment, please wrap it at 80 chars because it fits nicely that way.

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+			// only do this on certain types of events
+			//obj.isStateOK = false // something made state dirty
+			if exit, err := obj.DoSend(processChan, ""); exit || err != nil {
+				return err // we exit or bubble up a NACK...
+			}
+		}
+	}
+}
+
+// CheckApply is run to check the state and, if apply is true, to apply the necessary changes to reach the desired
```
purpleidea: Please wrap to 80 chars in this case.

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+// state. This is run before Watch and again if watch finds a change occurring to the state.
+func (obj *NspawnRes) CheckApply(apply bool) (checkok bool, err error) {
+	log.Printf("%v[%v]: CheckApply(%t)", obj.Kind(), obj.GetName(), apply)
+
+	if obj.isStateOK { // cache the state
+		return true, nil
+	}
+
+	// This resource depends on systemd. Ensure that it's running.
+	// TODO: If systemd is not running, should we start it?
```
purpleidea: Remove these TODO's

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	// TODO: We do this in two different places. Should we do it in Init?
+	if !systemdUtil.IsRunningSystemd() {
+		return false, errors.New("Systemd is not running.")
+	}
+
+	// Connect to org.freedesktop.machine1.Manager
```
purpleidea: lowercase Connect

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	if err != nil {
+		return false, fmt.Errorf("Failed to connect to dbus: %s", err)
+	}
+
+	// Compare the current state with the desired state and perform the
+	// appropriate action
+	properties, err := conn.GetProperties(obj.GetName())
+	if err != nil {
+		if err.(dbus.Error).Name != "org.freedesktop.machine1.NoSuchMachine" {
+			log.Printf("GetProperties error: %s", err.Error())
```
purpleidea: Return errwrap.Wrapf(err, "the msg...") instead and don't log.

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if err != nil {
+		return false, fmt.Errorf("Failed to connect to dbus: %s", err)
+	}
+
+	// Compare the current state with the desired state and perform the
+	// appropriate action
+	properties, err := conn.GetProperties(obj.GetName())
+	if err != nil {
+		if err.(dbus.Error).Name != "org.freedesktop.machine1.NoSuchMachine" {
```
purpleidea: You must do err, ok := err.(dbus.Error) to check if it's the right type first.

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+		// If we could not successfully get the properties because there's no such machine, the machine
+		// is stopped.
+		if obj.State == stopped {
+			properties = make(map[string]interface{}, 1)
+			properties["State"] = stopped
+		}
+		// error if we need the image, ignore if we don't
+		_, err = conn.GetImage(obj.GetName())
+		if err != nil && obj.State != stopped {
+			return false, fmt.Errorf("No machine nor image named '%s'", obj.GetName())
```
purpleidea: errwrap please

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	properties, err := conn.GetProperties(obj.GetName())
+	if err != nil {
+		if err.(dbus.Error).Name != "org.freedesktop.machine1.NoSuchMachine" {
+			log.Printf("GetProperties error: %s", err.Error())
+			return false, err
+		}
+		// If we could not successfully get the properties because there's no such machine, the machine
+		// is stopped.
+		if obj.State == stopped {
+			properties = make(map[string]interface{}, 1)
```
purpleidea: We probably don't want to be doing this. What was the goal?

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+			if exit, err := obj.DoSend(processChan, ""); exit || err != nil {
+				return err // we exit or bubble up a NACK...
+			}
+		}
+	}
+}
+
+// CheckApply is run to check the state and, if apply is true, to apply the necessary changes to reach the desired
+// state. This is run before Watch and again if watch finds a change occurring to the state.
+func (obj *NspawnRes) CheckApply(apply bool) (checkok bool, err error) {
```
purpleidea: We don't seem to look at the apply variable anywhere... It's required, see other resources for how it's used.

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	// If the state matches, we're done.
+	if properties["State"] == obj.State {
+		return true, nil
+	}
+
+	if obj.State == running {
+		// Start the machine using svc resource
+		var serviceName = fmt.Sprintf(nspawnServiceTemplate, obj.GetName())
+		log.Printf("Start service: %s", serviceName)
+		service, err := NewSvcRes(serviceName, "running", "")
```
purpleidea: This should be switched to use SvcRes{ } and Init() instead... It should probably be a private field in the Nspawn struct with it's methods mapped through to all the equivalents here. (eg: Init in init, Watch in Watch, and so on.)

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+		if err != nil {
+			return false, err
+		}
+		return service.CheckApply(true)
```
purpleidea: Mapping this through is good, but you should pass the apply variable...

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+func (obj *NspawnUID) IFF(uid ResUID) bool {
+	res, ok := uid.(*NspawnUID)
+	if !ok {
+		return false
+	}
+	return obj.name == res.name
+}
+
+// NspawnResAutoEdges holds the state of the auto edge generator.
+type NspawnResAutoEdges struct {
```
purpleidea: All this autoedges code isn't being used, so remove it. (Should probably look how it looks in noop.go)

resources/nspawn.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	// TODO: depending on if the systemd service api allows batching, we
+	// might be able to build this, although not sure how useful it is...
+	// it might just eliminate parallelism be bunching up the graph
+	return false // not possible atm
+}
+
+// Compare two resources and return if they are equivalent.
+func (obj *NspawnRes) Compare(res Res) bool {
+	return true
```
purpleidea: Needs to be implemented... See noop.go for example

examples/nspawn2.yaml
joejulian: Not necessarily. I used nspawn1.yaml to start it, then nspawn2.yaml to stop it. If you want it to be 2, though, it doesn't matter to me.

resources/nspawn.go
joejulian: Ah, yes, good point.

resources/nspawn.go
joejulian: When I removed it:
```
*NspawnRes does not implement Res (missing AutoEdges method)
```

purpleidea:
@joejulian When I mentioned:

```
it's methods mapped through to all the equivalents here. (eg: Init in init, Watch in Watch, and so on.)
```

What I meant is that the Svc resource you're embedding, need to have it's Init() happen in the Nspawn Init(). As a result, the svc resource probably gets a private field inside of the Nspawn struct. Similarly, the Svc watch should be plumbed up into the Nspawn watch loop, shouldn't it? Otherwise how do we know when the Svc changes state? It seems you're getting that via an nspawn match, so why not reuse the svc if that's how this should work?

Lastly, remember the `apply` variable in the CheckApply function specified whether to actually do things, or to only check. So if false, CheckApply should never change the state, only report on if it's correct or not.

Cheers!

joejulian:
> What I meant is that the Svc resource you're embedding, need to have it's Init() happen in the Nspawn Init(). As a result, the svc resource probably gets a private field inside of the Nspawn struct.

Oh, duh

> Similarly, the Svc watch should be plumbed up into the Nspawn watch loop, shouldn't it? Otherwise how do we know when the Svc changes state? It seems you're getting that via an nspawn match, so why not reuse the svc if that's how this should work?

We don't have to care about watching the svc state because the dbus messages for machine1 do that.

> Lastly, remember the apply variable in the CheckApply function specified whether to actually do things, or to only check. So if false, CheckApply should never change the state, only report on if it's correct or not.

Right, fixed that already.

resources/nspawn.go
joejulian: removed capitalization and punctuation from all remarks this leaves them looking utterly unreadable it's just like trying to read the text messages of a teenager

resources/nspawn.go
joejulian: done

resources/nspawn.go
joejulian: Thanks, checked or passed for the subordinate resource to check.

resources/nspawn.go
joejulian: done

resources/nspawn.go
joejulian: :point_up:

resources/nspawn.go
joejulian: gone

resources/nspawn.go
joejulian: But I don't care what the err from GetImage was. This is a compound error representing the results of both GetMachine and GetImage so it is purposefully not returning the err from either.

resources/nspawn.go
joejulian: done

resources/nspawn.go
joejulian: done

examples/nspawn2.yaml
purpleidea: I suppose changing it to 2 would be better. That way during the demo, we can live edit the file to change the state, which is pretty cool since it will notice live.

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+		BaseRes: BaseRes{
+			Name: name,
+		},
+		State: state,
+	}
+	return obj, obj.Init()
+}
+
+// Validate params
+func (obj *NspawnRes) Validate() error {
```
purpleidea: Hook up obj.svc.Validate too please.

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+
+	// this resource depends on systemd ensure that it's running
+	if !systemdUtil.IsRunningSystemd() {
+		return fmt.Errorf("Systemd is not running.")
+	}
+
+	// create a private message bus
+	bus, err := util.SystemBusPrivateUsable()
+	if err != nil {
+		return fmt.Errorf("Failed to connect to bus: %s", err)
```
purpleidea: use errwrap please, eg:
```
return errwrap.Wrapf(err, "Failed to connect to bus")
```

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	// add a match rule to match messages going through the message bus
+	bus.BusObject().Call("org.freedesktop.DBus.AddMatch", 0, fmt.Sprintf(
```
purpleidea: Does this have a return error we should check?

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+		case <-Startup(startup):
+			cuid.SetConverged(false)
+			send = true
+
+		// process org.freedesktop.machine1 events for this resource's
+		// name
+		case event := <-buschan:
+			if event.Body[0] == obj.GetName() {
+				if event.Name == machineNew {
+					log.Printf("Machine Started: %#v",
```
purpleidea: This and the below log message can be on the same line if you want

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+			// we avoid sending events on unpause
+			if exit, send = obj.ReadEvent(&event); exit {
+				return nil // exit
+			}
+
+		case <-Startup(startup):
+			cuid.SetConverged(false)
+			send = true
+
+		// process org.freedesktop.machine1 events for this resource's
```
purpleidea: the 80 char limit is a suggestion, but if you can make it looks balanced (eg: have the text on one line, have on the other) than it's better keeping it as a hard requirement. I'd balance or have it go long. Sorry nit, I probably didn't explain this well on the first review.

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	// this resource depends on systemd ensure that it's running
+	if !systemdUtil.IsRunningSystemd() {
+		return false, errors.New("Systemd is not running.")
+	}
+
+	// connect to org.freedesktop.machine1.Manager
+	conn, err := machined.New()
+
```
purpleidea: Why this empty newline when New() is matched with the below error?

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil {
+		if err, ok := err.(dbus.Error); ok && err.Name !=
+			"org.freedesktop.machine1.NoSuchMachine" {
+			return false, err
+		}
+		exists = false
+		// if we could not successfully get the properties because
+		// there's no such machine the machine is stopped
+		// error if we need the image ignore if we don't
+		_, err = conn.GetImage(obj.GetName())
```
purpleidea: Does this have to be nested in here instead of inlined right after this block?

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+
+	// if the machine doesn't exist and is supposed to be stopped or
+	// the state matches we're done
+	if !exists && obj.State == stopped || properties["State"] == obj.State {
+		return true, nil
+	}
+
+	if obj.State == running {
+		// start the machine using svc resource
+		var serviceName = fmt.Sprintf(nspawnServiceTmpl,
```
purpleidea: single line is fine here for code...

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+			obj.GetName())
+		log.Printf("Start service: %s", serviceName)
+		obj.svc.Name = serviceName
```
purpleidea: I think that this stuff needs to be done in Init() otherwise the svc Init() could fail if it needed them. Similarly so Validate doesn't fail. (It's currently not wired up, but coming soon.)

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+func (obj *NspawnRes) Compare(res Res) bool {
+	switch res.(type) {
+	case *NspawnRes:
+		res := res.(*NspawnRes)
+		if !obj.BaseRes.Compare(res) {
+			return false
+		}
+		if obj.Name != res.Name {
+			return false
+		}
```
purpleidea: Because we're using a composite resource, I think we need to sub-call Compare on the obj.svc too. Sorry that none of this was documented anywhere, you're the first person writing a composite resource but we'll be able to look at you as a future model, so let's get it perfect right away! Hurrah :)

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+		obj.svc.Name = serviceName
+		obj.svc.State = running
+		return obj.svc.CheckApply(apply)
+
+	}
+	if obj.State == stopped {
+		// terminate the machine with
+		// org.freedesktop.machine1.Manager.KillMachine
+		log.Printf("Stopping nspawn machine: %s", obj.GetName())
+		if apply {
```
purpleidea: inline please, eg:

```
if !apply {
	return false, nil
}
```

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+			if err := conn.KillMachine(obj.GetName()); err != nil {
+				return false, err
+			}
+		}
+		return true, nil
```
purpleidea: What situation gets us here? This means "no error and state was as we expected and I made no changes".

resources/nspawn.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// GetUIDs includes all params to make a unique identification of this object
+// most resources only return one although some resources can return multiple
+func (obj *NspawnRes) GetUIDs() []ResUID {
+	x := &NspawnUID{
+		BaseUID: BaseUID{name: obj.GetName(), kind: obj.Kind()},
+		name:    obj.Name, // svc name
+	}
+	return []ResUID{x}
```
purpleidea: Here you need to append any of the GetUIDs of the svc resource. Again, sorry not mentioned previously, you're the pioneer!

joejulian:
No, completely different signal. We want to use this one for future development.

resources/nspawn.go
joejulian: Ok, but then this would need to be done in CheckState because we don't even have the service name nor what state we want at this point.

resources/nspawn.go
joejulian: done

resources/nspawn.go
joejulian: done

resources/nspawn.go
joejulian: Nope, that was a lack of go knowledge.

resources/nspawn.go
joejulian: Ok, so true means nothing changed and false means we changed something?

joejulian: For posterity, if the "check" portion of CheckApply shows nothing needs changed, checkok is true. If a state change is needed in the "apply" portion, checkok is false.

resources/nspawn.go
```diff
@@ -0,0 +1,308 @@
...
+type NspawnRes struct {
+	BaseRes `yaml:",inline"`
+	State   string `yaml:"state"`
+	svc     SvcRes
+}
+
+// Init runs some startup code for this resource
+func (obj *NspawnRes) Init() error {
+	var serviceName = fmt.Sprintf(nspawnServiceTmpl, obj.GetName())
+	log.Printf("Start service: %s", serviceName)
```
purpleidea: Remove this log msg please, it doesn't happen here anyways.

resources/nspawn.go
```diff
@@ -0,0 +1,308 @@
...
+
+	// end of state checking. if we're here, checkok is false
+	if !apply {
+		return false, nil
+	}
+
+	if obj.State == running {
+		// start the machine using svc resource
+		return obj.svc.CheckApply(apply)
+
```
purpleidea: extra newline here to kill plz

resources/nspawn.go
```diff
@@ -0,0 +1,308 @@
...
+		if !obj.BaseRes.Compare(res) {
+			return false
+		}
+		if obj.Name != res.Name {
+			return false
+		}
+	default:
+		return false
+	}
+	return obj.svc.Compare(&obj.svc)
```
purpleidea: This was a tricky one, but I think it should actually be:

```
if !obj.svc.Compare(res.svc) {
    return false
}
```
inside of the case `*NspawnRes:` where we've decoded res into an NspawnRes.

resources/nspawn.go
```diff
@@ -0,0 +1,308 @@
...
+	BaseUID
+	name string // the machine name
+}
+
+// IFF aka if and only if they are equivalent, return true. If not, false
+func (obj *NspawnUID) IFF(uid ResUID) bool {
+	res, ok := uid.(*NspawnUID)
+	if !ok {
+		return false
+	}
```
purpleidea: I apologize, but I forget to ask you to chain in the IFF of the composite Svc.
I think this one is:

```
if !obj.svc.IFF(res) {
    return false
}
```
(untested)

joejulian: There's no access to the NspawnRes object here, so no access to svc.

If I'm reading this right, https://github.com/purpleidea/mgmt/blob/master/resources/resources.go#L158-L166 , we don't need to nest it because the fact that GetUIDs returns both it's own and the one for the svc resource means that both will be checked with their own IFF calls.

joejulian:
This solves the race condition with mgmt trying to kill containers before they've started up enough to accept the signal and retries if systemd recognizes and prevents a deadlock.

Let me know if you hate the retry loop and would rather I just pass the retry meta. This loop only retries under specific conditions as opposed to retrying regardless of the failure cause.

PR #92, opened 2016-11-10
resources: Add new hostname resource

vinzenz:
This resource allows to set and watch the hostname on a system utilizing systemd-hostnamed

### DIFF
--- a/resources/hostname.go
+++ b/resources/hostname.go
@@ -1,0 +1,326 @@
+// Mgmt
+// Copyright (C) 2013-2016+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"encoding/gob"
+	"errors"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/purpleidea/mgmt/event"
+	"github.com/purpleidea/mgmt/util"
+
+	"github.com/godbus/dbus"
+	errwrap "github.com/pkg/errors"
+)
+
+// ErrResourceInsufficientParameters is returned when the configuration of the resource
+// is insufficient for the resource to do any useful work.
+var ErrResourceInsufficientParameters = errors.New(
+	"Insufficient parameters for this resource")
+
+func init() {
+	gob.Register(&HostnameRes{})
+}
+
+const (
+	hostname1Path  = "/org/freedesktop/hostname1"
+	hostname1Iface = "org.freedesktop.hostname1"
+	dbusAddMatch   = "org.freedesktop.DBus.AddMatch"
+)
+
+// HostnameRes is a resource that allows setting and watching the hostname.
+//
+// StaticHostname is the one configured in /etc/hostname or a similar file.
+// It is chosen by the local user. It is not always in sync with the current
+// host name as returned by the gethostname() system call.
+//
+// TransientHostname is the one configured via the kernel's sethostbyname().
+// It can be different from the static hostname in case DHCP or mDNS have been
+// configured to change the name based on network information.
+//
+// PrettyHostname is a free-form UTF8 host name for presentation to the user.
+//
+// Hostname is the fallback value for all 3 fields above, if only Hostname is
+// specified, it will set all 3 fields to this value.
+type HostnameRes struct {
+	BaseRes           `yaml:",inline"`
+	Hostname          string `yaml:"hostname"`
+	PrettyHostname    string `yaml:"pretty_hostname"`
+	StaticHostname    string `yaml:"static_hostname"`
+	TransientHostname string `yaml:"transient_hostname"`
+
+	conn *dbus.Conn
+}
+
+// NewHostnameRes is a constructor for this resource. It also calls Init() for you.
+func NewHostnameRes(name, staticHostname, transientHostname, prettyHostname string) (*HostnameRes, error) {
+	obj := &HostnameRes{
+		BaseRes: BaseRes{
+			Name: name,
+		},
+		PrettyHostname:    prettyHostname,
+		StaticHostname:    staticHostname,
+		TransientHostname: transientHostname,
+	}
+	return obj, obj.Init()
+}
+
+// Init runs some startup code for this resource.
+func (obj *HostnameRes) Init() error {
+	obj.BaseRes.kind = "Hostname"
+	if obj.PrettyHostname == "" {
+		obj.PrettyHostname = obj.Hostname
+	}
+	if obj.StaticHostname == "" {
+		obj.StaticHostname = obj.Hostname
+	}
+	if obj.TransientHostname == "" {
+		obj.TransientHostname = obj.Hostname
+	}
+	return obj.BaseRes.Init() // call base init, b/c we're overriding
+}
+
+// Validate if the params passed in are valid data.
+// FIXME: where should this get called ?
+func (obj *HostnameRes) Validate() error {
+	if obj.PrettyHostname == "" && obj.StaticHostname == "" && obj.TransientHostname == "" {
+		return ErrResourceInsufficientParameters
+	}
+	return nil
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *HostnameRes) Watch(processChan chan event.Event) error {
+	if obj.IsWatching() {
+		return nil // TODO: should this be an error?
+	}
+	obj.SetWatching(true)
+	defer obj.SetWatching(false)
+	cuid := obj.converger.Register()
+	defer cuid.Unregister()
+
+	var startup bool
+	Startup := func(block bool) <-chan time.Time {
+		if block {
+			return nil // blocks forever
+			//return make(chan time.Time) // blocks forever
+		}
+		return time.After(time.Duration(500) * time.Millisecond) // 1/2 the resolution of converged timeout
+	}
+
+	// if we share the bus with others, we will get each others messages!!
+	bus, err := util.SystemBusPrivateUsable() // don't share the bus connection!
+	if err != nil {
+		return errwrap.Wrap(err, "Failed to connect to bus")
+	}
+	defer bus.Close()
+	callResult := bus.BusObject().Call(
+		"org.freedesktop.DBus.AddMatch", 0,
+		fmt.Sprintf("type='signal',path='%s',interface='org.freedesktop.DBus.Properties',member='PropertiesChanged'", hostname1Path))
+	if callResult.Err != nil {
+		return errwrap.Wrap(callResult.Err, "Failed to subscribe to DBus events for hostname1")
+	}
+
+	signals := make(chan *dbus.Signal, 10) // closed by dbus package
+	bus.Signal(signals)
+
+	var send = false // send event?
+	var dirty = false
+
+	for {
+		obj.SetState(ResStateWatching) // reset
+		select {
+		case <-signals:
+			cuid.SetConverged(false)
+			send = true
+			dirty = true
+
+		case event := <-obj.Events():
+			cuid.SetConverged(false)
+			// we avoid sending events on unpause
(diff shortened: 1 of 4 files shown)


resources/hostname.go
```diff
@@ -0,0 +1,259 @@
...
+	signals := make(chan *dbus.Signal, 10)
+	bus.Signal(signals)
+
+	var send = false // send event?
+	var exit = false
+	var dirty = false
+	var closed = false
+
+	defer func() {
+		if !closed {
```
purpleidea: What's this all about?

resources/hostname.go
```diff
@@ -0,0 +1,259 @@
...
+			if exit, err := obj.DoSend(processChan, ""); exit || err != nil {
+				return err // we exit or bubble up a NACK...
+			}
+		}
+	}
+}
+
+func (obj *HostnameRes) updateProperty(object dbus.BusObject, property, setterName string) (bool, error) {
+	if hostnameProperty, err := object.GetProperty("org.freedesktop.hostname1." + property); err == nil {
+		switch hostname := hostnameProperty.Value().(type) {
```
purpleidea: Is there any chance Value() can be nil, thus causing a panic?

resources/hostname.go
```diff
@@ -0,0 +1,259 @@
...
+				log.Printf("Changing %s: %s => %s", property, hostname, obj.Hostname)
+				if err := object.Call("org.freedesktop.hostname1."+setterName, 0, obj.Hostname, false).Err; err != nil {
+					return false, errwrap.Wrapf(err, "failed to call org.freedesktop.hostname1.%s", setterName)
+				}
+				return true, nil
+			}
+		default:
+			return false, fmt.Errorf("Received unexpected type as %s value, expected string got '%t'", property, hostname)
+		}
+	} else {
```
purpleidea: Inline this above please. Eg:

resources/hostname.go
purpleidea: Instead of a switch like this, we can inline it all I think, eg:

IOW: if err != nil: return
if !string return
etc...

resources/hostname.go
```diff
@@ -0,0 +1,259 @@
...
+		return false, errwrap.Wrapf(err, "failed to get org.freedesktop.hostname1.%s", property)
+	}
+	return false, nil
+}
+
+// CheckApply method for Hostname resource. Does nothing, returns happy!
+func (obj *HostnameRes) CheckApply(apply bool) (checkok bool, err error) {
```
purpleidea: This implementation doesn't seem to respect the apply variable. This is input from the engine telling you whether we should just run the Check part (apply == false) or CheckApply (apply == true).

The bool value returned ("checkok") is true if the state was good and changes aren't needed for both scenarios. IOW, if we made changes successfully (aka apply==true) we'd return (false, nil)

resources/hostname.go
```diff
@@ -0,0 +1,259 @@
...
+	log.Printf("%v[%v]: CheckApply(%t)", obj.Kind(), obj.GetName(), apply)
+
+	conn, err := util.SystemBusPrivateUsable()
+	if err != nil {
+		return false, errwrap.Wrap(err, "Failed to connect to the private system bus")
+	}
+	defer conn.Close()
+
+	hostnameObject := conn.Object(hostname1Iface, hostname1Path)
```
purpleidea: What's going on in the following 20 lines? I find it a bit confusing to read, and as a result, I'm lost.

resources/hostname.go
vinzenz: I am not sure, it seems like that dbus can close the channel, and if so it sends a nil event object, and then I won't try to close the channel that is closed already

resources/hostname.go
vinzenz: Nil is a valid in dbus terms? besides that I expect the code to fail if not, but well you never know - So if the wind moves the right direction and at the right speed it might be nil ;) 
I will add a check, but I don't know if it actually can be nil

resources/hostname.go
vinzenz: well it sets the StaticHostname and Hostname fields on the hostname1 object
However it won't fail on one of them failing and reports both errors or one error. Otherwise the implementation itself is in the updateProperty function

I wasn't really sure how to handle that scenario, so I went for this way for now.
Anyway that's what the review is for ;)

purpleidea: What are the two fields for? StaticHostname vs Hostname ?
Maybe this deserves a comment?

resources/hostname.go
purpleidea: I don't actually know if it can be nil, but I figured I'd ask to see if you looked into it or not :)

resources/hostname.go
```diff
@@ -0,0 +1,259 @@
...
+		return time.After(time.Duration(500) * time.Millisecond) // 1/2 the resolution of converged timeout
+	}
+
+	// if we share the bus with others, we will get each others messages!!
+	bus, err := util.SystemBusPrivateUsable() // don't share the bus connection!
+	if err != nil {
+		return errwrap.Wrap(err, "Failed to connect to bus")
+	}
+	defer bus.Close()
+	bus.BusObject().Call("org.freedesktop.DBus.AddMatch", 0,
```
purpleidea: Does this return an error we should check?

resources/hostname.go
purpleidea: Whaa?

A close on a closed channel will panic. Did you see that happen?

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+		StaticHostname:    staticHostname,
+		TransientHostname: transientHostname,
+	}
+	return obj, obj.Init()
+}
+
+// Init runs some startup code for this resource.
+func (obj *HostnameRes) Init() error {
+	obj.BaseRes.kind = "Hostname"
+	if obj.Hostname != "" {
```
purpleidea: Is this first check really necessary?

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+		if obj.TransientHostname == "" {
+			obj.TransientHostname = obj.Hostname
+		}
+	}
+	return obj.BaseRes.Init() // call base init, b/c we're overriding
+}
+
+// Validate if the params passed in are valid data.
+// FIXME: where should this get called ?
+func (obj *HostnameRes) Validate() error {
```
purpleidea: Do we want to check that we've set at least one hostname?

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+	signals := make(chan *dbus.Signal, 10)
+	bus.Signal(signals)
+
+	var send = false // send event?
+	var dirty = false
+	var closed = false
+
+	// Ensure that we only close the channel when it wasn't closed already
+	defer func() {
+		if !closed {
```
purpleidea: What about this? Did we decide if it was necessary to check !closed ?

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+			send = true
+		}
+
+		// do all our event sending all together to avoid duplicate msgs
+		if send {
+			startup = true // startup finished
+			send = false
+			if dirty {
+				dirty = false
+				obj.isStateOK = false // something made state dirty
```
purpleidea: `isStateOK` is unused. Either remove, or add the check for this in CheckApply. See other resources for reference.

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+	if propertyValue == expectedValue {
+		return true, nil
+	}
+
+	// Values don't match but shouldn't be changed either => false
+	if !apply {
+		return false, nil
+	}
+
+	// Attempting to apply the changes
```
purpleidea: nitpick; single line comments are generally not complete sentences. (which is okay!) As a result they shouldn't start with a capital letter in most cases. (There are a few other instances of these to fix too.)

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+	log.Printf("Changing %s: %s => %s", property, propertyValue, expectedValue)
+	if err := object.Call("org.freedesktop.hostname1."+setterName, 0, expectedValue, false).Err; err != nil {
+		return false, errwrap.Wrapf(err, "failed to call org.freedesktop.hostname1.%s", setterName)
+	}
+
+	// All good changes should now be applied again
+	return true, nil
```
purpleidea: Since we did make changes (successfully) this should be `return false, nil`. This is assuming you're going for the same pattern as the CheckApply signature.

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+	defer conn.Close()
+
+	hostnameObject := conn.Object(hostname1Iface, hostname1Path)
+	checkok = true
+	if obj.PrettyHostname != "" {
+		updated, err := updateHostnameProperty(hostnameObject, obj.PrettyHostname, "PrettyHostname", "SetPrettyHostname", apply)
+		if err != nil {
+			return false, err
+		}
+		checkok = checkok && updated
```
purpleidea: Nicely done!

I used a slightly different pattern in file.go:

```
	checkOK = true

	if c, err := obj.contentCheckApply(apply); err != nil {
		return false, err
	} else if !c {
		checkOK = false
	}

	// TODO
	//if c, err := obj.chmodCheckApply(apply); err != nil {
	//	return false, err
	//} else if !c {
	//	checkOK = false
	//}

	// TODO
	//if c, err := obj.chownCheckApply(apply); err != nil {
	//	return false, err
	//} else if !c {
	//	checkOK = false
	//}

	// if we did work successfully, or are in a good state, then state is ok
	if apply || checkOK {
		obj.isStateOK = true
	}
	return checkOK, nil // w00t
```

but I think your version is nicer! I should probably fix mine. The only thing missing is you probably need to add:

```
	// if we did work successfully, or are in a good state, then state is ok
	if apply || checkOK {
		obj.isStateOK = true
	}
```
and I think the variable should probably be "checkOK"... I forget if golint complains or not.

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+	return []ResUID{x}
+}
+
+// GroupCmp returns whether two resources can be grouped together or not.
+func (obj *HostnameRes) GroupCmp(r Res) bool {
+	_, ok := r.(*HostnameRes)
+	if !ok {
+		return false
+	}
+	return true // Hostname resources can always be grouped together!
```
purpleidea: Hrmm... Interesting. Um, so technically no, but on the other hand, if you have more than one hostname resource this is a problem. This should be false I think. If you disagree, tell me why.

resources/hostname.go
vinzenz: probably not

resources/hostname.go
vinzenz: probably a good idea if nothing is set, there's nothing ever todo

resources/hostname.go
```diff
@@ -0,0 +1,326 @@
...
+	gob.Register(&HostnameRes{})
+}
+
+const (
+	hostname1Path  = "/org/freedesktop/hostname1"
+	hostname1Iface = "org.freedesktop.hostname1"
+	dbusAddMatch   = "org.freedesktop.DBus.AddMatch"
+)
+
+// HostnameRes is a resource that allows setting and watching the hostname
```
vinzenz: Done

resources/hostname.go
vinzenz: Done

resources/hostname.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/purpleidea/mgmt/event"
+	"github.com/purpleidea/mgmt/util"
+
+	"github.com/godbus/dbus"
+	errwrap "github.com/pkg/errors"
+)
+
+// ErrResourceInsufficientParameters is returned when the configuration of the resource
+// is insufficient for the resource to do any useful work.
+var ErrResourceInsufficientParameters = errwrap.New(
```
purpleidea: no need for errwrap. Just use fmt.Errorf or errors.New

resources/hostname.go
```diff
@@ -0,0 +1,321 @@
...
+	if !ok {
+		return false, fmt.Errorf("Received unexpected type as %s value, expected string got '%T'", property, propertyValue)
+	}
+
+	// expected value and actual value match => checkOk
+	if propertyValue == expectedValue {
+		return true, nil
+	}
+
+	// Nothing to do anymore
```
purpleidea: nit, lcase please

resources/hostname.go
```diff
@@ -0,0 +1,321 @@
...
+		checkOK = checkOK && propertyCheckOK
+	}
+	if obj.TransientHostname != "" {
+		propertyCheckOK, err := updateHostnameProperty(hostnameObject, obj.TransientHostname, "Hostname", "SetHostname", apply)
+		if err != nil {
+			return false, err
+		}
+		checkOK = checkOK && propertyCheckOK
+	}
+
```
purpleidea: Still missing a:

```
	// if we did work successfully, or are in a good state, then state is ok
	if apply || checkOK {
		obj.isStateOK = true
}
```
otherwise we don't cache the state when we fix it.

PR #95, opened 2016-11-14
nspawn: bump go-systemd commit

joejulian:
joejulian/go-systemd was updated to provide more machine1 features and
was rebased to the latest master

### DIFF
--- a/resources/nspawn.go
+++ b/resources/nspawn.go
@@ -136,6 +136,10 @@ func (obj *NspawnRes) Watch(processChan chan event.Event) error {
 	buschan := make(chan *dbus.Signal, 10)
 	bus.Signal(buschan)
 
+	// testing shows that if a machine fails to terminate the first
+	// time it will never succeed until 90 seconds has passed.
+	tickChan := time.NewTicker(time.Second * 91).C
+
 	var send = false
 	var exit = false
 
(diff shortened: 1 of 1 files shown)


purpleidea:
@joejulian Thanks for your patches... Here's a quick review before bed, I can be more thorough tomorrow if you want.

1) I have to NACK this as is, because there's more than one thing happening per commit. Please keep each commit as an individual item, that way they can be individually critiqued separately. For example, you change the "nor" spelling thing in the same patch as the other stuff.

2) This default state thing you added `default:` in the select, probably breaks the converged timeout feature. Try it and see if it works, my guess is that it does not.

3) This comment and code:

```
+			// systemd will return an EDEADLK if the
+			// termination occurs before the container is
+			// fully started. We will try again since
+			// isStateOK is false. 
```

The second part isn't true. What will cause it to check again? Certainly not mgmt's core. Only an event, or an error combined with a retry meta parameter will do that. So if indeed there is a state transition event we need to watch for (perhaps "container is now up") then we should ignore the EDEADLK and call this again when we get the "container is up" event. Make sense? (Assuming this analysis is correct.) If systemd has no event generated for when the container is ready (and thus ready to be killed) then it's something needed upstream.

joejulian:
Yeah, I'm not going to submit C code to the systemd project. I'll just abandon this since the upstream project doesn't allow this to work as needed.

PR #99, opened 2016-11-22
misc: Prefer dnf over yum when present

vinzenz:
Just a small fix to prefer DNF when present.

This will avoid warnings issues by the wrapper script.


purpleidea:
Looks like a perfect patch. Merged, thanks!

PR #101, opened 2016-11-23
Openpgp feature

gzsierra:
Sorry for the wait.

Here is the "basic" [OpenPGP](https://godoc.org/golang.org/x/crypto/openpgp) setup.

What's working: 
- [X] Generate Key Pair 
- [X] Import Key Pair 
- [X] Save Key Pair 
- [X] Encrypt/Decrypt messages 
- [X] Write to file

### DIFF
--- a/mgmtmain/cli.go
+++ b/mgmtmain/cli.go
@@ -96,6 +96,11 @@ func run(c *cli.Context) error {
 	obj.NoCaching = c.Bool("no-caching")
 	obj.Depth = uint16(c.Int("depth"))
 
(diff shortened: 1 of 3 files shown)


purpleidea:
Hey @gzsierra Thanks for the patch. I'm going to start with some light review to get some discussion going...

The main NACK here is the `instanceX` stuff. Is this something accidental you left in, or was it meant to be configurable later? LMK and perhaps ping me in IRC so we can chat about the patch!

Cheers!

PR #105, opened 2016-12-22
Convert to use github.com/libvirt/libvirt-go

berrange:
Convert the code to use the new libvirt Go bindings.

NB, this PR is compile tested only - someone should check functionality - in particular the connection auth stuff

### DIFF
--- a/resources/virt.go
+++ b/resources/virt.go
@@ -27,8 +27,8 @@ import (
 
 	"github.com/purpleidea/mgmt/event"
 
+	"github.com/libvirt/libvirt-go"
 	errwrap "github.com/pkg/errors"
-	"github.com/rgbkrk/libvirt-go"
 )
 
 func init() {
@@ -60,7 +60,7 @@ type VirtRes struct {
 	URI        string             `yaml:"uri"`       // connection uri, eg: qemu:///session
 	State      string             `yaml:"state"`     // running, paused, shutoff
 	Transient  bool               `yaml:"transient"` // defined (false) or undefined (true)
-	CPUs       uint16             `yaml:"cpus"`
+	CPUs       uint               `yaml:"cpus"`
 	Memory     uint64             `yaml:"memory"` // in KBytes
 	OSInit     string             `yaml:"osinit"` // init used by lxc
 	Boot       []string           `yaml:"boot"`   // boot order. values: fd, hd, cdrom, network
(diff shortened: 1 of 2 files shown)


resources/virt.go
```diff
@@ -120,12 +120,29 @@ func (obj *VirtRes) Validate() error {
 	return nil
 }
 
-func (obj *VirtRes) connect() (conn libvirt.VirConnection, err error) {
+func (obj *VirtRes) connect() (conn *libvirt.Connect, err error) {
 	if obj.Auth != nil {
-		conn, err = libvirt.NewVirConnectionWithAuth(obj.URI, obj.Auth.Username, obj.Auth.Password)
+		callback := func(creds []*libvirt.ConnectCredential) {
+			for _, cred := range creds {
```
purpleidea: Since I'm unfamiliar with this new callback mechanism, is the idea to just modify the values passed in?? A small comment here or in the patch would be appreciated.

resources/virt.go
```diff
@@ -290,7 +298,7 @@ func (obj *VirtRes) attrCheckApply(apply bool) (bool, error) {
 	}
 
 	// check cpus
-	if domInfo.GetNrVirtCpu() != obj.CPUs {
+	if (uint16)(domInfo.NrVirtCpu) != obj.CPUs {
```
purpleidea: Should obj.CPUs be a uint16 instead? Why shouldn't type match SetVCPUs?

berrange:
Branch refreshed

purpleidea:
@berrange Thanks! I'll look at this shortly.

purpleidea:
@berrange Would you mind rebasing to git master?
(Or may I do this on your behalf?)

Cheers!

berrange:
Sure, go ahead & rebase at will, or do whatever else you want to this patch :-)

purpleidea:
@berrange Rebased and merged as https://github.com/purpleidea/mgmt/commit/0b416e44f86f05c8a1714bec6a7883896030143d

Thanks so much for the patch!

BTW: Here is a demo of what we can now do with your lovely libvirt library and mgmt:
https://www.youtube.com/watch?v=MmpwOQAb_SE&feature=youtu.be&t=1790

I'd like to eventually run a `watch free -m` or something similar with `watch cat /proc/cpuinfo | grep ...` inside a vm, and then change the ram or vcpu count with mgmt and see it dynamically change on the fly. Is libvirt+a linux vm up to the task of doing this?

Cheers!

Issue #98, opened 2016-11-21
System package dependencies in installation documentation


LrsK:
I tried installing mgmt by following the quickstart guide on an Ubuntu 16.04 LTS machine.  I get the following error:

```
➜  ~ go get -u github.com/purpleidea/mgmt
# github.com/rgbkrk/libvirt-go
dev/go/src/github.com/rgbkrk/libvirt-go/cfuncs.go:6:29: fatal error: libvirt/libvirt.h: No such file or directory
compilation terminated.
```

After running ``` sudo apt-get install libvirt-dev ``` this message disappears, but when attempting ``` go get ./... ``` I get arrors about pcap.h and systemd/sd-journal.h. After ``` sudo apt-get install libpcap-dev libsystemd-dev ``` these errors went away. 

Should these dependencies be listed in the docs, or are they so system dependent that the user will have to figure this out for their system?

purpleidea:
Technically these deps are listed in the deps script.

https://github.com/purpleidea/mgmt/blob/master/misc/make-deps.sh#L21

You're right that this could be clearer. Any recommendations? (Maybe a patch?)

PR #106, opened 2016-12-22
Changed wording of quick start section. Issue #98.

LrsK:
Issue #98 discusses some wording in the quick start guide. Attempted to make it clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ export GOPATH=$HOME/gopath
 go get -u github.com/purpleidea/mgmt
 cd $GOPATH/src/github.com/purpleidea/mgmt
 ```
-* Get the remaining golang deps with `go get ./...`, or run `make deps` if you're comfortable with how we install them.
+* Run `make deps` to install system and golang dependencies. Take a look at `misc/make-deps.sh` for details.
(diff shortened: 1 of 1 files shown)


purpleidea:
@LrsK Awesome, thanks! Could you make two really small changes please:

s/Go/golang/ so it's less ambiguous for new readers, and could you add a "readme: " prefix to your commit message?

Thanks!!

LrsK:
Did that work? :)

purpleidea:
@LrsK The changes are perfect, but this is now two commits instead of one.
You want to rebase these into a single commit. This is good practice if you've never done it before, but I can advise how if you need the help. Please LMK.

LrsK:
I've never done a rebase before, so that was fun to try. I messed up a few times too, so I got to try it several times. ;)

purpleidea:
@LrsK Great! Glad you tried out rebasing! Pro-tip: for simple single patch fixes, when you made the changes I asked for the first time around, instead of doing `git commit` there is a rebase shortcut: `git commit --amend` which will merge the new changes into the last commit. Then you can force push up here.

HTH and thanks for the patch! Send us some more!

PR #107, opened 2017-01-22
docs: Fix headers formating

goberghen:
Spaces between «#» and text

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -572,7 +572,7 @@ To report any bugs, please file a ticket at: [https://github.com/purpleidea/mgmt
 
 ## Authors
 
-Copyright © 2013-2016+ James Shubin and the project contributors
+Copyright (C) 2013-2016+ James Shubin and the project contributors
 
 Please see the
 [AUTHORS](https://github.com/purpleidea/mgmt/tree/master/AUTHORS) file
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -1,4 +1,4 @@
-#mgmt
+# mgmt
 
 <!--
 Mgmt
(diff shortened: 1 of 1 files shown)


docs/documentation.md
```diff
@@ -570,9 +570,9 @@ You can follow along [on my technical blog](https://ttboj.wordpress.com/).
 
 To report any bugs, please file a ticket at: [https://github.com/purpleidea/mgmt/issues](https://github.com/purpleidea/mgmt/issues).
 
-##Authors
+## Authors
 
-Copyright (C) 2013-2016+ James Shubin and the project contributors
+Copyright © 2013-2016+ James Shubin and the project contributors
```
purpleidea: I'd rather not make this `(C)` change, unless you have a good reason why.

goberghen: > Can you comment on why we need to make this change?

Well, it's a proper markdown syntax. Some markdown implementations will not convert `#Header` headers without whitespaces. CommonMark for example.
>I'd rather not make this (C) change, unless you have a good reason why.
I thought it should be proper copyright symbol too, since it's 2017 and we have unicode.

purpleidea: @goberghen Can you point to the spec that mentions needing that space?

The (C) is more accessible for anyone reading it, therefore for this reason I will leave it as is.

goberghen: @purpleidea,
https://guides.github.com/features/mastering-markdown/
http://spec.commonmark.org/0.27/
https://daringfireball.net/projects/markdown/basics — original markdown.pl
Headers with spaces works in every markdown implementations, it's the right syntax.

purpleidea:
@goberghen Okay, you've convinced me there should be a space!

If so, could you also include the change to the other markdown files in docs/* and README.md too please? This should all be in a single commit. Thanks!

purpleidea:
PS: please change the commit message to be something like:

docs: The message goes here

goberghen:
@purpleidea, can you close this pull request? I'll clone the repo on my computer, make changes in 1 commit and send you new pull request.
Don't know how to make it from in GitHub interface. There is already 2 commits.

purpleidea:
@goberghen I can tell you how if you want...

purpleidea:
And for future reference, you can update a branch with a force push.

goberghen:
Done.

> And for future reference, you can update a branch with a force push.

I know how to do it with git itself, but for github I think I should read documentation more.

PR #110, opened 2017-02-02
implement file attributes

mildred:
Add owner which must be username or uid of the file owner, group which is
the group name or gid of the file, and mode which is the octal unix file
permissions.

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -247,6 +247,15 @@ The exec resource can execute commands on your system.
 The file resource manages files and directories. In `mgmt`, directories are
 identified by a trailing slash in their path name. File have no such slash.
 
+It has the following properties:
+
+- `path`: file path (directories have a trailing slash here)
+- `content`: raw file content
+- `state`: either `exists` (the default value) or `absent`
+- `mode`: octal unix file permissions
+- `owner`: username or uid for the file owner
+- `group`: group name or gid for the file group
+
 #### Path
 
 The path property specifies the file or directory that we are managing.
--- a/resources/file.go
+++ b/resources/file.go
@@ -27,9 +27,12 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/user"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/purpleidea/mgmt/event"
 	"github.com/purpleidea/mgmt/recwatch"
@@ -51,6 +54,9 @@ type FileRes struct {
 	Content    *string `yaml:"content"` // nil to mark as undefined
 	Source     string  `yaml:"source"`  // file path for source content
 	State      string  `yaml:"state"`   // state: exists/present?, absent, (undefined?)
+	Owner      string  `yaml:"owner"`
+	Group      string  `yaml:"group"`
+	Mode       string  `yaml:"mode"`
 	Recurse    bool    `yaml:"recurse"`
 	Force      bool    `yaml:"force"`
 	path       string  // computed path
@@ -102,12 +108,53 @@ func (obj *FileRes) Validate() error {
 		return fmt.Errorf("Can't specify Content when creating a Dir.")
 	}
 
+	if obj.Mode != "" {
+		if _, err := obj.mode(); err != nil {
+			return err
+		}
+	}
+
+	if _, err := obj.uid(); obj.Owner != "" && err != nil {
+		return err
+	}
+
+	if _, err := obj.gid(); obj.Group != "" && err != nil {
+		return err
+	}
+
 	// XXX: should this specify that we create an empty directory instead?
 	//if obj.Source == "" && obj.isDir {
 	//	return fmt.Errorf("Can't specify an empty source when creating a Dir.")
 	//}
 
 	return obj.BaseRes.Validate()
+}
+
+// mode returns the file permission specified on the graph. It doesn't handle
+// the case where the mode is not specified. The caller should check obj.Mode is
+// not empty.
+func (obj *FileRes) mode() (os.FileMode, error) {
+	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	if err != nil {
+		return os.FileMode(0), errwrap.Wrapf(err, "Mode should be an octal number (%s)", obj.Mode)
+	}
+	return os.FileMode(m), nil
+}
+
+// uid returns the user id for the owner specified in the yaml file graph.
+// Caller should first check obj.Owner is not empty
+func (obj *FileRes) uid() (int, error) {
+	u2, err2 := user.LookupId(obj.Owner)
+	if err2 == nil {
+		return strconv.Atoi(u2.Uid)
+	}
+
+	u, err := user.Lookup(obj.Owner)
+	if err == nil {
+		return strconv.Atoi(u.Uid)
+	}
+
+	return -1, errwrap.Wrapf(err, "Owner lookup error (%s)", obj.Owner)
 }
 
 // Init runs some startup code for this resource.
@@ -614,6 +661,99 @@ func (obj *FileRes) contentCheckApply(apply bool) (checkOK bool, _ error) {
 	return checkOK, nil
 }
 
+// chmodCheckApply performs a CheckApply for the file permissions.
+func (obj *FileRes) chmodCheckApply(apply bool) (checkOK bool, _ error) {
+	log.Printf("%s[%s]: chmodCheckApply(%t)", obj.Kind(), obj.GetName(), apply)
+
+	if obj.State == "absent" {
+		// File is absent
+		return true, nil
+	}
+
+	if obj.Mode == "" {
+		// No mode specified, everything is ok
+		return true, nil
+	}
+
+	mode, err := obj.mode()
+	if err != nil {
+		return false, err
+	}
+
+	st, err := os.Stat(obj.Path)
+	if err != nil {
+		return false, err
+	}
+
+	// Nothing to do
+	if st.Mode() == mode {
+		return true, nil
+	}
+
+	// Not clean but don't apply
+	if !apply {
+		return false, nil
+	}
+
+	err = os.Chmod(obj.Path, mode)
+	return false, err
+}
+
+// chownCheckApply performs a CheckApply for the file ownership.
+func (obj *FileRes) chownCheckApply(apply bool) (checkOK bool, _ error) {
+	var expectedUID, expectedGID int
+	log.Printf("%s[%s]: chownCheckApply(%t)", obj.Kind(), obj.GetName(), apply)
+
+	if obj.State == "absent" {
+		// File is absent or no owner specified
+		return true, nil
+	}
+
+	st, err := os.Stat(obj.Path)
+	if err != nil {
+		return false, err
+	}
+
+	stUnix, ok := st.Sys().(*syscall.Stat_t)
+	if !ok {
+		// Not unix
+		panic("No support for your platform")
+	}
+
+	if obj.Owner != "" {
+		expectedUID, err = obj.uid()
+		if err != nil {
+			return false, err
+		}
+	} else {
+		// Nothing specified, no changes to be made, expect same as actual
+		expectedUID = int(stUnix.Uid)
+	}
+
+	if obj.Group != "" {
+		expectedGID, err = obj.gid()
+		if err != nil {
+			return false, err
+		}
+	} else {
+		// Nothing specified, no changes to be made, expect same as actual
+		expectedGID = int(stUnix.Gid)
+	}
+
+	// Nothing to do
+	if int(stUnix.Uid) == expectedUID && int(stUnix.Gid) == expectedGID {
+		return true, nil
+	}
+
+	// Not clean, but don't apply
+	if !apply {
+		return false, nil
+	}
+
+	err = os.Chown(obj.Path, expectedUID, expectedGID)
+	return false, err
+}
+
 // CheckApply checks the resource state and applies the resource if the bool
 // input is true. It returns error info and if the state check passed or not.
 func (obj *FileRes) CheckApply(apply bool) (checkOK bool, _ error) {
@@ -638,19 +778,17 @@ func (obj *FileRes) CheckApply(apply bool) (checkOK bool, _ error) {
 		checkOK = false
 	}
 
-	// TODO
-	//if c, err := obj.chmodCheckApply(apply); err != nil {
-	//	return false, err
-	//} else if !c {
-	//	checkOK = false
-	//}
+	if c, err := obj.chmodCheckApply(apply); err != nil {
+		return false, err
+	} else if !c {
+		checkOK = false
+	}
 
-	// TODO
-	//if c, err := obj.chownCheckApply(apply); err != nil {
-	//	return false, err
-	//} else if !c {
-	//	checkOK = false
-	//}
+	if c, err := obj.chownCheckApply(apply); err != nil {
+		return false, err
+	} else if !c {
+		checkOK = false
+	}
 
 	return checkOK, nil // w00t
 }
--- a/resources/file_attrs.go
+++ b/resources/file_attrs.go
@@ -1,0 +1,43 @@
+// Mgmt
+// Copyright (C) 2013-2016+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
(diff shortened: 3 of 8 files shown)


purpleidea:
@mildred Awesome to have a first patch from you, thanks!

Before I do a full review, let's get this passing... It looks like on golang 1.6 we get:

```
resources/file.go:165: undefined: user.LookupGroup
resources/file.go:170: undefined: user.LookupGroupId
```

Do you know what this is about?
One more "first contributor" nitpick, could you amend the commit message to have a prefix of:
"resources: file: <your message>"

Thanks - Full review coming after this builds :)

mildred:
Probably `user.LookupGroup` and `user.LookupGroupId` were added on go 1.7 ... What is the version requirement ?

purpleidea:
On Thu, Feb 2, 2017 at 3:35 PM, mildred <notifications@github.com> wrote:
>
> Probably user.LookupGroup and user.LookupGroupId were added on go 1.7 ... What is the version requirement ?

We're 1.6 or greater for the time being. I'm happy to move to 1.7, but
not until it's common in the distros. Mine (F24) is still on 1.6.4 for
example. Is there an alternate way to do this, or could we copy the
library code in temporarily until we're 1.7+ ?

Cheers!

mildred:
Do you have any objections to using https://github.com/hnakamur/groupon go 1.6 ?

purpleidea:
@mildred This looks fine to me, if it will solve the 1.6 problem. We can move to 1.7 in a few months (6?) and change this out. Maybe just include a TODO: replace with stdlib sort of comment?
Link: https://github.com/hnakamur/group

Thanks!

mildred:
I used [build constraints](https://golang.org/pkg/go/build/) which I think is better for that :) I could finally make them work (they need to be before the `package` keyword.

purpleidea:
Having a look now-- quick review: why is there a submodule? (We can just get the latest version from git master without this)

resources/file_attrs_oldgo.go
```diff
@@ -0,0 +1,45 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// +build !go1.7
```
purpleidea: Since this will be used on 1.8 as well, shouldn't this be labelled 1.6 and the other be !1.6 (so that new things work on 1.8 and etc...)

purpleidea:
Looks like tests failed for two reasons:

1) FAIL: Commit message should match the following regex: '^\([a-z0-9]\(\(, \)\|[a-z0-9]\)\+[a-z0-9]: \)\+[A-Z0-9][^:]\+[^:.]$'

eg:

prometheus: Implement rest api

resources: svc: Fix a race condition with reloads

2) yamlfmt failed. Run make  yamlfmt to get the canonical version of the change you made to the yaml file.

resources/file_attrs_oldgo.go
mildred: documentation says *"go1.7", from Go version 1.7 onward*. So this will exclude 1.7, 1.8, ...

mildred:
> Having a look now-- quick review: why is there a submodule? (We can just get the latest version from git master without this)

I just saw other dependencies in submodules and thought this was the way to bundle them. I can remove that provided it is in the GOPATH (go get would work fine here)

I'll update the PR soon...

mildred:
> yamlfmt failed. Run make yamlfmt to get the canonical version of the change you made to the yaml file.

I did not changed yaml files. Running `make yamlfmt` results in changes in (almost?) all yaml files. basically, multistrings introduced by `|` are replaced with `'...'`. Should I include that in my last commit, although it doesn't relate to what I did ?

mildred:
Well, a yaml file used for test had permissions in octal `0614`. formatyaml converted this to base 10. I replaced it by a yaml string `'0614'` which is identical since the go structure uses a string for that value. Hopefully travis won't disagree with that.

mildred:
I fixed some linter warnings in the last commit. Will that be enough?

resources/file.go
```diff
@@ -110,6 +130,32 @@ func (obj *FileRes) Validate() error {
...
 }
 
+func (obj *FileRes) mode() (os.FileMode, error) {
+	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	if err != nil {
+		return os.FileMode(0), fmt.Errorf("Mode should be an octal number (%v): %v", obj.Mode, err.Error())
+	}
+	return os.FileMode(m), nil
+}
+
```
purpleidea: Needs short comment please about what this function does. Follow the standard format of:

// uid does blah blah blah.

resources/file.go
```diff
@@ -110,6 +130,32 @@ func (obj *FileRes) Validate() error {
 	return obj.BaseRes.Validate()
 }
 
+func (obj *FileRes) mode() (os.FileMode, error) {
+	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	if err != nil {
+		return os.FileMode(0), fmt.Errorf("Mode should be an octal number (%v): %v", obj.Mode, err.Error())
```
purpleidea: %s and errwrap instead please. Same for the other places that do this as well.

resources/file.go
```diff
@@ -110,6 +130,32 @@ func (obj *FileRes) Validate() error {
...
+	}
+	return os.FileMode(m), nil
+}
+
+func (obj *FileRes) uid(defaultVal int) (int, error) {
+	if obj.Owner == "" {
+		return defaultVal, nil
+	}
+
+	u, err := user.Lookup(obj.Owner)
```
purpleidea: Assuming I'm understand the logic correctly (a comment would help!) we assume the string is a username, and if so return the uid? If not fallback to assuming it's a uid??

If so wouldn't we want to do the direct uid lookup first, as a number could be a username, a username couldn't be an id?

resources/file.go
```diff
@@ -110,6 +130,32 @@ func (obj *FileRes) Validate() error {
...
+	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	if err != nil {
+		return os.FileMode(0), fmt.Errorf("Mode should be an octal number (%v): %v", obj.Mode, err.Error())
+	}
+	return os.FileMode(m), nil
+}
+
+func (obj *FileRes) uid(defaultVal int) (int, error) {
+	if obj.Owner == "" {
+		return defaultVal, nil
```
purpleidea: I don't understand what this defaultVal stuff is about...

purpleidea:
@mildred Thanks for the patch! When you have a moment, could you please look at the review comments, and send me some updates! Cheers :)

purpleidea:
(you might also want to rebase to the newest git master)

resources/file.go
mildred: function uid returns the uid resolved for the current platform specified in the graph. If no uid was specified, return defaultVal

resources/file.go
mildred: Yes, we check first for username then user id, we could (and perhaps should) do the contrary.

resources/file.go
```diff
@@ -614,6 +660,81 @@ func (obj *FileRes) contentCheckApply(apply bool) (checkOK bool, _ error) {
...
+	st, err := os.Stat(obj.Path)
+	if err != nil {
+		return false, err
+	}
+
+	if st.Mode() == mode {
+		return true, nil
+	}
+
+	if apply {
```
mildred: no problem

resources/file.go
```diff
@@ -614,6 +660,81 @@ func (obj *FileRes) contentCheckApply(apply bool) (checkOK bool, _ error) {
...
+		// File is absent or no owner specified
+		return true, nil
+	}
+
+	st, err := os.Stat(obj.Path)
+	if err != nil {
+		return false, err
+	}
+
+	if stUnix, ok := st.Sys().(*syscall.Stat_t); ok {
```
mildred: `!ok` means mgmt is not running on an unix system. We should probably panic there until we support other operating systems. panic nd not error because it should never happen. This is not a normal error that can occur, but a truly exceptional condition that represents a bug (lack of support for the current platform)

resources/file.go
```diff
@@ -614,6 +660,81 @@ func (obj *FileRes) contentCheckApply(apply bool) (checkOK bool, _ error) {
...
+			return false, err
+		}
+
+		expectedGID, err := obj.gid(int(stUnix.Gid))
+		if err != nil {
+			return false, err
+		}
+
+		if int(stUnix.Uid) != expectedUID || int(stUnix.Gid) != expectedGID {
+			if apply {
```
mildred: ok

resources/file_attrs.go
```diff
@@ -0,0 +1,44 @@
...
+
+package resources
+
+import (
+	"fmt"
+	"os/user"
+	"strconv"
+)
+
+func (obj *FileRes) gid(defaultVal int) (int, error) {
```
mildred: Ok, i'll add: gid returns the group id specified in the graph or defaultVal if no value is specified. This function resolves the value in the graph to a numerical id for the current platform.

purpleidea: But why is this helpful to return a "default value" ?

resources/file.go
purpleidea: Then return early with panic(blah)

resources/file_attrs.go
mildred: I'll remove this default value. This avoids an extra check outside of the function in case the group is not set. Maybe not that helpful.

resources/file_attrs.go
```diff
@@ -0,0 +1,44 @@
...
+	if err == nil {
+		return strconv.Atoi(g.Gid)
+	}
+
+	g2, err2 := user.LookupGroupId(obj.Group)
+	if err2 == nil {
+		return strconv.Atoi(g2.Gid)
+	}
+
+	return defaultVal, fmt.Errorf("Group lookup error (%v): %v", obj.Group, err.Error())
```
mildred: ok

resources/file.go
```diff
@@ -110,6 +130,37 @@ func (obj *FileRes) Validate() error {
 	return obj.BaseRes.Validate()
 }
 
+// mode return the file permission specified on the graph. It doesn't handle the
```
purpleidea: nit: mode returns* (missing an s)

resources/file.go
```diff
@@ -614,6 +665,94 @@ func (obj *FileRes) contentCheckApply(apply bool) (checkOK bool, _ error) {
...
+		return false, err
+	}
+
+	stUnix, ok := st.Sys().(*syscall.Stat_t)
+	if !ok {
+		// Not unix
+		panic("No support for your platform")
+	}
+
+	expectedUID := int(stUnix.Uid)
```
purpleidea: is this expected or actual? It looks like actual... (similarly elsewhere...)

mildred:
I have prometheus-3.sh failing, don't know why...

resources/file.go
```diff
@@ -110,6 +130,37 @@ func (obj *FileRes) Validate() error {
...
+// empty.
+func (obj *FileRes) mode() (os.FileMode, error) {
+	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	if err != nil {
+		return os.FileMode(0), errwrap.Wrapf(err, "Mode should be an octal number (%s)", obj.Mode)
+	}
+	return os.FileMode(m), nil
+}
+
+// uid returns the user id for the owner specified in the graph. Caller should
```
purpleidea: function comments are sentences, and should end with a period. Also, "the owner specified in the graph" -- isn't this the owner specified by the file?

resources/file.go
```diff
@@ -110,6 +130,37 @@ func (obj *FileRes) Validate() error {
...
+// first check obj.Owner is not empty
+func (obj *FileRes) uid() (int, error) {
+	if obj.Owner == "" {
+		return -1, nil
```
purpleidea: What? Why...

The code is a bit unclear... Can you have a re-read through the whole thing and make sure it all makes sense?

mildred:
This one seems to fail:

```
# One CheckApply for a File ; in noop mode.
curl 127.0.0.1:9233/metrics | grep '^mgmt_checkapply_total{apply="false",errorful="false",eventful="true",kind="File"} 1$'
```

purpleidea:
@mildred Don't worry about this stuff. @roidelapluie can have a look, but it is possibly being caused by a race that I am working on fixing. Sorry about that!

docs/documentation.md
```diff
@@ -247,6 +247,15 @@ The exec resource can execute commands on your system.
 The file resource manages files and directories. In `mgmt`, directories are
 identified by a trailing slash in their path name. File have no such slash.
 
+It has the following properties:
+
+- `path`: file path (directories have a trailing slash here)
+- `content`: raw file content
+- `state`: either `exists` (the default value) or `absent`
+- `mode`: octal unix file permissions
```
roidelapluie: specify that we expect a "string"

resources/file.go
```diff
@@ -102,6 +108,20 @@ func (obj *FileRes) Validate() error {
 		return fmt.Errorf("Can't specify Content when creating a Dir.")
 	}
 
+	if obj.Mode != "" {
+		if _, err := obj.mode(); err != nil {
+			return err
```
roidelapluie: return err vs errwrap.Wrapf? @purpleidea ?

roidelapluie:
could maybe have some more tests?

maybe add:

```
if ! timeout 1 sudo true;
echo "skipping because no sudo access"
exit 0
fi
```
so you can run mgmt as root and test the uid/gid (should work on travis)

resources/file.go
purpleidea: if it would provide useful context, we always like to errwrap :) -- whether it makes sense really depends on what mode() does.

docs/documentation.md
mildred: Works if not a string too

resources/file.go
mildred: obj.mode() already uses errwrap if needed (unless I missed something)

docs/documentation.md
mildred: but I should add test for that perhaps

resources/file.go
mildred: Owner specified by the yaml file graph

resources/file.go
mildred: It should never happen, the owner is checking that value. I like to be sure.

mildred: Better not put 0 which is root

resources/file.go
mildred: it is expected, but in case the owner is not specified, the default value is actual so it detect no changes

mildred:
Updated with tests for owner and fixes requested

test/shell/file-owner.sh
```diff
@@ -0,0 +1,24 @@
...
+wait $pid	# get exit status
+e=$?
+
+ls -l /tmp/mgmt
+
+test -e /tmp/mgmt/f1
+test -e /tmp/mgmt/f2
+test $(stat -c%U:%G /tmp/mgmt/f1) = root:root
+test $(stat -c%u:%g /tmp/mgmt/f2) = 1:2
+
```
roidelapluie: If I had one last nip it would be to add:

sudo -A chown 5 /tmp/mgmt/f1

test $(stat -c%U:%G /tmp/mgmt/f1) = root:root

purpleidea:
@mildred Merged, w00t!

Thanks for the patch! Sorry for the long review. I realize now in retrospect that I completely didn't understand why you were doing the -1 stuff, which I know get, but I think the patch is a lot nicer like this!

Hope to have mildred{1..4} on a future patch! Ping me if you're looking for something to work on!

Cheers
James

purpleidea:
Merged as https://github.com/purpleidea/mgmt/commit/8c2c55216417e7f83c001ca9499943db972f2df1

test/shell/file-owner.sh
mildred: It won't work because we are waiting for mgmt to terminate before we check the files.

mildred:
Thank you, and no problem for the defaultVal. The patch is nicer that way. I'm waiting for you on #143 now 😉

purpleidea:
Looking!

PR #111, opened 2017-02-06
point to the correct puppet-guide path in docs.

TomRitserveldt:


purpleidea:
@TomRitserveldt Awesome, thanks for the fix!

Cheers
James

PR #112, opened 2017-02-06
TODO: Point to current file improvement issue

daenney:


daenney:
Also... it feels that `TODO.md` is a bit outdated? Is all that's there still up for grabs?

PR #123, opened 2017-02-08
Prometheus: Add mgmt specific metrics

roidelapluie:


pgraph/actions.go
```diff
@@ -414,10 +415,26 @@ func (g *Graph) Worker(v *Vertex) error {
 						playback = true
 						log.Printf("%s[%s]: CheckApply errored: %v", v.Kind(), v.GetName(), e)
 						if retry == 0 {
+							if obj.Prometheus() != nil {
+								if err := obj.Prometheus().UpdateState(fmt.Sprintf("%v[%v]", v.Kind(), v.GetName()), v.Kind(), prometheus.ResStateHardFail); err != nil {
+									// TODO: error correctly
+									log.Printf("%s[%s]: Prometheus.UpdateState() errored: %v", v.Kind(), v.GetName(), err)
+									return
```
roidelapluie: @purpleidea what do we want to do here? how to error correctly?

purpleidea: prom. should be running around the existing code and 
                    instrumenting it, but i would assume it shouldn't be 
                    interferring with behavior or things like that... right?

purpleidea: i think if you're experiencing failures at 
                    this sort of reporting level, it might make sense to: 1) 
                    report the errors back through the struct. 2) Have an error 
                    channel function that we hook up to in main, and 3) 
                    optionally fail or ignore based on some option the errors 
                    based on user desires.

purpleidea:
Why closed?

roidelapluie:
@purpleidea I reopen this with small parts ;; without requiring your fix first. I will start with mgmt_checkapply_total.

Issue #124, opened 2017-02-09
moving the directory where a file belongs breaks mgmt


roidelapluie:
if we move the parent directory of a file mgmt watches, mgmt will still watch the old file.

purpleidea:
I think this is related or the same as https://github.com/purpleidea/mgmt/issues/120

roidelapluie:
@purpleidea I do not think so. In #120 mgmt notifies that something is happening. #120 does not break the file functionality, it just adds some overhead.

This one breaks it (mgmt does not do what we exect.)

purpleidea:
1.7. New fanotify FAN_RENAME flag

This release adds a FAN_RENAME event to the fanotify interfaces to report extra info records about new and old parent+name details. This new event was designed as a replacement for the "inotify way" of joining the MOVED_FROM/MOVED_TO events using a cookie

purpleidea:
https://kernelnewbies.org/Linux_5.17#New_fanotify_FAN_RENAME_flag

purpleidea:
fanotify

    (FEATURED) Add FAN_RENAME event to report extra info records about new and old parent+name details. This new event was designed as a replacement for the "inotify way" of joining the MOVED_FROM/MOVED_TO events using a cookie [commit](https://git.kernel.org/linus/e54183fa7047c15819bc155f4c58501d9a9a3489), [commit](https://git.kernel.org/linus/2d9374f095136206a02eb0b6cd9ef94632c1e9f7), [commit](https://git.kernel.org/linus/1a9515ac9e55e68d733bab81bd408463ab1e25b1), [commit](https://git.kernel.org/linus/3cf984e950c1c3f41d407ed31db33beb996be132), [commit](https://git.kernel.org/linus/3982534ba5ce45e890b2f5ef5e7372c1accd14c7), [commit](https://git.kernel.org/linus/2bfbcccde6e7a787feabad4645f628f963fe0663), [commit](https://git.kernel.org/linus/7326e382c21e9c23c89c88369afdc90b82a14da8), [commit](https://git.kernel.org/linus/8cc3b1ccd930fe6971e1527f0c4f1bdc8cb56026)

    Add a new FAN_REPORT_TARGET_FID flag, which supports reporting child info in directory fanotify events [commit](https://git.kernel.org/linus/d61fd650e9d206a71fda789f02a1ced4b19944c4)

PR #125, opened 2017-02-09
inotify: test the behaviour of inotify on parent dir moves

roidelapluie:
This is a test for #124. It is disabled until #124 is fixed, so it can
already me merged.

### DIFF
--- a/examples/deep-dirs.yaml
+++ b/examples/deep-dirs.yaml
@@ -1,0 +1,9 @@
+---
+graph: mygraph
+resources:
+  file:
+  - name: file1
+    path: "/tmp/mgmt/a/b/c/f1"
+    content: |
+      i am f1
+    state: exists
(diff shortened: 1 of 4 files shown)


test/shell/move-upper-dir.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -e
+
+# TEST FOR #124 --- Disabled for now
```
purpleidea: Can you add a FIXME: prefix here so scripts I use to find issues will see this please?

purpleidea: Can you change the commit message prefix to "test: file:" please or "test: resources: file:" ... ?
The shell script could have a file- prefix as well.

Thanks

purpleidea:
W00t! Thanks for the super helpful patch!

purpleidea:
<fireworks emoji>

PR #128, opened 2017-02-09
Augeas resource

roidelapluie:
I open this pull request to get continuous feedback during all my work.

### DIFF
--- a/resources/augeas.go
+++ b/resources/augeas.go
@@ -1,0 +1,307 @@
+// Mgmt
+// Copyright (C) 2013-2016+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"encoding/gob"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/purpleidea/mgmt/event"
+	"github.com/purpleidea/mgmt/recwatch"
+
+	errwrap "github.com/pkg/errors"
+	// FIXME: we vendor go/augeas because master requires augeas 1.6.0
+	// and libaugeas-dev-1.6.0 is not yet available in a PPA.
+	"honnef.co/go/augeas"
+)
+
+func init() {
+	gob.Register(&AugeasRes{})
+}
+
+// AugeasRes is a resource that enables you to use the augeas resource.
+// Currently only allows you to change simple files (e.g sshd_config).
+type AugeasRes struct {
+	BaseRes `yaml:",inline"`
+
+	// File is the path to the file targetted by this resource.
+	// If specified, mgmt will watch that file.
+	File string `yaml:"file"`
+
+	// Lens is the lens  used by this resource. If specified, mgmt
+	// will lower the augeas overhead by only loeading that lens.
+	Lens string `yaml:"lens"`
+
+	// Sets is a list of changes that will be applied to the file, in the form of
+	// ["path", "value"]. mgmt will run augeas.Get() before augeas.Set(), to
+	// prevent changing the file when it is not needed.
+	Sets []AugeasSet `yaml:"sets"`
+
+	recWatcher *recwatch.RecWatcher // used to watch the changed files
+}
+
+// AugeasSet represents a key/value pair of settings to be applied.
+type AugeasSet struct {
+	Path  string `yaml:"path"`  // The relative path to the value to be changed.
+	Value string `yaml:"value"` // The value to be set on the given Path.
+}
+
+// NewAugeasRes is a constructor for this resource. It also calls Init() for you.
+func NewAugeasRes(name string) (*AugeasRes, error) {
+	obj := &AugeasRes{
+		BaseRes: BaseRes{
+			Name: name,
+		},
+	}
+	return obj, obj.Init()
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *AugeasRes) Default() Res {
+	return &AugeasRes{}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *AugeasRes) Validate() error {
+	if !strings.HasPrefix(obj.File, "/") {
+		return fmt.Errorf("File should start with a slash.")
+	}
+	if obj.Lens != "" && !strings.HasSuffix(obj.Lens, ".lns") {
+		return fmt.Errorf("Lens should have a .lns suffix.")
+	}
+	if (obj.Lens == "") != (obj.File == "") {
+		return fmt.Errorf("File and Lens must be specified together.")
+	}
+	return obj.BaseRes.Validate()
+}
+
+// Init initiates the resource.
+func (obj *AugeasRes) Init() error {
+	obj.BaseRes.kind = "Augeas"
+	return obj.BaseRes.Init() // call base init, b/c we're overriding
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+// Taken from the File resource.
+// FIXME: DRY - This is taken from the file resource
+func (obj *AugeasRes) Watch(processChan chan *event.Event) error {
+	var err error
+	obj.recWatcher, err = recwatch.NewRecWatcher(obj.File, false)
+	if err != nil {
+		return err
+	}
+	defer obj.recWatcher.Close()
+
+	// notify engine that we're running
+	if err := obj.Running(processChan); err != nil {
+		return err // bubble up a NACK...
+	}
+
+	var send = false // send event?
+	var exit *error
+
+	for {
+		if obj.debug {
+			log.Printf("%s[%s]: Watching: %s", obj.Kind(), obj.GetName(), obj.File) // attempting to watch...
+		}
+
+		select {
+		case event, ok := <-obj.recWatcher.Events():
+			if !ok { // channel shutdown
+				return nil
+			}
+			if err := event.Error; err != nil {
+				return errwrap.Wrapf(err, "Unknown %s[%s] watcher error", obj.Kind(), obj.GetName())
+			}
+			if obj.debug { // don't access event.Body if event.Error isn't nil
+				log.Printf("%s[%s]: Event(%s): %v", obj.Kind(), obj.GetName(), event.Body.Name, event.Body.Op)
+			}
+			send = true
+			obj.StateOK(false) // dirty
+
+		case event := <-obj.Events():
+			if exit, send = obj.ReadEvent(event); exit != nil {
+				return *exit // exit
+			}
+			//obj.StateOK(false) // dirty // these events don't invalidate state
+		}
+
+		// do all our event sending all together to avoid duplicate msgs
+		if send {
+			send = false
+			obj.Event(processChan)
+		}
+	}
+}
+
+// checkApplySet runs CheckApply for one element of the AugeasRes.Set
+func (obj *AugeasRes) checkApplySet(apply bool, ag *augeas.Augeas, set AugeasSet) (bool, error) {
+	fullpath := fmt.Sprintf("/files/%v/%v", obj.File, set.Path)
+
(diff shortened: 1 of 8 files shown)


purpleidea:
@roidelapluie Just ping me everytime you want a quick look. Cheers

NOTE: You'll probably find it useful to use the recwatch package. See file.go for the equivalent watching code. That way bug fixes to that package benefit everyone.

roidelapluie:
@purpleidea you can review now :)

Objective #1 is to have example/augeas1.yaml working ; this is the case.

I would like to get this working before I add more features to this resource.

resources/augeas.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *AugeasRes) Default() Res {
+	return &AugeasRes{}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *AugeasRes) Validate() error {
+	if strings.HasPrefix(obj.File, "/") == false {
```
purpleidea: if !str....

resources/augeas.go
```diff
@@ -0,0 +1,246 @@
...
+			return false, errwrap.Wrapf(err, "augeas: Error while loading!")
+		}
+	}
+
+	for _, set := range obj.Sets {
+		// sets:
+		path, value := set[0], set[1]
+		fullpath := fmt.Sprintf("/files/%v/%v", obj.File, path)
+		if getValue, agerr := ag.Get(fullpath); agerr != nil || value != getValue {
+			// note: augeas.Get throws an error is thrown if the path does not exist
```
purpleidea: grammar?

resources/augeas.go
```diff
@@ -0,0 +1,246 @@
...
+			checkOK = false
+			if seterr := ag.Set(fullpath, value); seterr != nil {
+				return false, errwrap.Wrapf(agerr, "Error while setting augeas value!")
+			}
+		}
+	}
+
+	if checkOK == false {
```
purpleidea: !x instead of x == false

resources/augeas.go
```diff
@@ -0,0 +1,246 @@
...
+		// do all our event sending all together to avoid duplicate msgs
+		if send {
+			send = false
+			obj.Event(processChan)
+		}
+	}
+}
+
+// CheckApply method for Augeas resource.
+func (obj *AugeasRes) CheckApply(apply bool) (checkOK bool, err error) {
```
purpleidea: You're missing the ver important !apply pattern. This is how noop works. See other resources for examples.
Also, where does your resource ever return true, nil?

roidelapluie: Fixed

resources/augeas.go
```diff
@@ -0,0 +1,249 @@
...
+	gob.Register(&AugeasRes{})
+}
+
+// AugeasRes is a resource that enables you to use the augeas resource.
+// Currently only allows you to change simple files (e.g sshd_config).
+type AugeasRes struct {
+	BaseRes    `yaml:",inline"`
+	File       string               `yaml:"file"` // the file targetted by this resource. If specified, mgmt will watch that file. Otherwhise, mgmt will poll the resource.
+	Lens       string               `yaml:"lens"` // the lens used by this resource. If specified, mgmt will watch that file.
+	Sets       [][]string           `yaml:"sets"` // A list of changes that will be applied to the file, in the form of ["path", "value"]. mgmt will run augeas.Get() before augeas.Set(), to prevent changing the file when it is not needed.
```
purpleidea: The comments after struct members will get used automatically in the godoc which is great! For extremely long comments, I *think* you can have them right above the struct member in question (and on multiple lines) and they should still show up. Have a quick look into this if you can.

resources/augeas.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	if obj.Lens != "" && !strings.HasSuffix(obj.Lens, ".lns") {
+		return fmt.Errorf("Lens should have a .lns suffix.")
+	}
+	return obj.BaseRes.Validate()
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+// Taken from the File resource.
+// FIXME: DRY - This is taken from the file resource
```
purpleidea: Do you want to use the *exact* same code as the file resource? I think using the recwatch module is pretty good as is. We're also able to do composite resources, although I'm not convinced this is what we want. We should talk/think about this more. In particular the case where you want to augeas a file, but also pick say a file mode or something.

resources/augeas.go
```diff
@@ -0,0 +1,249 @@
...
+		if send {
+			send = false
+			obj.Event(processChan)
+		}
+	}
+}
+
+// CheckApply method for Augeas resource.
+func (obj *AugeasRes) CheckApply(apply bool) (checkOK bool, err error) {
+	log.Printf("%s[%s]: CheckApply: %s", obj.Kind(), obj.GetName(), obj.File) // attempting to watch...
```
purpleidea: Not attempting to watch... That happens in Watch()

resources/augeas.go
```diff
@@ -0,0 +1,249 @@
...
+	opts := augeas.None
+	if obj.Lens != "" {
```
purpleidea: Can you add a small comment here explaining this part for augeas dummies like me?

resources/augeas.go
```diff
@@ -0,0 +1,249 @@
...
+		log.Printf("%s[%s]: NOLOAD NOWAY: %s", obj.Kind(), obj.GetName(), obj.File) // attempting to watch...
```
purpleidea: s/NOLOAD NOWAY/Augeas: Some message/

resources/augeas.go
```diff
@@ -0,0 +1,249 @@
...
+		opts = augeas.NoModlAutoload                                                // if the lens is specified, we can speed up augeas by not loading everything.
+	}
+
+	// Initiate augeas
```
purpleidea: nit, short comments are all lcase with the exceptions of special words like YAML. Long comments use proper sentences and punctuation.

resources/augeas.go
```diff
@@ -0,0 +1,249 @@
...
+			// state is not okay, no work done, exit, but without error
+			if apply {
+				if seterr := ag.Set(fullpath, value); seterr != nil {
+					return false, errwrap.Wrapf(agerr, "Error while setting augeas value!")
+				}
+			}
+		}
+	}
+
+	if !checkOK && apply {
```
purpleidea: I'd recommend you stop using the named return values to set and test and instead just flow through the function normally... Doing so will allow you to inline some of the above code and so it won't be as nested. Eg: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow (also applies to returning early)

resources/augeas.go
roidelapluie: Should not have been commited :)

resources/augeas.go
roidelapluie: What do you mean? In this line I need both conditions

resources/augeas.go
roidelapluie: done

resources/augeas.go
roidelapluie: Fixed

resources/augeas.go
roidelapluie: Same code is used in password.go as well

resources/augeas.go
roidelapluie: done

resources/augeas.go
roidelapluie: I indented less

resources/augeas.go
purpleidea: @roidelapluie What I meant is that we might be able to refactor this at some point. We can discuss another time.

resources/augeas.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// AugeasRes is a resource that enables you to use the augeas resource.
+// Currently only allows you to change simple files (e.g sshd_config).
+type AugeasRes struct {
+	BaseRes `yaml:",inline"`
+
+	// File is the path to the file targetted by this resource. If specified, mgmt will watch that file. Otherwhise,
+	// mgmt will poll the complete resource.
+	// TODO: implement polling
```
purpleidea: No need to implement polling. This is provided by our engine.

resources/augeas.go
```diff
@@ -0,0 +1,271 @@
...
+func init() {
+	gob.Register(&AugeasRes{})
+}
+
+// AugeasRes is a resource that enables you to use the augeas resource.
+// Currently only allows you to change simple files (e.g sshd_config).
+type AugeasRes struct {
+	BaseRes `yaml:",inline"`
+
+	// File is the path to the file targetted by this resource. If specified, mgmt will watch that file. Otherwhise,
```
purpleidea: Wrap this to about 80 chars please and remove the polling mention.

resources/augeas.go
purpleidea: This doesn't look fixed. I'd expect your resource to look like this...

https://github.com/purpleidea/mgmt/blob/master/docs/resource-guide.md#example-4

resources/augeas.go
purpleidea: See https://github.com/purpleidea/mgmt/blob/master/docs/resource-guide.md#example-4 or ping me in IRC if you have more questions...

purpleidea:
Okay, finally did a more thorough review, hope you don't mind see it as a patch. https://github.com/purpleidea/mgmt/tree/hack/augeas-review

Please "steal" anything you want from this assuming you agree and rebase into your patch.

Thanks!

roidelapluie:
@purpleidea care to review? I integrated your patch

resources/augeas.go
```diff
@@ -0,0 +1,309 @@
...
+		}
+	}
+}
+
+// checkApplySet runs CheckApply for one element of the AugeasRes.Set
+func (obj *AugeasRes) checkApplySet(apply bool, ag *augeas.Augeas, set AugeasSet) (checkOK bool, err error) {
+	checkOK = true
+	fullpath := fmt.Sprintf("/files/%v/%v", obj.File, set.Path)
+
+	if getValue, _ := ag.Get(fullpath); set.Value == getValue {
```
purpleidea: I think this should be

`if getValue, _ := ag.Get(fullpath); set.Value == getValue && err == nil {`

followed by:

`else if err != nil {`
return false, errwrap(err, ....)

resources/augeas.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	if !apply {
+		// If noop, we can return here directly. We return with
+		// nil even if err is not nil because it does not mean
+		// there is an error.
+		return false, nil
+	}
+
+	if err = ag.Set(fullpath, set.Value); err != nil {
```
purpleidea: `err :=`

resources/augeas.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	// Initiate augeas
+	ag, err := augeas.New("/", "", opts)
+	if err != nil {
+		return false, errwrap.Wrapf(err, "augeas: error while initializing")
+	}
+	defer ag.Close()
+
+	checkOK = true
```
purpleidea: Please move this to right before it is actually used (in the Sets range)

resources/augeas.go
```diff
@@ -0,0 +1,309 @@
...
+
+	for _, set := range obj.Sets {
+		if setCheckOK, err := obj.checkApplySet(apply, &ag, set); err != nil {
+			return false, errwrap.Wrapf(err, "augeas: error during CheckApply of one Set")
+		} else if !setCheckOK {
+			checkOK = false
+		}
+	}
+
+	if checkOK || !apply {
```
purpleidea: Be explicit
`return checkOK, nil`

purpleidea: A small comment here wouldn't hurt either eg: (`if state is correct or we can't apply, return early`)

resources/augeas.go
```diff
@@ -0,0 +1,309 @@
...
+		return
+	}
+
+	log.Printf("%s[%s]: changes needed, saving", obj.Kind(), obj.GetName())
+	if err = ag.Save(); err != nil {
+		return false, errwrap.Wrapf(err, "augeas: error while saving augeas values")
+	}
+
+	// FIXME: Workaround for https://github.com/dominikh/go-augeas/issues/13
```
purpleidea: I thought I mentioned this: Use os.Exists to check file state instead of this recursive thing.

resources/augeas.go
roidelapluie: Nope. I added a comment about this:

```
   // We do not check for errors because errors are also thrown when
   // the path does not exist.
```

resources/augeas.go
roidelapluie: I removed the named return then

resources/augeas.go
roidelapluie: Removed the named return ; added  comment

resources/augeas.go
roidelapluie: Fixed.

PR #131, opened 2017-02-09
macOS testing support

NeuralSandwich:
Implementing basic support for macOS.

-[x] make deps now installs libvirt support through homebrew
-[ ] graph0 successfully runs

### DIFF
--- a/test/shell/t1.sh
+++ b/test/shell/t1.sh
@@ -9,7 +9,7 @@
 set -o errexit
 set -o pipefail
 
-timeout --kill-after=3s 1s ./mgmt --help # hello world!
+$timeout --kill-after=3s 1s ./mgmt --help # hello world!
 pid=$!
 wait $pid	# get exit status
 exit $?
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -14,13 +14,14 @@ sudo_command=$(which sudo)
 YUM=`which yum 2>/dev/null`
 DNF=`which dnf 2>/dev/null`
 APT=`which apt-get 2>/dev/null`
+BREW=`which brew 2>/dev/null`
 
 # if DNF is available use it
 if [ -x "$DNF" ]; then
 	YUM=$DNF
 fi
 
-if [ -z "$YUM" -a -z "$APT" ]; then
+if [ -z "$YUM" -a -z "$APT" -a -z "$BREW" ]; then
 	echo "The package managers can't be found."
 	exit 1
 fi
@@ -34,6 +35,10 @@ if [ ! -z "$APT" ]; then
 	$sudo_command $APT install -y libvirt-dev || true
 	$sudo_command $APT install -y libaugeas-dev || true
 	$sudo_command $APT install -y libpcap0.8-dev || true
+fi
+
+if [ ! -z "$BREW" ]; then
+	$BREW install libvirt || true
 fi
 
 if [ $travis -eq 0 ]; then
--- a/test/shell/file-move-upper-dir.sh
+++ b/test/shell/file-move-upper-dir.sh
@@ -7,7 +7,7 @@ exit 0
 mkdir -p /tmp/mgmt/a/b/c/
 
 # run empty graph, with prometheus support
-timeout --kill-after=20s 15s ./mgmt run --tmp-prefix --yaml=examples/deep-dirs.yaml &
+$timeout --kill-after=20s 15s ./mgmt run --tmp-prefix --yaml=examples/deep-dirs.yaml &
 pid=$!
 sleep 5s	# let it converge
 
--- a/test/shell/prometheus-1.sh
+++ b/test/shell/prometheus-1.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 # run empty graph, with prometheus support
-timeout --kill-after=20s 15s ./mgmt run --tmp-prefix --prometheus &
+$timeout --kill-after=20s 15s ./mgmt run --tmp-prefix --prometheus &
 pid=$!
 sleep 5s	# let it converge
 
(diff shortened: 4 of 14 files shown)


roidelapluie:
I guess it won't work with mac due to its usage of inotify, libvirt, systemd ...

NeuralSandwich:
libvirt supports macOS. I'd like to be able to run mgmt from a mac and configure remote machines. It would also be nice to support some basic operations on a mac as well.

For inotify, macOS has FSSEvents API. Systemd there is launchd/launchctl. This is probably going to be a very slow process but even it is just the use case where running mgmt to setup an initial master node to do the setup would be good.

ffrank:
LMK when there is anything testible, I will try with my work machine.

NeuralSandwich:
@ffrank Will do, currently I am getting:

> actions.go:564: File[file1]: Watch errored: Unknown File[file1] watcher error: Unknown error: lstat /tmp/mgmt-hello-world: no such file or directory

Then everything hangs. Also sending SIGINT, will hang as well.

> 07:35:26 etcd.go:746: Trace: Etcd: CbLoop: Event: FinishLoop
^C07:36:26 cli.go:133: Interrupted by ^C
07:36:26 main.go:548: Destroy...
07:36:26 main.go:394: Main: GAPI exited
07:36:26 main.go:379: Main: Waiting...
07:36:26 actions.go:632: Noop[noop1]: Exited

purpleidea:
@NeuralSandwich Keep in mind if the remote machines are GNU/Linux you don't need the MacOSX support directly. In any case, I think we can accept patches for the alternate features with build tags. Eg: launchd could be an alternative svc implementation.

purpleidea:
> Then everything hangs. Also sending SIGINT, will hang as well.

This is probably an indication of one or two things:

1) Could be an unresolved bug that I'm working on fixing atm.
2) Could be a problem with the resource on OSX. Resources can cause the engine to hang if they don't shutdown cleanly. This might be OSX specific. Hopefully this article helps: https://ttboj.wordpress.com/2016/02/15/debugging-golang-programs/

purpleidea:
@roidelapluie 
RE: This...
> Please use $timeout not $TIMEOUT. Capitalized varnames should be reserved for bash internals.

Is it wrong to use capitals for large constants which are static without the script? I've done it this way in the past, is there a source that says this is wrong?

roidelapluie:
Well TMOUT is a bash internal. Confusing.

roidelapluie:
http://unix.stackexchange.com/questions/42847/are-there-naming-conventions-for-variables-in-shell-scripts#comment58850_42847

test/shell/t1.sh
```diff
@@ -9,7 +9,13 @@
 set -o errexit
 set -o pipefail
 
-timeout --kill-after=3s 1s ./mgmt --help # hello world!
+if [[ $(uname) == "Darwin" ]] ; then
```
purpleidea: why don't we put this code in the util.sh script and then we get it everywhere for free?
Then we could actually have it be a $TIMEOUT (in caps) to be used everywhere.

NeuralSandwich: Could do, I didn't want to just go ahead and shove it into shared code. Especially since the test are still failing.

purpleidea: It is shared code, and *you're* part of it!

NeuralSandwich: Moved the detection for timeout to util/shell. There is a slight issue, with it living in test/util.sh individually test in test/shell/ cannot be run without $timeout being defined.

test/test-shell.sh
```diff
@@ -50,7 +50,7 @@ for i in $DIR/test/shell/*.sh; do
 			[ -n "${failures}" ] && echo "$failures" && echo "$LINE"
 			echo "Script: $ii"
 			# if we see 124, it might be the exit value of timeout!
-			[ $e -eq 124 ] && echo "Exited: $e (timeout?)" || echo "Exited: $e"
+			[ $e -eq 124 ] && echo "Exited: $e ($timeout?)" || echo "Exited: $e"
```
purpleidea: Looks like a lazy search/replace here :P

purpleidea:
@NeuralSandwich 

RE:

> Moved the detection for timeout to util/shell. There is a slight issue, with it living in test/util.sh individually test in test/shell/ cannot be run without $timeout being defined.

I didn't understand this comment sorry... Can you rephrase please?
Also, your two commits should be squashed into one and rebased to current git master.
LMK if you're not sure how to do this.

Thanks!

test/test-shell.sh
NeuralSandwich: Oops yeah, I'll blame my cat distracting me.

NeuralSandwich:
@purpleidea So by using ```$timeout``` for the shell tests (which is defined in ```test/util.sh```) these tests cannot be run by directly (```bash test/shell/t1.sh```). They will always have to be run by ```test/test-shell.sh```, since that script includes ```test/util.sh``` which defines $timeout.

Just making a note of it as this changes their usability. These tests probably aren't called individually so probably isn't an issue but ¯\\\_(ツ)_/¯.

Sorry about not rebasing, I forgot to pull from upstream. Squishing these commits into might be a little confusing since each is mostly separate but I am happy to be inventive in the commit message.

purpleidea:
@NeuralSandwich Sorry you can keep them as separate commits!

purpleidea:
@NeuralSandwich You just need to rebase and rename the commit message of the first one so it passes CI.

purpleidea:
@NeuralSandwich Merged! Thanks :)

Let us know if you'd like something else to work on! There's tons of great stuff to do.
https://github.com/purpleidea/mgmt/commit/610202097aeaaac0c30d2ae6c0422b3750edefce

ffrank:
I think we should source `util.sh` from any script that needs things defined in there.

PR #132, opened 2017-02-10
vagrant: add Vagrantfile and some basic configuration

ffrank:


Vagrantfile
```diff
@@ -0,0 +1,43 @@
+
```
roidelapluie: newlines?

vagrant/bashrc.mgmt
```diff
@@ -0,0 +1,5 @@
+
```
roidelapluie: newline?

vagrant/bashrc.mgmt
```diff
@@ -0,0 +1,5 @@
...
+# alias for simple access
+alias cdrepo='cd $GOPATH/src/github.com/purpleidea/mgmt'
+
+export GOPATH=~/go
```
roidelapluie: I would call this file mgmt.bashrc, not bashrc.mgmt

misc/make-deps.sh
```diff
@@ -2,6 +2,8 @@
 # setup a simple go environment
 XPWD=`pwd`
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )"	# dir!
+VAGRANT_MODE=
```
purpleidea: You should have this ="" for example. So bash with unset var warnings won't happen IIRC.

vagrant/bashrc.mgmt
purpleidea: Can you change to 'gopath' please.

vagrant/motd
```diff
@@ -0,0 +1,14 @@
+Welcome to the mgmt hacking VM.
+
+Please make sure that your authorized key for GitHub is added to the SSH agent
+on the host machine (see `ssh-add -l`).
+
+You can jump into the mgmt repo with the alias `cdrepo`.
+
+To get started, try:
+
+  cdrepo
```
purpleidea: cute idea. Nit: let's name this cdmgmt if you want to have an alias.

ffrank: Hah check. That is actually the alias I use on my laptop :fist_right: :fist_left:

purpleidea:
@ffrank I'm under the impression that in the vagrant environment *everything* should run as root to make everyone's life easier...

ffrank:
I kind of like that you can run unprivileged by default, so that you get some feedback on what does and does not Just Work. Of course, in the absence of DBUS and the ability to get root ad hoc for resources like `pkg`, this is less fun in the vagrant environment.

OK, I'll probably go for root then.

purpleidea:
@ffrank I actually don't disagree, I was more trying to avoid making the make-deps script more complicated. Up to you!

purpleidea:
... I'd actually rather just duplicate some of the deps code in the Vagrantfile directly, rather than go crazy about code reuse of make-deps.sh

ffrank:
Good point, I was thinking about how to do this best...and yes, complexity is not ideal. But I really want to avoid repitition.
First idea was to rip the make-deps.sh into two scripts, and use both from `make deps`, but maybe only one for vagrant. That might be cleaner after all.

purpleidea:
@ffrank Just ping me when ready for a new review since GH is lame and doesn't notify when patch is updated.

ffrank:
Pushed a rebased and adapted version, incorporating all the feedback. Thanks for that.

I still build the environment for the vagrant user. root would just feel weird for coding and interacting with github.

purpleidea:
@ffrank Thanks for the update! I know a few folks who are excited to have this upstream!

A few things... We already have a commit message check, sorry: https://github.com/purpleidea/mgmt/blob/master/test/test-commit-message.sh (patches welcome if you think it needs improvements) -- Actually it only works on travis, so adding the `git log` fallback would be a welcome patch (in a separate commit)

Secondly, as per: https://github.com/purpleidea/mgmt/pull/132#issuecomment-278908556 I think it's better to keep the make-deps.sh script as is, and just copy the relevant parts into the vagrantfile if needed. I think splitting this up isn't necessary for now, sorry.

ffrank:
Sorry, the commit check was WiP I abandoned when I realized @roidelapluie had built that already. Added it to this commit on accident.

Should be fine now.

purpleidea:
Rebased and merged!

PR #134, opened 2017-02-10
spec: Fix rpmlint warnings

ashcrow:
Minor fixes for ``rpmlint`` errors:

- Switched spces to tabs
- Added quiet switch to setup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ $(SRPM): $(SPEC) $(SOURCE)
 #
 $(SPEC): rpmbuild/ spec.in
 	@echo Running templater...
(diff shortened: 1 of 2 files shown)


purpleidea:
@ashcrow Thanks for your patch!

Is the tabs->spaces something that is required or suggested by rpmlint? If not, I think we're firmly in the tabs camp :P 

Cheers

ashcrow:
Nope, not firmly. It was due to there being a mix of tabs and spaces. I can switch them over to all tabs.

purpleidea:
@ashcrow w00t, merged! Want to help polish the rest of this RPM off?

Thanks!
James

PR #136, opened 2017-02-11
Hack/augeas review

roidelapluie:
@purpleidea this is to discuss

### DIFF
--- a/resources/augeas.go
+++ b/resources/augeas.go
@@ -1,0 +1,308 @@
+// Mgmt
+// Copyright (C) 2013-2016+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


resources/augeas.go
```diff
@@ -230,6 +236,7 @@ func (obj *AugeasRes) CheckApply(apply bool) (checkOK bool, err error) {
 		return false, errwrap.Wrapf(err, "augeas: error while saving augeas values")
 	}
 
+	// XXX: recurse? not a good idea. Instead test for file existence manually.
```
roidelapluie: recurse with apply = False. So it will never be called by the function that uses it.

This will not only catch if the file does not exist, but it will also catch other problems. e.g.:

```
set:
  - ['ForwardX11', 'no']
  - ['ForwardX11', 'yes']
```

That will be detected by this recurse call as an error. it is a great way to detect that the resource has been applied successfully. I am not sure I want to remove that even if the upstream bug was fixed.

Issue #141, opened 2017-02-14
Improved YAML format


mildred:
Until we have the perfect DSL, I would suggest to improve the YAML format, and I have one idea for that:

```
version: 2
graph: mygraph
resources:
  - name: file1
    autoedge: true
    retry: -1
    # the presence of the file key denotes a file resource
    file:
      path: /tmp/mgmt/file1
      content: |
        Content of my file
      owner: root
  - name: docker
    # The presence of a svc key denotes a svc resource
    svc:
      state: running
      startup: enabled
    after: # list of resource it depends on, this create edges
      - file file1 # format: <resource kind><space><resource name>
    before: # list of resources that depends on this
      - noop checkpoint
  - name: checkpoint
    noop: true
edges:
  # additional edges just as before
```

The implementation of this (which I started) is even simpler because it doesn't need the reflect package in `yamlgraph/gconfig.go` in function `NewGraphFromConfig()`. The `resources` key already contains the list.

roidelapluie:
I would rather propose:

```
  - name: file1
    autoedge: true
    retry: -1
    kind: file
    params:
      path: /tmp/mgmt/file1
      content: |
        Content of my file
      owner: root
    metaparams:
```

roidelapluie:
But this is great! Even if I would rater prefer us to focus on the DSL instead of this implementation point.

mildred:
There is just one catch with this format: a resource name cannot be one of `name`, `before`, `after`, or one of the meta parameters (`autoedge`, `autogroup`, `retry`, `delay`, `poll`, `limit`, `burst`). Except for `noop` which can be a special case. if this restriction is not wanted, the syntax suggested by @roidelapluie is a nice alternative.

mildred:
I admit I'm biased towards yaml. This is one thing that I love in ansible. Clear syntax, easy to learn, machine processable with any language...

roidelapluie:
I like yaml too, and as it is our interface with mgmt this makes sense to improve it.

purpleidea:
I'm not explicitly against changing the YAML format, but as I've mentioned, it is only a crutch for me until we have the language, and so I don't plan to spend any cycles on this. If someone feels *really* strongly about this, and the new format doesn't have any limitations, then patches welcome I suppose. But I'm going to close this issue.

roidelapluie:
Please @purpleidea stop closing issues all the time.

purpleidea:
[...]

PR #143, opened 2017-02-14
yamlgraph: Implement easy writing resource structure

mildred:
Discussed in #141

Add a version tag to the yaml graph. Version 2 has a new structure where
resources are no longer grouped by their kind. Resources can also reference
other resources in a before and after list, which will create edges. It is
no longer necessary to have the edges specified separately.

Version 1 (or no version) is still compatible, though it has been
refactored to avoid use of the reflect package, and use an extensible list
of registred resource kinds. This also has the benefit of removing the
empty VirtRes and AugeasRes struct types when compiling without libvirt and
libaugeas.


resources/resources.go
```diff
@@ -39,6 +39,20 @@ import (
 	"golang.org/x/time/rate"
 )
 
+var registredResources map[string]func() Res = map[string]func() Res{}
```
purpleidea: typo registered*

mildred:
I completely understand, especially since you explicitly said you didn't really want to focus on YAML right now.

purpleidea:
On Tue, Feb 14, 2017 at 3:37 PM, mildred <notifications@github.com> wrote:
>
> I completely understand, especially since you explicitly said you didn't really want to focus on YAML right now.

This isn't a NACK, rather a "blocked until feat/mgmt gets merged"
(hopefully this week) and a "needs ACK/review" from someone else. Ping
in channel for a review... :)

roidelapluie:
I will review that later this week. This is big.

mildred:
After using this a few days, It may still not be ready to merge. Changing the format should not be done lightly. I still think an improved YAML format could be grat to have, a sort of mid-ground between the graph dump we currently have and the full-fledged DSL. But I am no longer sure about this version of the format I wrote.

Still, there are things included in this PR that I think would be great to include. The part that parses the current format differently, getting rid of the virt_disabled.go and augead_disabled.go files, and stop using the reflect package for parsing.

Basically, it boils down to the code in `gconfig.go` in this PR, especially:

- `func (c *GraphConfig) UnmarshalYAML(...)`
- `func (r *ResourceV1) UnmarshalYAML(...)`
- `func (r *ResourceV1) Decode(kind string) (err error)`

This could be presented as a separate PR, and I think it could help with conditional compilation of augeas and libvirt.

purpleidea:
@mildred Okay, I had another quick look, and while I don't want to merge all of this, there are a few parts I would like to look into:

1) Using the resources without the explicit list and the reflection. That might be nice on its own.

2) Implementing this as a different GAPI. This means it can stand alone without needing to have a V1 vs. V2 flag. The API: https://github.com/purpleidea/mgmt/blob/master/gapi/gapi.go#L44 Yamlgraph example: https://github.com/purpleidea/mgmt/blob/master/yamlgraph/gapi.go#L31

(You could have like yamlgraphv2 or something...)

3) Lastly, the refactor I did of Init which isn't merge in git master needs to go in first :( Sorry about that. I hate when I'm the blocker.

The first two points can be worked on if you're willing. Having the hardcoded list in particular is a dumb thing that I did.

mildred:
This must be closed in favor of #163. Just have to find time for it (and for real this time)

PR #144, opened 2017-02-14
github: Add ISSUE_TEMPLATE

roidelapluie:
Let's add a ISSUE_TEMPLATE so people know that github issue is not the
preferred place to start discussions.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

PR #147, opened 2017-02-14
github: Show a friendly message in issues template

purpleidea:


purpleidea:
@roidelapluie How about this instead of #144 ?

roidelapluie:
why did you remove the header? My point is that if you close any non-bug issue, it is better to warn the users upfront.

purpleidea:
@roidelapluie Oh. That's what you were getting at...

PR #149, opened 2017-02-14
resources: file: MkdirAll parent directories

mildred:
File resource fail if their parent directories is not there. Have it create them all if required. Existing autoedges to them makes it possible to customize them if needed (owner, permissions, ...)

### DIFF
--- a/resources/file.go
+++ b/resources/file.go
@@ -58,6 +58,7 @@ type FileRes struct {
 	Group      string  `yaml:"group"`
 	Mode       string  `yaml:"mode"`
 	Recurse    bool    `yaml:"recurse"`
+	Parents    bool    `yaml:"parents"` // create parent directories?
 	Force      bool    `yaml:"force"`
 	path       string  // computed path
 	isDir      bool    // computed isDir
@@ -413,6 +414,13 @@ func (obj *FileRes) fileCheckApply(apply bool, src io.ReadSeeker, dst string, sh
 	}
 	if obj.debug {
 		log.Printf("fileCheckApply: Apply: %s -> %s", src, dst)
+	}
+
+	if obj.Parents {
+		err = os.MkdirAll(filepath.Dir(dst), os.ModePerm)
+		if err != nil {
+			return sha256sum, false, err
+		}
 	}
 
 	dstClose() // unlock file usage so we can write to it
--- a/test/shell/t2.sh
+++ b/test/shell/t2.sh
@@ -16,5 +16,6 @@ test -e /tmp/mgmt/f1
 test -e /tmp/mgmt/f2
 test -e /tmp/mgmt/f3
 test ! -e /tmp/mgmt/f4
+test -e /tmp/mgmt/a/long/long/path/f5
 
 exit $e
--- a/test/shell/t2.yaml
+++ b/test/shell/t2.yaml
@@ -24,6 +24,12 @@ resources:
     content: |
       i am f4 and i should not be here
     state: absent
+  - name: file5
+    path: "/tmp/mgmt/a/long/long/path/f5"
+    parents: true
+    content: |
+      i am f5
+    state: exists
 edges:
 - name: e1
   from:


purpleidea:
@mildred File resource *should* fail if parent dirs aren't there.
So should Dir resource (a file resource with trailing slash).

Mkdir -p like behaviour will happen if you use Force parameter. Is this what you wanted or is there still something missing?

mildred:
The force parameter doesn't create parent directories if they are not there, and there are some cases where you want this behaviour, but perhaps not by default (although I would consider it a sane default)...

I hesitate to use the force parameter for that, can it have another meaning that could overlap or cause harm (think you had a file with `force: true` and you no longer need it. You set it to `state: absent` but keep the force parameter)? I would much prefer `parents: true` for instance.

Should I change this PR to trigger this with a special parameter?

purpleidea:
> although I would consider it a sane default

So I disagree here, but I want to explain why, and get you thinking about it, and either agreeing or challenging it!

Safety is meant as a major cornerstone of the project. In other words, if we weren't explicit about something, better to fail with an error, and force the programmer to update their code, rather than being "helpful" (eg mkdir -p) and letting them get past it. This is because long term, that silent "helpfulness" might be masking an issue.

Do you still want to merge something here, or is it all covered in your other PR? LMK :)

roidelapluie:
Creating parent dirs would not be good IMO.

Interesting discussions: https://projects.puppetlabs.com/issues/86

roidelapluie:
( comment 16 https://projects.puppetlabs.com/issues/86#note-16 )

( comment 24 https://projects.puppetlabs.com/issues/86#note-24 )

mildred:
Ok, I agree it should not be the default, but there is currently no option to mkdir parent directories at all. What do you think of a `parents: true` property? Should I implement mkdirs conditionned to that property?

Or perhaps I missed something. I tried with `force: true` and it doesn't work.

mildred:
From https://projects.puppetlabs.com/issues/86 it seems that most of the inconsistency would come from the permissions of the automatically created directories that might be unpredictable and non consistent across hosts.

I suggest that if we enable the automatic creation of parents behind an option, the parent directories would always get the same permission set. Owner and group not modified (mgmt user) and file mode should always be `0777 & ~umask` where `umask` is the umask of the mgmt process, which should always be the same.

It doesn't solve anything. If we run the same graph on two hosts, with on in which a parent directory is automatically created and not the other, we possibly do not get the same permissions for that directory on the two hosts.

I don't see this as a problem, but rather an interesting feature: We prevented the graph from failing on the host where the automatic creation happened. We possibly prevented a service downtime or something of that order. But I can concieve that it might not always be good to silently fix a problem like that. We could have also added a security issue on the host where we created the directory...

It's an interesting problem indeed...

roidelapluie:
this is not an interesting features , it brings uncertainty. I do not want to offend you but those who want that do not understand config management.

What will happen with:

```
file:
  path: /etc/http/vhostd/foo.conf
```

vs

```
file:
  path: /etc/http/vhost.d/foo.conf
```


How much time can you lose with that typo if mgmt creates the parents for you? To me, that is a feature we simply do not want in mgmt.

IMO, if you want that, just create an exec or specify all the parents directories.

mildred:
Not even behind a flag? If so, let's close this PR.

purpleidea:
A few comments...

1) @roidelapluie It was cool to see the old puppet discussions! Maybe this is a warning for what's coming here. I hope I make the right decisions!

2) @mildred While old puppet discussions are interesting for comparison; we're NOT puppet, so we will make some choices which are different, and some which are the same, but for different reasons.

3) It is true however that automatically making parent dirs on a FILE (not a dir) is a dangerous thing. This is ACK for dirs with force, but not files.

mildred:
I have not received an answer for this comment I made:

> I hesitate to use the force parameter for that, can it have another meaning that could overlap or cause harm (think you had a file with `force: true` and you no longer need it. You set it to `state: absent` but keep the force parameter)? I would much prefer `parents: true` for instance.

That's why I changed the PR to use a `parents` parameter instead.

Also, @purpleidea if you ACK creation of parent directories in case of directory resources (files that end with /), what should be the mode of those directories ?

- `0777 & ~umask`: predictable and always the same
- the same mode as the directory being created, defaulting to `0777 & ~umask` if not specified

I have a local commit that does this on top of #151 once it is merged.

purpleidea:
@mildred 

RE: 
>    I hesitate to use the force parameter for that, can it have another meaning that could overlap or cause harm (think you had a file with force: true and you no longer need it. You set it to state: absent but keep the force parameter)? I would much prefer parents: true for instance.

If the force did more than one thing in this context, I would be tempted to agree, but AFAICT it doesn't, therefore the force is unambiguous and is just named differently from parents: true. Right?

> Also, @purpleidea if you ACK creation of parent directories in case of directory resources (files that end with /), what should be the mode of those directories ?

    0777 & ~umask: predictable and always the same
    the same mode as the directory being created, defaulting to 0777 & ~umask if not specified

Good question! We should decide what we think is best, why, after comparing with the existing systems as well (for fun) and add a test case confirming our decision.

roidelapluie:
Ok this can be a feature if it really makes you happy......

😢

mildred:
> If the force did more than one thing in this context, I would be tempted to agree, but AFAICT it doesn't, therefore the force is unambiguous and is just named differently from parents: true. Right?

Force is used to transform a file resource to a directory and the contrary. It allows to delete a file or directory. We might want to have a resource that we don't want to force but want to have its parent directories created. I would say this conflicts.

purpleidea:
> Force is used to transform a file resource to a directory and the contrary. It allows to delete a file or directory. We might want to have a resource that we don't want to force but want to have its parent directories created. I would say this conflicts.

@mildred So in my time writing Puppet, I can't remember being in a situation where I wanted to make sure of one case, but was afraid that the other could happen by accident. Having said that, it doesn't mean there isn't one!

If you can present a good example of an existing Puppet module where this functionality would have made things much safer, please LMK and I'll reconsider.

Hope that sounds fair!

mildred:
It makes sense to only allow MkdirAll on directories and not simple files, and you did that when you merged #151, It seems enough to me.

PR #151, opened 2017-02-14
resources: file: Allow creation of empty directories

mildred:
I could not get mgmt to create empty directories with the file resource, I believe this was not implemented hence this PR. If it is indeed implemented, the test and the doc should probably be updated.

### DIFF
--- a/resources/file.go
+++ b/resources/file.go
@@ -443,6 +443,49 @@ func (obj *FileRes) fileCheckApply(apply bool, src io.ReadSeeker, dst string, sh
 	return sha256sum, false, dstFile.Sync()
 }
 
+// dirCheckApply is the CheckApply operation for an empty directory
+func (obj *FileRes) dirCheckApply(apply bool) (bool, error) {
+	// Check if the path exists and is a directory
+	st, err := os.Stat(obj.path)
+	if err != nil && !os.IsNotExist(err) {
+		return false, errwrap.Wrapf(err, "Error checking file resource existence")
+	}
+
+	if err == nil && st.IsDir() {
+		return true, nil // Already a directory, nothing to do
+	}
+	if err == nil && !st.IsDir() && !obj.Force {
+		return false, fmt.Errorf("Can't force file into dir: %s", obj.path)
+	}
+
+	if !apply {
+		return false, nil
+	}
+
+	// The path exists and is not a directory
+	// Delete the file if force is given
+	if err == nil && !st.IsDir() {
+		log.Printf("dirCheckApply: Removing (force): %s", obj.path)
+		if err := os.Remove(obj.path); err != nil {
+			return false, err
+		}
+	}
+
+	// Create the empty directory
+	var mode os.FileMode
+	if obj.Mode != "" {
+		mode, err = obj.mode()
+		if err != nil {
+			return false, err
+		}
+	} else {
+		mode = os.ModePerm
+	}
+
+	err = os.Mkdir(obj.path, mode)
+	return false, err
+}
+
 // syncCheckApply is the CheckApply operation for a source and destination dir.
 // It is recursive and can create directories directly, and files via the usual
 // fileCheckApply method. It returns checkOK and error as is normally expected.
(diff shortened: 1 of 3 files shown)


resources/file.go
```diff
@@ -443,6 +443,46 @@ func (obj *FileRes) fileCheckApply(apply bool, src io.ReadSeeker, dst string, sh
 	return sha256sum, false, dstFile.Sync()
 }
 
+// dirCheckApply is the CheckApply operation for an empty directory
+func (obj *FileRes) dirCheckApply(apply bool) (bool, error) {
+	// Check if the path exists and is a directory
+	st, err := os.Stat(obj.path)
+	if err != nil && !os.IsNotExist(err) {
+		return false, err
```
purpleidea: Add errwrap please

resources/file.go
```diff
@@ -443,6 +443,46 @@ func (obj *FileRes) fileCheckApply(apply bool, src io.ReadSeeker, dst string, sh
...
+
+	// Create the empty directory
+	mode := os.ModePerm
+	if obj.Mode != "" {
+		mode, err = obj.mode()
+		if err != nil {
+			return false, err
+		}
+	}
+	err = os.MkdirAll(obj.path, mode)
```
purpleidea: When creating a dir, think of it as Force allows us to MkdirAll where as without force we're only allowed to Mkdir (no automatic parents).

This is safer, because you want to explicitly say you want parents or not.

roidelapluie: We should never use MkdirAll (see #149 )

resources/file.go
mildred: Not sure errwrap can bring much extra information here...

resources/file.go
purpleidea: No, I think MkdirAll is okay with force, and without we use Mkdir. @mildred LMK

mildred:
@purpleidea  and now?

resources/file.go
```diff
@@ -443,6 +443,49 @@ func (obj *FileRes) fileCheckApply(apply bool, src io.ReadSeeker, dst string, sh
 	return sha256sum, false, dstFile.Sync()
 }
 
+// dirCheckApply is the CheckApply operation for an empty directory
+func (obj *FileRes) dirCheckApply(apply bool) (bool, error) {
+	// Check if the path exists and is a directory
+	st, err := os.Stat(obj.path)
+	if err != nil && !os.IsNotExist(err) {
+		return false, errwrap.Wrapf(err, "Error checking file resource existance")
```
purpleidea: existence*

resources/file.go
```diff
@@ -443,6 +443,49 @@ func (obj *FileRes) fileCheckApply(apply bool, src io.ReadSeeker, dst string, sh
...
+	if obj.Mode != "" {
+		mode, err = obj.mode()
+		if err != nil {
+			return false, err
+		}
+	} else {
+		mode = os.ModePerm
+	}
+
+	err = os.Mkdir(obj.path, mode)
```
purpleidea: Don't we want?

```
if obj.Force {
    return false, os.MkDirAll(obj.path, mode)
}

return false, os.Mkdir(obj.path, mode)
```

mildred: I planned to have this in #149, not directly here. And are we sure we want to mkdir all with the same unix permissions, or do we want the parent directories to have the mode `0777 & ~umask`?

purpleidea:
@mildred Hey! So I really wanted to merge your excellent patch, and so I did!

I ended up adding an additional patch with the mkdirall stuff. If you think it's wrong, we can back it out.

Lastly I patched a few nitpicks as well, basically due to capitalization. The general rule is: short one line comments aren't sentences so don't start with a capital, and don't end with a period. Doc strings, and longer comments are often sentences, so they do. Take this as friendly feedback for your next patch! Your code looks great!

Cheers
James

purpleidea:
Merged in: https://github.com/purpleidea/mgmt/commit/d0d62892c8ca386cd693fc21141d061ff030e87d
Small fixups by me in: https://github.com/purpleidea/mgmt/commit/d0d62892c8ca386cd693fc21141d061ff030e87d

mildred:
thanks

PR #153, opened 2017-02-15
misc: systemd unit: Split OPTS variable

mildred:
Systemd split variables when specified like `$VAR` and not when specified
like `${VAR}`. Since `OPTS` must contain multiple options, we must ensure
systemd will split it.

See also systemd.service(5)

### DIFF
--- a/misc/mgmt.service
+++ b/misc/mgmt.service
@@ -5,7 +5,7 @@ After=systemd-networkd.service
 Requires=systemd-networkd.service
 
 [Service]
-ExecStart=/usr/bin/mgmt run ${OPTS}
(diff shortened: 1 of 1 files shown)


mildred:
It would also be nice to provide a default value such as:

    Environment="OPTS=--yaml /etc/mgmt.yaml"

purpleidea:
@mildred I'm merging the first commit, but NACK-ing the second pending discussion.

purpleidea:
First patch merged in https://github.com/purpleidea/mgmt/commit/d170a523c3cbaa8df89e8ba8d16340420979bed9

mildred:
Ok, the second commit is not very important, we can close this PR

PR #154, opened 2017-02-15
documentation: Implement sphinx documentation

roidelapluie:
Licence moved to the bottom due to a read the docs bug.

To be used with Read The Docs.
DEMO: http://mgmtest.readthedocs.io/en/latest/


docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
purpleidea: 1) Can the Makefile stuff not be added to our main Makefile?

Otherwise like Makefile is:

> No newline at end of file

docs/Makefile
```diff
@@ -0,0 +1,20 @@
...
+SOURCEDIR     = .
+BUILDDIR      = _build
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
```
purpleidea: What does this do??

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+# directories to ignore when looking for source files.
+# This patterns also effect to html_static_path and html_extra_path
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store', 'venv']
+
+# The name of the Pygments (syntax highlighting) style to use.
+pygments_style = 'sphinx'
+
+# If true, `todo` and `todoList` produce output, else they produce nothing.
+todo_include_todos = False
+
```
purpleidea: double newline

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+# further.  For a list of options available for each theme, see the
+# documentation.
+#
+# html_theme_options = {}
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ['_static']
+
```
purpleidea: doubled

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+
+# -- Options for HTMLHelp output ------------------------------------------
+
+# Output file base name for HTML help builder.
+htmlhelp_basename = 'mgmtdoc'
+
```
purpleidea: doubled...

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+
+# -- Options for Texinfo output -------------------------------------------
+
+# Grouping the document tree into Texinfo files. List of tuples
+# (source start file, target name, title, author,
+#  dir menu entry, description, category)
+texinfo_documents = [
+    (master_doc, 'mgmt', u'mgmt Documentation',
+     author, 'mgmt', 'One line description of project.',
```
purpleidea: "Next Generation Config Mgmt" ?

docs/Makefile
roidelapluie: That Makefile is autogenerated by sphinx. I would not change it.

docs/Makefile
roidelapluie: That Makefile is autogenerated by sphinx. I would not change it.

docs/conf.py
roidelapluie: That file is autogenerated by sphinx. I would not change it.

docs/conf.py
roidelapluie: That file is autogenerated by sphinx. I would not change it.

docs/conf.py
roidelapluie: because it is before another section

docs/conf.py
roidelapluie: doubled because before another section

roidelapluie:
> Few small comments. Big one is: Why do we have to move the comment headers? I don't know what the bug is, but I think just moving the #mgmt underneath the comments should fix it.

Everything after the comment header is not rendered.

> Secondly, why are the TOC gone?

TOC is autogenerated by sphinx.

docs/Makefile
roidelapluie: What it does is when you call make X it will run sphinx X. Like make html will call sphix html.

purpleidea:
@roidelapluie 

> Everything after the comment header is not rendered.

Then let's just remove it completely.

So RE: removing the TOC, my concern is the following: This would then mandate the viewing through readthedocs, and takes away of the experience for someone just viewing the markdown file directly, eg through the GH page. Is there a way we can get around this and make it less bad for the native viewer?

Thanks

roidelapluie:
So, first, we are not forcing people with RTD. We use Sphinx, so everyone can build the doc locally, even in PDF.

Second, creating and maintaining a TOC is error prone and is a robot job, not a human job.

Third, I do not think that this is so bad for native viewers. They can still read the text on github.

Fourth, the main advantage of Sphinx is that the documentation is combined. It is not individual documents anymore, but a consistent mgmt manual.

purpleidea:
@roidelapluie Alright, I'll ACK it. Please just make the change to remove the header completely, since at the bottom it looks absurd.

purpleidea:
(Fair warning if anyone else has any preference for or against RTD, speak up now. eg: @ffrank or others

roidelapluie:
Fixed

PR #157, opened 2017-02-16
Future!

purpleidea:


purpleidea:
Hey @roidelapluie @ffrank and others...

So I was planning on merging this shortly. Looking for a shoddy review, and some help. Some notes:

1) This is part one of fixing some bad cruft I had in the engine. This introduces some different cruft, but is less bad overall, and takes us in the right direction.

2) If a resource fatally errors, then it will hang the engine. I have to fix this, but I think the worker core needs some rethinking about how we send signals, and so on. See the commit messages.

3) Graph transitions hang! Gah! I thought I fixed this, but there are more races.

So long story short, I'd like to build off of this branch, except I'm having a bit of tunnel vision and exhaustion from this branch, and wanted to move on to the language stuff for a bit, so if someone wants to look into this, please do!

roidelapluie:
@purpleidea to me, 2. is blocking and should prevent to merge to master.

purpleidea:
> @purpleidea to me, 2. is blocking and should prevent to merge to master.

@roidelapluie I actually agree, but here's the thing. Currently git master does some *different* things, which are bad, but aren't necessarily getting noticed.

For example, we call init TWICE for each resource, which is a big problem for more complex resources where this would init twice. This fixes that issue, but introduces others. etc... This now has a 1-1 mapping b/w resource init and close. Which virt resource needed for example to not double free.

ffrank:
I'll see if I can make some time over the weekend to dig into this one.

purpleidea:
I cherrypicked 3 obvious good commits onto master and removed them from here... Hopefully this makes things easier for now.

examples/virt4.yaml
```diff
@@ -3,7 +3,13 @@ graph: mygraph
 resources:
   virt:
   - name: mgmt4
+    meta:
```
roidelapluie: @purpleidea why is that needed Why not a virt5 example?

ffrank:
Throwing the towel at this review for now.

I don't think this mode of working will be very beneficial after all. Let me preface my thoughts with two observations:
1. @purpleidea writes good commit messages that confer intent.
2. At this stage of development, it's unavoidable that @purpleidea throws large swaths of code into master in order to get the project into a GA state.

Now, to put this kind of code review into some perspective: In Puppet, some core devs convene in a weekly hangout to look at what community contributors have submitted. There are at least two sets of eyes on it, with discussion on what they think of a given submission. Patches are usually only a few dozen lines of code, and even then it can be challenging to get an overview (despite being quite familiar with the code base). It's way more effective if the author can join the hangout and explain their reasoning.

Here, the tables are completely turned. The reviewers know little about the code base, we work on our own, and the patches are huge. The effort/benefit ratio is, er, not ideal in this scenario.

TL;DR I don't think it makes sense for outside contributors to try and review such monsters on our own. I guess @purpleidea is on his own with these.
It *would* be very helpful for the project if you could sometimes isolate some smaller changes that you need, submit them as simple PRs, and discuss them with us in a hangout. This will allow us to get involved and learn about the code. For the larger chunks, I fear you'll have to stick to business as usual.

purpleidea:
On Mon, Feb 20, 2017 at 12:28 PM, Julien Pivotto <notifications@github.com>
wrote:

> why is that needed? that file should meed out standards and not be hidden
> under the carpet.



Because it embeds a yaml file... :(

purpleidea:
I need to clean this up and fix things before it's ready for proper review. Thanks everyone for the comments so far.

resources/virt.go
```diff
@@ -108,15 +143,71 @@ func (obj *VirtRes) Init() error {
...
+
+	// check for hard to change properties
+	dom, err := obj.conn.LookupDomainByName(obj.GetName())
+	if err == nil {
+		defer dom.Free()
+	} else if !isNotFound(err) {
+		return errwrap.Wrapf(err, "%s[%s]: Could not lookup on init", obj.Kind(), obj.GetName())
+	}
+
+	if err == nil {
```
purpleidea: @ffrank That block was stuff we only do when we have a working machine. If we init and the machine doesn't exist yet, then we don't need to query it to get initial state.

PR #161, opened 2017-02-20
project: Add $me to AUTHORS

roidelapluie:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,4 +6,5 @@ This list is sorted alphabetically by first name.
 
 Felix Frank
 James Shubin
+Julien Pivotto
 Paul Morgan


purpleidea:
This needs rebasing to git master so your sig matches on such an important commit! :) Otherwise, ACK.

roidelapluie:
what about git merge ... rebasing does not scale

purpleidea:
@roidelapluie

w00t! Merged :)

RE:

> what about git merge ... rebasing does not scale

In general I try to avoid merges when they're not meaningful. For large early forks off of an old branch which then comes around and turns back into git master it might make sense. Fortunately they haven't been needed yet, and we're able to keep our history cleaner as a result :)

PR #163, opened 2017-02-20
yamlgraph: refactor parsing

mildred:
Subset of #143 with just refactoring of YAML parser

Avoid use of the reflect package, and use an extensible list of registred
resource kinds. This also has the benefit of removing the empty VirtRes and
AugeasRes struct types when compiling without libvirt and libaugeas.

### DIFF
--- a/resources/resources.go
+++ b/resources/resources.go
@@ -38,6 +38,24 @@ import (
 	"golang.org/x/time/rate"
 )
 
+var registeredResources = map[string]func() Res{}
+
+// RegisterResource registers a new resource by providing a constructor
+// function that returns a resource object ready to be unmarshalled from YAML.
+func RegisterResource(name string, creator func() Res) {
+	registeredResources[name] = creator
+}
+
+// NewEmptyNamedResource returns an empty resource object from a registered
+// type, ready to be unmarshalled.
+func NewEmptyNamedResource(name string) (Res, error) {
+	fn, ok := registeredResources[name]
+	if !ok {
+		return nil, fmt.Errorf("no resource named %s available", name)
+	}
+	return fn(), nil
+}
+
 //go:generate stringer -type=ResState -output=resstate_stringer.go
 
 // The ResState type represents the current activity state of each resource.
--- a/yamlgraph2/gconfig.go
+++ b/yamlgraph2/gconfig.go
@@ -1,0 +1,301 @@
+// Mgmt
+// Copyright (C) 2013-2017+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Package yamlgraph2 provides the facilities for loading a graph from a yaml file.
+package yamlgraph2
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"strings"
+
+	"github.com/purpleidea/mgmt/pgraph"
+	"github.com/purpleidea/mgmt/resources"
+
+	"gopkg.in/yaml.v2"
+)
+
+type collectorResConfig struct {
+	Kind    string `yaml:"kind"`
+	Pattern string `yaml:"pattern"` // XXX: Not Implemented
+}
+
+// Vertex is the data structure of a vertex.
+type Vertex struct {
+	Kind string `yaml:"kind"`
+	Name string `yaml:"name"`
+}
+
+// Edge is the data structure of an edge.
+type Edge struct {
+	Name   string `yaml:"name"`
+	From   Vertex `yaml:"from"`
+	To     Vertex `yaml:"to"`
+	Notify bool   `yaml:"notify"`
+}
+
+// ResourceData are the parameters for resource format.
+type ResourceData struct {
+	Name string               `yaml:"name"`
+	Meta resources.MetaParams `yaml:"meta"`
+}
+
+// Resource is the object that unmarshalls resources.
+type Resource struct {
+	ResourceData
+	unmarshal func(interface{}) error
+	resource  resources.Res
+}
+
+// Resources is the object that unmarshalls list of resources.
+type Resources struct {
+	Resources map[string][]Resource `yaml:"resources"`
+}
+
+// GraphConfigData contains the graph data for GraphConfig.
+type GraphConfigData struct {
+	Graph     string               `yaml:"graph"`
+	Collector []collectorResConfig `yaml:"collect"`
+	Edges     []Edge               `yaml:"edges"`
+	Comment   string               `yaml:"comment"`
+	Remote    string               `yaml:"remote"`
+}
+
+// GraphConfig is the data structure that describes a single graph to run.
+type GraphConfig struct {
+	GraphConfigData
+	ResList []resources.Res
+}
+
+// UnmarshalYAML unmarshalls the complete graph.
+func (c *GraphConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	// Unmarshal the graph data, except the resources
+	if err := unmarshal(&c.GraphConfigData); err != nil {
+		return err
+	}
+
+	// Unmarshal resources
+	var list Resources
+	list.Resources = map[string][]Resource{}
+	if err := unmarshal(&list); err != nil {
+		return err
+	}
+
+	// Finish unmarshalling by giving to each resource its kind
+	// and store each resource in the graph
+	for kind, resList := range list.Resources {
+		for _, res := range resList {
+			err := res.Decode(kind)
+			if err != nil {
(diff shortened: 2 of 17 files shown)


purpleidea:
@mildred This failed due to:

```
FAIL: Commit message should match the following regex: '^\([a-z0-9]\(\(, \)\|[a-z0-9]\)\+[a-z0-9]: \)\+[A-Z0-9][^:]\+[^:.]$'

eg:

prometheus: Implement rest api

resources: svc: Fix a race condition with reloads
```

(Basically, capitalize the first word after the `prefix:`, although a more descriptive commit message would be welcome too!)

purpleidea:
@mildred Okay, this is quite cool, and I'd like to merge it except three things (sorry for that!)

1) Can you implement this as a new GAPI so that we have your new system running in parallel with the old one? Should be easy to do, I can explain if it's not obvious. Just LMK--

2) I actually am having a bit of trouble understanding some of it, but I will try and get better at reading code. Hints welcome!

3) Was thinking of merging this after: https://github.com/purpleidea/mgmt/tree/feat/future (assuming it passes some manual smoke tests tomorrow) -- The main change is the Init() stuff moved, so this patch should reflect those changes as well.

Thanks!

mildred:
> I actually am having a bit of trouble understanding some of it, but I will try and get better at reading code. Hints welcome!

It's actually quite simple. When an object to unmarshal implements the method `UnmarshalYAML`, this method is called instead of using the yaml annotation in the struct. This method is called with a function (the `unmarshal` parameter) that can be called any number of times.

To decode the graph, we decompose `GraphConfig` (that implements the UnmarshalYAML interface) and `GraphConfigData` (that contains the yaml annotations). When `GraphConfig.UnmarshalYAML` is called, it decodes first the `GraphConfigData` (all but resources) then `Resources`. This is two steps, because in PR #143, there was a branch here for the yaml format version. This can probably be merged in a single operation in the context of this PR.

Lastly, `GraphConfig.UnmarshalYAML` finishes the decoding of each resource individually, after telling each resource its type. More on that below.

unmarshalling `map[string]Resource` leads to `Resource.UnmarshalYAML` being called. Unfortunately, the resource object don't know yet its type, so it can only decode common parameters (`name` and `meta`, which it does using the `ResourceData` helper type that is annotated). It then stores the `unmarshal` function for later use when the resource type is known.

When `map[string]Resource` is partially decoded, the `GraphConfig.UnmarshalYAML` function goes through the map and invoke `Resource.Decode` on each resource, giving the resource type as function argument.

`Resource.Decode` can then use the previously stored `unmarshal` function to decode type specific parameters. It asks the `resources` package for a new resource of the given type, and decodes the object. It then does some basic initialisation (`SetName`, `SetKind` and sets the meta parameters). The resource is then ready for graph insertion.

The changes in `NewGraphFromConfig` are basically just removing use of the `reflect` package to use the stored resources generated by the mentionned process. The code is otherwise identical.

Changes in `resources/resources.go` contains a public function for each different resource to register a constructor function that creates an empty resource of the kind. And the function `NewEmptyNamedResource` that constructs such a resource given its type.

mildred:
> Can you implement this as a new GAPI so that we have your new system running in parallel with the old one? Should be easy to do, I can explain if it's not obvious. Just LMK--

Why a separate GAPI. The benefits of not having `*_disabled.go` files are nullified in that case. It should be fairly simple to do however. I'd just copy the `yamlgraph` package and give it a new name.

> Was thinking of merging this after: https://github.com/purpleidea/mgmt/tree/feat/future (assuming it passes some manual smoke tests tomorrow) -- The main change is the Init() stuff moved, so this patch should reflect those changes as well.

After basic checks, it doesn't seem my changes are incompatible with that.

purpleidea:
@mildred The feat/future stuff has been merged in git master.

We can merge your work as a new GAPI, and cleanup the *_disabled.go stuff in a future commit if everyone prefers the new format. Does that work?

mildred:
Sorry for the delay, I don't always read my GitHub notifications. Coding this as a separate GPAI would work, I just need to get around and do this.

purpleidea:
LMK if you want help. Cheers

mildred:
@purpleidea done with the rebase, also changed the GAPI to have two versions of yaml parser. Added additional command line flag `--yaml2` to trigger the new GAPI.

purpleidea:
@mildred can you rebase this into a single patch so it's easier to review and passes the tests?

mildred:
done

purpleidea:
@mildred This failed because the commit message body (not the prefix) needs to start with a capital:

eg:  `yamlgraph: Refactor parsing` not  `yamlgraph: refactor parsing` although a more descriptive message about what this does would be preferred.

I'll start a quick review anyways...

mildred:
This version should work

resources/resources.go
```diff
@@ -38,6 +38,24 @@ import (
 	"golang.org/x/time/rate"
 )
 
+var registeredResources = map[string]func() Res{}
+
+// RegisterResource registers a new resource by providing a constructor
+// function that returns a resource object ready to be unmarshalled from YAML
```
purpleidea: nit: trailing period needed.

puppet/puppet.go
```diff
@@ -26,7 +26,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/purpleidea/mgmt/yamlgraph"
+	yamlgraph "github.com/purpleidea/mgmt/yamlgraph2"
```
purpleidea: Nack on this one line at least for now. Let's keep things stable until we deprecate yamlv1 and make sure felix is okay with the change first since his puppet-mgmtgraph depends on this format.

remote/remote.go
```diff
@@ -64,7 +64,7 @@ import (
 	cv "github.com/purpleidea/mgmt/converger"
 	"github.com/purpleidea/mgmt/util"
 	"github.com/purpleidea/mgmt/util/semaphore"
-	"github.com/purpleidea/mgmt/yamlgraph"
+	yamlgraph "github.com/purpleidea/mgmt/yamlgraph2"
```
purpleidea: Nack for now. Let's play with the v2 code first, sorry and leave this one-liner out for now.

resources/resources.go
```diff
@@ -38,6 +38,24 @@ import (
...
+func RegisterResource(name string, creator func() Res) {
+	registeredResources[name] = creator
+}
+
+// NewEmptyNamedResource returns an empty resource object from a registered
+// type, ready to be unmarshalled
```
purpleidea: nit: trailing period needed.

resources/resources.go
```diff
@@ -38,6 +38,24 @@ import (
...
+func NewEmptyNamedResource(name string) (Res, error) {
+	fn, ok := registeredResources[name]
+	if !ok {
+		return nil, fmt.Errorf("No resource named %s available", name)
```
purpleidea: golint says that error strings should not start with capital or end with period. sorry, small change needed.

yamlgraph2/gconfig.go
```diff
@@ -0,0 +1,303 @@
...
+
+// Edge is the data structure of an edge.
+type Edge struct {
+	Name   string `yaml:"name"`
+	From   Vertex `yaml:"from"`
+	To     Vertex `yaml:"to"`
+	Notify bool   `yaml:"notify"`
+}
+
+// ResourceData are the parameters for resource format
```
purpleidea: A bunch of these need trailing periods, sorry...

yamlgraph2/gconfig.go
```diff
@@ -0,0 +1,303 @@
...
+// GraphConfig is the data structure that describes a single graph to run.
+type GraphConfig struct {
+	GraphConfigData
+	ResList []resources.Res
+}
+
+// UnmarshalYAML unmarshalls the complete graph.
+func (c *GraphConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	// Unmarshal the graph data, except the resources
+	err := unmarshal(&c.GraphConfigData)
```
purpleidea: replace with one line version:

if err := unmarshal(...); err != nil

yamlgraph2/gconfig.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Unmarshal resources
+	var list Resources
+	list.Resources = map[string][]Resource{}
+	err = unmarshal(&list)
```
purpleidea: one line...

mildred:
I'm not sure we are agreeing on the same thing, this patch doesn't change at all the yaml format. It just changes the way it is handled internally (that's why I didn't want to split into a separate GAPI) so we don't need to use the reflect package and have empty *_disabled.go files for resource that are optional to compile.

resources/resources.go
```diff
@@ -187,8 +205,8 @@ type Res interface {
 
 // BaseRes is the base struct that gets used in every resource.
 type BaseRes struct {
-	Name       string           `yaml:"name"`
```
mildred: yes, sorry for that. I didn't think about that.

purpleidea:
@mildred Yes, as discussed in IRC, sorry my brain is indeed fried today. Reviewing updated patch...

purpleidea:
@mildred Patch looks great, thank you!

I'd like to add a few additional comments:

This is really excellent, and thank you for hanging in there and finishing this.

In particular, I'm especially pleased, because I always wanted to get rid of the reflect part, but didn't previously figure out how.

Lastly, I hope you feel particularly proud because the "RegisterResource into global map" pattern wasn't one I had considered, and it works so well I've already used the same idea as part of the upcoming language branch I'm working on! So IOW thanks for making future code better too!

MERGED :)

purpleidea:
PS: Was there something we wanted to change around the virt_disabled.go and augeas_disabled.go ? (I forget)

mildred:
yes, but they are still needed so long that we keep yamlv1

purpleidea:
got it, right :)

PR #164, opened 2017-02-20
Author

ffrank:
Based on #161


purpleidea:
@ffrank Read your commit message! You've contributed some important parts, so I think you deserve a place here! Welcome to the club!

(Don't know how @roidelapluie's commit ended up in here too, but I'm waiting for him to rebase against master for that one so the sig matches on such an important commit!)

PR #165, opened 2017-02-21
Debian packaging: the beginning

dsx:
- It can build .deb package
- It can register mgmt.service with systemd and start it
- Tests are disabled
- It needs more work


debian/changelog
```diff
@@ -0,0 +1,6 @@
+mgmt (0.0.1-1) unstable; urgency=low
```
purpleidea: Awesome dsx! One small comment-- Can we instead generate this from git log, similarly to what is done here: https://github.com/purpleidea/mgmt/blob/master/Makefile#L191 rather than hardcoding this... We don't expect to be the distro upstream package at least in this sort of "commit the changelog" way.

ffrank:
This looks super promising :+1:

purpleidea:
@dsx You've also got two instances of trailing whitespace (causes fails) and one missing newline. a `git show` should highlight these for you.

.gitignore
```diff
@@ -9,3 +9,11 @@ mgmt
 mgmt.static
 mgmt.iml
 rpmbuild/
+
+# Build stuff
```
roidelapluie: Move that to debian/.gitignore?

debian/mgmt.install
```diff
@@ -0,0 +1 @@
+mgmt usr/bin
```
roidelapluie: Can we also install the docs and the license?

debian/control
```diff
@@ -0,0 +1,15 @@
...
+Maintainer: Egor M. (dsx) <dsx@droidnest.org>
+Build-Depends:
+               debhelper (>= 8.0.0),
+               dh-golang,
+               dh-systemd,
+               golang-go,
+               
+Package: mgmt
+Architecture: amd64
+# Depends: ${shlibs:Depends} ${misc:Depends}
```
roidelapluie: That should not be commented

debian/control
```diff
@@ -0,0 +1,15 @@
+Source: mgmt
+Maintainer: Egor M. (dsx) <dsx@droidnest.org>
+Build-Depends:
+               debhelper (>= 8.0.0),
+               dh-golang,
+               dh-systemd,
+               golang-go,
+               
+Package: mgmt
+Architecture: amd64
```
roidelapluie: Should be any

debian/control
```diff
@@ -0,0 +1,15 @@
+Source: mgmt
+Maintainer: Egor M. (dsx) <dsx@droidnest.org>
+Build-Depends:
+               debhelper (>= 8.0.0),
```
roidelapluie: debhelper (>= 10.0.0),

as we have 10 in debian/compat

debian/mgmt.service
```diff
@@ -0,0 +1 @@
+../misc/mgmt.service
```
roidelapluie: Do not do this, instead, install ../misc/mgmt.service directly within mgmt.install

dsx: Why not? dh_systemd_{enable,start} will automagically install and start all $packagename.service files.

roidelapluie: Oh wait. Is this a symlink? Didn't understand that :)

roidelapluie: Still it should belong to install...

«If your service file is not provided by upstream and you ship it as part of your debian package, name it debian/$pkg.service and dh_installinit will install it to the correct location. For more information please refer to the dh_installinit(1) man page. »

In this case it is provided by mgmt, so install is the place.

dsx: > name it debian/$pkg.service

That's exactly what is happening here. There is debian/mgmt.service file which is being automatically installed.

roidelapluie: «If your service file is not provided by upstream»

dsx: Ok, apparently I can't read :)

I'll see how it is done in Debian and do the same thing. Personally, I don't see any point in manually installing and then explicitly calling some dh_* when it can be automagically handled with one symlink.

roidelapluie: The trick is that this should be installed by make install :)

roidelapluie: @dsx no nood to call manually dsx_. It it is in install that well be called automatically.

purpleidea:
@dsx Can you rebase this into a single patch for review please?
It should have a commit message like: "debian: Add packaging" or similar since our CI checks for commit message format.

dsx:
@purpleidea sure. I'll rebase when I get to it again.

purpleidea:
@dsx Thanks again for all your work. An updated version of this is in https://github.com/purpleidea/mgmt/pull/306 so I'm going to close this one. Head over there if you have any review comments.
Thanks!

PR #166, opened 2017-02-22
New future!

purpleidea:
After a bunch of re-working, I think this is ready for git master. I hope I don't eat those words, since I didn't go through every live demo that I usually run, but I think it's probably pretty good.

Comments welcome before I merge, and get ready for a proper release!

Hopefully this should also unblock everyone!

### DIFF
--- a/pgraph/actions.go
+++ b/pgraph/actions.go
@@ -316,7 +316,11 @@ func (g *Graph) Worker(v *Vertex) error {
 	// the Watch() function about which graph it is
 	// running on, which isolates things nicely...
 	obj := v.Res
-	obj.SetWorking(true) // gets set to false in Res.Close() method at end...
+
+	// run the init (should match 1-1 with Close function if this succeeds)
+	if err := obj.Init(); err != nil {
+		return errwrap.Wrapf(err, "could not Init() resource")
+	}
 
 	lock := &sync.Mutex{} // lock around processChan closing and sending
 	finished := false     // did we close processChan ?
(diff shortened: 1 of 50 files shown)


purpleidea:
FYI: I read through everything in https://github.com/purpleidea/mgmt/pull/157 and I think this resolves all of those issues. Thanks for your early reviews. /cc @roidelapluie @ffrank

purpleidea:
BTW, I also added new tests! w00t.

purpleidea:
FYI: I'm obviously not planning on merging the last commit which only fails due to the commit message of XXX (which I did on purpose to fail it)

Comments about all this? Fair warning...

resources/virt.go
```diff
@@ -490,24 +532,306 @@ func (obj *VirtRes) CheckApply(apply bool) (bool, error) {
...
+		return false, errwrap.Wrapf(err, "stateCheckApply failed")
+	} else if !c {
+		checkOK = false
+	}
+
+	// FIXME: should we wait to ensure machine is booted before continuing?
+	// it may be useful to wait for guest agent to hotplug some ram or cpu!
+
 	if !apply {
 		return false, nil
```
purpleidea: I changed this in https://github.com/purpleidea/mgmt/commit/c480bd94db5147fec09a336fb67eac5333091524 Thanks for the comment!

pgraph/actions.go
```diff
@@ -306,6 +306,149 @@ func (obj *SentinelErr) Error() string {
...
+	obj := v.Res
+	running := false
+	done := make(chan struct{})
+	playback := false                      // do we need to run another one?
+	_, wcuid, pcuid := obj.ConvergerUIDs() // get extra cuids (worker, process)
+
+	waiting := false
+	var timer = time.NewTimer(time.Duration(math.MaxInt64)) // longest duration
+	if !timer.Stop() {
+		<-timer.C // unnecessary, shouldn't happen
```
purpleidea: Great question! Actually the timer docs seem to say that this is needed, but in practice it never seems to happen. So your guess is as good as mine if I'm using timer correctly, but it does seem to work as expected.

purpleidea:
@roidelapluie RE: The feedback guide and that smaller patches are easier to review-- I completely agree, but I posted this here since I was trying to balance that with the comments that you and ffrank made about not just pushing to master. In any case, I apologize for it being big, but as we get closer and closer to minimally viable, I'm sure it will happen less.

PR #168, opened 2017-02-26
compilation: Make build the default target

roidelapluie:
It is really strange that whe I run make, it does not build mgmt. This
commit makes build the default target, without moving the target,
therefore we keep as much as we can the order of the file.

This also removes the confusion for designers that would run "make"
instead of "make art", whose work would be disrupted when we add a --
let's say -- make alpharelese command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,8 @@ ifneq ($(GOTAGS),)
     BUILD_FLAGS = -tags '$(GOTAGS)'
 endif
 
+default: build
+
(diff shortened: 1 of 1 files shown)


purpleidea:
LGTM, merged:

FYI: Our Makefile needs fixing, because it only currently looks at `main.go` for the deps, when it should really look at all the packages to know if we're out of date or not. Know what I mean? @roidelapluie

PR #169, opened 2017-02-27
Implement semaphore metaparam

purpleidea:


purpleidea:
@roidelapluie @ffrank Here's my attempt to involve you both in a small review :)

I'm pretty confident about the patch, and I think the feature is dope (it's been on my TODO list for a while) and I've been really itching to write it. It was actually very fun to write! I needed this to enjoy coding again after the nasty race hunting I recently landed in git master.

In any case, reviews welcome, and barring any red flags I'll intend to merge this to git master in your time in the morning or so. Realistically it's more fun to ping you folks so you can see the feature!

purpleidea:
PS: The patch contains docs, an example, and a test, so I think this is extra dope!

purpleidea:
FYI: The failure seems to be related to a possible bug I thought was completely squashed, but is apparently appearing for some reason... Should have virtually no effect on this patch review. Will try to look into that bug this week.

purpleidea:
I've pushed an updated version which should hopefully remedy this issue. At this point, it's late and I think fresh eyes should review in the morning! Night!

roidelapluie:
Did you cover the package groups?

purpleidea:
@roidelapluie How does this relate to autogrouping??

roidelapluie:
--sem 2

packages: [httpd, nginx, perl]

file: foo

Will it do:

| (httpd nginx) |
| (perl) foo |


or

| (httpd nginx perl) foo |


(group)

purpleidea:
@roidelapluie Oh -- good question actually. It would do the second thing.

Although if the semaphore was resource specific, I didn't look at how two resource semaphores are grouped. They should probably be merged. That is probably a TODO. Want to do it?

roidelapluie:
Funny because I would expect the first thing.

purpleidea:
On Mon, Feb 27, 2017 at 12:15 PM, Julien Pivotto
<notifications@github.com> wrote:
>
> Funny because I would expect the first thing.



Why?

purpleidea:
@roidelapluie Got any reasons why it should be the first things before I write the patch for the 2nd thing?

roidelapluie:
As a user, my resource is a package. Not a group of packages. If I put semaphore to 1 to replicate the behaviour of Puppet (like you specify in the doc) I expect it will just do one at the same time.

purpleidea:
@roidelapluie Maybe to be more correct it might be better to say: "--sema 1" && "--no-autogroup" for pure puppet semantics.

The second option doesn't of course exist yet, but I am happy to accept a patch for it.

Having said that, autogrouping never does something dangerous, so turning three packages (with sema:1) isn't negative in some way.

On the other hand, if X (with sem->A:1) and Y (with sem->B:4) are grouped, you would expect them to now require BOTH semaphores, right?

roidelapluie:
I expect that sema takes precedence over autogrouping. I would consider the opposite as a bug. Otherwise, sema is just «lying»

purpleidea:
@roidelapluie The reason I don't think it's lying, is because I see an autogrouped resource as a _single_ resource. But I understand your point. I think the only sane thing to do is document this, in particular because the code to do it the first way would be extremely complex I think.

Any other comments on this? @ffrank or others?

purpleidea:
@roidelapluie I've thought more about your "first thing" approach and I've decided that while I don't think it's the correct behaviour for _semaphores_ (in the context of this patch) I have not decided that it's not necessarily wrong in all cases to have that behaviour.

I personally don't think getting that behaviour is a high priority or if it would be very useful, but if it was something that folks wanted, I would instead see it as a "max grouping count" feature. In other words, I don't think it's illegitimate to be able to specify `autogroup:true` and autogroup_max:4.

If you really want this feature, I'm happy to help you understand the autogrouping code if you'd like to implement it. Hope that all sounds reasonable to you.

PR #171, opened 2017-02-28
testing: Remove 1.8 from allowed failure

roidelapluie:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ matrix:
   fast_finish: true
   allow_failures:
     - go: tip
-    - go: 1.8
 notifications:
   irc:
     channels:


purpleidea:
In principle, I like the idea, but I don't have access to a 1.8 machine yet, so I'm unfortunately going to NACK until I do, because if I cause an issue there I can't debug.

As an aside, I don't think we've *ever* had a failure that was 1.8 only...

roidelapluie:
Well I am running go 1.8 only

purpleidea:
@roidelapluie I mean, the tests are still running... And I always check to see if they failed on 1.8 and they don't. (Unless they fail everywhere)

purpleidea:
As soon as Fedora 26 comes out (IIRC will have 1.8) I will add this, sound reasonable?

roidelapluie:
so you do the work of a machine :)

purpleidea:
3 months: https://fedoraproject.org/wiki/Releases/26/ChangeSet#Golang_1.8

purpleidea:
Now fixed in master: https://github.com/purpleidea/mgmt/commit/74e04e81d5158af6db635321e4251c6c4423ae4d

PR #180, opened 2017-03-11
pgraph: Move the BackPoke to before the semaphores

purpleidea:
I can't think of a reason we should grab a semaphore before backpoking.
The semaphore is intended to block around the actual work in CheckApply,
not the dependency resolution of the correct vertex.

### DIFF
--- a/pgraph/actions.go
+++ b/pgraph/actions.go
@@ -169,6 +169,18 @@ func (g *Graph) Process(v *Vertex) error {
 	defer obj.SetState(resources.ResStateNil) // reset state when finished
 	obj.SetState(resources.ResStateProcess)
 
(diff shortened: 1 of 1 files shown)


purpleidea:
@roidelapluie @ffrank or anyone else would like to do a quick review. Cheers!

purpleidea:
Merging due to no comments.

roidelapluie:
Was ok.

ffrank:
I currently lack the insight into the logic flow to make a sensible comment on this kind of change.
You *seem* to have thought it through so :)

purpleidea:
@ffrank I know you might today, but I was trying to get you to want to dive into looking into it in case you were curious. Next time :)

Issue #175, opened 2017-03-04
[LOVE] pgraph->DeleteEdge needs test cases


purpleidea:
This is a beginner patch for a new golang/mgmt contributor!

The DeleteEdge methood in pgraph (currently at: https://github.com/purpleidea/mgmt/blob/master/pgraph/pgraph.go#L192 ) needs some tests!

Please write some! Tests should go in pgraph_test.go

Happy hacking!

taksuyu:
Seems doable. I'll give it a shot.

purpleidea:
@taksuyu Any progress on this? Thanks

taksuyu:
I may have to just push out something simple. Been a bit busy with stuff. Sorry about that.

taksuyu:
https://github.com/purpleidea/mgmt/pull/181

dreamer-nitj:
I will take it up

dreamer-nitj:
How do i run the test case.. i get this error:

`go run ./pgraph/pgraph_test.go 
go run: cannot run *_test.go files (./pgraph/pgraph_test.go)`

purpleidea:
`go test github.com/purpleidea/mgmt/pgraph`

OR

`go test github.com/purpleidea/mgmt/pgraph -v`

OR

`go test github.com/purpleidea/mgmt/pgraph -run TestCaseName`

purpleidea:
Closed in https://github.com/purpleidea/mgmt/commit/db6972638d6663d6bca02595776f092a9a6ce430

PR #181, opened 2017-03-15
test: Add a test for DeleteEdge

taksuyu:
This should close https://github.com/purpleidea/mgmt/issues/175

### DIFF
--- a/pgraph/pgraph_test.go
+++ b/pgraph/pgraph_test.go
@@ -627,6 +627,56 @@ func TestPgraphT11(t *testing.T) {
 	}
 }
 
+func TestPgraphDeleteEdge1(t *testing.T) {
+	g1 := NewGraph("g1")
+	g2 := NewGraph("g2")
+	v1 := NV("v1")
+	v2 := NV("v2")
+	e1 := NewEdge("e1")
+
+	g1.AddEdge(v1, v2, e1)
+	g1.DeleteEdge(e1)
+
+	g2.AddVertex(v1, v2)
+
+	runGraphCmp(t, g1, g2)
+}
+
+func TestPgraphDeleteEdge2(t *testing.T) {
+	g1 := NewGraph("g1")
+	g2 := NewGraph("g2")
+	v1 := NV("v1")
+	v2 := NV("v2")
+	e1 := NewEdge("e1")
+
+	g1.AddEdge(v1, v2, e1)
+	g1.DeleteEdge(e1)
+	g1.AddEdge(v1, v2, e1)
+
+	g2.AddEdge(v1, v2, e1)
+
+	runGraphCmp(t, g1, g2)
+}
+
+func TestPgraphDeleteEdge3(t *testing.T) {
+	g1 := NewGraph("g1")
+	g2 := NewGraph("g2")
+	v1 := NV("v1")
+	v2 := NV("v2")
+	v3 := NV("v3")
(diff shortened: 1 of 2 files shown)


pgraph/pgraph_test.go
```diff
@@ -627,6 +627,45 @@ func TestPgraphT11(t *testing.T) {
...
+	G.AddEdge(v5, v6, e5)
+	G.AddEdge(v1, v6, e6)
+
+	num := G.NumEdges()
+	edges := [6]*Edge{e1, e2, e3, e4, e5, e6}
+
+	for _, edge := range edges {
+		G.DeleteEdge(edge)
+
+		if G.NumEdges() >= num {
```
purpleidea: should we do a != cmp to be exact instead?

purpleidea:
Commit message should start with:

`pgraph: Message`

or `pgraph: test: Message`

pgraph/pgraph_test.go
taksuyu: Probably a good idea to make sure `DeleteEdge` isn't adding anything to the structure.

I'll introduce a variable to store the previous `G` to do so.

taksuyu: Actually, that may not be the best idea. Adding a `Vertex` would qualify as different.

taksuyu:
I'll make that change

purpleidea:
@MohamedAshiqrh Want to do a quick review to get warmed up?

pgraph/pgraph_test.go
```diff
@@ -627,6 +627,50 @@ func TestPgraphT11(t *testing.T) {
...
+			t.Errorf("deleting an edge failed")
+		}
+
+		num--
+	}
+
```
MohamedAshiqrh: @taksuyu @purpleidea we dont test failure cases is it? what if a deleted edge is tried to delete again?

taksuyu: @MohamedAshiqrh If that was the case, the amount of edges would be above 0 later on in the test. It could try to delete multiple edges I suppose. I could change it to something like
`if G.NumEdges () == (num -1)`

taksuyu:
I got a much simpler change coming, give me a bit

taksuyu:
I made them simpler and added some variant to it.

pgraph/pgraph_test.go
```diff
@@ -627,6 +627,56 @@ func TestPgraphT11(t *testing.T) {
...
+	v2 := NV("v2")
+	v3 := NV("v3")
+	e1 := NewEdge("e1")
+	e2 := NewEdge("e2")
+	save := *G
+
+	G.AddEdge(v1, v2, e1)
+	G.AddEdge(v2, v3, e2)
+
+	G.DeleteEdge(e1)
```
purpleidea: Here you change both `G` and `save`. I think this is a bit confusing! Please keep this ultra simple. Basic tests first test the core things in an extremely straight forward and boring way.

You should be creating `g1` and `g2` in all your tests. You can then perform some operations on `g2` so that it looks like `g1`. Then you run the cmp function on `g1` and `g2`. Isn't that simpler?

taksuyu: Sure..

purpleidea:
Closing because of inactivity and rot. Please feel free to resubmit as a new PR if it's fixed up. Thanks

PR #185, opened 2017-03-16
Our GAPI needs a tweak #174

### LINKED ISSUE #174: [LOVE] Our GAPI needs a tweak
This is intended for a new mgmt contributor or for someone with very few patches. If nobody claims this, but you are *really* itching to write this, that's okay too!

In mgmt we have a GAPI (graph API) which any frontend can implement to generate graphs that the mgmt engine will run. Currently we have a raw yaml graph input, and a puppet code input. When the mgmt language is built, it will use this API as well.

(Links to code are at an arbitrary git commit so that they don't break, but obviously you'll want to do this on git master and some things might have moved slightly if we've patches those files. Ping me if you can't find something.)

Here is the interface:
https://github.com/purpleidea/mgmt/blob/73ccbb69ea1d74ab5fd96b2096d5902313588a59/gapi/gapi.go#L44

The `Next()` function spits out events when there are new graphs available. In particular, when new "exported resources" are available (from other machines) we refresh our local graphs.

That stream of events is currently located here:
https://github.com/purpleidea/mgmt/blob/73ccbb69ea1d74ab5fd96b2096d5902313588a59/lib/main.go#L377

But it should actually be hidden inside the GAPI itself, since not every GAPI might necessarily want to Watch for exported resources.

So, that function should be moved out of there, and into the GAPI, eg here:
https://github.com/purpleidea/mgmt/blob/73ccbb69ea1d74ab5fd96b2096d5902313588a59/yamlgraph/gapi.go#L79

but to wrap it nicely, we'd like to actual have a generic interface for that watch called `ResWatch` which should go here:

https://github.com/purpleidea/mgmt/blob/73ccbb69ea1d74ab5fd96b2096d5902313588a59/gapi/gapi.go#L29

The implementation of that function will probably USE the WatchAll method, and go in here:

https://github.com/purpleidea/mgmt/blob/73ccbb69ea1d74ab5fd96b2096d5902313588a59/etcd/world.go#L30

Hopefully this is clear enough, if you have questions please let me know!

Happy hacking!

snowsky:
Fixes #174

### DIFF
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -2096,36 +2096,6 @@ func Leader(obj *EmbdEtcd) (string, error) {
 	return "", fmt.Errorf("members map is not current") // not found
 }
 
-// WatchAll returns a channel that outputs a true bool when activity occurs
-// TODO: Filter our watch (on the server side if possible) based on the
-// collection prefixes and filters that we care about...
-func WatchAll(obj *EmbdEtcd) chan bool {
-	ch := make(chan bool, 1) // buffer it so we can measure it
-	path := fmt.Sprintf("/%s/exported/", NS)
-	callback := func(re *RE) error {
-		// TODO: is this even needed? it used to happen on conn errors
-		log.Printf("Etcd: Watch: Path: %v", path) // event
-		if re == nil || re.response.Canceled {
-			return fmt.Errorf("watch is empty") // will cause a CtxError+retry
-		}
-		// we normally need to check if anything changed since the last
-		// event, since a set (export) with no changes still causes the
-		// watcher to trigger and this would cause an infinite loop. we
-		// don't need to do this check anymore because we do the export
-		// transactionally, and only if a change is needed. since it is
-		// atomic, all the changes arrive together which avoids dupes!!
-		if len(ch) == 0 { // send event only if one isn't pending
-			// this check avoids multiple events all queueing up and then
-			// being released continuously long after the changes stopped
-			// do not block!
-			ch <- true // event
-		}
-		return nil
-	}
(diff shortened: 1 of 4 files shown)


etcd/etcd.go
```diff
@@ -2096,36 +2096,6 @@ func Leader(obj *EmbdEtcd) (string, error) {
 	return "", fmt.Errorf("members map is not current") // not found
 }
 
-// WatchAll returns a channel that outputs a true bool when activity occurs
-// TODO: Filter our watch (on the server side if possible) based on the
-// collection prefixes and filters that we care about...
-func WatchAll(obj *EmbdEtcd) chan bool {
```
purpleidea: Why was this moved/removed from here? At most it should be renamed...

gapi/gapi.go
```diff
@@ -23,6 +23,16 @@ import (
 	"github.com/purpleidea/mgmt/resources"
 )
 
+// World is an interface to the rest of the different graph state. It allows
```
purpleidea: Why these changes??

lib/main.go
```diff
@@ -356,12 +356,17 @@ func (obj *Main) Run() error {
 	}
 
 	var gapiChan chan error // stream events are nil errors
+	world := &etcd.World{
```
purpleidea: This isn't even gofmt-ed. You should run `make gofmt` on your code before committing. We have a test case that catches badly formatted code.

purpleidea:
(In case you didn't already, you might want to also make sure your patch is rebased to the latest git master.) Eg: git checkout master; git pull origin master; git checkout your/feature/branch; git rebase master

purpleidea:
Now fixed in https://github.com/purpleidea/mgmt/commit/6fd5623b1f97cbfd77399a78d2aec14ae986ff2e

PR #188, opened 2017-05-16
exec: Add autoEdges between ExecRes and PkgRes

roidelapluie:
No Tests Yet

### DIFF
--- a/resources/exec.go
+++ b/resources/exec.go
@@ -331,11 +331,39 @@ type ExecUID struct {
 	// TODO: add more elements here
 }
 
-// AutoEdges returns the AutoEdge interface. In this case no autoedges are used.
+// ExecResAutoEdges holds the state of the auto edge generator.
+type ExecResAutoEdges struct {
+	edges []ResUID
+}
+
+// Next returns the next automatic edge.
+func (obj *ExecResAutoEdges) Next() []ResUID {
+	return obj.edges
+}
+
+// Test gets results of the earlier Next() call, & returns if we should continue!
+func (obj *ExecResAutoEdges) Test(input []bool) bool {
+	return false // Never keep going
+	// TODO: We could return false if we find as many edges as the number of different path in cmdFiles()
+}
+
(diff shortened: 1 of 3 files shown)


purpleidea:
Upstream blocker for this is fixed. Details in: https://github.com/purpleidea/mgmt/issues/189#issuecomment-305944917

Interested in fixing up and rebasing? I also made a test template in file_test.go which works and a modified version would work nicely as exec_test.go

Thanks!

purpleidea:
@roidelapluie Last ping... You want to fixup and merge, rebase-to-master and transfer to me, abandon or should I close?

roidelapluie:
@purpleidea hi

resources/exec.go
```diff
@@ -485,3 +513,27 @@ func (w *wrapWriter) Write(p []byte) (int, error) {
 func (w *wrapWriter) String() string {
 	return w.Buffer.String()
 }
+
+// cmdFiles returns all the potential files/commands this command might need
```
purpleidea: trailing period since docstrings are full sentences please.

resources/exec.go
```diff
@@ -485,3 +513,27 @@ func (w *wrapWriter) Write(p []byte) (int, error) {
...
+func (obj *ExecRes) cmdFiles() []string {
+	var paths []string
+	if obj.Shell != "" {
+		paths = append(paths, obj.Shell)
+	} else {
+		cmdSplit := strings.Fields(obj.Cmd)
+		paths = append(paths, cmdSplit[0])
```
purpleidea: the logic is right in that we won't get an empty "" because obj.Cmd is in Validate() but it could be a " " space, so we still need to check we have at least `len(cmdSplit) > 0` to avoid a panic. Similarly, logic is good below but needs these checks I think.

roidelapluie: fixed

resources/exec.go
roidelapluie: fixed

purpleidea:
Merged, thank you @roiedelapluie!

PR #197, opened 2017-06-13
hcl: added hil string interpolation to hcl frontend

ChrisMcKenzie:
This adds string interpolation to fields in hcl frontend, allowing for declarative `Receiver` support and automatically creates edges for resources that use values from other resources.

*Example*
```
resource "file" "file1" {
  path = "/tmp/mgmt-hello-world"
  content = "${exec.sleep.Output}"
  state = "exists"
}

resource "exec" "sleep" {
 cmd = "echo hello"
}
```

The following will create a file at `/tmp/mgmt-hello-world` with `hello` as the contents.

*Gotchas:*
- currently you need to capitalize the field name (i.e. `Output`)

### DIFF
--- a/hcl/parse.go
+++ b/hcl/parse.go
@@ -25,7 +25,9 @@ import (
 
 	"github.com/hashicorp/hcl"
 	"github.com/hashicorp/hcl/hcl/ast"
+	"github.com/hashicorp/hil"
 	"github.com/purpleidea/mgmt/gapi"
+	hv "github.com/purpleidea/mgmt/hil"
 	"github.com/purpleidea/mgmt/pgraph"
 	"github.com/purpleidea/mgmt/resources"
 )
@@ -68,6 +70,7 @@ type Resource struct {
 	resource resources.Res
 	Meta     resources.MetaParams
 	deps     []*Edge
+	rcv      map[string]*hv.ResourceVariable
 }
 
 type key struct {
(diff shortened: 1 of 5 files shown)


purpleidea:
> currently you need to capitalize the field name (i.e. Output)

Just heading to bed, so will do a further review in the morning, but I have written a currently unmerged function to figure out the capitalization so you can all lcase the inputs. Will merge it if it would help you.

purpleidea:
Quick review, looks good though! Thanks again!

ChrisMcKenzie:
Yeah I guess I could have just done `strings.ToUpper(string(field[0])) + field[1:]` but if you have a func for doing lets do that then it will be a bit more consistent.

purpleidea:
On Tue, Jun 13, 2017 at 12:07 AM, Christopher McKenzie <
notifications@github.com> wrote:

> Yeah I guess I could have just done strings.ToUpper(string(field[0])) +
> field[1:] but if you have a func for doing lets do that then it will be a
> bit more consistent.
>

Indeed. Actually that was my first thought too, but it fails if the param
is named "SomeFooThing", because you don't know where the word breaks are.
So it's coming soon :)

hcl/parse.go
```diff
@@ -296,6 +322,41 @@ func loadResourcesHcl(list *ast.ObjectList) ([]*Resource, error) {
 			})
 		}
 
+		var config map[string]interface{}
+		if err := hcl.DecodeObject(&config, item.Val); err != nil {
+			log.Printf("HCLParse: unable to decode body: %v", err)
```
purpleidea: For uniformity, let's keep all the hcl related log messages in the form: `hcl: Message`

resources/sendrecv.go
```diff
@@ -173,6 +173,11 @@ type Send struct {
 	Changed bool // set to true if this key was updated, read only!
 }
 
+// SetRecv allows for adding Recv's to the object
```
purpleidea: Nit: doc comments should be proper sentences and end with a period.

purpleidea:
@ChrisMcKenzie Ping? Want to rebase and fix these tiny things?

ChrisMcKenzie:
yeah sorry, I will have some fixes later today, yesterday was a busy day for me.

purpleidea:
@ChrisMcKenzie No worries, sorry for the ping.

ChrisMcKenzie:
@purpleidea got those changes in!

purpleidea:
@ChrisMcKenzie Happy to merge as-is, but fixes don't look committed. And the period is still missing :) https://github.com/purpleidea/mgmt/pull/197/commits/8102e0a4686e0a50935a8e127219429a5a057b8d#diff-174b7c16184138988d9289c53e21ea24R176

Why not be consistent with this: https://github.com/purpleidea/mgmt/blob/master/hcl/gapi.go#L91 "hcl" vs "HCL"?

purpleidea:
I patched some sillyness on top in https://github.com/purpleidea/mgmt/commit/5a3bd3ca67a19f1746c80f62babe3f2b8aa8c165

Everything else in your patch was great, and this is an awesome way to use send->recv.

Thanks!

ChrisMcKenzie:
@purpleidea that's what I get for being in a hurry! sorry about that!

Issue #72, opened 2016-09-22
implement an http resource


purpleidea:
We should have an http resource. Different names welcome. It should probably support long-polling to watch.

@purpleidea note: is the send/recv data passing between resources design something valuable for this?

ffrank:
FWIW, **without** data passing, this could really only store HTTP responses to files. In this case, it would probably be awkward that this is an alternative way to manage files.

purpleidea:
On Wed, Sep 28, 2016 at 6:02 PM, Felix Frank <notifications@github.com> wrote:
>
> FWIW, without data passing, this could really only store HTTP responses to files. In this case, it would probably be awkward that this is an alternative way to manage files.



Without data passing it could still take a file path and store the
output there... Not saying I necessarily prefer that, just being open
to different ideas.

ChrisMcKenzie:
I have started working on this!

purpleidea:
@ChrisMcKenzie Am particularly excited. Feel free to push a non-compiling WIP branch in case you want me to look at something early.

ChrisMcKenzie:
I am starting to get a hold on the `Res` interface, I am trying to think about how the `Watch` might be implemented I feel like in this case it should be a noop, originally I though about polling but that seems like it would cause issues!

purpleidea:
@ChrisMcKenzie What do you think about:

> It should probably support long-polling to watch.

As a backup:

> It could use the Poll metaparam.

ChrisMcKenzie:
I think I like the `metaparam` approach best as it can be disabled if need be.

purpleidea:
Since I did some reading of the HTTP libs recently, I thought I should mention that this should probably look at `*Modified-Since` headers perhaps.

purpleidea:
FWIW, this should probably be called `http:client`. We have the `http:server` stuff merged, so this makes sense. Having a client resource that can download a file for us would be really neat.

purpleidea:
@frebib asks in channel:

> This is an interesting one, and raises a possible question with mgmt's design. What is the difference between a http:client resource with send/recv and a http client function in mcl? Could you argue that they do the same thing? A resource by nature is to "enforce a state", but it's being somewhat abused here to do something else. It's widely accepted that having multiple ways to do the same thing are frowned upon, so should this be a good example of where to define what a resource should and shouldn't be used for?

Here's my answer:

This is an excellent point.

[...]

PR #198, opened 2017-06-25
WIP: http resource

ChrisMcKenzie:
This is my first pass on the http resource (#72) currently it does a naive check of the body on an interval defined by the user. I think this still needs a bit of work but it is a start.

Also I realize that this is currently in its own package I developed it that way so that I could run the tests a bit more easily. I am not sure what your take on resource being in there own packages but this helped a bunch in testing. will move back to resources if you are not on board with my thinking.

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,8 @@ import (
 	"os"
 
 	mgmt "github.com/purpleidea/mgmt/lib"
+
+	_ "github.com/purpleidea/mgmt/resources/http"
 )
 
 // These constants are some global variables that are used throughout the code.
--- a/resources/http/http.go
+++ b/resources/http/http.go
@@ -1,0 +1,304 @@
+// Mgmt
+// Copyright (C) 2013-2017+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -22,6 +22,8 @@ import (
 	"os"
 
 	mgmt "github.com/purpleidea/mgmt/lib"
+
+	_ "github.com/purpleidea/mgmt/resources/http"
```
purpleidea: I think this should go in resources.go instead. (So that people importing `resources` get it too.) If you could add a comment like: `// resource registration` or something please.

resources/http/http.go
```diff
@@ -0,0 +1,304 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
```
purpleidea: If it's supposed to be in its own package, this should be `http` right?
(Interesting that this works though... I wonder what the golang semantics are... Eg: same namespace, different package? I honestly don't know)

resources/http/http_test.go
```diff
@@ -0,0 +1,66 @@
...
+	pass, err := res.CheckApply(false)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if res.(*HTTPRes).ShaSum != responseSum {
+		t.Errorf("expected ShaSum to equal %s got %s", responseSum, res.(*HTTPRes).ShaSum)
+	}
+
+	fmt.Println(pass)
```
purpleidea: Tests should use t.Logf instead. Otherwise the output will be shown out of order.

resources/http/http.go
ChrisMcKenzie: Lol, yeah it totally should be http!

purpleidea:
Quick ping to see if you're alive and/or need any pointers. Thanks!

purpleidea:
This is stale, so closing.

PR #200, opened 2017-07-04
set the etcd max transaction limit to a reasonable default

nitinprakash96:


purpleidea:
@nitinprakash96 Thanks for the patch! This value should be a constant somewhere at the top of the file, and then used here.

You'll also need your commit messages to start with `etcd: The commit message` for this to pass.

Thanks!

purpleidea:
@nitinprakash96 I have stupidly forgotten that this patch depends on a newer version of etcd, and we are currently vendoring an older one. As a result, it can't be merged until we bump etcd. I can cherry-pick it and merge down the road if you're interested.

If so, please make sure you run `go fmt` on your code (`make gofmt` will fix it) and rebase everything into a single patch. (ping if you need tips on how to do this). Lastly the comment should be on the constant, not the usage.

And your patch failed because commit messages need to be capitalized. Eg: `etcd: Set the` not `etcd: set the`

Thanks!

nitinprakash96:
@purpleidea Ah I see now why the patch fails to build! I'll rebase and wait for further instructions.

nitinprakash96:
The build fails possibly due to the older version of etcd the we're using. Is that so @purpleidea ?

purpleidea:
On Wed, Jul 5, 2017 at 1:40 PM, Nitin Prakash <notifications@github.com> wrote:
>
> The build fails possibly due to the older version of etcd the we're using. Is that so @purpleidea ?



Correct. Plus those other small things you need to fix up. If you fix
them, then I will merge it when we bump the version.

Sorry about this, my bad for not realizing earlier when I made the ticket.

nitinprakash96:
Okay so I tried running the tests manually on my system and apparently all of the fail is caused by the older version. For example, when running gometalinter,
```
gometalinter --disable-all --enable=goimports --enable=golint --enable=gotype --enable=misspell --enable=unconvert .
```
outputs:
```
etcd/etcd.go:1660:2:warning: invalid operation: cfg (variable of type *github.com/purpleidea/mgmt/vendor/github.com/coreos/etcd/embed.Config) has no field or method MaxTxnOps (unconvert)
```

or while running go test,
```
go test github.com/purpleidea/mgmt
```
outputs:
```
# github.com/purpleidea/mgmt/etcd
etcd/etcd.go:1660: cfg.MaxTxnOps undefined (type *embed.Config has no field or method MaxTxnOps)
```

Other than that there was just `gofmt` which I fixed as per your previous suggestion. But still if there's any solution could you help me out?

purpleidea:
On Thu, Jul 6, 2017 at 8:57 AM, Nitin Prakash <notifications@github.com>
wrote:

> But still if there's any solution could you help me out?



You're good for now. Best is to work on another patch. What interests you?

nitinprakash96:
I'll take a look into other love issues and see if I can fix anything.

purpleidea:
@nitinprakash96 Sorry for the delay... If you'd like to rebase your patch, it should work now :)

Also I'd recommend removing the uint() wrapper from the constant. If needed add it below where the value is actually used!

Thanks!

purpleidea:
Turns out I'm stupid and this isn't in a released version yet... So we'll have ti wait for https://github.com/coreos/etcd/commit/ae7ddfb4838f30da1ef04ee48cb89170d7c3268f to be in a released version.

purpleidea:
Fixed in https://github.com/purpleidea/mgmt/commit/24cb2e6450bae2cfecfa7322695d6b6eeec14c19

Issue #199, opened 2017-07-01
[LOVE] Implement graph isomorphism


purpleidea:
This is an issue for a beginner to medium level golang programmer. It also requires an understanding of basic graph algorithms and the willingness to figure out how to implement a graph isomorphism algorithm. This is an algorithm that determines if two graphs are equivalent or not.

In mgmt we have an internal pgraph (graph) library. We currently have a `GraphCmp` algorithm: https://github.com/purpleidea/mgmt/blob/master/pgraph/pgraph.go#L501 which takes a VertexCmp and EdgeCmp method as input, and compares two graphs.

It has some limitations, in that it fails if there are duplicate vertices in either graph.
We'd like to replace this with a proper implementation based on a known algorithm,

This patch should also include some new tests!

Please leave a comment if you're going to work on this patch!

Thanks and happy hacking

purpleidea:
Here are a few misc. references btw:

https://github.com/networkx/networkx/blob/master/networkx/algorithms/isomorphism/isomorphvf2.py
https://www3.cs.stonybrook.edu/~algorith/implement/nauty/implement.shtml
http://www.dharwadker.org/tevet/isomorphism/
https://github.com/igraph/igraph/tree/master/src/bliss

The idea is to implement something straight forward that works, not to implement the most efficient 100,000 vertex capable monster.

AdnanLFC:
As I said, I'd like to give it a shot. Any additional discussion needed before I start working on this?

purpleidea:
@AdnanLFC I think you're good to go! Feel free to ping if you come across any implementation or design questions that you'd like answered.

Thanks!

purpleidea:
@AdnanLFC Any chance you're still working on this? Please let us know either way, thanks!

AdnanLFC:
Got pulled away by some other things... :disappointed: However, I might give it another try these days.

keur:
I'll take this.

purpleidea:
@keur Awesome-- one disclosure: this is a hard problem (you've been warned) and it's not 100% clear what's blocking us. Last I looked, this might be helpful...

https://github.com/purpleidea/mgmt/blob/f342e06ef0d0fc173dcb5320e9a376f14d561868/lang/interpret_test.go#L324

https://github.com/purpleidea/mgmt/blob/f342e06ef0d0fc173dcb5320e9a376f14d561868/lang/lang_test.go#L884

[...]

PR #201, opened 2017-07-28
pgraph: test: Added tests for DeleteEdge

AdnanLFC:
Potential solution for #175

### DIFF
--- a/pgraph/pgraph_test.go
+++ b/pgraph/pgraph_test.go
@@ -686,3 +686,84 @@ func TestSort1(t *testing.T) {
 		t.Errorf(str)
 	}
 }
+
+func TestDeleteEdge1(t *testing.T) {
+	g, _ := NewGraph("g")
+
+	v1 := NV("v1")
+	v2 := NV("v2")
+	v3 := NV("v3")
+
+	e1 := NE("e1")
+	e2 := NE("e2")
+	e3 := NE("e3")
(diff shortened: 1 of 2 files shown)


pgraph/pgraph_test.go
```diff
@@ -686,3 +686,84 @@ func TestSort1(t *testing.T) {
...
+	g.DeleteEdge(e5)
+	g.DeleteEdge(e6)
+
+	ie := g.IncomingGraphEdges(v1)
+	oe := g.OutgoingGraphEdges(v1)
+
+	if !reflect.DeepEqual(ie, []Edge{e7}) {
+		res := ""
+		for _, e := range ie {
+			res += e.String() + " "
```
purpleidea: Patch LGTM, thanks!

I will make one nitpick for future code (no need to change this patch) which is that the foo + " " bit, leaves a trailing space at the end. Fine for tests, as long as you're aware of the limitation for general code :)

Thanks again, and LMK if you have any questions. If you are planning to work on https://github.com/purpleidea/mgmt/issues/199 as you mentioned in IRC, please leave a comment in the issue.

Cheers!

AdnanLFC: Thanks for the tip, will keep that in mind. :)

purpleidea:
Now merged, thanks again and congrats on first patch! (Sorry for merge delay.)

PR #206, opened 2017-08-19
Problem: Msg Priority field lacks validation

dkliban:
Solution: validate that Msg Priority field is one of the following values:
"Emerg", "Alert", "Crit", "Err", "Warning", "Notice", "Info", "Debug"

### DIFF
--- a/resources/msg.go
+++ b/resources/msg.go
@@ -69,6 +69,18 @@ func (obj *MsgRes) Validate() error {
 			return fmt.Errorf("fields cannot begin with _")
 		}
 	}
+	switch obj.Priority {
+	case "Emerg":
+	case "Alert":
+	case "Crit":
+	case "Err":
+	case "Warning":
+	case "Notice":
+	case "Info":
+	case "Debug":
+	default:
+		return fmt.Errorf("invalid Priority '%s'", obj.Priority)
+	}
 	return obj.BaseRes.Validate()
 }
 
@@ -94,7 +106,6 @@ func (obj *MsgRes) updateStateOK() {
 }
 
 // JournalPriority converts a string description to a numeric priority.
-// XXX: Have Validate() make sure it actually is one of these.
 func (obj *MsgRes) journalPriority() journal.Priority {
 	switch obj.Priority {
 	case "Emerg":
--- a/resources/msg_test.go
+++ b/resources/msg_test.go
@@ -1,0 +1,54 @@
+// Mgmt
+// Copyright (C) 2013-2017+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"testing"
+)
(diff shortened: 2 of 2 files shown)


resources/msg.go
```diff
@@ -69,6 +69,16 @@ func (obj *MsgRes) Validate() error {
 			return fmt.Errorf("fields cannot begin with _")
 		}
 	}
+        ValidPriorities := []string{"Emerg", "Alert", "Crit", "Err", "Warning", "Notice", "Info", "Debug"}
```
purpleidea: uppercase things are usually public. This should start with a lowercase name.

resources/msg.go
```diff
@@ -69,6 +69,16 @@ func (obj *MsgRes) Validate() error {
...
+        priorityIsValid := false
+        for _, element := range ValidPriorities {
+            if strings.Compare(obj.Priority, element) == 0 {
+                priorityIsValid = true
+            }
+        }
+        if priorityIsValid == false {
```
purpleidea: I think this whole approach could be simplified... How about just a switch statement: https://tour.golang.org/flowcontrol/9

resources/msg.go
```diff
@@ -69,6 +69,16 @@ func (obj *MsgRes) Validate() error {
 			return fmt.Errorf("fields cannot begin with _")
 		}
 	}
+        ValidPriorities := []string{"Emerg", "Alert", "Crit", "Err", "Warning", "Notice", "Info", "Debug"}
+        priorityIsValid := false
+        for _, element := range ValidPriorities {
+            if strings.Compare(obj.Priority, element) == 0 {
```
purpleidea: Please read the recommendation on this methods docs: https://golang.org/pkg/strings/#Compare

resources/msg.go
```diff
@@ -114,7 +123,6 @@ func (obj *MsgRes) journalPriority() journal.Priority {
 	case "Debug":
 		return journal.PriDebug
 	}
-	return journal.PriNotice
```
purpleidea: Why did you remove this?

resources/msg_test.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func TestMsgValidate2(t *testing.T) {
+        r1 := &MsgRes{
+                BaseRes: BaseRes{
+                        Name: "msg1",
+                        Kind: "msg",
+                        MetaParams: DefaultMetaParams,
+                },
+                Priority:   "Priority",
```
purpleidea: Tests look good! But maybe make the bad value slightly more obvious?

resources/msg.go
dkliban: i made the assumption that it was there because you didn't have validation and wanted it to default to Notice when a match was not found.

resources/msg.go
dkliban: ah yes!

resources/msg.go
dkliban: thanks

resources/msg.go
dkliban: much simpler :)

resources/msg.go
```diff
@@ -69,6 +69,18 @@ func (obj *MsgRes) Validate() error {
 			return fmt.Errorf("fields cannot begin with _")
 		}
 	}
+	switch priority := obj.Priority; priority {
```
purpleidea: why all this instead of just `switch obj.Priority {` ?

resources/msg.go
```diff
@@ -69,6 +69,18 @@ func (obj *MsgRes) Validate() error {
...
+	case "Emerg":
+	case "Alert":
+	case "Crit":
+	case "Err":
+	case "Warning":
+	case "Notice":
+	case "Info":
+	case "Debug":
+	default:
+		return fmt.Errorf("invalid Priority")
```
purpleidea: LGTM, maybe add a invalid Priority
```
"`%s`", obj.Priority)

to the end?

purpleidea:
@dkliban Hey! Merged :) Sorry I didn't do this sooner, I didn't realize you had pushed a new patch! GH is ridiculously lame here and doesn't send notifications.

Thanks! LMK if you're ready for something more interesting!

Issue #204, opened 2017-08-07
[LOVE] expand ~username type strings properly


purpleidea:
This should be a beginner level patch for someone who is either new-ish to golang, or new to mgmt. It is meant to encourage new programmers to get involved with mgmt. If you're going to work on this issue, please leave a comment below.

We have a virt resource in mgmt, As you can see here: https://github.com/purpleidea/mgmt/blob/master/resources/virt.go#L1173 the `expandHome` doesn't support the common `~username` extension that works in `cd`.

This patch would add the necessary additions, and optionally a small test or two in virt_test.go

If this takes you more than two hours after you've got the hello world example building, or if you have any questions, please ping us in #mgmtconfig on Freenode IRC.

Thanks and happy hacking!

purpleidea:
Fixed in https://github.com/purpleidea/mgmt/pull/210

PR #210, opened 2017-09-06
virt: Added support for ~username format into expandHome

artmello:
- Enabled expandHome to expand both ~/ and ~username/ paths
- Added some unit tests for expandHome
- Fixed issue #204

### DIFF
--- a/resources/virt.go
+++ b/resources/virt.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"os/user"
 	"path"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


purpleidea:
@artmello Thanks for your patch! Doing a quick review now...

resources/virt.go
```diff
@@ -1168,15 +1169,25 @@ func isNotFound(err error) bool {
...
+		if err != nil {
+			return p, fmt.Errorf("can't expand ~ into home directory")
+		}
+		return path.Join(usr.HomeDir, p[len("~/"):]), nil
 	}
-	usr, err := user.Current()
-	if err != nil {
-		return p, fmt.Errorf("can't expand ~ into home directory")
+
+	r, _ := regexp.Compile("~([^/]+)/")
```
purpleidea: Do we need to check this error? I'd recommend putting a short comment above the regexp explaining the logic if you'd like.

More of a nitpick really, but overall patch looks great, thanks!

purpleidea:
@artmello Please rebase into a single patch and remove the merge commit. Ping if you need tips on doing this. Left one review item: https://github.com/purpleidea/mgmt/pull/210#discussion_r137346877

purpleidea:
@artmello Patch looks pretty perfect, but now there's one new pattern you need to learn...

Instead of:

```
	if err != nil {
		return p, fmt.Errorf("can't compile regexp: %s", err)
	}

```

You can do:

```
	if err != nil {
		return p, errwrap.Wrapf(err, "can't compile regexp")
	}
```

This passes through the previous error as an error, instead of forcing into a string so early!

Please fix this up, and rebase, and it looks perfect for merging :)

purpleidea:
Patch looks perfect, thank you and welcome to the project!

Rebased to master and merged as 80b8f8740fb91152e5e63cd288845094acc26412

PR #217, opened 2017-09-11
WIP Add a User Resource

jonathangold:
For early review.
Don't Merge.

### DIFF
--- a/resources/user.go
+++ b/resources/user.go
@@ -1,0 +1,310 @@
+// Mgmt
+// Copyright (C) 2013-2017+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"fmt"
+	"log"
+	"os/exec"
+	"os/user"
+	"strconv"
+	"syscall"
+
+	"github.com/purpleidea/mgmt/recwatch"
+
+	errwrap "github.com/pkg/errors"
+)
+
+func init() {
+	RegisterResource("user", func() Res { return &UserRes{} })
+}
+
+const passwdFile = "/etc/passwd"
+
+// UserRes is a user account resource.
+type UserRes struct {
+	BaseRes           `yaml:",inline"`
+	State             string  `yaml:"state"` // state: exists, absent
+	UID               *uint32 `yaml:"uid"`
+	GID               *uint32 `yaml:"gid"`
+	HomeDir           *string `yaml:"homedir"`
+	AllowDuplicateUID bool    `yaml:"allowduplicateuid"`
+
+	recWatcher *recwatch.RecWatcher
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *UserRes) Default() Res {
+	return &UserRes{
+		BaseRes: BaseRes{
+			MetaParams: DefaultMetaParams, // force a default
+		},
+	}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *UserRes) Validate() error {
+	if obj.State != "exists" && obj.State != "absent" {
+		return fmt.Errorf("State must be 'exists' or 'absent'")
+	}
+	return obj.BaseRes.Validate()
+}
+
+// Init initializes the resource.
+func (obj *UserRes) Init() error {
+	return obj.BaseRes.Init() // call base init, b/c we're overriding
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *UserRes) Watch() error {
+	var err error
+	obj.recWatcher, err = recwatch.NewRecWatcher(passwdFile, false)
+	if err != nil {
+		return err
+	}
+	defer obj.recWatcher.Close()
+
+	// notify engine that we're running
+	if err := obj.Running(); err != nil {
+		return err // bubble up a NACK...
+	}
+
+	var send = false // send event?
+	var exit *error
+
+	for {
+		if obj.debug {
+			log.Printf("Watching: %s", passwdFile) // attempting to watch...
+		}
+
+		select {
+		case event, ok := <-obj.recWatcher.Events():
+			if !ok { // channel shutdown
+				return nil
+			}
+			if err := event.Error; err != nil {
+				return errwrap.Wrapf(err, "Unknown %s watcher error", obj)
+			}
+			if obj.debug { // don't access event.Body if event.Error isn't nil
+				log.Printf("%s: Event(%s): %v", obj, event.Body.Name, event.Body.Op)
+			}
+			send = true
+			obj.StateOK(false) // dirty
+
+		case event := <-obj.Events():
+			if exit, send = obj.ReadEvent(event); exit != nil {
+				return *exit // exit
+			}
+			//obj.StateOK(false) // dirty // these events don't invalidate state
+		}
+
+		// do all our event sending all together to avoid duplicate msgs
+		if send {
+			send = false
+			obj.Event()
+		}
+	}
+}
+
+// CheckApply method for User resource.
+func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
+	log.Printf("%s: CheckApply(%t)", obj, apply)
+
+	var exists = true
+	usr, err := user.Lookup(obj.GetName())
+	if err != nil {
+		if _, ok := err.(user.UnknownUserError); !ok {
+			return false, errwrap.Wrapf(err, "error looking up user")
+		}
+		log.Printf("the user: %s does not exist", obj.GetName())
+		exists = false
+	}
+
+	if obj.AllowDuplicateUID == false && obj.UID != nil {
+		existingUID, err := user.LookupId(strconv.Itoa(int(*obj.UID)))
+		if err != nil {
+			if _, ok := err.(user.UnknownUserIdError); !ok {
+				return false, errwrap.Wrapf(err, "error looking up UID")
+			}
+		} else if existingUID.Username != obj.GetName() {
+			return false, fmt.Errorf("the requested UID is already taken")
+		}
+	}
+
+	if obj.State == "absent" && !exists {
+		return true, nil
+	}
+
+	if usercheck := true; exists && obj.State == "exists" {
+		intUID, err := strconv.Atoi(usr.Uid)
+		if err != nil {
+			return false, errwrap.Wrapf(err, "error casting UID to int")
+		}
+		intGID, err := strconv.Atoi(usr.Gid)
+		if err != nil {
+			return false, errwrap.Wrapf(err, "error casting GID to int")
+		}
+		if obj.UID != nil && int(*obj.UID) != intUID {
+			usercheck = false
+		}
+		if obj.GID != nil && int(*obj.GID) != intGID {
+			usercheck = false
+		}
+		if obj.HomeDir != nil && *obj.HomeDir != usr.HomeDir {
+			usercheck = false
+		}
+		if usercheck {
+			return true, nil
+		}
+	}
+
+	if !apply {
+		return false, nil
+	}
+
+	var cmdName string
+	var args []string
+	if obj.State == "exists" {
+		if exists {
+			cmdName = "usermod"
+			log.Printf("modifying user: %s", obj.GetName())
+		} else {
+			cmdName = "useradd"
+			log.Printf("adding user: %s", obj.GetName())
+		}
+		if obj.AllowDuplicateUID {
+			args = append(args, "--non-unique")
+		}
+		if obj.UID != nil {
+			args = append(args, "-u", fmt.Sprintf("%d", *obj.UID))
+		}
+		if obj.GID != nil {
+			args = append(args, "-g", fmt.Sprintf("%d", *obj.GID))
+		}
+		if obj.HomeDir != nil {
+			args = append(args, "-d", *obj.HomeDir)
+		}
+	}
+	if obj.State == "absent" {
+		cmdName = "userdel"
+	}
+
+	args = append(args, obj.GetName())
+
+	cmd := exec.Command(cmdName, args...)
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+		Pgid:    0,
+	}
+	if err := cmd.Run(); err != nil {
+		return false, errwrap.Wrapf(err, "cmd failed to run")
+	}
+
+	return false, nil
+}
+
+// UserUID is the UID struct for UserRes.
+type UserUID struct {
+	BaseUID
+	name string
+}
+
+// UIDs includes all params to make a unique identification of this object.
+// Most resources only return one, although some resources can return multiple.
+func (obj *UserRes) UIDs() []ResUID {
+	x := &UserUID{
+		BaseUID: BaseUID{Name: obj.GetName(), Kind: obj.GetKind()},
+		name:    obj.Name,
+	}
+	return []ResUID{x}
+}
+
+// GroupCmp returns whether two resources can be grouped together or not.
+func (obj *UserRes) GroupCmp(r Res) bool {
+	_, ok := r.(*UserRes)
+	if !ok {
+		return false
+	}
+	return false
+}
+
+// Compare two resources and return if they are equivalent.
+func (obj *UserRes) Compare(r Res) bool {
+	// we can only compare UserRes to others of the same resource kind
+	res, ok := r.(*UserRes)
+	if !ok {
+		return false
+	}
+	if !obj.BaseRes.Compare(res) { // call base Compare
+		return false
+	}
+	if obj.Name != res.Name {
+		return false
+	}
+	if obj.State != res.State {
+		return false
+	}
+	if (obj.UID == nil) != (res.UID == nil) {
+		return false
+	}
+	if obj.UID != nil && res.UID != nil {
+		if *obj.UID != *res.UID {
+			return false
+		}
+	}
+	if (obj.GID == nil) != (res.GID == nil) {
+		return false
+	}
+	if obj.GID != nil && res.GID != nil {
+		if *obj.GID != *res.UID {
+			return false
+		}
+	}
+	if (obj.HomeDir == nil) != (res.HomeDir == nil) {
+		return false
+	}
+	if obj.HomeDir != nil && res.HomeDir != nil {
+		if *obj.HomeDir != *obj.HomeDir {
+			return false
+		}
+	}
+	if obj.AllowDuplicateUID != res.AllowDuplicateUID {
+		return false
+	}
+	return true
+}
+
+// UnmarshalYAML is the custom unmarshal handler for this struct.
+// It is primarily useful for setting the defaults.
+func (obj *UserRes) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type rawRes UserRes // indirection to avoid infinite recursion
+
+	def := obj.Default()      // get the default
+	res, ok := def.(*UserRes) // put in the right format
+	if !ok {
+		return fmt.Errorf("could not convert to UserRes")
+	}
+	raw := rawRes(*res) // convert; the defaults go here
+
+	if err := unmarshal(&raw); err != nil {
+		return err
+	}
+
+	*obj = UserRes(raw) // restore from indirection with type conversion!
+	return nil
+}
--- a/examples/user1.yaml
+++ b/examples/user1.yaml
@@ -1,0 +1,9 @@
+---
+graph: mygraph
+resources:
+  user:
+  - name: testuser
+    uid: 1002
+    gid: 100
+    state: exists
+edges: []
--- a/yamlgraph/gconfig.go
+++ b/yamlgraph/gconfig.go
@@ -68,6 +68,7 @@ type Resources struct {
 	Pkg      []*resources.PkgRes      `yaml:"pkg"`
 	Svc      []*resources.SvcRes      `yaml:"svc"`
 	Timer    []*resources.TimerRes    `yaml:"timer"`
+	User     []*resources.UserRes     `yaml:"user"`
 	Virt     []*resources.VirtRes     `yaml:"virt"`
 }
 


resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"fmt"
+	errwrap "github.com/pkg/errors"
```
purpleidea: Imports should be separated by a single line of whitespace in up to three "groups".

1st) Builtin packages
2nd) Internal packages
3rd) External packages

Here is an example of that: https://github.com/purpleidea/mgmt/blob/0df4824a56678a8f2173838ff58817b22feefee9/resources/resources.go#L21

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+)
+
+func init() {
+	RegisterResource("user", func() Res { return &UserRes{} })
+}
+
+//TODO
+type UserRes struct {
+	BaseRes `yaml:",inline"`
+	Gid     string `yaml:"gid"`
```
purpleidea: Nit: order these fields in a logical order. Probably something like: State, Uid, Gid, Home...

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+	HomeDir string `yaml:"homedir"`
+	State   string `yaml:"state"` // state: exists, absent
+	Uid     string `yaml:"uid"`
+	exec    *ExecRes
```
purpleidea: Probably also worth adding a newline between the end of the public fields and the start of the private ones.
For longer structs you might want to add newlines between different related groups of fields for example.

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+	if obj.State != "exists" && obj.State != "absent" {
+		return fmt.Errorf("State must be 'exists' or 'absent'")
+	}
+	return obj.BaseRes.Validate()
+}
+
+// Init initializes the resource.
+func (obj *UserRes) Init() error {
+	obj.exec = &ExecRes{}
+	obj.exec.Init()
```
purpleidea: Great, but you forgot to check return value of exec's Init()

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
+	log.Printf("%s: CheckApply(%t)", obj, apply)
+
+	var exists = true
+
+	usr, err := user.Lookup(obj.GetName())
+
+	log.Printf("%+v", usr)
+
+	if err != nil {
```
purpleidea: Not entirely true as per the docs:

>  Lookup looks up a user by username. If the user cannot be found, the returned error is of type UnknownUserError. 

So you only want to do that if the error is that type. Otherwise for "regular" errors we errwrap return an error!

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+		log.Printf("the user: %s does not exist", obj.GetName())
+		exists = false
+	}
+
+	if obj.State == "exists" && exists &&
```
purpleidea: This won't scale with the addition of more properties. Try and think of a better way to refactor this. Ping if you're out of ideas.

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if obj.State == "absent" && !exists {
+		return true, nil
+	}
+
+	if !apply {
+		return false, nil
+	}
+
+	var cmd string
```
purpleidea: This is good, building up the command line, but I think it might be more elegant to have a list of strings, and then join them in the end.

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+			cmd += " -u " + obj.Uid
+		}
+		obj.exec.Cmd = cmd
+	}
+
+	if obj.State == "absent" {
+		obj.exec.Cmd = "userdel " + obj.GetName()
+	}
+
+	if _, err := obj.exec.CheckApply(apply); err != nil {
```
purpleidea: why drop the first return value? You'll need that to know what to return if it's not an error.

resources/user.go
```diff
@@ -0,0 +1,212 @@
...
+		return false, errwrap.Wrapf(err, "nested exec failed %s", cmd)
+	}
+	return false, nil
+}
+
+// USERUID is the UID struct for UserRes.
```
purpleidea: CAPS

jonathangold:
Thanks. I'll get on this asap. PS: it does work as expected.

resources/user.go
jonathangold: Yeah. This is not very elegant, but I wanted to get it working. Maybe a for loop to iterate through the fields would work.

resources/user.go
jonathangold: I initially removed the check because it segfaults when I run it:. What am I doing wrong?

    if err := obj.exec.Init(); err != nil {
        return err
    }

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x482a2d]

goroutine 141 [running]:
sync.(*Mutex).Lock(0x0)
        /usr/lib/go-1.8/src/sync/mutex.go:48 +0x2d
github.com/purpleidea/mgmt/resources.(*BaseRes).ProcessExit(0xc4201d4000)
        /home/jon/gopath/src/github.com/purpleidea/mgmt/resources/resources.go:439 +0x32
github.com/purpleidea/mgmt/resources.(*BaseRes).Worker(0xc4201d4000, 0x0, 0x0)
        /home/jon/gopath/src/github.com/purpleidea/mgmt/resources/actions.go:515 +0xb1
github.com/purpleidea/mgmt/resources.(*MGraph).Start.func2(0xc420284780, 0x7fa389027350, 0xc4201d4000)
        /home/jon/gopath/src/github.com/purpleidea/mgmt/resources/mgraph.go:163 +0x17a
created by github.com/purpleidea/mgmt/resources.(*MGraph).Start
        /home/jon/gopath/src/github.com/purpleidea/mgmt/resources/mgraph.go:168 +0x716
jon@jon-ThinkPad-E470:~/mgmt (feat/user_res)$

resources/user.go
jonathangold: Why/where do I need to return anything if it's not an error?

jonathangold:
@purpleidea: everything is working as planned. The watch function should probably be moved to util as you suggested. I still need to make sure the console output makes sense, and address the way CheckApply checks the state of the user as per your previous comment.

resources/user.go
```diff
@@ -0,0 +1,242 @@
...
+
+// Init initializes the resource.
+func (obj *UserRes) Init() error {
+	return obj.BaseRes.Init() // call base init, b/c we're overriding
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *UserRes) Watch() error {
+	var err error
+	obj.recWatcher, err = recwatch.NewRecWatcher("/etc/passwd", false)
```
purpleidea: /etc/passwd should be a const at the top after the imports and init

resources/user.go
```diff
@@ -0,0 +1,242 @@
...
+	if err := obj.Running(); err != nil {
+		return err // bubble up a NACK...
+	}
+
+	var send = false // send event?
+	var exit *error
+
+	for {
+		if obj.debug {
+			log.Printf("Watching: /etc/passwd") // attempting to watch...
```
purpleidea: use the const...

resources/user.go
```diff
@@ -0,0 +1,242 @@
...
+func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
+	log.Printf("%s: CheckApply(%t)", obj, apply)
+
+	var exists = true
+
+	usr, err := user.Lookup(obj.GetName())
+
+	log.Printf("%+v", usr)
+
+	if err != nil && err == err.(user.UnknownUserError) {
```
purpleidea: 1) You forgot to return err if this is not of the special type...

2) The type conversion syntax is:

err, ok = err.(user.UnknownUserError)

if ok, then you can look at error or know it's now been converted to that type...

resources/user.go
purpleidea: Try just checking one property at a time, and marking as "dirty" (or just returning) if wrong (for checkOnly) or handling below. Make sense?

resources/user.go
```diff
@@ -0,0 +1,242 @@
...
+
+	}
+
+	if obj.State == "absent" {
+		cmd = "userdel"
+	}
+
+	args = append(args, obj.GetName())
+
+	usercmd := exec.Command(cmd, args...)
```
purpleidea: You'll need to add these: https://github.com/purpleidea/mgmt/blob/master/resources/exec.go#L206 IIRC

resources/user.go
```diff
@@ -0,0 +1,242 @@
...
+
+func init() {
+	RegisterResource("user", func() Res { return &UserRes{} })
+}
+
+// UserRes is a user account resource
+type UserRes struct {
+	BaseRes `yaml:",inline"`
+	State   string `yaml:"state"` // state: exists, absent
+	UID     string `yaml:"uid"`
```
purpleidea: Uid*
(to match same pattern as Gid)

purpleidea: Both of these should be the right kind of uint's (look up in the golang docs for the size, and see what /etc/passwd allows as max int) it's probably uint64, but I didn't check, not string

resources/user.go
```diff
@@ -0,0 +1,242 @@
...
+	var cmd string
+	var args []string
+	if obj.State == "exists" {
+		if !exists {
+			cmd = "useradd"
+		}
+		if exists {
+			cmd = "usermod"
+		}
+		if obj.UID != "" {
```
purpleidea: If these are now int's should they instead be *uint64 ? IOW, will we ever want to set the 0 uid in our resource or not?

jonathangold:
@purpleidea: I've implemented the changes. I'm not sure why test-headerfmt is failing all of a sudden. I had to change Uid and Gid to all caps since golint complains that Uid should be UID. Everything seems to be working properly, but I'm sure there are still a few things to clean up.

purpleidea:
On Thu, Sep 14, 2017 at 5:03 AM, jonathangold <notifications@github.com> wrote:
> @purpleidea: I've implemented the changes. I'm not sure why test-headerfmt
> is failing all of a sudden.

This is because we changed the license, so that header text needs to
match what's in main.go

> I had to change Uid and Gid to all caps since
> golint complains that Uid should be UID. Everything seems to be working
> properly, but I'm sure there are still a few things to clean up.

Sounds good, will review shortly.

jonathangold:
> This is because we changed the license, so that header text needs to match what's in main.go

Yeah, figured out as soon as I wrote my comment.

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/purpleidea/mgmt/recwatch"
+
+	errwrap "github.com/pkg/errors"
+)
+
+func init() {
+	RegisterResource("user", func() Res { return &UserRes{} })
+}
+
+const file = "/etc/passwd"
```
purpleidea: let's name this passwdFile or userPasswdFile or something. This will help avoid collisions since the namespace is shared with other resources and this is pretty unique.

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// Init initializes the resource.
+func (obj *UserRes) Init() error {
+	return obj.BaseRes.Init() // call base init, b/c we're overriding
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *UserRes) Watch() error {
+
```
purpleidea: nit: get rid of this extra nl please

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+	var exists = true
+
+	usr, err := user.Lookup(obj.GetName())
+	if err != nil {
+		if err == err.(user.UnknownUserError) {
+			log.Printf("the user: %s does not exist", obj.GetName())
+			exists = false
+		} else {
+			return false, fmt.Errorf("error looking up user")
```
purpleidea: errwrap since we're carrying the err

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+	}
+}
+
+// CheckApply method for User resource. Does nothing, returns happy!
+func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
+	log.Printf("%s: CheckApply(%t)", obj, apply)
+
+	var exists = true
+
```
purpleidea: kill this extra nl please

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+		if err == err.(user.UnknownUserError) {
+			log.Printf("the user: %s does not exist", obj.GetName())
+			exists = false
+		} else {
+			return false, fmt.Errorf("error looking up user")
+		}
+	}
+
+	if obj.AllowDuplicateUID == false {
+		existingUID, err := user.LookupId(strconv.Itoa(int(*obj.UID)))
```
purpleidea: you only want to do this if obj.UID != nil, because if its not pointing to anything, the user didn't specify the UID. Without this check, this would be a panic.

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+		if err == nil && existingUID.Username != obj.GetName() {
+			return false, fmt.Errorf("the uid: %d is already taken", *obj.UID)
+		}
+	}
+
+	if exists && obj.State == "exists" {
+		usercheck := true
+		if obj.GID != nil && strconv.Itoa(int(*obj.GID)) != usr.Gid {
```
purpleidea: 2x: similar need to check != nil before using the pointer...

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		if exists {
+			cmdName = "usermod"
+			log.Printf("modifying user: %s", obj.GetName())
+		}
+		if obj.AllowDuplicateUID == true {
+			args = append(args, "--non-unique")
+		}
+		if obj.UID != nil {
+			args = append(args, "-u", strconv.Itoa(int(*obj.UID)))
```
purpleidea: instead of all the `strconv.Itoa(int(x)))` everywhere, you might want to consider: fmt.Sprintf("%d", x) instead.

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+	return []ResUID{x}
+}
+
+// GroupCmp returns whether two resources can be grouped together or not.
+func (obj *UserRes) GroupCmp(r Res) bool {
+	_, ok := r.(*UserRes)
+	if !ok {
+		return false
+	}
+	return false // TODO: this is doable!
```
purpleidea: remove these two incorrect TODO comments unless I'm mistaken. I think it came from copy-pasta

resources/user.go
```diff
@@ -0,0 +1,275 @@
...
+// Compare two resources and return if they are equivalent.
+func (obj *UserRes) Compare(r Res) bool {
+	// we can only compare UserRes to others of the same resource kind
+	res, ok := r.(*UserRes)
+	if !ok {
+		return false
+	}
+	if !obj.BaseRes.Compare(res) { // call base Compare
+		return false
+	}
```
purpleidea: Add in the compare checks for the new struct fields :) (See other resources for an example... careful comparing the pointers)

resources/user.go
```diff
@@ -0,0 +1,298 @@
...
+	errwrap "github.com/pkg/errors"
+)
+
+func init() {
+	RegisterResource("user", func() Res { return &UserRes{} })
+}
+
+const passwdFile = "/etc/passwd"
+
+// UserRes is a user account resource
```
purpleidea: needs trailing period. (doc strings are sentences)

resources/user.go
```diff
@@ -0,0 +1,298 @@
...
+
+		// do all our event sending all together to avoid duplicate msgs
+		if send {
+			send = false
+			obj.Event()
+		}
+	}
+}
+
+// CheckApply method for User resource. Does nothing, returns happy!
```
purpleidea: Comment is old.

resources/user.go
```diff
@@ -0,0 +1,298 @@
...
+func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
+	log.Printf("%s: CheckApply(%t)", obj, apply)
+
+	var exists = true
+	usr, err := user.Lookup(obj.GetName())
+	if err != nil {
+		if err == err.(user.UnknownUserError) {
```
purpleidea: This doesn't make sense... You probably want:

```
if _, ok := err.(user.UnknownUserError); ok {
```

(check if err is of that type)

You can inline the else of this test with !ok here as well.

resources/user.go
```diff
@@ -0,0 +1,298 @@
...
+			log.Printf("the user: %s does not exist", obj.GetName())
+			exists = false
+		} else {
+			return false, errwrap.Wrapf(err, "error looking up user")
+		}
+	}
+
+	if obj.AllowDuplicateUID == false && obj.UID != nil {
+		existingUID, err := user.LookupId(strconv.Itoa(int(*obj.UID)))
```
purpleidea: You never handled the err case! If LookupId fails because of some other reason, then we need to error out CheckApply, otherwise the engine won't know something bad happened...

resources/user.go
```diff
@@ -0,0 +1,298 @@
...
+		if err == nil && existingUID.Username != obj.GetName() {
+			return false, fmt.Errorf("the uid: %d is already taken", *obj.UID)
+		}
+	}
+
+	if exists && obj.State == "exists" {
+		usercheck := true
+		if obj.GID != nil {
```
purpleidea: golang has short circuit operators, so you can onliner these two...

resources/user.go
```diff
@@ -0,0 +1,298 @@
...
+			if strconv.Itoa(int(*obj.GID)) != usr.Gid {
+				usercheck = false
+			}
+		}
+		if obj.UID != nil {
+			if strconv.Itoa(int(*obj.UID)) != usr.Uid {
+				usercheck = false
+			}
+		}
+		if obj.HomeDir != "" && obj.HomeDir != usr.HomeDir {
```
purpleidea: I just noticed that we might have three states:

1) HomeDir is some path "/home/james"
2) HomeDir is empty ""
3) Don't touch HomeDir (home dir is nil)

So I think this should be a *string... Do you agree?

resources/user.go
```diff
@@ -0,0 +1,298 @@
...
+	args = append(args, obj.GetName())
+
+	cmd := exec.Command(cmdName, args...)
+
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+		Pgid:    0,
+	}
+	if err := cmd.Run(); err != nil {
+		return true, nil
```
purpleidea: errwrap false, err ???

jonathangold:
@purpleidea: No worries. Just pushed the changes. All part of the learning process I suppose.

purpleidea:
@jonathangold linter failed: https://travis-ci.org/purpleidea/mgmt/jobs/276094028#L929

Also commit message should be something like: `resources: Add user resource

resources/user.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// CheckApply method for User resource.
+func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
+	log.Printf("%s: CheckApply(%t)", obj, apply)
+
+	var exists = true
+	usr, err := user.Lookup(obj.GetName())
+	if err != nil {
+		if _, ok := err.(user.UnknownUserError); ok {
```
purpleidea: Read previous review comment please...

resources/user.go
purpleidea: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

resources/user.go
```diff
@@ -0,0 +1,305 @@
...
+			if existingUID.Username != obj.GetName() {
+				return false, fmt.Errorf("the uid: %d is already taken", *obj.UID)
+			} else {
+				return false, errwrap.Wrapf(err, "error looking up uid")
+			}
+		}
+	}
+
+	if usercheck := true; exists && obj.State == "exists" {
+		if obj.GID != nil {
```
purpleidea: See previous review comment...

resources/user.go
```diff
@@ -0,0 +1,305 @@
...
+	RegisterResource("user", func() Res { return &UserRes{} })
+}
+
+const passwdFile = "/etc/passwd"
+
+// UserRes is a user account resource.
+type UserRes struct {
+	BaseRes           `yaml:",inline"`
+	State             string  `yaml:"state"` // state: exists, absent
+	AllowDuplicateUID bool    `yaml:"allowduplicateuid"`
```
purpleidea: Nitpick, but I'd recommend putting the option modifier last in the list, sorry. More logical in that it's a thing that's less commonly used. (Same order in Compare)

resources/user.go
```diff
@@ -0,0 +1,305 @@
...
+	if obj.Name != res.Name {
+		return false
+	}
+	if obj.State != res.State {
+		return false
+	}
+	if obj.AllowDuplicateUID != res.AllowDuplicateUID {
+		return false
+	}
+	if obj.UID != nil && res.UID != nil {
```
purpleidea: what happens if obj.UID is nil and res.UID is not nil?

the clean solution to this can be found in file.go

jonathangold:
Ok. I'll take a stab at it after dinner.

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	if obj.AllowDuplicateUID == false && obj.UID != nil {
+		existingUID, err := user.LookupId(strconv.Itoa(int(*obj.UID)))
+		if err != nil {
+			if _, ok := err.(user.UnknownUserIdError); !ok {
+				return false, errwrap.Wrapf(err, "error looking up UID")
+			}
+		} else if existingUID.Username != obj.GetName() {
+			return false, errwrap.Wrapf(err, "the requested UID is already taken")
```
purpleidea: what `err` are we wrapping here ;) ? don't you mean fmt.Errorf instead?
(more important to make sure you know the difference than to just fix)

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+	}
+
+	if usercheck := true; exists && obj.State == "exists" {
+		if obj.GID != nil && strconv.Itoa(int(*obj.GID)) != usr.Gid {
```
purpleidea: two small comments, which you don't *have* to fix for this to *work*, but if you're motivated you probably should. They're good learning examples of "even though it compiles and passes tests, it should probably be different".
i mention this as a positive mentoring tip, but only read further if you promise to remember:

1) in the struct and the code we follow the order: `uid`, `gid`, `homedir`... as a result, it's usually classy to be consistent where it makes sense to be. This is one case, except you did gid, and then uid.

2) After reading up on the docs, in turns out the usr.Uid is a string, since that's the common type for cross platform compat. While you're comparison is correct, and will probably work, you converted our int to a string and then compared the two. It's probably more correct to convert the string to and int, and then compare the ints.

Why is that? Well if the library returned an empty string, and we had a 400, then they'd not match, but we'd silently be eating the error, where as if you perform a string->int you'll have another error to check if it can't be converted, and you'll catch that lib error and return "can't convert uid to uint32" or whatever...

Make sense? If not, ping before patching, and make sure you get the error handling perfect :) I believe in you!

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+		if obj.HomeDir != nil && *obj.HomeDir != usr.HomeDir {
+			usercheck = false
+		}
+		if usercheck {
+			return true, nil
+		}
+	}
+
+	if obj.State == "absent" && !exists {
```
purpleidea: This is good! However, it's probably better to move this check above the "usercheck" section, because it's cheaper and we can handle the simple case first.

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+		if obj.GID != nil {
+			args = append(args, "-g", fmt.Sprintf("%d", *obj.GID))
+		}
+		if obj.HomeDir != nil {
+			args = append(args, "-d", *obj.HomeDir)
+		}
+
+	}
+
```
purpleidea: lose the nl since these two are related

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+	if !apply {
+		return false, nil
+	}
+
+	var cmdName string
+	var args []string
+	if obj.State == "exists" {
+		if !exists {
```
purpleidea: replace with: 
```
if exists {

} else {

}
```
instead of two separate if blocks

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+			cmdName = "useradd"
+			log.Printf("adding user: %s", obj.GetName())
+		}
+		if exists {
+			cmdName = "usermod"
+			log.Printf("modifying user: %s", obj.GetName())
+		}
+		if obj.AllowDuplicateUID == true {
```
purpleidea:
```
 +		if obj.AllowDuplicateUID == true { 
```

is identical to

```
 +		if obj.AllowDuplicateUID { 
```

so use the simpler version since there's no readability advantage :)

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+		if obj.UID != nil {
+			args = append(args, "-u", fmt.Sprintf("%d", *obj.UID))
+		}
+		if obj.GID != nil {
+			args = append(args, "-g", fmt.Sprintf("%d", *obj.GID))
+		}
+		if obj.HomeDir != nil {
+			args = append(args, "-d", *obj.HomeDir)
+		}
+
```
purpleidea: kill this nl

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	if obj.State == "absent" {
+		cmdName = "userdel"
+	}
+
+	args = append(args, obj.GetName())
+
+	cmd := exec.Command(cmdName, args...)
+
```
purpleidea: loose the nl

resources/user.go
```diff
@@ -0,0 +1,306 @@
...
+	if !obj.BaseRes.Compare(res) { // call base Compare
+		return false
+	}
+	if obj.Name != res.Name {
+		return false
+	}
+	if obj.State != res.State {
+		return false
+	}
+	if (obj.UID == nil) != (res.UID == nil) {
```
purpleidea: Perfect! :) :)

resources/user.go
jonathangold: My bad. I realize now there's no error to wrap.

resources/user.go
jonathangold: I promise to remember. :)

purpleidea:
I am *so* thrilled to MERGE this and to have it in mgmt.

Great work, and I'm sure your new skills will make for shorter review cycles next time. Thanks for your patience!

PR #222, opened 2017-09-17
resources: Refresh packagekit cache before install

### LINKED ISSUE #80: [LOVE] package not found on centos 7.x
cli used:
sudo ./mgmt run --file examples/pkg1.yaml --tmp-prefix

15:55:14 pkg.go:268: Pkg[powertop]: CheckApply(true)
15:55:16 pgraph.go:738: Pkg[powertop]: CheckApply errored: The FilterState method failed with: Can't find package named 'powertop'..
15:55:16 pgraph.go:881: Pkg[powertop]: Exited with failure: The FilterState method failed with: Can't find package named 'powertop'..

pkcon works:
[vagrant@testem01z1 mgmt] sudo pkcon install powertop
Resolving                     [=========================]
Querying                      [=========================]
Testing changes               [=========================]
Finished                      [=========================]
Installing                    [=========================]
Testing changes               [=========================]
Installing packages           [=========================]
Finished                      [=========================]

juanluisvaladas:
Fixes #80 
Packagekit has its own cache and if that cache is empty or invalid it won't refresh it automatically

### DIFF
--- a/resources/packagekit/packagekit.go
+++ b/resources/packagekit/packagekit.go
@@ -396,7 +396,11 @@ func (bus *Conn) InstallPackages(packageIDs []string, transactionFlags uint64) e
 	bus.matchSignal(ch, interfacePath, PkIfaceTransaction, signals)
 
 	obj := bus.GetBus().Object(PkIface, interfacePath) // pass in found transaction path
-	call := obj.Call(FmtTransactionMethod("InstallPackages"), 0, transactionFlags, packageIDs)
+	call := obj.Call(FmtTransactionMethod("RefreshCache"), 0, false)
+	if call.Err != nil {
+		return call.Err
+	}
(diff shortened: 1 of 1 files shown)


purpleidea:
@juanluisvaladas Patch looks great, thank you! Merged.

One quick question: Is there some way to detect if we *need* to refresh before we do so, so that we don't unnecessarily refresh if we're recently up-to-date?

juanluisvaladas:
@purpleidea  The method definition is RefreshCache (in  'b' force)
and the packageKit docs say
force:  If the caches should be cleaned and reloaded even if there is valid, up to date data.

I understand this only refreshes the cache if it's necessary.

Source: https://www.freedesktop.org/software/PackageKit/gtk-doc/Transaction.html#Transaction.RefreshCache

Issue #193, opened 2017-06-08
Mildred is missing from AUTHORS


purpleidea:
Hey @mildred ! I think you should be here:

https://github.com/purpleidea/mgmt/blob/master/AUTHORS

If you'd like to, please send a patch of course :)

Thanks!
James

purpleidea:
Ping @mildred please feel free to either send a patch for this or close. Thanks!

purpleidea:
Hey @mildred -- We'd love to have you in the AUTHORS file. I'm going to close this issue, but feel free to send a patch if you're ever interested!

Thanks again.

PR #223, opened 2017-09-21
project: Add Mildred to the AUTHORS list

justinclift:
See issue #193


purpleidea:
@justinclift Thank you for the patch, although this is intended for the actual contributor to write!

Send us some things, and you'll get invited here too one day :)

We've got issues tagged as LOVE for new contributors!

PR #224, opened 2017-09-27
resource: Adds HttpResource

dkliban:


resources/http.go
```diff
@@ -0,0 +1,258 @@
...
+				resp := httpResponse{0, "", err, nil}
+				select {
+				case httpResponseChan <- &resp:
+					return
+				case <-closeChan:
+					return
+				}
+			}
+			response_body, err := ioutil.ReadAll(res.Body)
+			if err != nil {
```
purpleidea: Isn't this case and the next part exactly the same?

resources/http.go
```diff
@@ -0,0 +1,258 @@
...
+
+// HTTPRes is an http resource for retrieving data over http
+type HTTPRes struct {
+	BaseRes        `yaml:",inline"`
+	URL            string            `yaml:"url"`
+	Timeout        int               `yaml:"timeout"` // the http timeout in seconds
+	Method         string            `yaml:"method"`
+	Header         map[string]string `yaml:"header"`
+	Body           string            `yaml:"body"`
+	responseBody   *string
```
purpleidea: add a newline separating the start of the private fields please

Also, what are these fields here for?

resources/http.go
```diff
@@ -0,0 +1,258 @@
...
+
+func init() {
+	RegisterResource("http", func() Res { return &HTTPRes{} })
+}
+
+// HTTPRes is an http resource for retrieving data over http
+type HTTPRes struct {
+	BaseRes        `yaml:",inline"`
+	URL            string            `yaml:"url"`
+	Timeout        int               `yaml:"timeout"` // the http timeout in seconds
```
purpleidea: Do we validate these? What if it's non-zero (uint instead?)

resources/http.go
```diff
@@ -0,0 +1,258 @@
...
+	return obj.BaseRes.Validate()
+}
+
+// Init runs some startup code for this resource.
+func (obj *HTTPRes) Init() error {
+	obj.client = &http.Client{
+		Timeout: time.Duration(obj.Timeout) * time.Second,
+	}
+
+	obj.SetKind("http")
```
purpleidea: not needed

resources/http.go
```diff
@@ -0,0 +1,258 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			body := bytes.NewBufferString(obj.Body)
+			req, err := http.NewRequest(obj.Method, obj.URL, body)
+			if err != nil {
+				resp := httpResponse{0, "", err, nil}
```
purpleidea: I think you can just do:
```
resp := httpResponse{err: err}
```
instead. Also it's preferable to do:

```
resp := &httpResponse{ ... }
```

instead of taking the pointer of resp. This way the object is created in it's correct pointer state at the source :)

This all applies in a few cases in the code...

resources/http.go
```diff
@@ -0,0 +1,258 @@
...
+			send = false
+			obj.Event()
+		}
+	}
+}
+
+// CheckApply checks the resource state and applies the resource if the bool
+// input is true. It returns error info and if the state check passed or not.
+func (obj *HTTPRes) CheckApply(apply bool) (bool, error) {
+	if val, exists := obj.Recv["URL"]; exists && val.Changed {
```
purpleidea: I think you should just take out all of this unused body for now and put // XXX: TODO or something...

resources/http.go
```diff
@@ -0,0 +1,258 @@
...
+	if !ok {
+		return false
+	}
+	if !obj.BaseRes.Compare(res) { // call base Compare
+		return false
+	}
+	if obj.Name != res.Name {
+		return false
+	}
+
```
purpleidea: You need to compare all the public fields from the struct...

resources/http_test.go
```diff
@@ -0,0 +1,40 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"testing"
+)
+
+func TestHttpWatch1(t *testing.T) {
```
purpleidea: Does this work or do something?

purpleidea:
Hey @dkliban I just pushed a patch: https://github.com/purpleidea/mgmt/commit/f9d452ad2c3d402847ed735ed2bba7934d4d78f3

It includes an example long-poll server and client that is race-free.

Sorry I didn't think about all this race stuff earlier, but this should provide a good working example on how to be race-free. This Watch method could support different server variants, like this "redirect" method, for example.

If you're still interested on finishing this up, we'd love to have it. I think the idea was a little too avant-garde for me to be a good mentor on, so sorry for that. If you've had it with this one, let me know and I or someone else can adopt it.

Thanks!
James

dkliban:
@purpleidea I would like to continue with this work. I am going to be traveling from Thursday to Sunday. Let's try to do a google hangout before then or on Monday.

purpleidea:
Anytime you like is fine with me. Although I am currently having trouble
sleeping right now, so probably anytime/day after 8 hours from now. Ping me
on IRC to schedule. Cheers!

dkliban:
I am not able to finish this work at this time. Closing the PR.

PR #225, opened 2017-09-30
resource: Add an Aws resource

jonathangold:
For feedback, don't merge

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -1,0 +1,653 @@
+// Mgmt
+// Copyright (C) 2013-2017+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	errwrap "github.com/pkg/errors"
+)
+
+func init() {
+	RegisterResource("aws:ec2", func() Res { return &AwsEc2Res{} })
+}
+
+// AwsPrefix is a const which gets prepended onto the instance name.
+const AwsPrefix = "mgmt:"
+
+// AwsRegions is a list of all AWS regions generated using ec2.DescribeRegions.
+// cn-north-1 and us-gov-west-1 are not returned, probably due to security.
+// List available at http://docs.aws.amazon.com/general/latest/gr/rande.html
+var AwsRegions = []string{
+	"ap-northeast-1",
+	"ap-northeast-2",
+	"ap-south-1",
+	"ap-southeast-1",
+	"ap-southeast-2",
+	"ca-central-1",
+	"cn-north-1",
+	"eu-central-1",
+	"eu-west-1",
+	"eu-west-2",
+	"sa-east-1",
+	"us-east-1",
+	"us-east-2",
+	"us-gov-west-1",
+	"us-west-1",
+	"us-west-2",
+}
+
+// AwsEc2Res is an AWS EC2 resource. In order to create a client session, your
+// AWS credentials must be present in ~/.aws - For detailed instructions see
+// http://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html
+type AwsEc2Res struct {
+	BaseRes `yaml:",inline"`
+	State   string `yaml:"state"` // state: running, stopped, terminated
+	Region  string `yaml:"region"`
+	Type    string `yaml:"type"` // the ec2 instance type, ie. t2.micro
+	ImageID string `yaml:"imageid"`
+
+	client *ec2.EC2 // client session for AWS API calls
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *AwsEc2Res) Default() Res {
+	return &AwsEc2Res{
+		BaseRes: BaseRes{
+			MetaParams: DefaultMetaParams, // force a default
+		},
+	}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *AwsEc2Res) Validate() error {
+	if obj.State != "running" && obj.State != "stopped" && obj.State != "terminated" {
+		return fmt.Errorf("state must be 'running', 'stopped' or 'terminated'")
+	}
+
+	// compare obj.Region to the list of available AWS endpoints.
+	validRegion := false
+	for _, region := range AwsRegions {
+		if obj.Region == region {
+			validRegion = true
+			break
+		}
+	}
+	if !validRegion {
+		return fmt.Errorf("region must be a valid AWS endpoint")
+	}
+
+	// check the instance type
+	// there is currently no api call to enumerate available instance types
+	if obj.Type == "" {
+		return fmt.Errorf("no instance type specified")
+	}
+
+	// check imageId against a list of available images
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(obj.Region),
+	})
+	if err != nil {
+		return errwrap.Wrapf(err, "error creating session")
+	}
+	client := ec2.New(sess)
+
+	imagesInput := &ec2.DescribeImagesInput{}
+	images, err := client.DescribeImages(imagesInput)
+	if err != nil {
+		return errwrap.Wrapf(err, "error describing images")
+	}
+	validImage := false
+	for _, image := range images.Images {
+		if obj.ImageID == *image.ImageId {
+			validImage = true
+			break
+		}
+	}
+	if !validImage {
+		return fmt.Errorf("imageid must be a valid ami available in the specified region")
+	}
+
+	return obj.BaseRes.Validate()
+}
+
+// Init initializes the resource.
+func (obj *AwsEc2Res) Init() error {
+	// create a client session for the AWS API
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(obj.Region),
+	})
+	if err != nil {
+		return errwrap.Wrapf(err, "error creating session")
+	}
+	obj.client = ec2.New(sess)
+
+	return obj.BaseRes.Init() // call base init, b/c we're overriding
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *AwsEc2Res) Watch() error {
+	send := false
+	var exit *error
+	if err := obj.Running(); err != nil {
+		return err
+	}
+	type chanStruct struct {
+		str string
+		err error
+	}
+	awsChan := make(chan *chanStruct)
+	closeChan := make(chan struct{})
+	ctx, cancel := context.WithCancel(context.TODO())
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		select {
+		case <-closeChan:
+			cancel()
+		}
+	}()
+	wg.Add(1)
+	defer wg.Wait()
+	go func() {
+		defer wg.Done()
+		defer close(awsChan)
+		for {
+			diInput := &ec2.DescribeInstancesInput{
+				Filters: []*ec2.Filter{
+					{
+						Name:   aws.String("tag:Name"),
+						Values: []*string{aws.String(obj.prependName())},
+					},
+					{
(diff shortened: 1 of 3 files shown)


jonathangold:
Hey @purpleidea. I didn't have as much time this week as I had hoped, but I got it working. I'm sure there are still some things to fix, but it works as expected.

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+	var exit *error
+	sess := session.Must(session.NewSession(&aws.Config{
+		Region: aws.String(obj.Region),
+	}))
+	svc := ec2.New(sess)
+	diInput := ec2.DescribeInstancesInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag:Name"),
+				Values: []*string{aws.String(obj.GetName())},
```
purpleidea: assuming I read this correctly as being a unique instance tag we're giving to our machines, didn't we want to add a special prefix like: `mgmt:` ? if so that prefix should be a public constant, and perhaps a small helper function should be used to wrap obj.GetName() when we want to get that string.

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+	if err := obj.Running(); err != nil {
+		return err
+	}
+	awschan := make(chan string)
+	go func() {
+		for {
+			diOutput, err := svc.DescribeInstances(&diInput)
+			if err != nil {
+				errwrap.Wrapf(err, "error describing instances")
```
purpleidea: return?

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+			}
+			var waitInput ec2.DescribeInstancesInput
+			if len(diOutput.Reservations) == 1 {
+				waitInput = ec2.DescribeInstancesInput{
+					InstanceIds: []*string{diOutput.Reservations[0].Instances[0].InstanceId},
```
purpleidea: Is there a guarantee that there will be at least one instances returned here? (otherwise this could panic)

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+				}
+			} else {
```
purpleidea: maybe best to inline this else right before the if statement?

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+		select {
+		case event := <-obj.Events():
+			if exit, send = obj.ReadEvent(event); exit != nil {
+				return *exit
+			}
+		case state := <-awschan:
+			log.Printf(state)
+			obj.StateOK(false)
+			send = true
+			if send {
```
purpleidea: this send should be after the select because it could trigger from obj.Events() as well... see other resources

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+			_, err := svc.TerminateInstances(&terminateInput)
+			if err != nil {
+				return false, errwrap.Wrapf(err, "error terminating instance")
+			}
+			svc.WaitUntilInstanceTerminated(&describeInput)
+			log.Println("instance terminated")
+		}
+	}
+	if len(diOutput.Reservations) < 1 && obj.State == "running" {
+		minCount := int64(1)
```
purpleidea: These are a bit awkward looks like you can instead use:

```
func (s *RunInstancesInput) SetMinCount(v int64) *RunInstancesInput
```

and chain these in...

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+				svc.WaitUntilInstanceStopped(&waitInput)
+				if *diOutput.Reservations[0].Instances[0].State.Name != "running" {
+					awschan <- "stopped"
+					svc.WaitUntilInstanceRunning(&waitInput)
+				}
+			}
+			if obj.State == "stopped" && len(diOutput.Reservations) == 1 {
+				svc.WaitUntilInstanceRunning(&waitInput)
+				if *diOutput.Reservations[0].Instances[0].State.Name != "stopped" {
+					awschan <- "running"
```
purpleidea: what if we close while being blocked here? we might need a close channel signal, follow the ``closeChan` pattern for an example, see: https://github.com/purpleidea/mgmt/pull/224/files#diff-e546bf8ca2594db1f20ccb778896dfb6R98

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+				Name:   aws.String("tag:Name"),
+				Values: []*string{aws.String(obj.GetName())},
+			},
+			{
+				Name:   aws.String("instance-state-name"),
+				Values: []*string{aws.String("running"), aws.String("pending"), aws.String("stopped"), aws.String("stopping")},
+			},
+		},
+	}
+	diOutput, err := svc.DescribeInstances(&diInput)
```
purpleidea: do we ever need to take about the `NextToken` ? https://godoc.org/github.com/aws/aws-sdk-go/service/ec2#DescribeInstancesOutput

resources/aws.go
jonathangold: Yes, if there is no instance to return, then len(diOutput.Reservations) == 0.

resources/aws.go
jonathangold: No, we are only interested in diOutput.Reservations[0].Instances[0]. If either array has more than one item, CheckApply will error. NextToken is for handling many instances at once as far as I can tell.

resources/aws.go
```diff
@@ -0,0 +1,334 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	errwrap "github.com/pkg/errors"
+)
+
+func init() {
+	RegisterResource("aws", func() Res { return &AwsRes{} })
```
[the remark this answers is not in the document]
jonathangold: I think aws:ec2 is prettier and more extensible when we bring more aws resources in. I'll see if it works.

resources/aws.go
jonathangold: done

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+	errwrap "github.com/pkg/errors"
+)
+
+func init() {
+	RegisterResource("aws:ec2", func() Res { return &AwsEc2Res{} })
+}
+// AwsPrefix is a const which gets prepended onto the instance name.
+const AwsPrefix = "mgmt:"
+
+// AwsRegions is a list of all AWS regions generated using ec2.DescribeRegions.
```
purpleidea: Optional: Add an http link to this comment so users can see these in their browser or figure out easily what's new/changed and write patches?

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+var AwsRegions = []string{
+	"ap-south-1",
```
purpleidea: Are these in a particular order, and if so please state, otherwise maybe this should be alphabetical?

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+func (obj *AwsEc2Res) Validate() error {
+	if obj.State != "running" && obj.State != "stopped" && obj.State != "terminated" {
+		return fmt.Errorf("state must be 'running', 'stopped' or 'terminated'")
+	}
+
+	// compare obj.Region to the list of available AWS endpoints.
+	validRegion := false
+	for _, region := range AwsRegions {
+		if obj.Region == region {
+			validRegion = true
```
purpleidea: you can add a break here

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+
+	imagesInput := &ec2.DescribeImagesInput{}
+	images, err := client.DescribeImages(imagesInput)
+	if err != nil {
+		return errwrap.Wrapf(err, "error describing images")
+	}
+	validImage := false
+	for _, image := range images.Images {
+		if obj.ImageID == *image.ImageId {
+			validImage = true
```
purpleidea: break

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+	// create a client session for the AWS API
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(obj.Region),
+	})
+	if err != nil {
+		return errwrap.Wrapf(err, "error creating session")
+	}
+	obj.client = ec2.New(sess)
+
+	obj.wg = &sync.WaitGroup{}
```
purpleidea: Since this is only used inside the Watch method, it doesn't need to be defined in the struct (or initialized here) and can be a local variable in watch, eg:  `wg := &sync.WaitGroup{}`

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+							aws.String("running"),
+							aws.String("stopping"),
+							aws.String("stopped"),
+						},
+					},
+				},
+			}
+			diOutput, err := obj.client.DescribeInstances(diInput)
+			if err != nil {
+				awsChan <- &chanStruct{
```
purpleidea: needs to be wrapped in select pattern with closeChan easy out so we don't ever block/deadlock on close
same for other instances of this
also, on error don't we want this loop to return?

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+						awsChan <- &chanStruct{
+							err: errwrap.Wrapf(err, "error describing instances"),
+						}
+					}
+					var stateName string
+					if len(stateOutput.Reservations) == 1 {
+						stateName = *stateOutput.Reservations[0].Instances[0].State.Name
+					}
+					if len(stateOutput.Reservations) == 0 || (len(stateOutput.Reservations) == 1 && stateName != "running") {
+						x := &chanStruct{
```
purpleidea: good, but you can remove the x temp variable and directly use the struct now that it's more clear what the syntax is

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+				return *exit
+			}
+		case msg, ok := <-awsChan:
+			if !ok {
+				return *exit
+			}
+			if err := msg.err; err != nil {
+				return err
+			}
+			log.Println("instance", msg.str)
```
purpleidea: want to make this message prettier please?

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+		for {
+			diInput := &ec2.DescribeInstancesInput{
+				Filters: []*ec2.Filter{
+					{
+						Name:   aws.String("tag:Name"),
+						Values: []*string{aws.String(obj.prependName())},
+					},
+					{
+						Name: aws.String("instance-state-name"),
+						Values: []*string{
```
purpleidea: you had a clever idea about looking at the requested state and adding the opposite string here, are we still doing that?

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+		}
+		if obj.State == "running" {
+			startInput := &ec2.StartInstancesInput{
+				InstanceIds: []*string{instanceID},
+			}
+			_, err := obj.client.StartInstances(startInput)
+			if err != nil {
+				return false, errwrap.Wrapf(err, "error starting instance")
+			}
+			obj.client.WaitUntilInstanceRunning(describeInput)
```
purpleidea: do these waits have a guaranteed timeout, and do they always return eventually??
eg: what if an aws region blows up while we're running it? Can we use the ctx version with a specific timeout that we choose and set as a constant, or if you think it's a good idea, as a struct option?

resources/aws_ec2.go
```diff
@@ -0,0 +1,535 @@
...
+			obj.client.WaitUntilInstanceTerminated(describeInput)
+			log.Println("instance terminated")
+		}
+	}
+	if len(diOutput.Reservations) < 1 && obj.State == "running" {
+		runParams := &ec2.RunInstancesInput{
+			ImageId:      aws.String(obj.ImageID),
+			InstanceType: aws.String(obj.Type),
+		}
+		runParams.SetMinCount(1)
```
purpleidea: why not just define these in the struct instead of using the method to do so?

resources/aws_ec2.go
```diff
@@ -0,0 +1,348 @@
...
+type AwsEc2UID struct {
+	BaseUID
+	name string
+}
+
+// UIDs includes all params to make a unique identification of this object.
+// Most resources only return one, although some resources can return multiple.
+func (obj *AwsEc2Res) UIDs() []ResUID {
+	x := &AwsEc2UID{
+		BaseUID: BaseUID{Name: obj.GetName(), Kind: obj.GetKind()},
```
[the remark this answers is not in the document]
purpleidea: ping

resources/aws_ec2.go
jonathangold: They're in the order that DescribeRegions outputs.

resources/aws_ec2.go
jonathangold: that happens lower down (stoppedInput/runningInput), this request gets us the instance ids we need for the WaitUntil requests.

resources/aws_ec2.go
```diff
@@ -0,0 +1,602 @@
...
+		}
+		if obj.State == "running" {
+			startInput := &ec2.StartInstancesInput{
+				InstanceIds: []*string{instanceID},
+			}
+			_, err := obj.client.StartInstances(startInput)
+			if err != nil {
+				return false, errwrap.Wrapf(err, "error starting instance")
+			}
+			obj.client.WaitUntilInstanceRunningWithContext(ctx, describeInput)
```
purpleidea: This is a clever, but unfortunately not correct in certain cases. Better to check the error, and you can also test if it's of type context whatever eg: https://golang.org/pkg/context/#pkg-variables or nil or a regular error unrelated to the context.

purpleidea:
I think that this is READY FOR MERGE :)

Thanks for all your hard work.

Of course there are some things we still want to fix, but that is with any code, so let's get this in git master for now, and work on the rest in another patch. It will be easier to review and iterate that way.

Congrats!

PR #226, opened 2017-10-11
util: DRY create PathWatch function

jonathangold:
Moved the path/file watch code into a util function. Removed debug messages as discussed.

### DIFF
--- a/resources/augeas.go
+++ b/resources/augeas.go
@@ -25,8 +25,6 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


purpleidea:
Sorry I've been AFK on this one. I need to think about it more... (This is me doing a bad job as a maintainer.) Thanks for the patch though!

purpleidea:
I'm going to close this for now, since it didn't turn out to be very important. Thanks again!

PR #227, opened 2017-10-16
resources: Improve nspawn resource

jonathangold:
resource is working.

### DIFF
--- a/examples/nspawn1.yaml
+++ b/examples/nspawn1.yaml
@@ -2,6 +2,6 @@
 graph: mygraph
 resources:
   nspawn:
-  - name: mgmt-nspawn1
+  - name: nspawn1
     state: running
 edges: []
--- a/resources/nspawn.go
+++ b/resources/nspawn.go
@@ -21,13 +21,16 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strconv"
+	"unicode"
 
 	"github.com/purpleidea/mgmt/util"
 
+	systemdDbus "github.com/coreos/go-systemd/dbus"
+	machined "github.com/coreos/go-systemd/machine1"
 	systemdUtil "github.com/coreos/go-systemd/util"
 	"github.com/godbus/dbus"
 	errwrap "github.com/pkg/errors"
-	machined "github.com/purpleidea/go-systemd/machine1"
 )
 
 const (
@@ -47,8 +50,8 @@ func init() {
 type NspawnRes struct {
 	BaseRes `yaml:",inline"`
 	State   string `yaml:"state"`
-	// We're using the svc resource to start the machine because that's
-	// what machinectl does. We're not using svc.Watch because then we
+	// We're using the svc resource to start and stop the machine because
+	// that's what machinectl does. We're not using svc.Watch because then we
 	// would have two watches potentially racing each other and producing
 	// potentially unexpected results. We get everything we need to monitor
 	// the machine state changes from the org.freedesktop.machine1 object.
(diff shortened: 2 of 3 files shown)


resources/nspawn.go
```diff
@@ -284,6 +301,9 @@ func (obj *NspawnRes) Compare(r Res) bool {
 	if obj.Name != res.Name {
 		return false
 	}
+	if obj.State != res.State {
```
purpleidea: Nice catch!

examples/nspawn1.yaml
```diff
@@ -2,6 +2,6 @@
 graph: mygraph
 resources:
   nspawn:
-  - name: mgmt-nspawn1
+  - name: nspawnexample
```
purpleidea: nit: Mind naming this `nspawn1` instead?

resources/nspawn.go
```diff
@@ -221,10 +238,10 @@ func (obj *NspawnRes) CheckApply(apply bool) (checkOK bool, err error) {
 		}
 	}
 	if obj.State == stopped {
```
purpleidea: As per the improved code, can both of these branches be collapsed into a single one?

resources/nspawn.go
```diff
@@ -175,7 +192,7 @@ func (obj *NspawnRes) CheckApply(apply bool) (checkOK bool, err error) {
 	// compare the current state with the desired state and perform the
 	// appropriate action
 	var exists = true
-	properties, err := conn.GetProperties(obj.GetName())
+	properties, err := conn.DescribeMachine(obj.GetName())
```
purpleidea: Awesome!

resources/nspawn.go
```diff
@@ -126,6 +142,7 @@ func (obj *NspawnRes) Watch() error {
 	var send = false
 	var exit *error
 
+	defer close(buschan)
```
purpleidea: This was an insightful thing to add-- one other thing we need to check first: how do we stop the receiving of signals on that channel *first*-- that's something that I think we forgot to do in v1 of the patch. Without that, we'll cause a panic if we receive a signal after this closes.

resources/nspawn.go
```diff
@@ -65,29 +66,44 @@ func (obj *NspawnRes) Default() Res {
...
-		running: {},
-	}
-	if _, exists := validStates[obj.State]; !exists {
+	if obj.State != running && obj.State != stopped {
 		return fmt.Errorf("invalid state: %s", obj.State)
 	}
-
-	if err := obj.svc.Validate(); err != nil { // composite resource
-		return errwrap.Wrapf(err, "validate failed for embedded svc")
+	for _, char := range obj.GetName() {
```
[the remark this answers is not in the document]
jonathangold: Looks like it's pretty straightforward to escape/un-escape the machine name properly.
https://github.com/coreos/go-systemd/blob/master/unit/escape.go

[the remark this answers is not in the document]
purpleidea: Nice find!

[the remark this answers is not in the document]
jonathangold: @purpleidea This was fixed in systemd v230-231. See:
https://github.com/systemd/systemd/issues/4677#issuecomment-260880731
My computer is running v229 (Ubuntu LTS strikes again.) I think it's safe to get rid of this check, and maybe add a warning comment for versions < 231. 
From the docs, the name has a max length of 64 chars, which I'll add a check for.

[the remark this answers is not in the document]
purpleidea: Even more fantastic of a discovery! AIUI this means dashes *are* allowed in machine names, as long as we're using newer systemd. If this is true, please add a systemd version check in the Validate method and let's allow dashes if we can!

resources/nspawn.go
```diff
@@ -65,29 +68,83 @@ func (obj *NspawnRes) Default() Res {
...
+	if err != nil {
+		return nil, err
+	}
+	svc := res.(*SvcRes)
+	svc.State = obj.State
+	return svc, nil
+}
+
+// systemdVersion uses dbus to check which version of systemd is installed
+func systemdVersion() (int, error) {
```
purpleidea: should this be a uint since it's never negative? Possibly even a uint16?

resources/nspawn.go
```diff
@@ -21,13 +21,16 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strconv"
+	"unicode"
 
 	"github.com/purpleidea/mgmt/util"
 
+	systemDbus "github.com/coreos/go-systemd/dbus"
```
purpleidea: Just checking: did you mean `systemdDbus` instead? (note the extra d)

resources/nspawn.go
```diff
@@ -65,29 +68,83 @@ func (obj *NspawnRes) Default() Res {
...
+	bus, err := systemDbus.NewSystemdConnection()
+	if err != nil {
+		return 0, errwrap.Wrapf(err, "failed to connect to bus")
+	}
+	// get the systemd version
+	verString, err := bus.GetManagerProperty("Version")
+	if err != nil {
+		return 0, errwrap.Wrapf(err, "could not get version property")
+	}
+	bus.Close()
```
purpleidea: I think this needs to be a `defer` right after the error check of it's creation. Why? Suppose `bus.GetManagerProperty("Version")` fails, then we'd return without having ever closed the bus. This is why defer is so useful.

resources/nspawn.go
```diff
@@ -65,29 +68,83 @@ func (obj *NspawnRes) Default() Res {
...
+	// lose the surrounding quotes
+	verNum := verString[1 : len(verString)-1]
```
purpleidea: If length of verString is 0 this will panic. Probably rare, but probably a good idea to check so that this can never panic. Alternatively, you might consider using https://golang.org/pkg/strconv/#Unquote instead. Up to you if it's a good fit.

resources/nspawn.go
```diff
@@ -115,7 +172,7 @@ func (obj *NspawnRes) Watch() error {
 	if err := call.Err; err != nil {
 		return err
 	}
-	buschan := make(chan *dbus.Signal, 10)
```
purpleidea: I'm not sure about this one. Are you?
At minimum add a TODO to look into this more (whether we give the channel a buffer or not)
Try repeatedly sending signals to see if it drops one without a buffer.
Eg: https://github.com/purpleidea/mgmt/blob/master/resources/packagekit/packagekit.go#L42

(I think eventually someone, hopefully not us, should dig into the implementation to understand why it does that and if there is a fix...)

resources/nspawn.go
```diff
@@ -126,6 +183,7 @@ func (obj *NspawnRes) Watch() error {
 	var send = false
 	var exit *error
 
+	defer bus.Close()
```
purpleidea: see other comment about bus.Close()

resources/nspawn.go
```diff
@@ -65,29 +68,88 @@ func (obj *NspawnRes) Default() Res {
...
+	// get the systemd version
+	verString, err := bus.GetManagerProperty("Version")
+	if err != nil {
+		return 0, errwrap.Wrapf(err, "could not get version property")
+	}
+	// lose the surrounding quotes
+	verNum, err := strconv.Unquote(verString)
+	if err != nil {
+		return 0, errwrap.Wrapf(err, "error unqoting version number")
```
purpleidea: typo

resources/nspawn.go
```diff
@@ -65,29 +68,88 @@ func (obj *NspawnRes) Default() Res {
...
+	res, err := NewNamedResource("svc", fmt.Sprintf(nspawnServiceTmpl, obj.GetName()))
+	if err != nil {
+		return nil, err
+	}
+	svc := res.(*SvcRes)
+	svc.State = obj.State
+	return svc, nil
+}
+
+// systemdVersion uses dbus to check which version of systemd is installed
```
purpleidea: end docstring with a period please

resources/nspawn.go
```diff
@@ -65,29 +68,88 @@ func (obj *NspawnRes) Default() Res {
...
+	verNum, err := strconv.Unquote(verString)
+	if err != nil {
+		return 0, errwrap.Wrapf(err, "error unqoting version number")
+	}
+	// cast to uint16
+	ver64, err := strconv.ParseUint(verNum, 10, 16)
+	if err != nil {
+		return 0, errwrap.Wrapf(err, "error casting systemd version number")
+	}
+	ver := uint16(ver64)
```
purpleidea: you can just return uint16(...), nil

resources/nspawn.go
```diff
@@ -115,8 +177,9 @@ func (obj *NspawnRes) Watch() error {
 	if err := call.Err; err != nil {
 		return err
 	}
-	buschan := make(chan *dbus.Signal, 10)
-	bus.Signal(buschan)
+	// TODO: verify that dbus doesn't drop messages without a chan buffer
```
purpleidea: I think I was wrong. I'm sorry. I think this needs to be buffered to reduce the chance of deadlock due to issues in the godbus implementation. So please add the buffer back and a relevant comment asking about if the implementation can deadlock if we close when there's still a message not read from the channel.

Alternatively, the harder thing to do is to make sure we drain the channel on close-- I think it's probably better we leave that to a future patch. I want to merge this thing.

resources/nspawn.go
```diff
@@ -126,9 +189,12 @@ func (obj *NspawnRes) Watch() error {
 	var send = false
 	var exit *error
 
+	defer close(busChan)
```
purpleidea: Nice!

purpleidea:
Patch LGTM, just waiting for CI. Thanks!

purpleidea:
MERGED :)

PR #229, opened 2017-10-20
resources: Add a group resource

jonathangold:
There are a few features to implement in the future (user list, and group name change,) but the core functionality is all here.

### DIFF
--- a/resources/group.go
+++ b/resources/group.go
@@ -1,0 +1,294 @@
+// Mgmt
+// Copyright (C) 2013-2017+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os/exec"
+	"os/user"
+	"strconv"
+	"syscall"
+
+	"github.com/purpleidea/mgmt/recwatch"
+
+	errwrap "github.com/pkg/errors"
+)
+
+func init() {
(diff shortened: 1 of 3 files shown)


resources/group.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	// if the group exists and no GID is specified, we are done
+	if obj.State == "exists" && exists && obj.GID == nil {
+		return true, nil
+	}
+	// if the group doesn't exist and should be absent, we are done
+	if obj.State == "absent" && !exists {
+		return true, nil
+	}
+
```
purpleidea: where's the `if !apply` part? (so that noop works correctly)

resources/group.go
```diff
@@ -0,0 +1,272 @@
...
+		// if the group exists and has the correct GID, we are done
+		if obj.State == "exists" && *obj.GID == uint32(existingGID) {
+			return true, nil
+		}
+	}
+	// if the group exists and no GID is specified, we are done
+	if obj.State == "exists" && exists && obj.GID == nil {
+		return true, nil
+	}
+	// if the group doesn't exist and should be absent, we are done
```
purpleidea: I think this check should be moved up as high as possible, since it's a short `easy out`. It also then might simplify the other situations because we know of !exists then also not "absent".

The check above this one should come next.

resources/group.go
```diff
@@ -0,0 +1,272 @@
...
+	if obj.State == "absent" && !exists {
+		return true, nil
+	}
+
+	var cmdName string
+	args := []string{obj.GetName()}
+
+	if obj.State == "exists" && exists {
```
purpleidea: These first two if statements can be combined and simplified... Give it a try.

resources/group.go
```diff
@@ -0,0 +1,272 @@
...
+			return false, fmt.Errorf("the requested GID belongs to another group")
+		}
+		// get the existing group's GID
+		existingGID, err := strconv.ParseUint(group.Gid, 10, 32)
+		if err != nil {
+			return false, errwrap.Wrapf(err, "error casting existing GID")
+		}
+		// check if existing group has the wrong GID
+		if *obj.GID != uint32(existingGID) {
+			log.Printf("%s: Inconsistent GID: %s", obj, obj.GetName())
```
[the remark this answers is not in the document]
jonathangold: No. If the group has the wrong gid, we can change it to the desired one.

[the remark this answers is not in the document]
purpleidea: aha! okay, awesome-- please add that info as a comment under the first one.

purpleidea:
Fantastic work, and I am excited about the autoedges code that's coming next. I'm slightly jealous that I'm not writing it because I expect it will be incredibly fun, but I know it's in good hands!

MERGED

PR #230, opened 2017-10-22
resources: Enhancements to user and group

emmmmygold:
I also cleaned up the log messages and removed a couple of unnecessary ones. Let me know if you think that should be in a separate pull request.

### DIFF
--- a/resources/group.go
+++ b/resources/group.go
@@ -41,7 +41,7 @@ const groupFile = "/etc/group"
 type GroupRes struct {
 	BaseRes `yaml:",inline"`
 	State   string  `yaml:"state"` // state: exists, absent
-	GID     *uint32 `yaml:"gid"`
+	GID     *uint32 `yaml:"gid"`   // the group's gid
 
 	recWatcher *recwatch.RecWatcher
 }
@@ -130,7 +130,6 @@ func (obj *GroupRes) CheckApply(apply bool) (checkOK bool, err error) {
 		if _, ok := err.(user.UnknownGroupError); !ok {
 			return false, errwrap.Wrapf(err, "error looking up group")
 		}
-		log.Printf("%s: Group not found: %s", obj, obj.GetName())
 		exists = false
 	}
 	// if the group doesn't exist and should be absent, we are done
@@ -225,6 +224,26 @@ func (obj *GroupRes) CheckApply(apply bool) (checkOK bool, err error) {
 type GroupUID struct {
 	BaseUID
 	name string
+	gid  *uint32
+}
+
+// IFF aka if and only if they are equivalent, return true. If not, false.
+func (obj *GroupUID) IFF(uid ResUID) bool {
+	res, ok := uid.(*GroupUID)
+	if !ok {
+		return false
+	}
+	if obj.gid != nil && res.gid != nil {
+		if *obj.gid != *res.gid {
+			return false
+		}
+	}
+	if obj.name != "" && res.name != "" {
+		if obj.name != res.name {
+			return false
+		}
+	}
+	return true
 }
 
 // UIDs includes all params to make a unique identification of this object.
@@ -233,6 +252,7 @@ func (obj *GroupRes) UIDs() []ResUID {
 	x := &GroupUID{
 		BaseUID: BaseUID{Name: obj.GetName(), Kind: obj.GetKind()},
 		name:    obj.Name,
+		gid:     obj.GID,
 	}
 	return []ResUID{x}
 }
--- a/resources/user.go
+++ b/resources/user.go
@@ -19,10 +19,13 @@ package resources
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os/exec"
 	"os/user"
+	"sort"
 	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/purpleidea/mgmt/recwatch"
@@ -39,11 +42,13 @@ const passwdFile = "/etc/passwd"
 // UserRes is a user account resource.
 type UserRes struct {
 	BaseRes           `yaml:",inline"`
-	State             string  `yaml:"state"` // state: exists, absent
-	UID               *uint32 `yaml:"uid"`
-	GID               *uint32 `yaml:"gid"`
-	HomeDir           *string `yaml:"homedir"`
-	AllowDuplicateUID bool    `yaml:"allowduplicateuid"`
+	State             string   `yaml:"state"`             // state: exists, absent
+	UID               *uint32  `yaml:"uid"`               // uid must be unique unless AllowDuplicateUID is true
+	GID               *uint32  `yaml:"gid"`               // gid of the user's primary group
+	Group             *string  `yaml:"group"`             // name of the user's primary group
+	Groups            []string `yaml:"groups"`            // list of supplemental groups
+	HomeDir           *string  `yaml:"homedir"`           // path to the user's home directory
+	AllowDuplicateUID bool     `yaml:"allowduplicateuid"` // allow duplicate uid
 
 	recWatcher *recwatch.RecWatcher
 }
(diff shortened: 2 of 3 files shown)


resources/group.go
```diff
@@ -225,6 +224,20 @@ func (obj *GroupRes) CheckApply(apply bool) (checkOK bool, err error) {
...
+
+// IFF aka if and only if they are equivalent, return true. If not, false.
+func (obj *GroupUID) IFF(uid ResUID) bool {
+	res, ok := uid.(*GroupUID)
+	if !ok {
+		return false
+	}
+	var eq bool
+	if obj.gid != nil && res.gid != nil {
+		eq = *obj.gid == *res.gid
```
purpleidea: cute, but i think this is overly complex
just return true/false where appropriate

resources/group.go
```diff
@@ -225,6 +224,22 @@ func (obj *GroupRes) CheckApply(apply bool) (checkOK bool, err error) {
...
+	gid  *uint32
+}
+
+// IFF aka if and only if they are equivalent, return true. If not, false.
+func (obj *GroupUID) IFF(uid ResUID) bool {
+	res, ok := uid.(*GroupUID)
+	if !ok {
+		return false
+	}
+	var g, n bool
```
purpleidea: I'm not sure if this logic is right. First of all, don't we want this to return true if both valid pairs (gid & name) are true, out of whichever are present?

I think the code can be simplified by checking for different and returning false early, only to return true at the end if all is good.

resources/user.go
```diff
@@ -196,12 +201,27 @@ func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
 		if obj.GID != nil {
 			args = append(args, "-g", fmt.Sprintf("%d", *obj.GID))
 		}
+		if obj.Group != nil {
+			args = append(args, "-g", *obj.Group)
```
purpleidea: we should probably check for invalid chars in this obj.Group string in Validate(). Even though it's directly being passed into argv, it may change, thus exposing us for a cli injection attack. Best to filter out bogus things like spaces, quotes, and anything else that's illegal.

resources/user.go
```diff
@@ -196,12 +201,27 @@ func (obj *UserRes) CheckApply(apply bool) (checkOK bool, err error) {
...
+		}
+		if obj.Groups != nil {
+			var groups string
+			for i, group := range obj.Groups {
```
purpleidea: use the strings.Join method instead
since this gets joined with comma, we'll definitely have to validate this list as well in Validate

resources/user.go
```diff
@@ -224,6 +244,70 @@ type UserUID struct {
 	name string
 }
 
+// UserResAutoEdges holds the state of the auto edge generator.
+type UserResAutoEdges struct {
+	UIDs    []ResUID
+	pointer int
+}
+
+// AutoEdges returns the the user's group.
```
purpleidea: Technically this comment is wrong. Can you spiffy it up ever so slightly? In the best case scenario it should explain the logic of what it's doing and why, but I won't force that on you if you're really not in the mood.

resources/group.go
```diff
@@ -225,6 +224,22 @@ func (obj *GroupRes) CheckApply(apply bool) (checkOK bool, err error) {
...
+	if obj.gid != nil && res.gid != nil {
+		g = *obj.gid == *res.gid
+	}
+	n = obj.name == res.name
+	// return true if gid or name are the same
+	return g || n
```
purpleidea: I'm going to make a giant guess here: you probably settled on `g || n` instead of `g && n` to get it to work. But here's why the && (which would have been correct) wouldn't work in your testing, it's subtle: since we have some situations where we can only build an autoedge with a gid, (we have no knowledge in the user what name the group is) but we *always* know on the group side what the name is (that's the primary key) so the IFF function will in some scenarios be coming empty string to group name, and fail. So since an empty group name is invalid (please add that check for `group` and `groups` to Validate) we actually need to do a if `obj.name != "" && res.name != "" {  if .. != ... { return false }   }` similar to how we do the uid check?

In case none of this made sense, then maybe I'm just wrong for lack of sleep and failure to remember how my autoedges code works :P

resources/user.go
```diff
@@ -275,6 +359,9 @@ func (obj *UserRes) Compare(r Res) bool {
 			return false
 		}
 	}
+	if !reflect.DeepEqual(obj.Groups, res.Groups) {
```
purpleidea: please do the compare properly without using reflect, sorry :(

purpleidea (changes requested):
Nicely done! I think you've knocked out the autoedges stuff very quickly. One comment to make for general learning purposes, keep in mind that here you used the `users` Autoedges method, and the `group`'s GroupUID struct, however it's also valid to do the opposite or a combination of the two! My guess is that this is probably correct, but I just want to make sure that you think it over once!

Few fixes mentioned, and I don't mind if you squash this all into a single patch. (It's only slightly cheating, but not a major deal.)

resources/group.go
```diff
@@ -225,6 +224,27 @@ func (obj *GroupRes) CheckApply(apply bool) (checkOK bool, err error) {
...
+// IFF aka if and only if they are equivalent, return true. If not, false.
+func (obj *GroupUID) IFF(uid ResUID) bool {
+	res, ok := uid.(*GroupUID)
+	if !ok {
+		return false
+	}
+	eq := true
+	if obj.gid != nil && res.gid != nil {
+		if *obj.gid != *res.gid {
+			eq = false
```
purpleidea: if we know this is false, why not just return false` now and avoid the eq var?
(same point below and at end)

resources/user.go
```diff
@@ -59,8 +63,31 @@ func (obj *UserRes) Default() Res {
 
 // Validate if the params passed in are valid data.
 func (obj *UserRes) Validate() error {
+	const badChars string = " \"{}()<>&*‘|=?;[]$–#~!.”%%/\\:+,`"
```
purpleidea: can you either add a comment on where you got the spec for badChars, or make it obvious in a different way?
eg: whitelist of a-z + specialchars?

resources/user.go
```diff
@@ -59,8 +63,31 @@ func (obj *UserRes) Default() Res {
...
+		if *obj.Group == "" {
+			return fmt.Errorf("group cannot be empty string")
+		}
+	}
+	if obj.Groups != nil {
+		for _, group := range obj.Groups {
+			if strings.ContainsAny(group, badChars) {
+				return fmt.Errorf("groups list contains invalid character(s)")
+			}
+			if group == "" {
```
purpleidea: nit: do the cheap test first, also makes it easier to read in case user is not sure what happens if you pass in an empty string...

resources/user.go
```diff
@@ -275,6 +391,29 @@ func (obj *UserRes) Compare(r Res) bool {
...
 		}
 	}
+	if (obj.Groups == nil) != (res.Groups == nil) {
+		return false
+	}
+	if obj.Groups != nil && res.Groups != nil {
+		if len(obj.Groups) != len(res.Groups) {
+			return false
+		}
+		// make a map with each group as key and 1 as value
```
purpleidea: there's a much simpler algorithm which is more correct...

resources/user.go
```diff
@@ -224,6 +274,72 @@ type UserUID struct {
 	name string
 }
 
+// UserResAutoEdges holds the state of the auto edge generator.
+type UserResAutoEdges struct {
+	UIDs    []ResUID
+	pointer int
+}
+
+// AutoEdges returns edges from the user resource to each group found in
```
purpleidea: Can you add a comment explaining the reverse stuff please? It would be very helpful. Thanks!

purpleidea (changes requested):
Patch looks great, Few fixups needed. Consider squashing into a single commit. Thanks!

resources/user.go
```diff
@@ -39,11 +42,13 @@ const passwdFile = "/etc/passwd"
 // UserRes is a user account resource.
 type UserRes struct {
 	BaseRes           `yaml:",inline"`
-	State             string  `yaml:"state"` // state: exists, absent
-	UID               *uint32 `yaml:"uid"`
-	GID               *uint32 `yaml:"gid"`
-	HomeDir           *string `yaml:"homedir"`
-	AllowDuplicateUID bool    `yaml:"allowduplicateuid"`
+	State             string   `yaml:"state"` // state: exists, absent
+	UID               *uint32  `yaml:"uid"`
```
purpleidea: sorry about this, I should have mentioned it earlier, but can you add some short docs as comment at the end of some of the parameters? if you think it's important to have a long doc string, then it can be put above the field in full sentence form. These show up in the docs.

resources/user.go
```diff
@@ -224,6 +279,75 @@ type UserUID struct {
...
 
+// UserResAutoEdges holds the state of the auto edge generator.
+type UserResAutoEdges struct {
+	UIDs    []ResUID
+	pointer int
+}
+
+// AutoEdges returns edges from the user resource to each group found in
+// its definition. The groups can be in any of the three applicable fields
+// (GID, Group and Groups.) If the user exists reversed ensures the edge
```
purpleidea: I think you're missing a comma after `If the user exists`

purpleidea (commented):
Looks great! Two small doc suggestions, otherwise perfect, thank you!

purpleidea:
Patch LGTM :)

purpleidea:
w00t!

resources/user.go
ross710: Addressed in `3688560`/`5045af3`. Added the `bootstrap resolves same-pump conflict by authorization time` test and extended `fuelCheck()` with a `dispenserAuthorizationTime` param (backed by an `ImmutablePayment` metadata referenceId), covering the `ImmutableCheck` bootstrap path.

_Generated with Claude Code._

PR #231, opened 2017-10-29
etcd: Add advertise urls to cli

jonathangold:
I think that covers it.

### DIFF
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -176,12 +176,14 @@ type EmbdEtcd struct { // EMBeddeD etcd
 	exitchan    chan struct{}
 	exitTimeout <-chan time.Time
 
-	hostname   string
-	memberID   uint64            // cluster membership id of server if running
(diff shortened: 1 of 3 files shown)


etcd/etcd.go
```diff
@@ -1651,8 +1655,8 @@ func (obj *EmbdEtcd) StartServer(newCluster bool, peerURLsMap etcdtypes.URLsMap)
 	cfg := embed.NewConfig()
 	cfg.Name = memberName // hostname
 	cfg.Dir = obj.dataDir
-	cfg.ACUrls = obj.clientURLs
-	cfg.APUrls = peerURLs
+	cfg.ACUrls = obj.advertiseClientURLs
```
purpleidea: Want to put these below the other two so that everything is in the same order throughout?
Also, do we want to do an `if len(...) > 0` check before we do the assignment, and if not default to the obj.clientUrls/peerUrls, or does that happen automatically with etcd internally?

purpleidea:
Merged, Thanks!

PR #232, opened 2017-10-29
etcd: Plumbed in the new cli flags for advertise urls

jonathangold:
@purpleidea: Got mgmt working though NAT!

### DIFF
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -400,9 +400,13 @@ func (obj *EmbdEtcd) Startup() error {
 	// if i am alone and will have to be a server...
 	if !obj.noServer && bootstrapping {
 		log.Printf("Etcd: Bootstrapping...")
+		surls := obj.serverURLs
+		if len(obj.advertiseServerURLs) > 0 {
+			surls = obj.advertiseServerURLs
+		}
 		// give an initial value to the obj.nominate map we keep in sync
 		// this emulates Nominate(obj, obj.hostname, obj.serverURLs)
-		obj.nominated[obj.hostname] = obj.serverURLs // initial value
+		obj.nominated[obj.hostname] = surls // initial value
 		// NOTE: when we are stuck waiting for the server to start up,
 		// it is probably happening on this call right here...
 		obj.nominateCallback(nil) // kick this off once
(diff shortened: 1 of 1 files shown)


etcd/etcd.go
```diff
@@ -1439,15 +1443,22 @@ func (obj *EmbdEtcd) nominateCallback(re *RE) error {
...
 			// nominate self so "member" list is correct for peers to see
-			Nominate(obj, obj.hostname, obj.serverURLs)
+			Nominate(obj, obj.hostname, surls)
 			// XXX: if this fails, where will we retry this part ?
 		}
 
 		// advertise client urls
-		if curls := obj.clientURLs; len(curls) > 0 {
+		curls := obj.clientURLs
+		if len(obj.advertiseClientURLs) > 0 {
```
purpleidea: subtle, but I think this len(advClientUrls) should happen nested *within* the earlier len(curls) because if there are no curls specified this will add advertise urls, when it shouldn't.

etcd/etcd.go
```diff
@@ -1439,15 +1443,22 @@ func (obj *EmbdEtcd) nominateCallback(re *RE) error {
 			obj.endpoints[obj.hostname] = addresses // now we have some!
 			// client connects to one of the obj.endpoints servers...
 			log.Printf("Etcd: Addresses are: %s", addresses)
-
```
purpleidea: why delete this newline?

jonathangold:
@purpleidea: I'll take a look soon.

etcd/etcd.go
```diff
@@ -1440,14 +1448,22 @@ func (obj *EmbdEtcd) nominateCallback(re *RE) error {
...
+			}
 			// XXX: just put this wherever for now so we don't block
 			// nominate self so "member" list is correct for peers to see
-			Nominate(obj, obj.hostname, obj.serverURLs)
+			Nominate(obj, obj.hostname, surls)
 			// XXX: if this fails, where will we retry this part ?
 		}
 
 		// advertise client urls
-		if curls := obj.clientURLs; len(curls) > 0 {
```
purpleidea: you actually don't need to change this first line, it's more concise and idiomatic and happens to do what we want AFAICT.

PR #234, opened 2017-10-30
resources: Added LaunchScript field to AWS:EC2

jonathangold:
This commit adds a UserData field to EC2 which allows the user to specify a script to run at first launch.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -19,6 +19,7 @@ package resources
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"log"
 	"sync"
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -541,6 +544,14 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 		}
 		runParams.SetMinCount(1)
 		runParams.SetMaxCount(1)
+		if obj.UserData != "" {
+			userDataBytes, err := ioutil.ReadFile(obj.UserData)
```
purpleidea: Is there a reason why this is a filepath instead of the actual string contents?

resources/aws_ec2.go
```diff
@@ -65,11 +67,12 @@ var AwsRegions = []string{
...
-	State   string `yaml:"state"` // state: running, stopped, terminated
-	Region  string `yaml:"region"`
-	Type    string `yaml:"type"` // the ec2 instance type, ie. t2.micro
-	ImageID string `yaml:"imageid"`
+	BaseRes  `yaml:",inline"`
+	State    string `yaml:"state"`    // state: running, stopped, terminated
+	Region   string `yaml:"region"`   // region must match an element of AwsRegions
+	Type     string `yaml:"type"`     // type of ec2 instance, ie. t2.micro
+	ImageID  string `yaml:"imageid"`  // imageid must be available on the chosen region
+	UserData string `yaml:"userdata"` // userdata is a path to a script run on first startup
```
purpleidea: While AWS calls it this, feel free to rename this field to something better if you think that's suitable. Not saying it necessarily is in this case, but I'm open to ideas.

resources/aws_ec2.go
jonathangold: @purpleidea I think the most likely use case is to have a bash script saved to a file, rather than inline, no?

resources/aws_ec2.go
jonathangold: I think LaunchScript would be appropriate.

purpleidea:
00:15 < purpleidea> jonathangold: failed because ci test did not expected a 
                    colon in the middle, sorry.
00:16 < purpleidea> jonathangold: maybe change commit message to AwsEc2 ?
00:16 < purpleidea> OR
00:16 < purpleidea> resource: aws: ec2: Add UserData field

purpleidea:
Merged, thanks!

PR #236, opened 2017-11-03
resource: aws: ec2: Added options for HTTP endpoint to watch

jonathangold:
@purpleidea I realize this is a HUUUUGE commit, so I'd suggest mostly ignoring the stuff that's changed in the Watch() method at first for review. All I did was add an if/else to choose the watch functionality. The other changes were just updating some variables to use the ones added to the struct.

The methods I added are: 
- makeSnsTopic makes an sns topic
- makeCloudWatchRule makes a cloudwatch rule
- targetCloudWatchRule targets the rule to the topic
- snsAuthorize which authorizes cloudwatch to publish to sns.
- snsServer which returns the http server and postHandler() which listens for posts from SNS.
- snsSubscribe subscribes the endpoint to the topic after the snsServer server is running
- Close removes the subscription. topic and cloudwatch rule

I think a hangout would be good so we can go through the code together, but I think it's pretty good. ;)
The next step, imo is to replace the original watch with [SQS long-polling](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html), which will use the same cloudwatch events, and will solve the terminating/stopping issue.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -59,6 +59,16 @@ const (
 	SnsSubscriptionProto = "http"
 	// SnsServerShutdownTimeout is the maximum number of seconds to wait for the http server to shutdown gracefully.
 	SnsServerShutdownTimeout = 30
+	// SnsPolicy is the topic atribute that defines the security policy for the topic.
+	SnsPolicy = "Policy"
+	// SnsPolicySid is the friendly name of the policy statement.
+	SnsPolicySid = CwePrefix + "publish"
+	// SnsPolicyEffect allows the action(s) defined in the policy statement.
+	SnsPolicyEffect = "Allow"
+	// SnsPolicyService is the cloudwatch events security principal that we are granting the permission to.
+	SnsPolicyService = "events.amazonaws.com"
+	// SnsPolicyAction is the specific permission we are granting in the policy.
+	SnsPolicyAction = "SNS:Publish"
 	// CwePrefix gets prepended onto the cloudwatch rule name.
 	CwePrefix = Ec2Prefix + "cw-"
 	// CweRuleName is the name of the rule created by makeCloudWatchRule.
@@ -144,6 +154,33 @@ type AwsEc2Res struct {
 type chanStruct struct {
 	event awsEc2Event
 	err   error
+}
+
+// snsPolicy denotes the structure of sns security policies.
+type snsPolicy struct {
+	Version   string      `json:"Version"`
+	ID        string      `json:"Id"`
+	Statement []statement `json:"Statement"`
+}
+
+// statement denotes the structure of sns security policy statements.
+type statement struct {
+	Sid       string      `json:"Sid"`
+	Effect    string      `json:"Effect"`
+	Principal principal   `json:"Principal"`
+	Action    interface{} `json:"Action"`
+	Resource  string      `json:"Resource"`
+	Condition *struct {
+		StringEquals *struct {
+			AWSSourceOwner *string `json:"AWS:SourceOwner,omitempty"`
+		} `json:"StringEquals,omitempty"`
+	} `json:"Condition,omitempty"`
+}
+
+// principal describes the aws or service account principal.
+type principal struct {
+	AWS     string `json:"AWS,omitempty"`
+	Service string `json:"Service,omitempty"`
 }
 
 // cloudWatchRule denotes the structure of cloudwatch rules.
@@ -288,6 +325,10 @@ func (obj *AwsEc2Res) Init() error {
 		// target cloudwatch rule to sns topic
 		if err := obj.cweTargetRule(obj.snsTopicArn, CweTargetID, CweTargetJSON, CweRuleName); err != nil {
 			return errwrap.Wrapf(err, "error targeting cloudwatch rule")
+		}
+		// authorize cloudwatch to publish on sns
+		if err := obj.snsAuthorizeCloudWatch(obj.snsTopicArn); err != nil {
+			return errwrap.Wrapf(err, "error authorizing cloudwatch for sns")
 		}
 	}
 
@@ -1023,6 +1064,94 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
 		return err
 	}
 	log.Printf("%s: Subscription Confirmed", obj)
+	return nil
+}
+
+// snsProcessEvents unmarshals instance state-change notifications and, if the
+// event matches the instance we are watching, returns an awsEc2Event.
+func (obj *AwsEc2Res) snsProcessEvent(message, instanceName string) (awsEc2Event, error) {
+	// unmarshal the message
+	var msg postMsg
+	err := json.Unmarshal([]byte(message), &msg)
+	if err != nil {
+		return awsEc2EventNone, err
+	}
+	// Check if the instance id in the message matches the name of the
+	// instance we're watching.
+	diInput := &ec2.DescribeInstancesInput{
+		InstanceIds: []*string{aws.String(msg.InstanceID)},
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag:Name"),
+				Values: []*string{aws.String(instanceName)},
+			},
+		},
+	}
+	diOutput, err := obj.client.DescribeInstances(diInput)
+	if err != nil {
+		return awsEc2EventNone, err
+	}
+	// if diOutput returns any reservations, return an awsEc2Event.
+	if len(diOutput.Reservations) != 0 {
+		switch msg.State {
+		case "running":
+			return awsEc2EventInstanceRunning, nil
+		case "stopped":
+			return awsEc2EventInstanceStopped, nil
+		case "terminated":
+			return awsEc2EventInstanceTerminated, nil
+		}
+	}
+	return awsEc2EventNone, nil
+}
+
+// snsAuthorize adds the necessary permission for cloudwatch to publish to the SNS topic.
+func (obj *AwsEc2Res) snsAuthorizeCloudWatch(topicArn string) error {
+	// get the topic attributes, including the security policy
+	gaInput := &sns.GetTopicAttributesInput{
+		TopicArn: aws.String(topicArn),
+	}
+	attrs, err := obj.snsClient.GetTopicAttributes(gaInput)
+	if err != nil {
+		return err
+	}
+	// get the existing security policy
+	pol := attrs.Attributes[SnsPolicy]
+	// unmarshal the current sns security policy
+	var policy snsPolicy
+	err = json.Unmarshal([]byte(*pol), &policy)
+	if err != nil {
+		return err
+	}
+	// construct a policy statement
+	permission := statement{
+		Sid:    SnsPolicySid,
+		Effect: SnsPolicyEffect,
+		Principal: principal{
+			Service: SnsPolicyService,
+		},
+		Action:   SnsPolicyAction,
+		Resource: topicArn,
+	}
+	// add the new policy statement to the existing one(s)
+	policy.Statement = append(policy.Statement, permission)
+	// marshal the updated policy
+	newPolicyBytes, err := json.Marshal(policy)
+	if err != nil {
+		return err
+	}
+	// update topic attributes with the new policy
+	newPolicy := string(newPolicyBytes)
+	saInput := &sns.SetTopicAttributesInput{
+		AttributeName:  aws.String(SnsPolicy),
+		AttributeValue: aws.String(newPolicy),
+		TopicArn:       aws.String(topicArn),
+	}
+	_, err = obj.snsClient.SetTopicAttributes(saInput)
+	if err != nil {
+		return err
+	}
+	log.Printf("%s: Authorized Target", obj)
 	return nil
 }
 
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -91,9 +91,11 @@ const (
 type awsEc2Event uint8
 
 const (
-	awsEc2EventWatchReady awsEc2Event = iota
+	awsEc2EventNone awsEc2Event = iota
+	awsEc2EventWatchReady
 	awsEc2EventInstanceStopped
 	awsEc2EventInstanceRunning
+	awsEc2EventInstanceTerminated
 	awsEc2EventInstanceExists
 )
 
@@ -200,6 +202,12 @@ type postData struct {
 	Type    string `json:"Type"`
 	Token   string `json:"Token"`
 	Message string `json:"Message"`
+}
+
+// postMsg is used to unmarshal the postData message if it's an event notification.
+type postMsg struct {
+	InstanceID string `json:"instance-id"`
+	State      string `json:"state"`
 }
 
 // Default returns some sensible defaults for this resource.
@@ -960,6 +968,7 @@ func (obj *AwsEc2Res) snsListener(addr string) (net.Listener, error) {
 }
 
 // snsPostHandler listens for posts on the SNS Endpoint.
+// TODO: download pem and check message against signature
 func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
 	if req.Method != "POST" {
 		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
@@ -1000,6 +1009,28 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
 			event: awsEc2EventWatchReady,
 		}:
 		case <-obj.closeChan:
+		}
+	}
+	// process cloudwatch event notifications.
+	if post.Type == "Notification" {
+		event, err := obj.snsProcessEvent(post.Message, obj.prependName())
+		if err != nil {
+			select {
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "error confirming subscription"),
+			}:
+			case <-obj.closeChan:
+			}
+			return
+
+		}
+		if event != awsEc2EventNone {
+			select {
+			case obj.awsChan <- &chanStruct{
+				event: event,
+			}:
+			case <-obj.closeChan:
+			}
 		}
 	}
 }


jonathangold:
@purpleidea: split it into more manageable commits, and as a bonus, got really good with rebasing. :) FYI, it's the "Added sns functions to watch" commit that throws off the diff, as it indents the old watch. But nothing changed beyond lines 201-208 and the closing brace after the old watch. Btw I'll fix the past tense messages during the review phase.

jonathangold:
@purpleidea: split it into more manageable commits, and as a bonus, got really good with rebasing. :) FYI, it's the "Added sns functions to watch" commit that throws off the diff, as it indents the old watch. But nothing changed beyond lines 201-208 and the closing brace after the old watch. Btw I'll fix the past tense messages during the review phase.

jonathangold:
One thing that's missing is a timeout to shut the server down if there's a configuration error (ports not open, etc...) As of right now, it just hangs in that case, waiting for some post data.

resources/aws_ec2.go
```diff
@@ -20,25 +20,46 @@ package resources
...
 func init() {
 	RegisterResource("aws:ec2", func() Res { return &AwsEc2Res{} })
 }
 
-// AwsPrefix is a const which gets prepended onto the instance name.
-const AwsPrefix = "mgmt:"
+const (
+	// AwsPrefix is a const which gets prepended onto the instance name.
+	AwsPrefix = "mgmt:"
+	// waitTimeout is the duration of the timeout context in CheckApply
```
purpleidea: add "in seconds" or whatever unit to comment please

resources/aws_ec2.go
```diff
@@ -20,25 +20,46 @@ package resources
...
+	waitTimeout = 400
+	// topicName is the name of the sns topic created by makeSnsTopic
+	topicName = "mgmtconfigEC2events"
```
purpleidea: Does this work when we have multiple aws resources running simultaneously or is there a topic collision? In any case, It would nice if we built up these constants from a logical mgmt prefix. Eg:

```
const (
    AwsPrefix = "_mgmt"
    SnsPrefix = AwsPrefix + ":sns"
)
```

Or whatever makes sense logically and aesthetically. Same for the rule below.

resources/aws_ec2.go
```diff
@@ -20,25 +20,46 @@ package resources
...
+	// ruleName is the name of the cloudwatch topic created by makeCloudWatchRule
+	ruleName = "mgmtconfigEC2state"
+	// eventPattern describes the events the cloudwatch rule should detect
+	eventPattern = `{"source":["aws.ec2"],"detail-type":["EC2 Instance State-change Notification"],"detail":{"state":["running","stopped","terminated"]}}`
+	// snsPermission describes the security rule to allow cloudwatch to publish on sns
+	snsPermission = `{"Sid":"AllowSNSforCloudWatch","Effect":"Allow","Principal":{"Service":"events.amazonaws.com"},"Action":["SNS:Publish"],"Resource":"%s"}`
+	detail        = "$.detail"
```
purpleidea: What are these? Add comments and/or clarify their var names. Probably don't need to call them `fooConst`, we know they're constant, specify the things we don't know. Eg: "snsProto" or similar.

resources/aws_ec2.go
```diff
@@ -66,17 +87,31 @@ var AwsRegions = []string{
...
 	// UserData is used to run bash and cloud-init commands on first launch.
 	// See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html
 	// for documantation and examples.
 	UserData string `yaml:"userdata"`
 
-	client *ec2.EC2 // client session for AWS API calls
+	client    *ec2.EC2              // client session for AWS API calls
+	snsClient *sns.SNS              // client session for SNS API calls
+	cweClient *cwe.CloudWatchEvents // client session for SNS API calls
+	srv       *http.Server          // http server for SNS endpont
```
purpleidea: s/srv/snsServer/ like snsClient ?

maybe group these things into logical chunks separate by whitespace?
you've got a copy+pasta error on one of the comments here.

resources/aws_ec2.go
```diff
@@ -66,17 +87,31 @@ var AwsRegions = []string{
...
-	State   string `yaml:"state"`   // state: running, stopped, terminated
-	Region  string `yaml:"region"`  // region must match an element of AwsRegions
-	Type    string `yaml:"type"`    // type of ec2 instance, ie. t2.micro
-	ImageID string `yaml:"imageid"` // imageid must be available on the chosen region
+	BaseRes     `yaml:",inline"`
+	State       string `yaml:"state"`       // state: running, stopped, terminated
+	Region      string `yaml:"region"`      // region must match an element of AwsRegions
+	Type        string `yaml:"type"`        // type of ec2 instance, ie. t2.micro
+	ImageID     string `yaml:"imageid"`     // imageid must be available on the chosen region
+	SnsEndpoint string `yaml:"snsendpoint"` // the public url endpoint to subscribe to events, ie. http://server:12345/
```
purpleidea: While it is entirely correct that this is an SnsEndpoint, this is *our* public API that we're building. So if we can abstract away the weird names of things that amazon invents and instead present more logical interfaces, this is a good idea IMO. For example, we might instead want to call this `WatchEndpoint`, `WebHookEndpoint` or something else clever. I need to think on it, but maybe you have an idea of something that doesn't include "SNS".

resources/aws_ec2.go
```diff
@@ -66,17 +87,31 @@ var AwsRegions = []string{
...
 	// for documantation and examples.
 	UserData string `yaml:"userdata"`
 
-	client *ec2.EC2 // client session for AWS API calls
+	client    *ec2.EC2              // client session for AWS API calls
+	snsClient *sns.SNS              // client session for SNS API calls
+	cweClient *cwe.CloudWatchEvents // client session for SNS API calls
+	srv       *http.Server          // http server for SNS endpont
+	awsChan   chan *chanStruct      // channel for aws events
+	closeChan chan struct{}         // a channel to receive signals shut down
```
purpleidea: please clarify grammar on these two comments

resources/aws_ec2.go
```diff
@@ -66,17 +87,31 @@ var AwsRegions = []string{
...
+	topicArn  string                // the topicArn we need to delete the topic when we're shutting down
+}
+
+// a struct for passing data around watch
```
purpleidea: please follow the doc convention even though this struct is private. it also should be a sentence and end with a .

resources/aws_ec2.go
```diff
@@ -158,60 +200,36 @@ func (obj *AwsEc2Res) Init() error {
...
-				case awsChan <- &chanStruct{
-					err: errwrap.Wrapf(err, "error describing instances"),
-				}:
-				case <-closeChan:
-				}
-				return
-			}
-			if obj.State == "running" {
-				stoppedInput := &ec2.DescribeInstancesInput{
+	if obj.SnsEndpoint != "" {
```
purpleidea: below here in `Watch()` patch is basically unreadable as a single commit. Might have to review individual commits.
If I understand correctly: We're basically doing Watch type (1) or Watch type (2) and they otherwise don't share much code, correct? Where (1) is the long poll, (2) is the sns thing, and maybe in the future (3) is an sqs long poller. If that's the case, it's probably best to have three top level Watch methods, like `WatchLongPoll`, `WatchSns` and `WatchSqs` and the main watch calls the appropriate one as needed.

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
 func (obj *AwsEc2Res) prependName() string {
 	return AwsPrefix + obj.GetName()
 }
+
+// snsServer returns an http server used to listen for sns messages
```
purpleidea: end with a .

resources/aws_ec2.go
```diff
@@ -66,17 +87,31 @@ var AwsRegions = []string{
...
-	Region  string `yaml:"region"`  // region must match an element of AwsRegions
-	Type    string `yaml:"type"`    // type of ec2 instance, ie. t2.micro
-	ImageID string `yaml:"imageid"` // imageid must be available on the chosen region
+	BaseRes     `yaml:",inline"`
+	State       string `yaml:"state"`       // state: running, stopped, terminated
+	Region      string `yaml:"region"`      // region must match an element of AwsRegions
+	Type        string `yaml:"type"`        // type of ec2 instance, ie. t2.micro
+	ImageID     string `yaml:"imageid"`     // imageid must be available on the chosen region
+	SnsEndpoint string `yaml:"snsendpoint"` // the public url endpoint to subscribe to events, ie. http://server:12345/
+	SnsPort     uint16 `yaml:"snsport"`     // the local port connected to the above url, ie. 23456
```
purpleidea: I think you've simplified the API to only specify a listen port, why not allow the whole connection addr? Eg: why not `SnsListenAddr` or something. Of course it's fine if you find only an int that you turn it into fmt.Sprintf(":%d, theint), but otherwise maybe they only want to listen on the 24.12.2.5 network instead of 0.0.0.0 ?

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+// snsServer returns an http server used to listen for sns messages
+func (obj *AwsEc2Res) snsServer() *http.Server {
+	port := fmt.Sprintf(":%d", int(obj.SnsPort))
+	srv := &http.Server{Addr: port}
+	log.Printf("%s: Starting SNS Endpoint", obj)
+	//rand.Seed(time.Now().UTC().UnixNano())
+	http.HandleFunc("/", obj.postHandler)
+	log.Printf("%s: Listening on %s", obj, port)
+	go func() {
```
purpleidea: anytime we have go routines, I expect to see waitgroups somewhere...

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		}
+	}()
+	return srv
+}
+
+// postHandler listens for posts on the SNS Endpoint
+// TODO: download pem and check message against signature
+func (obj *AwsEc2Res) postHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method == "POST" {
+		post, _ := ioutil.ReadAll(req.Body)
```
purpleidea: apparently there is a cleaner way to read that streams instead. In case this is an easy change and it's better, can you look into it?

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		if err != nil {
+			select {
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "error unmarshaling policy"),
+			}:
+			case <-obj.closeChan:
+			}
+			return
+		}
+		if p.Type == "SubscriptionConfirmation" {
```
purpleidea: Is this a constant somewhere in the aws API?

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+			}
+		}
+	} else {
+		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
+	}
+}
+
+// makeSnsTopic creates a topic on aws sns
+func (obj *AwsEc2Res) makeSnsTopic() {
+	//make sns client
```
purpleidea: nit: space after //

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+				return
+			}
+		}
+	} else {
+		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
+	}
+}
+
+// makeSnsTopic creates a topic on aws sns
+func (obj *AwsEc2Res) makeSnsTopic() {
```
purpleidea: this is basically a method with a bunch of side effects... Why isn't it a regular method that returns `error` ?
In the caller we can these do this `select` stuff or not, depending on the way it's called and if needed...
same comment for a few other functions...

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		select {
+		case obj.awsChan <- &chanStruct{
+			err: errwrap.Wrapf(err, "error getting topic attributes"),
+		}:
+		case <-obj.closeChan:
+		}
+		return
+	}
+	// unmarshal the current sns publishing policy
+	type statement struct {
```
purpleidea: are these structs already defined in the aws lib somewhere?

resources/aws_ec2.go
```diff
@@ -663,3 +714,365 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		}
+		return
+	}
+	log.Printf("%s: Authorized Target", obj)
+}
+
+// Close cleans up when we're done. This is needed to delete some of the AWS
+// objects created for the SNS endpoint.
+func (obj *AwsEc2Res) Close() error {
+	var errors error
```
purpleidea: please pick a different var name as this is a common package import. Maybe `errs` or `errList` ?

resources/aws_ec2.go
jonathangold: I guess it makes sense to have a unique topic for each resource, ie. SnsPrefix + obj.prependName(), that way, we don't see events from other instances.

resources/aws_ec2.go
jonathangold: it originally was just sns, but that name is taken by the api. Will do.

resources/aws_ec2.go
jonathangold: To clarify, return the error and do if err = makeSnsTopic() { handle error } where it's called?

purpleidea: I think that pattern would be more correct. It also means each of these small worker functions have fewer side effects, and the large error handling sections using select would be de-duplicated.

resources/aws_ec2.go
jonathangold: Nope. :( Not that I could find, anyway. I did find "sns" though.

resources/aws_ec2.go
jonathangold: Can't find these in in sns or aws. Not sure where else they could be.

resources/aws_ec2.go
```diff
@@ -20,25 +20,49 @@ package resources
...
 
 func init() {
 	RegisterResource("aws:ec2", func() Res { return &AwsEc2Res{} })
 }
 
-// AwsPrefix is a const which gets prepended onto the instance name.
-const AwsPrefix = "mgmt:"
+const (
+	// AwsPrefix is a const which gets prepended onto object names.
+	AwsPrefix = "_mgmt-"
```
purpleidea: I kind of liked the colon. Can we use that instead of a dash? Or is it better. Everything else looks great!

FYI: one liner comments don't need to be full sentences, but it's okay if they are. No changes needed.

resources/aws_ec2.go
```diff
@@ -66,17 +90,85 @@ var AwsRegions = []string{
...
-	Region  string `yaml:"region"`  // region must match an element of AwsRegions
-	Type    string `yaml:"type"`    // type of ec2 instance, ie. t2.micro
-	ImageID string `yaml:"imageid"` // imageid must be available on the chosen region
+	BaseRes         `yaml:",inline"`
+	State           string `yaml:"state"`           // state: running, stopped, terminated
+	Region          string `yaml:"region"`          // region must match an element of AwsRegions
+	Type            string `yaml:"type"`            // type of ec2 instance, ie. t2.micro
+	ImageID         string `yaml:"imageid"`         // imageid must be available on the chosen region
+	WatchEndpoint   string `yaml:"watchendpoint"`   // the public url endpoint to subscribe to events, ie. http://server:12345/
+	WatchListenAddr string `yaml:"watchlistenaddr"` // the local port connected to the above url, ie. 23456
```
purpleidea: s/ the local port/the local listen string/ ?? or similar (the comment)

huge nit: pick `ie:` `i.e.:` or `ex:` `ie.`, isn't a thing, eg: https://en.wiktionary.org/wiki/i.e.

resources/aws_ec2.go
```diff
@@ -66,17 +90,85 @@ var AwsRegions = []string{
...
 	// UserData is used to run bash and cloud-init commands on first launch.
 	// See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html
 	// for documantation and examples.
 	UserData string `yaml:"userdata"`
 
-	client *ec2.EC2 // client session for AWS API calls
```
purpleidea: I think the logical grouping in this struct isn't 100% please either re-arrange or explain in comments here to me why it makes sense

resources/aws_ec2.go
```diff
@@ -66,17 +90,85 @@ var AwsRegions = []string{
...
+
+	topicArn string // the topicArn needed to identify the sns topic
+}
+
+// snsPolicy denotes the structure of sns security policies.
+type snsPolicy struct {
+	Version   string      `json:"Version"`
+	ID        string      `json:"Id"`
+	Statement []statement `json:"Statement"`
+}
```
purpleidea: 1 nl's between structs, struct comments at the top are nice too.

resources/aws_ec2.go
```diff
@@ -158,30 +257,111 @@ func (obj *AwsEc2Res) Init() error {
...
 		select {
-		case <-closeChan:
-			cancel()
+		case <-obj.closeChan:
+			obj.cancel()
 		}
 	}()
-	wg.Add(1)
-	defer wg.Wait()
+	if obj.WatchEndpoint != "" {
```
purpleidea: This part is good! However, do you think instead we should have these functions have the same signature as the main Watch and just do `return obj.longpollWatch()` for example? IOW is the rest of the Watch code sufficiently similar that it's worth sharing?

resources/aws_ec2.go
jonathangold: No colons allowed in topic names (or cloudwatch rules afaict) :( Tried slashes and pipes too.

purpleidea: SO *that's* valuable information that should be in a source comment!

jonathangold: gotcha

resources/aws_ec2.go
jonathangold: My groups were:
- client objects
- server objects
- channels
- waitgroups (combine with channels?)
- context stuff
- data
Did you have something else in mind in particular?

resources/aws_ec2.go
```diff
@@ -174,22 +257,95 @@ func (obj *AwsEc2Res) Init() error {
...
 	return obj.longpollWatch()
 }
 
+// snsWatch uses amazon cloudwatch events and simple notification service to
+// detect ec2 instance state changes.
+func (obj *AwsEc2Res) snsWatch() error {
+	send := false
+	var exit *error
+	var err error
+	if err = obj.snsMakeTopic(); err != nil {
```
jonathangold: @purpleidea: This is clearly a more sensible pattern. Not sure what I was thinking.

purpleidea: If all of these functions have side effects, eg they are reading or setting some value in the struct, then it's much harder to follow what's happening. In this case, if it's not needed that the variable be in the struct, have the values get passed in as arguments, and output and then passed to the next function.

In some cases, it's even fine to get the output, and then in this visible code store it in the struct after the function ends.

jonathangold: That makes sense. Gotta get back into the haskell/functional programming mindset. Thanks

PR #239, opened 2017-11-08
Refactor channels and waitgroup

jonathangold:
Moved the waitgroup and channels into the resource struct to prepare for additional watch methods.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -85,6 +85,7 @@ type AwsEc2Res struct {
 
 	awsChan   chan *chanStruct // channel used to send events and errors to Watch()
 	closeChan chan struct{}    // channel used to cancel context when it's time to shut down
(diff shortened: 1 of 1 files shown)


purpleidea:
I think this LGTM :)

resources/aws_ec2.go
```diff
@@ -85,6 +85,8 @@ type AwsEc2Res struct {
 
 	awsChan   chan *chanStruct // channel used to send events and errors to Watch()
 	closeChan chan struct{}    // channel used to cancel context when it's time to shut down
+
```
purpleidea: No reason these can't be all together afaik

resources/aws_ec2.go
```diff
@@ -168,6 +170,8 @@ func (obj *AwsEc2Res) Init() error {
 	obj.awsChan = make(chan *chanStruct)
 	obj.closeChan = make(chan struct{})
```
purpleidea: -nl?

PR #240, opened 2017-11-08
Add http server to Aws:Ec2

jonathangold:
Add an http server that will ultimately be used as an endpoint for amazon SNS.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -21,7 +21,10 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"io/ioutil"
 	"log"
+	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
@@ -74,8 +77,10 @@ type AwsEc2Res struct {
 	BaseRes `yaml:",inline"`
 	State   string `yaml:"state"`   // state: running, stopped, terminated
 	Region  string `yaml:"region"`  // region must match an element of AwsRegions
-	Type    string `yaml:"type"`    // type of ec2 instance, ie. t2.micro
+	Type    string `yaml:"type"`    // type of ec2 instance, eg: t2.micro
 	ImageID string `yaml:"imageid"` // imageid must be available on the chosen region
+
+	WatchListenAddr string `yaml:"watchlistenaddr"` // the local address or port that the sns listens on, eg: 10.0.0.0:23456 or 23456
 	// UserData is used to run bash and cloud-init commands on first launch.
 	// See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html
 	// for documantation and examples.
@@ -676,4 +681,30 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 func (obj *AwsEc2Res) prependName() string {
 	return AwsPrefix + obj.GetName()
+}
+
+// snsServer returns an http server used to listen for sns messages.
+func (obj *AwsEc2Res) snsServer() *http.Server {
+	addr := obj.WatchListenAddr
+	// if addr is a port
+	if _, err := strconv.Atoi(obj.WatchListenAddr); err == nil {
+		addr = fmt.Sprintf(":%s", obj.WatchListenAddr)
+	}
+	handler := http.HandlerFunc(obj.snsPostHandler)
+	return &http.Server{
+		Addr:    addr,
+		Handler: handler,
+	}
+}
+
+// snsPostHandler listens for posts on the SNS Endpoint.
+func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method != "POST" {
+		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
+		return
+	}
+	post, _ := ioutil.ReadAll(req.Body)
+	if obj.debug {
+		log.Printf("%s: Post: %s", obj, string(post))
+	}
 }


resources/aws_ec2.go
```diff
@@ -71,11 +74,12 @@ var AwsRegions = []string{
...
-	State   string `yaml:"state"`   // state: running, stopped, terminated
-	Region  string `yaml:"region"`  // region must match an element of AwsRegions
-	Type    string `yaml:"type"`    // type of ec2 instance, ie. t2.micro
-	ImageID string `yaml:"imageid"` // imageid must be available on the chosen region
+	BaseRes         `yaml:",inline"`
+	State           string `yaml:"state"`           // state: running, stopped, terminated
+	Region          string `yaml:"region"`          // region must match an element of AwsRegions
+	Type            string `yaml:"type"`            // type of ec2 instance, eg: t2.micro
+	ImageID         string `yaml:"imageid"`         // imageid must be available on the chosen region
+	WatchListenAddr string `yaml:"watchlistenaddr"` // the local address or port connected to the above url, eg: 10.0.0.0:23456 or 23456
```
purpleidea: what "above url" ?
(i realize this might come later, but yeah, maybe reword?)

resources/aws_ec2.go
```diff
@@ -677,3 +681,40 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+			case <-obj.closeChan:
+			}
+			return
+		}
+	}()
+	return srv
+}
+
+// postHandler listens for posts on the SNS Endpoint.
+func (obj *AwsEc2Res) postHandler(w http.ResponseWriter, req *http.Request) {
```
purpleidea: since this is just used by sns, maybe `snsPostHandler` ?

resources/aws_ec2.go
```diff
@@ -677,3 +681,40 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	if req.Method == "POST" {
+		post, _ := ioutil.ReadAll(req.Body)
+		log.Printf(string(post))
+	} else {
```
purpleidea: can we flip this and have an early return?
eg req.Method != "POST" { err, return }

resources/aws_ec2.go
```diff
@@ -677,3 +681,40 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	// if addr is a port
+	if _, err := strconv.Atoi(obj.WatchListenAddr); err == nil {
+		addr = fmt.Sprintf(":%s", obj.WatchListenAddr)
+	}
+	srv := &http.Server{Addr: addr}
+	log.Printf("%s: Starting SNS Endpoint", obj)
+	http.HandleFunc("/", obj.postHandler)
+	log.Printf("%s: Listening on %s", obj, addr)
+	obj.wg.Add(1)
+	go func() {
```
purpleidea: so this part needs some discussion...
i'm tempted to say that this `run` part of the server should be a different method to keep the awsChan stuff out of this.
eg: this should *only* build the server... meaning we might rename to `snsServer` for example.
perhaps we could merge this without this goroutine until it's clear how best it should get called?
happy to review that as a second patch in this PR if that's what comes next. LMK what you think. (don't just blindly agree with me, disagreeing is fine)

jonathangold: I think that's fine. Separation of concerns and all that. I'll remove it from this commit, and add a second one with the listener.

resources/aws_ec2.go
```diff
@@ -677,3 +681,27 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+}
+
+// snsPostHandler listens for posts on the SNS Endpoint.
+func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method != "POST" {
+		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
+		return
+	}
+	post, _ := ioutil.ReadAll(req.Body)
+	log.Printf(string(post))
```
purpleidea: stale debug message?

resources/aws_ec2.go
```diff
@@ -677,3 +681,27 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+// snsServer returns an http server used to listen for sns messages.
+func (obj *AwsEc2Res) snsServer() *http.Server {
+	addr := obj.WatchListenAddr
+	// if addr is a port
+	if _, err := strconv.Atoi(obj.WatchListenAddr); err == nil {
+		addr = fmt.Sprintf(":%s", obj.WatchListenAddr)
+	}
+	srv := &http.Server{Addr: addr}
+	log.Printf("%s: Starting SNS Endpoint", obj)
```
purpleidea: Are these two log messages still in the same place and/or relevant/useful?

resources/aws_ec2.go
```diff
@@ -677,3 +681,27 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 }
+
+// snsServer returns an http server used to listen for sns messages.
+func (obj *AwsEc2Res) snsServer() *http.Server {
+	addr := obj.WatchListenAddr
+	// if addr is a port
+	if _, err := strconv.Atoi(obj.WatchListenAddr); err == nil {
+		addr = fmt.Sprintf(":%s", obj.WatchListenAddr)
+	}
+	srv := &http.Server{Addr: addr}
```
purpleidea: you can return the struct directly, no need for the variable

resources/aws_ec2.go
```diff
@@ -677,3 +681,27 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	log.Printf("%s: Starting SNS Endpoint", obj)
+	http.HandleFunc("/", obj.snsPostHandler)
```
purpleidea: weird that this HandleFunc isn't associated with the particular server at all... Is this right? How does this work? What if we had two different http servers running on different ports? isn't that allowed?

purpleidea: https://golang.org/src/net/http/server.go?#L2345 is needed AFAICT

resources/aws_ec2.go
```diff
@@ -677,3 +681,30 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	// if addr is a port
+	if _, err := strconv.Atoi(obj.WatchListenAddr); err == nil {
+		addr = fmt.Sprintf(":%s", obj.WatchListenAddr)
+	}
+	handler := http.HandlerFunc(obj.snsPostHandler)
+	return &http.Server{
+		Addr:    addr,
+		Handler: handler,
+	}
+
```
purpleidea: nit: kill this extra nl please

resources/aws_ec2.go
```diff
@@ -677,3 +681,30 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+// snsPostHandler listens for posts on the SNS Endpoint.
+func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method != "POST" {
+		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
+		return
+	}
+	post, _ := ioutil.ReadAll(req.Body)
+	if obj.debug {
+		log.Printf(string(post))
```
purpleidea: do you want to add some sort of prefix to this log message so we know where it came from?

resources/aws_ec2.go
```diff
@@ -71,11 +74,12 @@ var AwsRegions = []string{
...
-	State   string `yaml:"state"`   // state: running, stopped, terminated
-	Region  string `yaml:"region"`  // region must match an element of AwsRegions
-	Type    string `yaml:"type"`    // type of ec2 instance, ie. t2.micro
-	ImageID string `yaml:"imageid"` // imageid must be available on the chosen region
+	BaseRes         `yaml:",inline"`
+	State           string `yaml:"state"`           // state: running, stopped, terminated
+	Region          string `yaml:"region"`          // region must match an element of AwsRegions
+	Type            string `yaml:"type"`            // type of ec2 instance, eg: t2.micro
+	ImageID         string `yaml:"imageid"`         // imageid must be available on the chosen region
+	WatchListenAddr string `yaml:"watchlistenaddr"` // the local address or port that the sns listens on, eg: 10.0.0.0:23456 or 23456
```
purpleidea: Let's add one nl before this WatchListenAddr

Might as well keep the basic components in the first chunk. Sorry for the late thought, just trying to think ahead on how to visualize and deal with all the complexity. Thanks!

resources/aws_ec2.go
```diff
@@ -677,3 +682,29 @@ func (obj *AwsEc2Res) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	}
+}
+
+// snsPostHandler listens for posts on the SNS Endpoint.
+func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method != "POST" {
+		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
+		return
+	}
+	post, _ := ioutil.ReadAll(req.Body)
```
purpleidea: why don't we check this error? (sorry)
more importantly, what do we even ever do in this method? it seems basically useless... (or is something coming later?)

jonathangold: Yeah. This is just so it has something to do. Once snsSubscribe is done, this will use json.Decoder to read the post, and we can get rid of the ioutil stuff.

PR #241, opened 2017-11-09
Add snsMakeTopic

jonathangold:
For review after the http server patch is merged. Next is the making and confirming the subscription.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -33,6 +33,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/sns"
+	multierr "github.com/hashicorp/go-multierror"
 	errwrap "github.com/pkg/errors"
 )
 
@@ -44,6 +46,12 @@ const (
 	// AwsPrefix is a const which gets prepended onto object names. We can only use
 	// alphanumeric chars, underscores and hyphens for sns topics and cloud watch rules.
 	AwsPrefix = "_mgmt-"
+	// Ec2Prefix is added to the names of sns and cloudwatch objects.
+	Ec2Prefix = AwsPrefix + "ec2-"
+	// SnsPrefix gets prepended onto the sns topic.
+	SnsPrefix = Ec2Prefix + "sns-"
+	// SnsTopicName is the name of the sns topic created by snsMakeTopic.
+	SnsTopicName = SnsPrefix + "events"
 	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
 	waitTimeout = 400
 )
@@ -87,6 +95,11 @@ type AwsEc2Res struct {
 	UserData string `yaml:"userdata"`
 
 	client *ec2.EC2 // client session for AWS API calls
+
+	snsClient *sns.SNS // client for AWS SNS API calls
+	// snsTopicArn requires looping through every topic to get,
+	// so we save it here when we create the topic instead.
+	snsTopicArn string
 
 	awsChan   chan *chanStruct // channel used to send events and errors to Watch()
 	closeChan chan struct{}    // channel used to cancel context when it's time to shut down
@@ -174,6 +187,25 @@ func (obj *AwsEc2Res) Init() error {
 	obj.awsChan = make(chan *chanStruct)
 	obj.closeChan = make(chan struct{})
 	obj.wg = &sync.WaitGroup{}
+
+	// if we are using sns watch
+	if obj.WatchListenAddr != "" {
+		// make sns client
+		snsSess, err := session.NewSession(&aws.Config{
+			Region: aws.String(obj.Region),
+		})
+		if err != nil {
+			return errwrap.Wrapf(err, "error creating sns session")
+		}
+		obj.snsClient = sns.New(snsSess)
+		// make the sns topic
+		snsTopicArn, err := obj.snsMakeTopic()
+		if err != nil {
+			return errwrap.Wrapf(err, "error making sns topic")
+		}
+		// save the topicArn for later use
+		obj.snsTopicArn = snsTopicArn
+	}
 
 	return obj.BaseRes.Init() // call base init, b/c we're overriding
 }
@@ -656,6 +688,9 @@ func (obj *AwsEc2Res) Compare(r Res) bool {
 	if obj.UserData != res.UserData {
 		return false
 	}
+	if obj.WatchListenAddr != res.WatchListenAddr {
+		return false
+	}
 	return true
 }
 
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -44,6 +45,12 @@ const (
 	// AwsPrefix is a const which gets prepended onto object names. We can only use
 	// alphanumeric chars, underscores and hyphens for sns topics and cloud watch rules.
 	AwsPrefix = "_mgmt-"
+	// Ec2Prefix is added to the names of sns and cloudwatch objects.
+	Ec2Prefix = AwsPrefix + "ec2-"
+	// SnsPrefix gets prepended onto the sns topic.
+	SnsPrefix = Ec2Prefix + "sns-"
+	// topicName is the name of the sns topic created by snsMakeTopic.
+	topicName = SnsPrefix + "events"
```
purpleidea: Should this be private? If it's sns specific can you call it SnsTopicName? or snsTopicName

resources/aws_ec2.go
```diff
@@ -87,6 +94,9 @@ type AwsEc2Res struct {
 
 	client *ec2.EC2 // client session for AWS API calls
 
+	snsClient *sns.SNS // client session for SNS API calls
+	topicArn  string   // the unique id of the SNS topic
```
purpleidea: snsTopicArn

resources/aws_ec2.go
```diff
@@ -708,3 +718,30 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+// snsMakeTopic creates a topic on aws sns.
+func (obj *AwsEc2Res) snsMakeTopic() error {
+	// make sns client
+	snsSess, err := session.NewSession(&aws.Config{
+		Region: aws.String(obj.Region),
+	})
+	if err != nil {
+		return err
+	}
+	obj.snsClient = sns.New(snsSess)
```
purpleidea: do we use snsClient somewhere later outside of this function? IOW why does it need to be in the struct instead of returning it or only having it as a local var?

resources/aws_ec2.go
```diff
@@ -708,3 +718,30 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+		Name: aws.String(topicName),
+	}
+	topic, err := obj.snsClient.CreateTopic(topicInput)
+	if err != nil {
+		return err
+	}
+	log.Printf("%s: Created SNS Topic", obj)
+
+	if topic.TopicArn != nil {
+		obj.topicArn = *topic.TopicArn
```
purpleidea: why do we store this instead of returning it?

resources/aws_ec2.go
purpleidea: It's fine to do it this way as a side effect, just contrast with potentially returning it and then dealing with it that way instead. Both are valid, different advantages and disadvantages for both.

resources/aws_ec2.go
jonathangold: This will be used in snsSubscribe, snsAuthorize, cweTargetRule, and subscribe. I suppose I could return it and pass it in as an argument to those functions if you think that's cleaner. It is also used in Close() which means I'd have to get the topicArn again another way. There's sns.ListTopics, but that returns all topics, not just the one we're concerned with.

resources/aws_ec2.go
jonathangold: Is there a reason you think it should be public?

purpleidea: It makes sense for this to be public if an external non-mgmt tool might want to look up this value to know what key is expected on the aws side. Eg: so that it also doesn't conflict with it for example. If this isn't the case, then it can be private. (I genuinely don't know which is needed.)

resources/aws_ec2.go
purpleidea: I'm not really sure. But it might be worth rebasing the large PR with everything in it so I can get a slimmed down view of the big picture again.

resources/aws_ec2.go
jonathangold: Fair enough. Best to avoid collisions as much as possible.

resources/aws_ec2.go
```diff
@@ -175,6 +185,17 @@ func (obj *AwsEc2Res) Init() error {
 	obj.closeChan = make(chan struct{})
 	obj.wg = &sync.WaitGroup{}
 
+	if obj.WatchListenAddr != "" {
```
purpleidea: Does any of this have a matching close or cleanup part it should do that we need to do in `Close()` ?

resources/aws_ec2.go
```diff
@@ -708,3 +729,21 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
 		log.Printf("%s: Post: %s", obj, string(post))
 	}
 }
+
+// snsMakeTopic creates a topic on aws sns.
+func (obj *AwsEc2Res) snsMakeTopic() error {
```
purpleidea: any reason why the function signature isn't something like (string, error) where the string value is the topicArn I suppose? In particular, while looking at pr#242 I don't actually see this function being called anywhere. Should it happen in init? in checkapply?

resources/aws_ec2.go
```diff
@@ -175,6 +186,24 @@ func (obj *AwsEc2Res) Init() error {
...
+			return errwrap.Wrapf(err, "error creating sns session")
+		}
+		obj.snsClient = sns.New(snsSess)
+
+		// make the sns topic
+		topicArn, err := obj.snsMakeTopic()
+		if err != nil {
+			return errwrap.Wrapf(err, "error making sns topic")
+		}
+		obj.snsTopicArn = topicArn
```
purpleidea: struct variant is called `obj.snsTopicArn` so why not call the var `snsTopicArn` ?

resources/aws_ec2.go
```diff
@@ -708,3 +737,42 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	return topicArn, nil
+}
+
+// Close cleans up when we're done. This is needed to delete some of the AWS
+// objects created for the SNS endpoint.
+func (obj *AwsEc2Res) Close() error {
+	var errList error
+	if obj.WatchListenAddr != "" {
+		// delete the topic
+		dtInput := &sns.DeleteTopicInput{
```
purpleidea: one nit about asymmetry: you have a `snsMakeTopic` to do the making, where as the inverse to that is just inline in the close method. It's not a hard rule, but usually both would either be inline, or both would be in methods. I think the latter might be appropriate.

2nd thing: suppose `Init` runs, but returns early with `"error creating sns session"`, and then `Close` gets called-- will running this DeleteTopic stuff error because we didn't have a proper session? Probably good to make sure that code is safe in both scenarios. All the better for it to have it's own method to do so and grow in.

resources/aws_ec2.go
```diff
@@ -708,3 +737,42 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		return "", err
+	}
+	log.Printf("%s: Created SNS Topic", obj)
+
+	var topicArn string
+	if topic.TopicArn != nil {
+		topicArn = *topic.TopicArn
+	}
+	return topicArn, nil
```
purpleidea: Is returning an empty string for topicArn valid, or should that be an error scenario? Because the current code _could_ return ("", nil) -- Just FYI.

jonathangold: It would error out further down the line, but I can nip it in the bud here with an if ... != ""

resources/aws_ec2.go
jonathangold: will do.

jonathangold: It will error, but if there was an error creating the session in Init, it would error anyway. I'm not sure there's any benefit in creating it to delete objects that, by definition, can't have been made in the first place.

purpleidea: What I mean was, `Close` should not unnecessarily error trying to delete something via a session that does not exist because it was not created in `Init`. The solution isn't to try and create it in `Close`-- the solution is to skip over the deletion in `Close` if the session is nil.

resources/aws_ec2.go
```diff
@@ -708,3 +750,50 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	return nil
+}
+
+// Close cleans up when we're done. This is needed to delete some of the AWS
+// objects created for the SNS endpoint.
+func (obj *AwsEc2Res) Close() error {
+	var errList error
+	if obj.WatchListenAddr != "" {
+		// make sns client
+		snsClient, err := obj.snsClient()
```
purpleidea: Why didn't we just re-use the client from Init() ? You took that out... Maybe I did not explain clearly. Better explanation here: https://github.com/purpleidea/mgmt/pull/241#discussion_r150230274

resources/aws_ec2.go
```diff
@@ -708,3 +750,50 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	// make topic
+	topicInput := &sns.CreateTopicInput{
+		Name: aws.String(SnsTopicName),
+	}
+	topic, err := snsClient.CreateTopic(topicInput)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("%s: Created SNS Topic", obj)
+	return topic.TopicArn, nil
```
purpleidea: You've now created a scenario where you can return nil, nil. Is this what you want? It's legal, and exists in some situations, but typically that has a special meaning. What is it in this case? Will this cause a panic anywhere?

purpleidea:
Nice :)

PR #242, opened 2017-11-09
Aws make and confirm subscription

jonathangold:
For review after #241 is merged.
In https://github.com/purpleidea/mgmt/pull/242/commits/8ef2edf94ff8df091dbd1e967843da829ef1dbac I pass the errors to awsChan, as I can't return an error from a postHandler afaict. If there's a better way, I'm all ears.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -20,8 +20,8 @@ package resources
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -703,8 +742,128 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+		if err != nil {
+			select {
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "error confirming subscription"),
+			}:
+			case <-obj.closeChan:
+			}
+			return
+		}
+		log.Printf("%s: Subscription Confirmed", obj)
```
jonathangold: At this point, it's safe to say (once the cloudwatch code is added,) that we are officially watching the resource, so we can tell the engine we're running. We can do it here, or, to minimize side-effects, send a message to snsWatch and do obj.Running() there.

PR #243, opened 2017-11-10
Aws: start and stop the sns endpoint

jonathangold:
This patch starts and stops the sns endpoint, and sets up the event channel handling for snsWatch.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -52,6 +52,8 @@ const (
 	SnsPrefix = Ec2Prefix + "sns-"
 	// SnsTopicName is the name of the sns topic created by snsMakeTopic.
 	SnsTopicName = SnsPrefix + "events"
+	// SnsServerShutdownTimeout is the maximum number of seconds to wait for the http server to shutdown gracefully.
+	SnsServerShutdownTimeout = 30
 	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
 	waitTimeout = 400
 )
@@ -212,6 +214,9 @@ func (obj *AwsEc2Res) Init() error {
 
 // Watch is the primary listener for this resource and it outputs events.
 func (obj *AwsEc2Res) Watch() error {
+	if obj.WatchListenAddr != "" {
+		return obj.snsWatch()
+	}
 	return obj.longpollWatch()
 }
 
@@ -462,6 +467,73 @@ func (obj *AwsEc2Res) longpollWatch() error {
 			}
 		}
 	}()
+	for {
+		select {
+		case event := <-obj.Events():
+			if exit, send = obj.ReadEvent(event); exit != nil {
+				return *exit
+			}
+		case msg, ok := <-obj.awsChan:
+			if !ok {
+				return *exit
+			}
+			if err := msg.err; err != nil {
+				return err
+			}
+			log.Printf("%s: State: %s", obj, msg.str)
+			obj.StateOK(false)
+			send = true
+		}
+		if send {
+			send = false
+			obj.Event()
+		}
+	}
+}
+
+// snsWatch uses amazon cloudwatch events and simple notification service to
+// detect ec2 instance state changes.
+func (obj *AwsEc2Res) snsWatch() error {
+	send := false
+	var exit *error
+	defer obj.wg.Wait()
+	defer close(obj.closeChan)
+	// set up the sns endpoint
+	snsServer := obj.snsServer()
+	// shutdown the sns endpoint when we're done
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.TODO(), SnsServerShutdownTimeout*time.Second)
+		defer cancel()
+		if err := snsServer.Shutdown(ctx); err != nil {
+			if err != context.Canceled {
+				log.Printf("%s: error stopping sns endpoint: %s", obj, err)
+				return
+			}
+			log.Printf("%s: sns server shutdown cancelled", obj)
+		}
+	}()
+	obj.wg.Add(1)
+	// start the endpoint
+	go func() {
+		defer obj.wg.Done()
+		defer close(obj.awsChan)
+		if err := snsServer.ListenAndServe(); err != nil {
+			// when we shut down
+			if err == http.ErrServerClosed {
+				log.Printf("%s: Stopped SNS Endpoint", obj)
+				return
+			}
+			// any other error
+			select {
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "sns server error"),
+			}:
+			case <-obj.closeChan:
+			}
+		}
+	}()
+	log.Printf("%s: Started SNS Endpoint", obj)
+	// process events
 	for {
 		select {
 		case event := <-obj.Events():


resources/aws_ec2.go
```diff
@@ -454,6 +497,61 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	// subscribe endpoint to the topic
+	if err = obj.snsSubscribe(snsClient); err != nil {
+		return errwrap.Wrapf(err, "error subscribing to sns topic")
+	}
+	// process events
+	for {
+		select {
+		case event := <-obj.Events():
+			if exit, send = obj.ReadEvent(event); exit != nil {
+				close(obj.closeChan)
```
purpleidea: I noticed this, and then wondered if we had made this possible mistake before, and I think this is yes, so I opened https://github.com/purpleidea/mgmt/issues/245

resources/aws_ec2.go
```diff
@@ -486,6 +489,67 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	// set up the sns endpoint
+	snsEndpoint := obj.snsServer()
+	// shutdown the sns endpoint when we're done
+	obj.wg.Add(1)
+	go func() {
+		defer obj.wg.Done()
+		select {
+		case <-obj.closeChan:
+			if err := snsEndpoint.Shutdown(nil); err != nil {
+				log.Printf("error stopping sns endpoint: %s", err)
```
jonathangold: Is there a better way to handle this error? I can't pass it to awsChan, since the channel's not being read from anymore, and I can't return it because it's in a goroutine...

jonathangold:
@purpleidea: this is ready for review. If there's a better way to handle the error on line 508, I'm happy to change it. I can't return it, as it's in a goroutine, and I can't send it to awsChan, since it's not being read anymore. I had originally shut the server down in Close() but I think this is better, just not sure about the error handling.

jonathangold:
FWIW, I don't think that the engine needs to respond to the error and do anything, since it will only happen when we're exiting. What do you think?

resources/aws_ec2.go
```diff
@@ -486,6 +489,68 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
 
+// snsWatch uses amazon cloudwatch events and simple notification service to
+// detect ec2 instance state changes.
+func (obj *AwsEc2Res) snsWatch() error {
+	send := false
+	var exit *error
+	defer obj.wg.Wait()
+	defer close(obj.closeChan)
+	// set up the sns endpoint
+	snsEndpoint := obj.snsServer()
```
purpleidea: why are we calling this `snsEndpoint` instead of `snsServer` ? (which is also the name of the function)

resources/aws_ec2.go
```diff
@@ -486,6 +489,68 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	// shutdown the sns endpoint when we're done
```
purpleidea: If the object of this is to run as this Watch method is shutting down, then I believe this whole method can be replace with a defer instead of this equally behaving, yet more heavyweight approach. Do you agree?

resources/aws_ec2.go
```diff
@@ -486,6 +489,68 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+			}
+			log.Printf("%s: Stopped SNS Endpoint", obj)
+		}
+	}()
+	obj.wg.Add(1)
+	// start the endpoint
+	go func() {
+		defer obj.wg.Done()
+		defer close(obj.awsChan)
+		if err := snsEndpoint.ListenAndServe(); err != nil {
```
purpleidea: If you trace this `ListenAndServe` method, eventually if it doesn't error initially, you eventually get here:

https://golang.org/src/net/http/server.go?s=87921:87976#L2676

You don't always need to trace the code, (starwars: UTSL) because it was mentioned in the docs: ` ListenAndServe always returns a non-nil error.`, but tracing told us why if you follow the code to `Serve`. Fun!

So this *will* error... We probably don't want to always complain if the "error" is a "we shutdown now" error. So possibly some error parsing/detection is needed here.

resources/aws_ec2.go
```diff
@@ -486,6 +489,68 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	defer close(obj.closeChan)
+	// set up the sns endpoint
+	snsEndpoint := obj.snsServer()
+	// shutdown the sns endpoint when we're done
+	obj.wg.Add(1)
+	go func() {
+		defer obj.wg.Done()
+		select {
+		case <-obj.closeChan:
+			if err := snsEndpoint.Shutdown(nil); err != nil {
```
purpleidea: Seeing this misc `nil` here was a red flag. I hope you're reading the API docs instead of doing copy+pasta from stackoverflow ;) -- https://golang.org/pkg/net/http/#Server.Shutdown It says: `Make sure the program doesn't exit and waits instead for Shutdown to return.`

So looks like we probably want to add a context with a timeout here instead of `nil` and do the safe thing and wait.

resources/aws_ec2.go
jonathangold: good point. iirc, at first, snsEndpoint was in the struct, so it couldn't share a name with the method. I'll change it.

resources/aws_ec2.go
jonathangold: This was my first approach. Hence the question about not using closeChan. If I don't need it, I think I should remove closeChan from the struct in a new commit before this one.

resources/aws_ec2.go
jonathangold: Cool. Will do. BTW, I got this from the somewhat cryptic HTTP docs, not stack :). Was aware of the context, but thought it was unnecessary (I was thinking of WithCancel, not timeout.) I'll add it.

resources/aws_ec2.go
jonathangold: Weird, I guess I didn't notice, since snsWatch has stopped reading from the channel when it shuts down. I suppose if it returns early for some reason, we'll have to deal with it. I'll spend some time in the source and fix.

resources/aws_ec2.go
purpleidea: No, you still need closeChan to shutdown the other goroutine :) Just fix the things in the comments, the rest is perfect!

resources/aws_ec2.go
purpleidea: > snsWatch has stopped reading from the channel when it shuts down I suppose if it returns early for some reason, we'll have to deal with it.

Yup. Or I think more likely is that it can't start up due to the port being in use maybe?

resources/aws_ec2.go
```diff
@@ -486,6 +491,69 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+		}
+	}()
+	obj.wg.Add(1)
+	// start the endpoint
+	go func() {
+		defer obj.wg.Done()
+		defer close(obj.awsChan)
+		if err := snsServer.ListenAndServe(); err != nil {
+			// when we shut down
+			if err == http.ErrServerClosed || err == context.Canceled {
```
jonathangold: @purpleidea: I finished the changes, but wanted to check two things with you about this error.
1. Is the error checking correct? as I understand it http returns the context error (context.Canceled) if it's there.
2. I think this is the correct place to print the "Stopped SNS Endpoint" message, even though it's more logical in the defer func at first glance, but would print even if there was an http error. What do you think?

jonathangold: I think I know the answer to 2 and 3, but having some insecurity :P
1 ?
2 yes
3 this is good

resources/aws_ec2.go
```diff
@@ -486,6 +491,72 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	// shutdown the sns endpoint when we're done
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.TODO(), SnsServerTimeout*time.Second)
+		defer cancel()
+		if err := snsServer.Shutdown(ctx); err != nil {
+			if err != context.Canceled {
+				log.Printf("%s: error stopping sns endpoint: %s", obj, err)
+				return
+			}
+			log.Printf("%s: sns server shutdown timed out", obj)
```
purpleidea: should be server shutdown `cancelled` instead of timed out, since context might get cancelled by function.

resources/aws_ec2.go
```diff
@@ -486,6 +491,72 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+		}
+	}()
+	obj.wg.Add(1)
+	// start the endpoint
+	go func() {
+		defer obj.wg.Done()
+		defer close(obj.awsChan)
+		if err := snsServer.ListenAndServe(); err != nil {
+			if err != http.ErrServerClosed {
```
purpleidea: why did you nest this, when it was elegant before :)

resources/aws_ec2.go
```diff
@@ -54,6 +54,8 @@ const (
 	SnsTopicName = SnsPrefix + "events"
 	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
 	waitTimeout = 400
+	// SnsServerTimeout is the duration in seconds of the timout context for the sns server shutdown.
+	SnsServerTimeout = 30
```
purpleidea: SnsServerShutdownTimeout i think would be better

resources/aws_ec2.go
```diff
@@ -52,6 +52,8 @@ const (
 	SnsPrefix = Ec2Prefix + "sns-"
 	// SnsTopicName is the name of the sns topic created by snsMakeTopic.
 	SnsTopicName = SnsPrefix + "events"
+	// SnsServerShutdownTimeout is the duration in seconds of the timout context for the sns server shutdown.
```
purpleidea: timout -> timeout

also, keep in mind you can rephrase user facing comments to real english as opposed to literal technical descriptions. eg: "SnsServerShutdownTimeout is the maximum number of seconds to wait for the http server to shutdown gracefully."

resources/aws_ec2.go
```diff
@@ -486,6 +491,74 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+			log.Printf("%s: sns server shutdown cancelled", obj)
+		}
+	}()
+	obj.wg.Add(1)
+	// start the endpoint
+	go func() {
+		defer obj.wg.Done()
+		defer close(obj.awsChan)
+		if err := snsServer.ListenAndServe(); err != nil {
+			// when we shut down
```
purpleidea: you forgot something important here (within the next 5 lines), what is it? :)

jonathangold: return!

PR #244, opened 2017-11-10
Aws: Create and confirm sns subscription

jonathangold:
For review after #243 is merged. This patch is concerned with adding and confirming the sns subscription. Cleanup is handled by snsDeleteTopic().

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net"
 	"net/http"
 	"strconv"
 	"sync"
@@ -510,9 +511,17 @@ func (obj *AwsEc2Res) snsWatch() error {
 	var exit *error
 	defer obj.wg.Wait()
 	defer close(obj.closeChan)
-	// set up the sns endpoint
-	snsServer := obj.snsServer()
-	// shutdown the sns endpoint when we're done
+	// create the sns listener
+	// closing is handled by http.Server.Shutdown in the defer func below
+	listener, err := obj.snsListener(obj.WatchListenAddr)
+	if err != nil {
+		return errwrap.Wrapf(err, "error creating listener")
+	}
+	// set up the sns server
+	snsServer := &http.Server{
+		Handler: http.HandlerFunc(obj.snsPostHandler),
+	}
+	// close the listener and shutdown the sns server when we're done
 	defer func() {
 		ctx, cancel := context.WithTimeout(context.TODO(), SnsServerShutdownTimeout*time.Second)
 		defer cancel()
@@ -525,11 +534,11 @@ func (obj *AwsEc2Res) snsWatch() error {
 		}
 	}()
 	obj.wg.Add(1)
-	// start the endpoint
+	// start the sns server
 	go func() {
 		defer obj.wg.Done()
 		defer close(obj.awsChan)
-		if err := snsServer.ListenAndServe(); err != nil {
+		if err := snsServer.Serve(listener); err != nil {
 			// when we shut down
 			if err == http.ErrServerClosed {
 				log.Printf("%s: Stopped SNS Endpoint", obj)
@@ -802,18 +811,16 @@ func (obj *AwsEc2Res) prependName() string {
 	return AwsPrefix + obj.GetName()
 }
 
-// snsServer returns an http server used to listen for sns messages.
-func (obj *AwsEc2Res) snsServer() *http.Server {
-	addr := obj.WatchListenAddr
-	// if addr is a port
+// snsListener returns a listener bound to watchlistenaddr.
+func (obj *AwsEc2Res) snsListener(addr string) (net.Listener, error) {
 	if _, err := strconv.Atoi(obj.WatchListenAddr); err == nil {
 		addr = fmt.Sprintf(":%s", obj.WatchListenAddr)
 	}
-	handler := http.HandlerFunc(obj.snsPostHandler)
-	return &http.Server{
-		Addr:    addr,
-		Handler: handler,
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, err
 	}
+	return listener, nil
 }
 
 // snsPostHandler listens for posts on the SNS Endpoint.
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -20,8 +20,8 @@ package resources
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
@@ -53,6 +53,9 @@ const (
 	SnsPrefix = Ec2Prefix + "sns-"
 	// SnsTopicName is the name of the sns topic created by snsMakeTopic.
 	SnsTopicName = SnsPrefix + "events"
+	// SnsSubscriptionProto is used to tell sns that the subscriber uses the http protocol.
+	// TODO: add https support
+	SnsSubscriptionProto = "http"
 	// SnsServerShutdownTimeout is the maximum number of seconds to wait for the http server to shutdown gracefully.
 	SnsServerShutdownTimeout = 30
 	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
@@ -65,7 +68,7 @@ const (
 type awsEc2Event uint8
 
 const (
-	awsEc2EventServerReady awsEc2Event = iota
+	awsEc2EventWatchReady awsEc2Event = iota
 	awsEc2EventInstanceStopped
 	awsEc2EventInstanceRunning
 	awsEc2EventInstanceExists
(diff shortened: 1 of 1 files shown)


jonathangold:
@purpleidea This is ready for review after #243 is merged. I've followed the logic through several times, and I'm convinced it's correct (famous last words.) I know there's a question of if/how we want to recover from errors, but in normal operation I haven't encountered any. I'm not convinced there are any cases where we wouldn't want to notify the engine that something's wrong and shut down.

resources/aws_ec2.go
```diff
@@ -533,6 +544,11 @@ func (obj *AwsEc2Res) snsWatch() error {
 		}
 	}()
 	log.Printf("%s: Started SNS Endpoint", obj)
+	// subscribing the endpoint to the topic needs to happen after starting the
+	// http server so that the server can process the subscription confirmation.
+	if err = obj.snsSubscribe(obj.WatchEndpoint, obj.snsTopicArn); err != nil {
```
purpleidea: You can `err :=` instead of declaring `err` above.

resources/aws_ec2.go
```diff
@@ -846,13 +865,30 @@ func (obj *AwsEc2Res) snsDeleteTopic(topicArn string) error {
 	return nil
 }
 
+// snsSubscribe subscribes the endpoint to the sns topic.
+// Returning SubscriptionArn here is useless as it is still pending confirmation.
```
purpleidea: When is it no longer "pending confirmation" ? Do we need to wait for that to happen to be able to do something else?

resources/aws_ec2.go
```diff
@@ -52,6 +52,8 @@ const (
 	SnsPrefix = Ec2Prefix + "sns-"
 	// SnsTopicName is the name of the sns topic created by snsMakeTopic.
 	SnsTopicName = SnsPrefix + "events"
+	// httpProto is used to tell sns that the subscriber uses the http protocol.
+	SnsSubscriptionProto = "http"
```
purpleidea: is there an https version?
also comment should start with `SnsSubscriptionProto` instead.

resources/aws_ec2.go
```diff
@@ -829,9 +836,29 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
 		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
 		return
 	}
-	post, _ := ioutil.ReadAll(req.Body)
-	if obj.debug {
-		log.Printf("%s: Post: %s", obj, string(post))
+	// decode json
+	decoder := json.NewDecoder(req.Body)
+	var post postData
+	err := decoder.Decode(&post)
```
purpleidea: This can be in the `if` scope

resources/aws_ec2.go
```diff
@@ -829,9 +836,29 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		select {
+		case obj.awsChan <- &chanStruct{
+			err: errwrap.Wrapf(err, "error decoding post"),
+		}:
+		case <-obj.closeChan:
+		}
+		return
+	}
+	if post.Type == "SubscriptionConfirmation" {
```
purpleidea: This feels like an event saying "we're now ready" to watch events or something...
If my understanding is correct, that means this should be the obj.Running() runner, no?
Probably best to understand this a bit better, and maybe add a comment.

resources/aws_ec2.go
```diff
@@ -829,9 +836,29 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
-	post, _ := ioutil.ReadAll(req.Body)
-	if obj.debug {
-		log.Printf("%s: Post: %s", obj, string(post))
+	// decode json
+	decoder := json.NewDecoder(req.Body)
+	var post postData
+	err := decoder.Decode(&post)
+	if err != nil {
+		select {
+		case obj.awsChan <- &chanStruct{
```
purpleidea: If we didn't decode the message, we shut it all down? I'm okay with that, assuming aws doesn't sent erroneous messages often, but we should probably log some message to let us know their API changed...

resources/aws_ec2.go
jonathangold: Yeah. I have another commit going which adds a run field to chanStruct and uses it to signal snsWatch, which then runs obj.Running(). It might be simpler to just run it here so I don't need to modify chanStruct. What do you prefer?

resources/aws_ec2.go
jonathangold: There is, but it requres a valid ssl certificate, which I don't have. In practice we should be able to detect which protocol the user wants to use by the WatchEndpoint URL they specify. I will edit the first commit to change the comment.

resources/aws_ec2.go
jonathangold: Lol, seems like never, but if you rerun confirm subscription after a little bit, it returns the arn. I don't really need it though, as deleting the topic deletes the subscriptions as well. I just thought I should explain why the method doesn't return anything.

resources/aws_ec2.go
```diff
@@ -547,9 +570,16 @@ func (obj *AwsEc2Res) snsWatch() error {
 			if err := msg.err; err != nil {
 				return err
 			}
-			log.Printf("%s: State: %s", obj, msg.str)
-			obj.StateOK(false)
-			send = true
+			if msg.run {
```
purpleidea: This is a fine approach, but here's an alternate idea:

We currently have random string messages (which are all internal to us) -- Instead, please make a bunch of constant iota's and use those instead. This can be one as well. In fact, the first commit could be to switch all the existing code to use iota of uint8 or something.

resources/aws_ec2.go
```diff
@@ -829,9 +859,35 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
-	if obj.debug {
-		log.Printf("%s: Post: %s", obj, string(post))
+	// decode json
+	decoder := json.NewDecoder(req.Body)
+	var post postData
+	if err := decoder.Decode(&post); err != nil {
+		select {
+		case obj.awsChan <- &chanStruct{
+			err: errwrap.Wrapf(err, "error decoding incoming POST, check struct formatting"),
+		}:
```
purpleidea: Do we also want to reply with some HTTP error code to let AWS know they suck? (or at least that we didn't understand?) Pick something in 4xx that's not teapot. 400 is ok.

resources/aws_ec2.go
```diff
@@ -829,9 +859,35 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+		if err := obj.snsConfirmSubscription(obj.snsTopicArn, post.Token); err != nil {
+			select {
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "error confirming subscription"),
+			}:
+			case <-obj.closeChan:
+			}
+			return
+		}
+		// now that the subscription is confirmed, we can tell the engine we're running
```
purpleidea: Does running snsConfirmSubscription work immediately, or do we have to wait for some callBack from aws to know that happened? What a weird API.

jonathangold: There's no callback, but even if there's a short delay, we won't miss any messages.
http://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.html#SendMessageToHttp.retry
"By default, if the initial delivery of the message fails, Amazon SNS attempts up to three retries with a delay between failed attempts set at 20 seconds."

purpleidea: Okay, don't tell me, add something in the comments :)

resources/aws_ec2.go
```diff
@@ -554,6 +571,15 @@ func (obj *AwsEc2Res) snsWatch() error {
 		}
 	}()
 	log.Printf("%s: Started SNS Endpoint", obj)
+	// Subscribing the endpoint to the topic needs to happen after starting the
+	// http server, so that the server can process the subscription confirmation.
+	// We won't drop incoming connections from aws by this point, because we've
+	// already opened the server listener. In the worst case scenario the incoming
+	// aws connections will be accepted but will block until our http server finishes
+	// getting ready in it's goroutine.
```
purpleidea: it's -> its please, and wrap this whole paragraph to 80 chars please.

resources/aws_ec2.go
```diff
@@ -53,6 +53,8 @@ const (
 	SnsPrefix = Ec2Prefix + "sns-"
 	// SnsTopicName is the name of the sns topic created by snsMakeTopic.
 	SnsTopicName = SnsPrefix + "events"
+	// SnsSubscriptionProto is used to tell sns that the subscriber uses the http protocol.
```
purpleidea: Please add on the next line under the comment:

```// TODO: add https support```

resources/aws_ec2.go
purpleidea: What about this comment ^^^

purpleidea:
LGTM :) :)

PR #247, opened 2017-11-13
resources: aws: ec2: Change event messages to iota consts

jonathangold:
replaced the strings in awsChan with msgContents iota as discussed.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -58,6 +58,18 @@ const (
 	waitTimeout = 400
 )
 
+//go:generate stringer -type=awsEc2Event -output=awsec2event_stringer.go
+
+// awsEc2Event represents the contents of event messages sent via awsChan.
+type awsEc2Event uint8
+
+const (
+	awsEc2EventServerReady awsEc2Event = iota
+	awsEc2EventInstanceStopped
+	awsEc2EventInstanceRunning
+	awsEc2EventInstanceExists
+)
+
 // AwsRegions is a list of all AWS regions generated using ec2.DescribeRegions.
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -58,6 +58,18 @@ const (
 	waitTimeout = 400
 )
 
+//go:generate stringer -type=msgContents -output=msgcontents_stringer.go
+
+// msgContents represents the contents of messages sent via awsChan.
+type msgContents uint8
+
+const (
+	msgContentsRunMgmt         msgContents = iota // 0
```
purpleidea: remove the // 0 comments please

also please rename from `RunMgmt` to something else that doesn't have `mgmt` string in it. Maybe `ServerRunning` ??

Could we also name the type `awsEc2Event` or something similar please?

resources/aws_ec2.go
```diff
@@ -110,8 +122,8 @@ type AwsEc2Res struct {
 
 // chanStruct defines the type for a channel used to pass events and errors to watch.
 type chanStruct struct {
-	str string
-	err error
+	contents msgContents
```
purpleidea: how about `event` instead of `contents`?

resources/aws_ec2.go
```diff
@@ -547,7 +559,13 @@ func (obj *AwsEc2Res) snsWatch() error {
 			if err := msg.err; err != nil {
 				return err
 			}
-			log.Printf("%s: State: %s", obj, msg.str)
+			if msg.contents == msgContentsRunMgmt {
+				if err := obj.Running(); err != nil {
```
purpleidea: I think this part only happens in a later patch...

PR #248, opened 2017-11-14
 resources: aws: ec2: Use custom listener for snsServer 

jonathangold:
This patch replaces the call to Server.ListenAndServe() with Server.Serve(listener) in order to make sure the listener is up and running before we subscribe to the topic in a future patch.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net"
 	"net/http"
 	"strconv"
 	"sync"
@@ -510,9 +511,17 @@ func (obj *AwsEc2Res) snsWatch() error {
 	var exit *error
 	defer obj.wg.Wait()
 	defer close(obj.closeChan)
-	// set up the sns endpoint
-	snsServer := obj.snsServer()
-	// shutdown the sns endpoint when we're done
+	// create the sns listener
+	// closing is handled by http.Server.Shutdown in the defer func below
+	listener, err := obj.snsListener(obj.WatchListenAddr)
+	if err != nil {
+		return errwrap.Wrapf(err, "error creating listener")
+	}
+	// set up the sns server
+	snsServer := &http.Server{
+		Handler: http.HandlerFunc(obj.snsPostHandler),
+	}
+	// close the listener and shutdown the sns server when we're done
 	defer func() {
 		ctx, cancel := context.WithTimeout(context.TODO(), SnsServerShutdownTimeout*time.Second)
 		defer cancel()
@@ -525,11 +534,11 @@ func (obj *AwsEc2Res) snsWatch() error {
 		}
 	}()
 	obj.wg.Add(1)
-	// start the endpoint
+	// start the sns server
 	go func() {
 		defer obj.wg.Done()
 		defer close(obj.awsChan)
-		if err := snsServer.ListenAndServe(); err != nil {
+		if err := snsServer.Serve(listener); err != nil {
 			// when we shut down
 			if err == http.ErrServerClosed {
 				log.Printf("%s: Stopped SNS Endpoint", obj)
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -525,11 +526,16 @@ func (obj *AwsEc2Res) snsWatch() error {
 		}
 	}()
 	obj.wg.Add(1)
-	// start the endpoint
+	// create the sns listener
+	listener, err := obj.snsListener(obj.WatchListenAddr)
```
purpleidea: Out of curiosity: if you call shutdown and there's no listener, it doesn't panic, does it? (If it's easy to test, please do.)

purpleidea: Also, I think this needs to happen *before* the server creation stuff, and also it needs to defer some sort of close https://golang.org/pkg/net/#Listener ...

That way we (1) open listener, (defer listener close (4)) (2) open server (defer server close (3))

Makes sense?

resources/aws_ec2.go
```diff
@@ -802,16 +808,21 @@ func (obj *AwsEc2Res) prependName() string {
...
 	}
-	handler := http.HandlerFunc(obj.snsPostHandler)
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return listener, nil
+}
+
+// snsServer returns an http server used to listen for sns messages.
```
purpleidea: I agree with you now that this might be overkill for a phone function. You can kill it if you'd like.

resources/aws_ec2.go
jonathangold: If we close the listener manually, doesn't that defeat the purpose of the graceful shutdown? Server.Shutdown handles this for us, afaict. https://golang.org/src/net/http/server.go?s=47634:47680#L2492

Also Server.Shutdown(ctx) panics if the server isn't serving (ie. before running Server.Serve(listener)) so we'll have to make sure it is running before calling Shutdown(ctx). Not sure how to check that, though.

jonathangold: Also, I think as long as it happens before snsServer.Serve(listener) we're ok, as the server creation just uses an addr, and the listener doesn't come into play until we start the server.

purpleidea: Cool-- okay, just leave a comment in the source saying that the listener will get close by shutdown... (if we *never* serve) will the shutdown still close without panicing?

jonathangold: Unfortunately it does panic. :( I looked at the server type and it wasn't obvious how to check if it's serving. I suppose we could make a HTTP request and shutdown if it doesn't error, but that seems kind of sloppy.

resources/aws_ec2.go
```diff
@@ -510,29 +511,39 @@ func (obj *AwsEc2Res) snsWatch() error {
...
+	}
+	// shutdown the sns server when we're done
 	defer func() {
-		ctx, cancel := context.WithTimeout(context.TODO(), SnsServerShutdownTimeout*time.Second)
-		defer cancel()
-		if err := snsServer.Shutdown(ctx); err != nil {
-			if err != context.Canceled {
-				log.Printf("%s: error stopping sns endpoint: %s", obj, err)
-				return
+		if snsServer.ConnState != nil {
```
purpleidea: please add a comment explaining this connstate logic...

resources/aws_ec2.go
```diff
@@ -510,29 +511,39 @@ func (obj *AwsEc2Res) snsWatch() error {
...
+			if err := snsServer.Shutdown(ctx); err != nil {
+				if err != context.Canceled {
+					log.Printf("%s: error stopping sns endpoint: %s", obj, err)
+					return
+				}
+				log.Printf("%s: sns server shutdown cancelled", obj)
 			}
-			log.Printf("%s: sns server shutdown cancelled", obj)
 		}
+		listener.Close()
```
purpleidea: do we need to check the error?

purpleidea:
Cool, okay this has now been merged in feat/aws https://github.com/purpleidea/mgmt/tree/feat/aws
Future patches can be submitted against this branch (instead of git master) and I'll be responsible to keep this branch rebased on git master, until I finally merge it over there when it's "complete" [1].

[1] Where complete is some arbitrary finish of the upcoming features.

purpleidea:
(GH is too stupid AFAICT to tell that it's merged in a branch, sorry... So you'll have to use your eyeballs if you want to verify)

PR #249, opened 2017-11-16
Aws: Add cloudwatch rule and target sns

jonathangold:
Let me know if you want me to split this into 2 pulls.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -33,6 +33,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
+	cwe "github.com/aws/aws-sdk-go/service/cloudwatchevents"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/sns"
 	multierr "github.com/hashicorp/go-multierror"
@@ -58,6 +59,18 @@ const (
 	SnsSubscriptionProto = "http"
 	// SnsServerShutdownTimeout is the maximum number of seconds to wait for the http server to shutdown gracefully.
 	SnsServerShutdownTimeout = 30
+	// CwePrefix gets prepended onto the cloudwatch rule name.
+	CwePrefix = Ec2Prefix + "cw-"
+	// CweRuleName is the name of the rule created by makeCloudWatchRule.
+	CweRuleName = CwePrefix + "state"
+	// CweRuleSource describes the resource type to monitor for cloudwatch events.
+	CweRuleSource = "aws.ec2"
+	// CweRuleDetailType describes the specific type of events to trigger cloudwatch.
+	CweRuleDetailType = "EC2 Instance State-change Notification"
+	// CweTargetID is used to tell cloudwatch events to target the sns service.
+	CweTargetID = "sns"
+	// CweTargetJSON is the json field that cloudwatch will send to our endpoint so we don't get more than we need.
+	CweTargetJSON = "$.detail"
 	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
 	waitTimeout = 400
 )
@@ -120,6 +133,8 @@ type AwsEc2Res struct {
 	// so we save it here when we create the topic instead.
 	snsTopicArn string
 
+	cweClient *cwe.CloudWatchEvents // client for AWS CloudWatchEvents API calls
+
 	awsChan   chan *chanStruct // channel used to send events and errors to Watch()
 	closeChan chan struct{}    // channel used to cancel context when it's time to shut down
 	wg        *sync.WaitGroup  // waitgroup for goroutines in Watch()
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -858,16 +990,135 @@ func (obj *AwsEc2Res) snsDeleteTopic(topicArn string) error {
...
+	return nil
+}
+
+// cweDeleteRule removes the target and deletes the cloudwatch rule.
+func (obj *AwsEc2Res) cweDeleteRule(name string) error {
+	// delete the rule
+	drInput := &cwe.DeleteRuleInput{
+		Name: aws.String(name),
+	}
+	log.Printf("%s: Deleting CloudWatch Rule", obj)
```
jonathangold: I need to change this, and the one in cweRemoveTarget to past tense like the other messages. Will leave it as is until you review.

resources/aws_ec2.go
```diff
@@ -238,6 +265,30 @@ func (obj *AwsEc2Res) Init() error {
 		}
 		// save the topicArn for later use
 		obj.snsTopicArn = snsTopicArn
+
+		// make cloudwatch client
+		cweSess, err := session.NewSession(&aws.Config{
+			Region: aws.String(obj.Region),
+		})
+		if err != nil {
+			return errwrap.Wrapf(err, "error creating sns session")
```
purpleidea: is this error message correct? sns?

resources/aws_ec2.go
```diff
@@ -972,6 +1023,84 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
...
+}
+
+// cweMakeRule makes a cloud watch rule.
+func (obj *AwsEc2Res) cweMakeRule(name string, eventPattern string) error {
+	// make cloudwatch rule
+	putRuleInput := &cwe.PutRuleInput{
+		Name:         aws.String(name),
+		EventPattern: aws.String(eventPattern),
+	}
+	_, err := obj.cweClient.PutRule(putRuleInput)
```
purpleidea: this line can be merged with the `if`

resources/aws_ec2.go
```diff
@@ -972,6 +1023,84 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
...
+}
+
+// cweDeleteRule deletes the cloudwatch rule.
+func (obj *AwsEc2Res) cweDeleteRule(name string) error {
+	// delete the rule
+	drInput := &cwe.DeleteRuleInput{
+		Name: aws.String(name),
+	}
+	log.Printf("%s: Deleting CloudWatch Rule", obj)
+	if _, err := obj.cweClient.DeleteRule(drInput); err != nil {
```
purpleidea: No need to change, this is fine (and what I would have expected), however FYI as a fun learning exercise, the following code is actually equivalent! Do you know why?

```
	err := obj.cweClient.DeleteRule(drInput)
	return errwrap.Wrapf(err, "error deleting cloudwatch rule")
```

resources/aws_ec2.go
```diff
@@ -972,6 +1023,84 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
...
+		return errwrap.Wrapf(err, "error deleting cloudwatch rule")
+	}
+	return nil
+}
+
+// cweTargetRule configures cloudwatch to send events to sns topic.
+func (obj *AwsEc2Res) cweTargetRule(topicArn string, targetID string, inputPath string, ruleName string) error {
```
purpleidea: can be simplified as 

```
 +func (obj *AwsEc2Res) cweTargetRule(topicArn, targetID, inputPath, ruleName string) error { 
```

same in at least one other places.

purpleidea:
PS: the error is our commit message test failing which is too dumb to know that this is against a different branch. woops. You're welcome to fix this if you know how. I do not. https://github.com/purpleidea/mgmt/blob/master/test/test-commit-message.sh#L62

PPS: s/resource/resources/ in your message.

jonathangold:
@purpleidea sns is only responsible for delivering the messages to our endpoint. The messages themselves are generated by the cloudwatch event that we set up. IOW, Cloudwatch (the publisher) detects the state change, then publishes a message on sns which gets pushed to the endpoint (the subscriber) for processing. I will explain this wherever we decide is the most appropriate place. I think a godoc comment makes sense, maybe on the snsWatch method.

resources/aws_ec2.go
jonathangold: I haven't seen (maybe I just haven't noticed) this pattern before. This doesn't result in a compiler error since there are two return values and we're only assigning one?

purpleidea: Woops, typo. Fixed above. Check again...

jonathangold: @purpleidea That would return nil if there was no error, which I do right after. It would save 3 LoC, but I like the explicit-ness of having both returns. Good trick to have in the tool-bag though.

purpleidea:
Merged, thanks! (The failure was the commit message bug)

PR #250, opened 2017-11-16
docker: Bump to golang 1.8

xiu:
The containers were still using docker 1.6.

Also added a chown as I couldn't make build inside the development container due to permission issues.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,10 +1,10 @@
-FROM golang:1.6.2
+FROM golang:1.8
 
 MAINTAINER Michał Czeraszkiewicz <contact@czerasz.com>
 
 # Set the reset cache variable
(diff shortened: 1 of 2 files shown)


purpleidea:
@xiu Thanks for the patch! It LGTM... One disclaimer: I personally don't test any of the docker dev environment or running regularly, so YMMV. Would love to have you hacking on more patches or even a Docker resource https://github.com/purpleidea/mgmt/blob/master/docs/resource-guide.md if you're really into docker!

xiu:
@purpleidea I was playing with the containers in the intent of fixing some of the "good first issues" so you might see a bit more of me :)

purpleidea:
@xiu Merged, thank you!

Great-- We only tagged a small number of "good first issues", but I'm happy to make more if there's nothing that interests you. Feel free to join us in #mgmtconfig on Freenode IRC for more discussion.

PR #251, opened 2017-11-17
Aws: complete sns/cloudwatch feature

jonathangold:
Here are the last two pieces of the sns/cloudwatch puzzle. Feel free to review as one patch or to look at each commit individually, as the two are only tangentially related.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -59,6 +59,16 @@ const (
 	SnsSubscriptionProto = "http"
 	// SnsServerShutdownTimeout is the maximum number of seconds to wait for the http server to shutdown gracefully.
 	SnsServerShutdownTimeout = 30
+	// SnsPolicy is the topic attribute that defines the security policy for the topic.
+	SnsPolicy = "Policy"
+	// SnsPolicySid is the friendly name of the policy statement.
+	SnsPolicySid = CwePrefix + "publish"
+	// SnsPolicyEffect allows the action(s) defined in the policy statement.
+	SnsPolicyEffect = "Allow"
+	// SnsPolicyService is the cloudwatch events security principal that we are granting the permission to.
+	SnsPolicyService = "events.amazonaws.com"
+	// SnsPolicyAction is the specific permission we are granting in the policy.
+	SnsPolicyAction = "SNS:Publish"
 	// CwePrefix gets prepended onto the cloudwatch rule name.
 	CwePrefix = Ec2Prefix + "cw-"
 	// CweRuleName is the name of the rule created by makeCloudWatchRule.
@@ -144,6 +154,33 @@ type AwsEc2Res struct {
 type chanStruct struct {
 	event awsEc2Event
 	err   error
+}
+
+// snsPolicy denotes the structure of sns security policies.
+type snsPolicy struct {
+	Version   string         `json:"Version"`
+	ID        string         `json:"Id"`
+	Statement []snsStatement `json:"Statement"`
+}
+
+// snsStatement denotes the structure of sns security policy statements.
+type snsStatement struct {
+	Sid       string       `json:"Sid"`
+	Effect    string       `json:"Effect"`
+	Principal snsPrincipal `json:"Principal"`
+	Action    interface{}  `json:"Action"`
+	Resource  string       `json:"Resource"`
+	Condition *struct {
+		StringEquals *struct {
+			AWSSourceOwner *string `json:"AWS:SourceOwner,omitempty"`
+		} `json:"StringEquals,omitempty"`
+	} `json:"Condition,omitempty"`
+}
+
+// snsPrincipal describes the aws or service account principal.
+type snsPrincipal struct {
+	AWS     string `json:"AWS,omitempty"`
+	Service string `json:"Service,omitempty"`
 }
 
 // cloudWatchRule denotes the structure of cloudwatch rules.
@@ -288,6 +325,11 @@ func (obj *AwsEc2Res) Init() error {
 		// target cloudwatch rule to sns topic
 		if err := obj.cweTargetRule(obj.snsTopicArn, CweTargetID, CweTargetJSON, CweRuleName); err != nil {
 			return errwrap.Wrapf(err, "error targeting cloudwatch rule")
+		}
+		// authorize cloudwatch to publish on sns
+		// This gets cleaned up in Close(), when the topic is deleted.
+		if err := obj.snsAuthorizeCloudWatch(obj.snsTopicArn); err != nil {
+			return errwrap.Wrapf(err, "error authorizing cloudwatch for sns")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


purpleidea:
So the first commit in this list LGTM, so I've merged that to feat/aws ... reviewing the rest!

resources/aws_ec2.go
```diff
@@ -1031,6 +1072,56 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
...
+	}
+	attrs, err := obj.snsClient.GetTopicAttributes(gaInput)
+	if err != nil {
+		return err
+	}
+	// get the existing security policy
+	pol := attrs.Attributes[SnsPolicy]
+	// unmarshal the current sns security policy
+	var policy snsPolicy
+	err = json.Unmarshal([]byte(*pol), &policy)
```
purpleidea: err := and combine with below

resources/aws_ec2.go
```diff
@@ -146,6 +156,33 @@ type chanStruct struct {
...
 
+// snsPolicy denotes the structure of sns security policies.
+type snsPolicy struct {
+	Version   string      `json:"Version"`
+	ID        string      `json:"Id"`
+	Statement []statement `json:"Statement"`
+}
+
+// statement denotes the structure of sns security policy statements.
+type statement struct {
```
purpleidea: does it make sense to call this snsStatement to make it more obvious what it's related to?

resources/aws_ec2.go
```diff
@@ -146,6 +156,33 @@ type chanStruct struct {
...
+	Action    interface{} `json:"Action"`
+	Resource  string      `json:"Resource"`
+	Condition *struct {
+		StringEquals *struct {
+			AWSSourceOwner *string `json:"AWS:SourceOwner,omitempty"`
+		} `json:"StringEquals,omitempty"`
+	} `json:"Condition,omitempty"`
+}
+
+// principal describes the aws or service account principal.
```
purpleidea: does it make sense to call this snsPrincipal to make it more obvious what it's related to?

resources/aws_ec2.go
```diff
@@ -289,6 +326,10 @@ func (obj *AwsEc2Res) Init() error {
 		if err := obj.cweTargetRule(obj.snsTopicArn, CweTargetID, CweTargetJSON, CweRuleName); err != nil {
 			return errwrap.Wrapf(err, "error targeting cloudwatch rule")
 		}
+		// authorize cloudwatch to publish on sns
```
purpleidea: is there a corresponding close/unregister part?

resources/aws_ec2.go
```diff
@@ -1072,6 +1103,44 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
 	return nil
 }
 
+// snsProcessEvents unmarshals instance state-change notifications and, if the
+// event matches the instance we are watching, returns an awsEc2Event.
+func (obj *AwsEc2Res) snsProcessEvent(message, instanceName string) (awsEc2Event, error) {
+	// unmarshal the message
+	var msg postMsg
+	err := json.Unmarshal([]byte(message), &msg)
```
purpleidea: You can combine this error into the `err != nil` line.

resources/aws_ec2.go
```diff
@@ -1072,6 +1103,44 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
...
+		switch msg.State {
+		case "running":
+			return awsEc2EventInstanceRunning, nil
+		case "stopped":
+			return awsEc2EventInstanceStopped, nil
+		case "terminated":
+			return awsEc2EventInstanceTerminated, nil
+		}
+	}
+	return awsEc2EventNone, nil
```
purpleidea: Is this a possible valid scenario or should it be an error?

resources/aws_ec2.go
```diff
@@ -91,9 +91,11 @@ const (
 type awsEc2Event uint8
 
 const (
-	awsEc2EventWatchReady awsEc2Event = iota
+	awsEc2EventNone awsEc2Event = iota
```
purpleidea: What do we need a `None` for?

resources/aws_ec2.go
```diff
@@ -1007,6 +1016,28 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+		event, err := obj.snsProcessEvent(post.Message, obj.prependName())
+		if err != nil {
+			select {
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "error confirming subscription"),
+			}:
+			case <-obj.closeChan:
+			}
+			return
+
```
purpleidea: nill this nl please

resources/aws_ec2.go
```diff
@@ -1007,6 +1016,28 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+		}
+		if event != awsEc2EventNone {
```
purpleidea: ```if event == awsEc2EventNone``` can we just return early?

resources/aws_ec2.go
jonathangold: This is used as a zero value in the case that snsProcessEvent returns an error, or finds that the instance id does not match the resource we are watching. If we receive awsEc2EventNone, we can ignore the event.

resources/aws_ec2.go
jonathangold: This is fine. It means we processed the event, but the instance id didn't match our resource, so snsWatch can ignore the event, and we can continue to watch,

resources/aws_ec2.go
jonathangold: No. This goes away when we delete the topic.

purpleidea: Do you mind adding a comment about that here please?

resources/aws_ec2.go
```diff
@@ -1031,6 +1103,92 @@ func (obj *AwsEc2Res) snsConfirmSubscription(topicArn string, token string) erro
...
+		Sid:    SnsPolicySid,
+		Effect: SnsPolicyEffect,
+		Principal: snsPrincipal{
+			Service: SnsPolicyService,
+		},
+		Action:   SnsPolicyAction,
+		Resource: topicArn,
+	}
+	// add the new policy statement to the existing one
+	policy.Statement = []snsStatement{policy.Statement[0], permission}
```
purpleidea: can we do:

policy.Statement = append(policy.Statement, permission) ?

AND can we check that policy.Statement isn't nil first ? Remember even if the aws API should always have a value there we should not trust that.

jonathangold: @purpleidea That's what I had here before, but in the case that we don't exit cleanly and delete the topic, when we re-run SetTopicAttributes hangs indefinitely when you try to add a policy with duplicate statements. ie. []snsStatement{policy.Statement[0], permission, permission}
You're right that I need to add a check, and maybe the correct approach would be to compare any additional statements to 'permission' and if they match, return early since our job is done. What do you think?

purpleidea:
This branch needs a rebase, and then LMK if it's ready for final review.

purpleidea:
Cool :) All merged! Thanks!

jonathangold:
@purpleidea forgot to add the check to make sure policy.Statement isn't nil. I can add in another patch later. :P

purpleidea:
@jonathangold Write the patch and I'll rebase it in for you.

PR #254, opened 2017-11-19
travis: Allow travis builds to access target branches

jonathangold:
Finally works. Turns out travis sets remote.origin.fetch to a temporary version of the master branch, so we have to override with our own remote.origin.fetch before we fetch.


purpleidea:
Awesome detective work!

purpleidea:
Thanks again, merged!

PR #255, opened 2017-11-21
resources: aws: ec2: bugfixes

jonathangold:
I think this is now race free. 
Here's the control flow:

0. SIGINT is received from CTRL^C
1. Shutdown() is called
2. Server immediately returns http.ErrServerClosed
3. Open listener(s) and idle connection(s) are closed
4. Shutdown() returns any errors from closing the listeners or context
5. We return *exit


jonathangold:
The crucial difference afaict, is that step 5 (return *exit) turns into step 1 (right after ctrl^c) if we defer the function, which is not desirable. I have another idea, I'm gonna play with.

PR #260, opened 2017-11-23
build: Build mgmt on any go file change

roidelapluie:
Prior to this commit, running make would only rebuild mgmt when
main.go was changed. It means that make clean build was needed.

With this commit, any go file change in this directory will
trigger a new compilation.


purpleidea:
finally! \o/ merged :)

PR #256, opened 2017-11-21
resources: Add User/Group to ExecRes

xiu:
Fixes #221 

#214 - removed FileRes Go 1.6 specific code

This is still WIP, I'm still playing with the autoedges

PR #269, opened 2017-11-24
test: Switch to goimports instead of gofmt

xiu:
see https://github.com/purpleidea/mgmt/pull/256#issuecomment-346360414

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,9 @@ test:
 	./test.sh
 
 gofmt:
-	find . -maxdepth 3 -type f -name '*.go' -not -path './old/*' -not -path './tmp/*' -exec gofmt -w {} \;
+	# TODO: remove gofmt once goimports has a -s option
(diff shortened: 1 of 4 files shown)


purpleidea:
Patch LGTM, but can you add this change in Makefile as well (gofmt target binary, but leave the name the same) and also add the -s if you want?

xiu:
Updated, goimports doesn't have a -s option though. I can add a run of gofmt -w -s before goimports if you want

purpleidea:
@xiu Actually, I was hoping you'd remove that old comment of mine and add the -s :)
But yeah, we'd need to do it in two runs. I'm okay with that if you'd like to write the patch :) :)

xiu:
Ok, done :) I've just doubled the find command to keep it clear

purpleidea:
Patch LGTM-- the two files that are updated when you run this are simplifications we didn't previously check. You can commit those with this patch as well :)

Thanks!

purpleidea:
(rebased to master and merged)

Thank you!

purpleidea:
ac629404f499a34d71db71b995d79893d5aff565

PR #270, opened 2017-11-25
resources: AutoEdges from Exec to User/Group/File

xiu:
WIP

### DIFF
--- a/resources/exec.go
+++ b/resources/exec.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os/exec"
 	"os/user"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


resources/exec.go
```diff
@@ -392,6 +393,46 @@ func (obj *ExecRes) AutoEdges() (AutoEdge, error) {
 			},
 			path: x, // what matters
 		})
+		data = append(data, &FileUID{
+			BaseUID: BaseUID{
+				Name:     obj.GetName(),
+				Kind:     obj.GetKind(),
+				Reversed: &reversed,
+			},
+			path: x, // TODO: check that we handle relative paths in FileRes
```
purpleidea: This is a good question :P 
Would you like to open an issue ticket about this? We should probably check for this in Validate() since AFAIK relative paths don't make sense in file resources. Am I right ? Wrong? Let's discuss more in the issue.

xiu: Opened #271 in that regard :)

purpleidea:
@xiu Are you going to write part two?

purpleidea:
cc @frebib this might interest you.

purpleidea:
This has been replaced with https://github.com/purpleidea/mgmt/pull/660
Thanks again =D

PR #272, opened 2017-11-27
Aws/snsverify

jonathangold:
Here are the methods to verify sns message signatures.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -19,12 +19,16 @@ package resources
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
+	"regexp"
 	"strconv"
 	"sync"
 	"time"
@@ -69,6 +73,11 @@ const (
 	SnsPolicyService = "events.amazonaws.com"
 	// SnsPolicyAction is the specific permission we are granting in the policy.
 	SnsPolicyAction = "SNS:Publish"
+	// SnsCertURLRegex is used to make sure we only download certificates
+	// from amazon. This regex will match "https://sns.***.amazonaws.com/"
+	// where *** represents any combination of words and hyphens, and will
+	// match any aws region name, eg: ca-central-1.
+	SnsCertURLRegex = `(^https:\/\/sns\.([\w\-])+\.amazonaws.com\/)`
 	// CwePrefix gets prepended onto the cloudwatch rule name.
 	CwePrefix = Ec2Prefix + "cw-"
 	// CweRuleName is the name of the rule created by makeCloudWatchRule.
@@ -133,6 +142,12 @@ type AwsEc2Res struct {
 
 	WatchEndpoint   string `yaml:"watchendpoint"`   // the public url of the sns endpoint, eg: http://server:12345/
 	WatchListenAddr string `yaml:"watchlistenaddr"` // the local address or port that the sns listens on, eg: 10.0.0.0:23456 or 23456
+	// ErrorOnMalformedPost controls whether or not malformed HTTP post
+	// requests, that cause JSON decoder errors, will also make the engine
+	// shut down. If ErrorOnMalformedPost set to true and an error occurs,
+	// Watch() will return the error and the engine will shut down.
+	ErrorOnMalformedPost bool `yaml:"erroronmalformedpost"`
+
 	// UserData is used to run bash and cloud-init commands on first launch.
 	// See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html
 	// for documantation and examples.
@@ -199,9 +214,16 @@ type ruleDetail struct {
 
 // postData is the format of the messages received and decoded by snsPostHandler().
 type postData struct {
-	Type    string `json:"Type"`
-	Token   string `json:"Token"`
-	Message string `json:"Message"`
+	Type             string `json:"Type"`
+	MessageID        string `json:"MessageId"`
+	Token            string `json:"Token"`
+	TopicArn         string `json:"TopicArn"`
+	Message          string `json:"Message"`
+	SubscribeURL     string `json:"SubscribeURL"`
+	Timestamp        string `json:"Timestamp"`
+	SignatureVersion string `json:"SignatureVersion"`
+	Signature        string `json:"Signature"`
+	SigningCertURL   string `json:"SigningCertURL"`
 }
 
 // postMsg is used to unmarshal the postData message if it's an event notification.
@@ -920,13 +942,16 @@ func (obj *AwsEc2Res) Compare(r Res) bool {
 	if obj.ImageID != res.ImageID {
 		return false
 	}
-	if obj.UserData != res.UserData {
-		return false
-	}
 	if obj.WatchEndpoint != res.WatchEndpoint {
 		return false
 	}
 	if obj.WatchListenAddr != res.WatchListenAddr {
+		return false
+	}
+	if obj.ErrorOnMalformedPost != res.ErrorOnMalformedPost {
+		return false
+	}
+	if obj.UserData != res.UserData {
 		return false
 	}
 	return true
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -971,25 +1009,28 @@ func (obj *AwsEc2Res) snsListener(listenAddr string) (net.Listener, error) {
...
-		}:
-		case <-obj.closeChan:
-		}
+		log.Printf("%s: error decoding post: %s", obj, err)
 		return
 	}
+	// Verify the x509 signature. If there is an error verifying the
+	// signature, we print the error, ignore the event and return.
+	if err := obj.snsVerifySignature(post); err != nil {
+		http.Error(w, "Bad request", http.StatusBadRequest)
```
purpleidea: put log message first please

resources/aws_ec2.go
```diff
@@ -1020,7 +1061,7 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
 		if err != nil {
 			select {
 			case obj.awsChan <- &chanStruct{
-				err: errwrap.Wrapf(err, "error confirming subscription"),
+				err: errwrap.Wrapf(err, "errpr processing event"),
```
purpleidea: typo

resources/aws_ec2.go
```diff
@@ -69,6 +75,8 @@ const (
 	SnsPolicyService = "events.amazonaws.com"
 	// SnsPolicyAction is the specific permission we are granting in the policy.
 	SnsPolicyAction = "SNS:Publish"
+	// SnsCertURLRegex is used to make sure we only download certificates from amazon.
```
purpleidea: add a description describing in human the pattern we're looking for please.

resources/aws_ec2.go
```diff
@@ -971,25 +1009,28 @@ func (obj *AwsEc2Res) snsListener(listenAddr string) (net.Listener, error) {
...
 		return
 	}
-	// decode json
+	// Decode the post. If we are unable to decode the data, we print the
+	// error, ignore the post and return.
 	decoder := json.NewDecoder(req.Body)
 	var post postData
 	if err := decoder.Decode(&post); err != nil {
 		http.Error(w, "Bad request", http.StatusBadRequest)
-		select {
```
purpleidea: let's keep this in but wrap it in a condition some sort of something ErrorXXXStrict

resources/aws_ec2.go
```diff
@@ -85,6 +93,29 @@ const (
 	waitTimeout = 400
 )
 
+// snsCanonicalTemplate is a template that formats sns messages according to
+// amazon's canonical message format needed to check message signatures.
+// We use printf for the URL so we don't escape special characters.
+const snsCanonicalTemplate = `Message
```
purpleidea: sadly I think this is a good use case for:

`var str string
str += fmt.Sprintf(XXX)
if foo {
   str += fmt.Sprintf(XXX)

}
```

Also please add a link to the docs about this so people don't think we're insane. (amazon is insane)

**Comment by jonathangold** [comment]
@purpleidea See comment in IRC re `fmt.Sprintf` vs `buffer.Write` before you merge.

File: resources/aws_ec2.go [hunk 6]
```diff
@@ -1038,6 +1075,91 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	}
+	return parsedCert, nil
+}
+
+// snsCanonicalFormat formats post messages as required for signature
+// verification. For more information about this requirement see:
+// http://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.verify.signature.html
+func (obj *AwsEc2Res) snsCanonicalFormat(post postData) []byte {
+	var str string
+	str += fmt.Sprintf("Message\n")
```
purpleidea: A lot of these can be simplified, eg: 

` 	str += "Message\n"`

or 

`str += post.MessageID + "\n"` (if it's already a string)

or 

`str += string(post.MessageID) + "\n"` (if it isn't)

resources/aws_ec2.go
```diff
@@ -1038,6 +1075,91 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	str += fmt.Sprintf("%s\n", post.Timestamp)
+	if post.Token != "" {
+		str += fmt.Sprintf("Token\n")
+		str += fmt.Sprintf("%s\n", post.Token)
+	}
+	str += fmt.Sprintf("TopicArn\n")
+	str += fmt.Sprintf("%s\n", post.TopicArn)
+	str += fmt.Sprintf("Type\n")
+	str += fmt.Sprintf("%s\n", post.Type)
+	return []byte(str)
```
purpleidea: Assuming you want the function signature to stay the same, I think this is fine as is, modulo the above mentioned changes.

resources/aws_ec2.go
```diff
@@ -69,6 +73,10 @@ const (
 	SnsPolicyService = "events.amazonaws.com"
 	// SnsPolicyAction is the specific permission we are granting in the policy.
 	SnsPolicyAction = "SNS:Publish"
+	// SnsCertURLRegex is used to make sure we only download certificates
+	// from amazon. This regex will match "https://sns.***.amazonaws.com/"
```
purpleidea: Either comment or regex is wrong since there is a dash (`-`) you didn't talk about.

resources/aws_ec2.go
```diff
@@ -133,6 +141,12 @@ type AwsEc2Res struct {
 
 	WatchEndpoint   string `yaml:"watchendpoint"`   // the public url of the sns endpoint, eg: http://server:12345/
 	WatchListenAddr string `yaml:"watchlistenaddr"` // the local address or port that the sns listens on, eg: 10.0.0.0:23456 or 23456
+	// ErrorJSONStrict is a bool that controls weather json.Decode(post)
+	// errors are sent to the engine or not. If it is set to true and an
+	// error is produced, Watch() will return the error and the engine will
+	// shut down.
+	ErrorJSONStrict bool `yaml:"errorjsonstrict"`
```
purpleidea: Let's call this something without the word JSON. (The implementation should be allowed to change without the interface needing to.) How about ErrorOnInvalidMessages or similar (replace Messages with whatever they're called)

resources/aws_ec2.go
```diff
@@ -971,25 +995,38 @@ func (obj *AwsEc2Res) snsListener(listenAddr string) (net.Listener, error) {
...
 		return
 	}
-	// decode json
+	// Decode the post. If an error is produced we either ignore the post,
+	// or if ErrorJSONStrict is true, send the error through awsChan so
+	// Watch() can return the error and the engine can shut down.
 	decoder := json.NewDecoder(req.Body)
 	var post postData
 	if err := decoder.Decode(&post); err != nil {
-		http.Error(w, "Bad request", http.StatusBadRequest)
```
purpleidea: Why did you move this?
My comment about having log earlier was about log vs. http -- not "internal action" vs. "http". I think the http should happen first, and then we shutdown. Yes you can put the log message before that.

resources/aws_ec2.go
```diff
@@ -971,25 +995,38 @@ func (obj *AwsEc2Res) snsListener(listenAddr string) (net.Listener, error) {
...
+		if obj.ErrorJSONStrict {
+			select {
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "error decoding incoming POST, check struct formatting"),
+			}:
+			case <-obj.closeChan:
+			}
+			return
 		}
+		log.Printf("%s: error decoding post: %s", obj, err)
```
purpleidea: Don't these two need to be moved up to the top? If ErrorJSONStrict we don't get the log message!

resources/aws_ec2.go
jonathangold: I figured we wanted one or the other, but that makes sense. Changed.

resources/aws_ec2.go
jonathangold: I figured we wanted one or the other, but this makes sense. Changed.

jonathangold:
@purpleidea Changes pushed.

resources/aws_ec2.go
```diff
@@ -133,6 +142,12 @@ type AwsEc2Res struct {
 
 	WatchEndpoint   string `yaml:"watchendpoint"`   // the public url of the sns endpoint, eg: http://server:12345/
 	WatchListenAddr string `yaml:"watchlistenaddr"` // the local address or port that the sns listens on, eg: 10.0.0.0:23456 or 23456
+	// ErrorOnMalformedPost controls whether or not malformed HTTP post
+	// requests, that cause JSON decoder errors, will also make the engine
+	// shut down. If ErrorOnBadRequest set to true and an error occurs,
```
purpleidea: ErrorOnBadRequest ??? What's that?

resources/aws_ec2.go
```diff
@@ -1038,6 +1075,91 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	}
+	// convert the message to canonical form
+	message := obj.snsCanonicalFormat(post)
+	// decode the message signature from base64
+	signature, err := base64.StdEncoding.DecodeString(post.Signature)
+	if err != nil {
+		return errwrap.Wrapf(err, "error decoding string")
+	}
+	// check the signature against the message
+	err = cert.CheckSignature(x509.SHA1WithRSA, message, signature)
```
purpleidea: err := and group with the if please.

alternatively we could

return errwrap.Wrapf(cert.CheckSignature(x509.SHA1WithRSA, message, signature),  "error checking signature")

but that's too long and messy. Better to do the first fix and keep it obvious for crypto.

resources/aws_ec2.go
```diff
@@ -1038,6 +1075,91 @@ func (obj *AwsEc2Res) snsPostHandler(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		return nil, errwrap.Wrapf(err, "http get error")
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errwrap.Wrapf(err, "error reading post body")
+	}
+	// decode and the certificate and discard the rest
+	decodedCert, _ := pem.Decode(body)
```
purpleidea: what's the second arg? an error?

jonathangold: It's any junk data that may exist in the response after the pem certificate. I'll make that more clear in the comment.

resources/aws_ec2.go
jonathangold: whoops

jonathangold:
@purpleidea done.

PR #274, opened 2017-11-28
 resources: aws: ec2: CheckApply fix pending and stopping cases

jonathangold:
Found this bug while trying to fix the other one. :P

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -92,6 +92,9 @@ const (
 	CweTargetJSON = "$.detail"
 	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
 	waitTimeout = 400
+	// AwsErrIncorrectInstanceState is the error returned when an action
+	// cannot be completed due to the current instance state.
+	AwsErrIncorrectInstanceState = "IncorrectInstanceState"
 )
 
 //go:generate stringer -type=awsEc2Event -output=awsec2event_stringer.go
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -785,6 +785,13 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 	ctx, cancel := context.WithTimeout(context.TODO(), waitTimeout*time.Second)
 	defer cancel()
 	if len(diOutput.Reservations) == 1 {
+		// If the instance is in the process of starting up or stopping
+		// we cannot affect its state until it finishes and Watch sends
+		// an event.
+		initialState := *diOutput.Reservations[0].Instances[0].State.Name
```
purpleidea: Is it possible that it gets into this state *right* after we check this?

jonathangold: @purpleidea the short answer is yes. 
Instead I should be checking for the error that's produced where I try to start/stop/terminate the instances and return false, nil there.

resources/aws_ec2.go
```diff
@@ -798,6 +798,13 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 			}
 			_, err := obj.client.StartInstances(startInput)
 			if err != nil {
+				// If the instance is not in a state where it
+				// can be started, we can't do anything.
+				if aerr, ok := err.(awserr.Error); ok {
+					if aerr.Code() == "IncorrectInstanceState" {
```
purpleidea: If we're going to have all these absurd strings everywhere let's at least store them as `const` Maybe something like `AwsErrWhatever`...

jonathangold:
@purpleidea fixed. :)

PR #275, opened 2017-11-28
Aws/bugs

jonathangold:
This patch fixes 2 bugs in longpollWatch()

https://github.com/purpleidea/mgmt/commit/be50e3f05da54ea71636def9eeea4b5e6000f459 The first is that after 10-15 minutes, the waiters return with "ResourceNotReady: exceeded wait
attempts" which we need to ignore and continue.

https://github.com/purpleidea/mgmt/commit/152b9cbec30b271ab47ccec342e41f7b6f09a3f9 The second takes care of the concurrent waiters for stopped and terminated as discussed.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -466,46 +466,25 @@ func (obj *AwsEc2Res) longpollWatch() error {
 						},
 					}
 					log.Printf("%s: Watching: %s", obj, *diOutput.Reservations[0].Instances[0].InstanceId)
-					if err := obj.client.WaitUntilInstanceStoppedWithContext(ctx, waitInput); err != nil {
-						if aerr, ok := err.(awserr.Error); ok {
-							if aerr.Code() == request.CanceledErrorCode {
-								log.Printf("%s: Request cancelled", obj)
-							}
-							if aerr.Code() == request.WaiterResourceNotReadyErrorCode {
-								continue
-							}
-						}
-						select {
-						case obj.awsChan <- &chanStruct{
-							err: errwrap.Wrapf(err, "unknown error waiting for instance to stop"),
-						}:
-						case <-obj.closeChan:
-						}
-						return
-					}
-					stateOutput, err := obj.client.DescribeInstances(diInput)
+					// Wait for instance to stop or
+					// terminate concurrently.
+					event, err := longpollRunningWaiter(ctx, waitInput, obj.client)
 					if err != nil {
 						select {
 						case obj.awsChan <- &chanStruct{
-							err: errwrap.Wrapf(err, "error describing instances"),
+							err: errwrap.Wrapf(err, "internal waiter error"),
 						}:
 						case <-obj.closeChan:
 						}
 						return
 					}
-					var stateName string
-					if len(stateOutput.Reservations) == 1 {
-						stateName = *stateOutput.Reservations[0].Instances[0].State.Name
+					select {
+					case obj.awsChan <- &chanStruct{
+						event: event,
+					}:
+					case <-obj.closeChan:
 					}
-					if len(stateOutput.Reservations) == 0 || (len(stateOutput.Reservations) == 1 && stateName != "running") {
-						select {
-						case obj.awsChan <- &chanStruct{
-							event: awsEc2EventInstanceStopped,
-						}:
-						case <-obj.closeChan:
-							return
-						}
-					}
+					return
 				}
 			}
 			if obj.State == "stopped" {
@@ -641,10 +620,13 @@ func (obj *AwsEc2Res) longpollWatch() error {
 			}
 		case msg, ok := <-obj.awsChan:
 			if !ok {
-				return *exit
+				return fmt.Errorf("channel closed unexpectedly")
 			}
 			if err := msg.err; err != nil {
 				return err
+			}
+			if msg.event == awsEc2EventNone {
+				continue
 			}
 			log.Printf("%s: State: %v", obj, msg.event)
 			obj.StateOK(false)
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -394,6 +394,9 @@ func (obj *AwsEc2Res) longpollWatch() error {
 	}()
 	obj.wg.Add(1)
 	go func() {
+		// shorten some errors we need throughout
+		errCancelled := request.CanceledErrorCode
```
purpleidea: amazingly, this is actually a constant! https://godoc.org/github.com/aws/aws-sdk-go/aws/request#pkg-constants

let's not re-define it, let's just use it directly.

resources/aws_ec2.go
```diff
@@ -466,23 +466,47 @@ func (obj *AwsEc2Res) longpollWatch() error {
 						},
 					}
 					log.Printf("%s: Watching: %s", obj, *diOutput.Reservations[0].Instances[0].InstanceId)
-					if err := obj.client.WaitUntilInstanceStoppedWithContext(ctx, waitInput); err != nil {
-						if aerr, ok := err.(awserr.Error); ok {
-							if aerr.Code() == errCancelled {
-								log.Printf("%s: Request cancelled", obj)
+					// Wait for instance to stop or
+					// terminate concurrently.
+					wg := sync.WaitGroup{}
```
purpleidea: Cool!

jonathangold:
@purpleidea Agreed. I think that in the long-run, longpollWatch and CheckApply would both benefit from being broken down into helper funcs/methods, much like snsWatch is. At the very least, it would eliminate a bunch of channel selects and make for easier reading.
For now I'll just encapsulate the goroutines in their own method.

resources/aws_ec2.go
```diff
@@ -1014,6 +994,65 @@ func (obj *AwsEc2Res) prependName() string {
 	return AwsPrefix + obj.GetName()
 }
 
+// waitForStopAndTerminate waits for the instance to stop and waits for it to
+// terminate. If either waiter returns, the instance state is checked, and an
+// awsEc2Event is returned.
+func (obj *AwsEc2Res) waitForStopAndTerminate(ctx context.Context, waitInput *ec2.DescribeInstancesInput) (awsEc2Event, error) {
```
purpleidea: do we want to name this the absurd `waitUntilInstanceStoppedOrTerminatedWithContext` to match AWS naming conventions?

Maybe it could also have the same signature? as WaitUntilInstanceStoppedWithContext (so return error)

IOW we pretend it's identical to `WaitUntilInstanceStoppedWithContext` or `WaitUntilInstanceTerminatedWithContext` except it waits for either.

resources/aws_ec2.go
```diff
@@ -1014,6 +994,65 @@ func (obj *AwsEc2Res) prependName() string {
...
+	go func() {
+		defer wg.Done()
+		defer wgCancel()
+		if err := obj.client.WaitUntilInstanceTerminatedWithContext(wgCtx, waitInput); err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				if aerr.Code() == request.CanceledErrorCode || aerr.Code() == request.WaiterResourceNotReadyErrorCode {
+					return
+				}
+			}
+			select {
```
purpleidea: This shouldn't be part of this function if you agree with the above comment about matching the AWS API.

resources/aws_ec2.go
```diff
@@ -1014,6 +994,65 @@ func (obj *AwsEc2Res) prependName() string {
...
+			case obj.awsChan <- &chanStruct{
+				err: errwrap.Wrapf(err, "unknown error waiting for instance to terminate"),
+			}:
+			case <-obj.closeChan:
+			}
+		}
+	}()
+	wg.Wait()
+	// Check the instance state, and return the appropriate event.
```
purpleidea: As well, this should also be outside of this function if you agree with my initial API comment. Of course if this type of logic happens often enough throughout the code, it's fine to put it in a helper function!

resources/aws_ec2.go
```diff
@@ -1014,6 +994,78 @@ func (obj *AwsEc2Res) prependName() string {
...
+		defer wgCancel()
+		if err := obj.client.WaitUntilInstanceStoppedWithContext(wgCtx, waitInput); err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				if aerr.Code() == request.CanceledErrorCode || aerr.Code() == request.WaiterResourceNotReadyErrorCode {
+					return
+				}
+			}
+			select {
+			case errChan <- errwrap.Wrapf(err, "unknown error waiting for instance to stop"):
+			case <-obj.closeChan:
```
purpleidea: `obj.*` shouldn't be in here-- remember this is like a new AWS method. Actually it probably shouldn't even be part of `*AwsEc2Res`, but if you want to use a closeChan pattern, you need to pick a private var inside this function.

External to this function, which could get closed by the context, that context closing might trigger an internal close via the internal closeChan! :)

resources/aws_ec2.go
```diff
@@ -1014,6 +994,78 @@ func (obj *AwsEc2Res) prependName() string {
...
+					return
+				}
+			}
+			select {
+			case errChan <- errwrap.Wrapf(err, "unknown error waiting for instance to terminate"):
+			case <-obj.closeChan:
+			}
+		}
+	}()
+	select {
```
purpleidea: This pattern isn't correct-- the two go routines will start running, and then assuming something isn't immediately available on errChan, default will trigger mean we *will* return nil. The only catch is we'll only see that happen after the two go-routines close and the waitGroup lets us exit.

So you've actually just masked any chance of error.

jonathangold:
Sorry. I pushed this code prematurely. I'll work on it tomorrow and make sure the logic makes sense. :P

jonathangold:
Sorry. I pushed this code prematurely. I'll work on it tomorrow and make sure the logic makes sense. :P

resources/aws_ec2.go
```diff
@@ -1014,6 +994,91 @@ func (obj *AwsEc2Res) prependName() string {
...
+	defer close(errChan)
+	wg := sync.WaitGroup{}
+	defer wg.Wait()
+	closeChan := make(chan struct{})
+	wgCtx, wgCancel := context.WithCancel(ctx)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		select {
+		case <-wgCtx.Done():
```
jonathangold: @purpleidea Not %100 sure if this is the right approach to closing closeChan. Is there a chance that this could block? IOW, if there's an error, is it possible that wgCtx won't get cancelled so closeChan doesn't get closed?

resources/aws_ec2.go
```diff
@@ -1014,6 +994,91 @@ func (obj *AwsEc2Res) prependName() string {
...
+			}
+		}
+	}()
+	select {
+	case err, ok := <-errChan:
+		if !ok {
+			return fmt.Errorf("error reading from errChan")
+		}
+		return err
+	case <-wgCtx.Done():
```
purpleidea: what does the AWS api return if the context gets cancelled? :P

resources/aws_ec2.go
```diff
@@ -1014,6 +994,91 @@ func (obj *AwsEc2Res) prependName() string {
...
+			select {
+			case errChan <- errwrap.Wrapf(err, "unknown error waiting for instance to terminate"):
+			case <-closeChan:
+			}
+		}
+	}()
+	select {
+	case err, ok := <-errChan:
+		if !ok {
+			return fmt.Errorf("error reading from errChan")
```
purpleidea: this is not what this means, and not what we'd want to return if errChan closes-- we'd want to use that channel closed signal as a "return nil" (without error!)

resources/aws_ec2.go
```diff
@@ -1014,6 +994,91 @@ func (obj *AwsEc2Res) prependName() string {
...
+// or the other.
+func waitUntilInstanceStoppedOrTerminatedWithContext(ctx context.Context, waitInput *ec2.DescribeInstancesInput, c *ec2.EC2) error {
+	errChan := make(chan error)
+	defer close(errChan)
+	wg := sync.WaitGroup{}
+	defer wg.Wait()
+	closeChan := make(chan struct{})
+	wgCtx, wgCancel := context.WithCancel(ctx)
+	wg.Add(1)
+	go func() {
```
purpleidea: You can make this work without having a third goroutine AFAICT.

resources/aws_ec2.go
purpleidea: this is not what this means-- rephrase your error message please and explain why you want this scenario to be an error. I'm okay with this, but you have to tell me why.

resources/aws_ec2.go
```diff
@@ -1014,6 +994,91 @@ func (obj *AwsEc2Res) prependName() string {
...
+			select {
+			case errChan <- errwrap.Wrapf(err, "unknown error waiting for instance to stop"):
+			case <-closeChan:
+			}
+		}
+	}()
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		defer wgCancel()
```
purpleidea: what's this? you shouldn't do that here.
and you didn't close(closeChan) -- which is probably what you want to do here. you can cheat by doing this via a https://golang.org/pkg/sync/#Once if you want.

resources/aws_ec2.go
```diff
@@ -641,11 +618,14 @@ func (obj *AwsEc2Res) longpollWatch() error {
 			}
 		case msg, ok := <-obj.awsChan:
 			if !ok {
-				return *exit
+				return fmt.Errorf("channel closed unexpectedly")
```
jonathangold: If you want me to put this in another patch I don't mind. It came up when I was testing and I hadn't noticed.

purpleidea: Should this be an error or nil?

resources/aws_ec2.go
```diff
@@ -1014,6 +994,101 @@ func (obj *AwsEc2Res) prependName() string {
...
+		}
+	}
+	return awsEc2EventNone, nil
+}
+
+// waitUntilInstanceStoppedOrTerminatedWithContext combines the two waiters
+// required to trigger events if a running instance is stopped or terminated.
+// This function is needed, because the AWS api only provides waiters for one
+// or the other.
+func waitUntilInstanceStoppedOrTerminatedWithContext(ctx context.Context, waitInput *ec2.DescribeInstancesInput, c *ec2.EC2) error {
```
purpleidea: Nice implementation here! ;)

resources/aws_ec2.go
```diff
@@ -466,45 +466,22 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
-							err: errwrap.Wrapf(err, "unknown error waiting for instance to stop"),
-						}:
-						case <-obj.closeChan:
-						}
-						return
-					}
-					stateOutput, err := obj.client.DescribeInstances(diInput)
+					// Wait for instance to stop or
+					// terminate concurrently.
+					event, err := longpollRunningWaiter(ctx, waitInput, obj.client)
```
purpleidea: RE: https://github.com/jonathangold/mgmt/blob/de5caffce49a6d69156c8b1853670b4d18930ca0/resources/aws_ec2.go#L471-L485

are there any returns needed like the other ones in the other select's?

purpleidea:
Merged thanks!

PR #276, opened 2017-11-29
resources: virt: Update libvirt-xml target

jonathangold:
~~***This compiles, and in theory should work, but I haven't tested it***~~

Builds started failing due to go-libvirt-xml 6d97448. In that patch,
the DomainChannelTarget struct was changed from having a single type
field, to having an individual field for each virtualization type.

This patch updates the connection check in Init to reflect the changes
to go-libvirt-xml, so that builds no longer fail.

### DIFF
--- a/resources/virt.go
+++ b/resources/virt.go
@@ -185,9 +185,9 @@ func (obj *VirtRes) Init() error {
 
 		// guest agent: domain->devices->channel->target->state == connected?
 		for _, x := range domXML.Devices.Channels {
-			if x.Target.Type == "virtio" && strings.HasPrefix(x.Target.Name, "org.qemu.guest_agent.") {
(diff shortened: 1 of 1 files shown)


jonathangold:
I checked the XML marshaling in go-libvirt-xml and the output doesn't seem to have changed. I don't think any other changes are needed.

purpleidea:
@jonathangold LGTM -- but:

> This compiles, and in theory should work, but I haven't tested it

Want to test it? :)

jonathangold:
@purpleidea I do, but I won't have a chance until later tonight. :\

purpleidea:
@jonathangold That's cool!

jonathangold:
@purpleidea: this is now tested and good to go.

purpleidea:
https://github.com/purpleidea/mgmt/commit/406aa55667f11cf9fe1f59ca8d5da1d0d3e2bb5b

Merged thanks!!

PR #278, opened 2017-12-05
resources: aws: ec2: Add options to bootstrap mgmt

jonathangold:
Working for rhel and ubuntu images.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -33,6 +33,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
+	cwe "github.com/aws/aws-sdk-go/service/cloudwatchevents"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/sns"
 	multierr "github.com/hashicorp/go-multierror"
@@ -58,6 +59,18 @@ const (
 	SnsSubscriptionProto = "http"
 	// SnsServerShutdownTimeout is the maximum number of seconds to wait for the http server to shutdown gracefully.
 	SnsServerShutdownTimeout = 30
+	// CwePrefix gets prepended onto the cloudwatch rule name.
+	CwePrefix = Ec2Prefix + "cw-"
+	// CweRuleName is the name of the rule created by makeCloudWatchRule.
+	CweRuleName = CwePrefix + "state"
+	// CweRuleSource describes the resource type to monitor for cloudwatch events.
+	CweRuleSource = "aws.ec2"
+	// CweRuleDetailType describes the specific type of events to trigger cloudwatch.
+	CweRuleDetailType = "EC2 Instance State-change Notification"
+	// CweTargetID is used to tell cloudwatch events to target the sns service.
+	CweTargetID = "sns"
+	// CweTargetJSON is the json field that cloudwatch will send to our endpoint so we don't get more than we need.
+	CweTargetJSON = "$.detail"
 	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
 	waitTimeout = 400
 )
@@ -120,6 +133,8 @@ type AwsEc2Res struct {
 	// so we save it here when we create the topic instead.
 	snsTopicArn string
 
+	cweClient *cwe.CloudWatchEvents // client for AWS CloudWatchEvents API calls
+
 	awsChan   chan *chanStruct // channel used to send events and errors to Watch()
 	closeChan chan struct{}    // channel used to cancel context when it's time to shut down
 	wg        *sync.WaitGroup  // waitgroup for goroutines in Watch()
@@ -129,6 +144,18 @@ type AwsEc2Res struct {
 type chanStruct struct {
 	event awsEc2Event
 	err   error
+}
+
+// cloudWatchRule denotes the structure of cloudwatch rules.
+type cloudWatchRule struct {
+	Source     []string   `json:"source"`
+	DetailType []string   `json:"detail-type"`
+	Detail     ruleDetail `json:"detail"`
+}
+
+// ruleDetail is the structure of the detail field in cloudWatchRule.
+type ruleDetail struct {
+	State []string `json:"state"`
 }
 
 // postData is the format of the messages received and decoded by snsPostHandler().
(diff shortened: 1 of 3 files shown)


resources/aws_ec2.go
```diff
@@ -111,6 +113,36 @@ const (
...
+chmod +x "/opt/mgmtconfig/mgmt"
+
+curl "{{- .RemoteGraph -}}" -o graph.yaml
+
+internal=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
+external=$(curl http://169.254.169.254/latest/meta-data/public-ipv4)
+
+/opt/mgmtconfig/mgmt run --yaml graph.yaml --no-pgp --tmp-prefix \
+--seeds {{.Seed}} \
+--client-urls "http://${internal}:2379" \
```
jonathangold: Should we let the user define the client and server ports?

resources/aws_ec2.go
```diff
@@ -873,6 +931,15 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 			userData := base64.StdEncoding.EncodeToString([]byte(obj.UserData))
 			runParams.SetUserData(userData)
 		}
+		if obj.Bootstrap {
+			//TODO: Firewall rules (security groups)
```
purpleidea: silly, but this needs a space:

```
// TODO: blah
```

resources/aws_ec2.go
```diff
@@ -154,7 +186,11 @@ type AwsEc2Res struct {
 	// UserData is used to run bash and cloud-init commands on first launch.
 	// See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html
 	// for documantation and examples.
-	UserData string `yaml:"userdata"`
+	UserData    string `yaml:"userdata"`
+	Bootstrap   bool   `yaml:"bootstrap"`   // install mgmt on the aws instance and phone home
```
purpleidea: Can `Bootstrap` be a useful value instead of a bool? Just something to think about. Let's discuss here before you change it.

resources/aws_ec2.go
```diff
@@ -111,6 +113,36 @@ const (
 	awsEc2EventInstanceExists
 )
 
+// bootstrapScript is a bash script template for installing mgmt and connecting
+// back to the seed specified in the definition.
+const bootstrapScript string = `#!/bin/bash
+if [ -n "$(command -v apt)" ]
```
purpleidea: what's command?
what about `which` and `test -z`, eg: https://github.com/purpleidea/mgmt/blob/master/misc/make-deps.sh#L14

(which one is preferable?)

resources/aws_ec2.go
```diff
@@ -111,6 +113,36 @@ const (
...
+  yum install -y augeas-libs libvirt-libs
+fi
+
+mkdir -p "/opt/mgmtconfig"
+curl "{{- .Binary -}}" -o "/opt/mgmtconfig/mgmt"
+chmod +x "/opt/mgmtconfig/mgmt"
+
+curl "{{- .RemoteGraph -}}" -o graph.yaml
+
+internal=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
```
purpleidea: aws magic?? needs comments please

resources/aws_ec2.go
```diff
@@ -111,6 +113,36 @@ const (
...
+  apt update
+  apt install -y libaugeas0 libvirt0
+fi
+
+if [ -n "$(command -v yum)" ]
+then
+  yum install -y augeas-libs libvirt-libs
+fi
+
+mkdir -p "/opt/mgmtconfig"
```
purpleidea: maybe we can use `mktemp` instead of /opt/ ?

resources/aws_ec2.go
```diff
@@ -111,6 +113,36 @@ const (
...
+const bootstrapScript string = `#!/bin/bash
+if [ -n "$(command -v apt)" ]
+then
+  apt update
+  apt install -y libaugeas0 libvirt0
+fi
+
+if [ -n "$(command -v yum)" ]
+then
+  yum install -y augeas-libs libvirt-libs
```
purpleidea: tabs in shell scripts instead of spaces please

resources/aws_ec2.go
purpleidea: Good question. Let's say no for now. You can pull the values we want to use from etcd package (they are constants in our etcd package) https://github.com/purpleidea/mgmt/blob/master/etcd/etcd.go#L90
You can either split on these, or patch it to store them separately in etcd.go -- I think parsing is preferable.

resources/aws_ec2.go
```diff
@@ -1091,6 +1170,29 @@ func waitUntilInstanceStoppedOrTerminatedWithContext(ctx context.Context, waitIn
 	}
 }
 
+// bootstrapUserData creates a startup script which installs mgmt on the
+// instance, and connects to the specified seed URL.
+func (*AwsEc2Res) bootstrapUserData(binary, graph, seed string) (string, error) {
+	// create a template for the user data
+	buff := new(bytes.Buffer)
```
purpleidea: buff is sexy, buf is buffer :)

resources/aws_ec2.go
jonathangold: afaik, `command` is built into the shell whereas `which` is an external binary. Truthfully I don't think it really matters which one we use.

resources/aws_ec2.go
jonathangold: Yeah. I'll add a comment.

resources/aws_ec2.go
jonathangold: `mktemp -p "/tmp/mgmtconfig"`?

jonathangold: I'll play with it.

resources/aws_ec2.go
jonathangold: I suppose it's a redundant field. If binary, remotegraph and seed are set, we know they want to bootstrap. Did you have something particular in mind?

purpleidea: Not really, but I didn't think about it yet. Think on it, and let me know. I'll try and put myself in the mode and do the same.

resources/aws_ec2.go
purpleidea: mktemp -d will get you a temp dir instead of assuming that opt is writable or available.

resources/aws_ec2.go
purpleidea: Yeah, that's cool if this works! I had never even seen this before :) Is it something that works on bash and sh ?

jonathangold: It works in sh for me.

resources/aws_ec2.go
jonathangold: Unfortunately it's not possible since etcd imports resources so importing etcd isn't allowed. I added them as consts.

purpleidea:
As discussed, we're putting this feature on pause for now.

PR #279, opened 2017-12-06

jonathangold:
This patch simplifies longpollwatch by getting rid of some unnecessary
api calls and breaking the waiters out into their own functions.

The describeInstances() at the beginning of each case was eliminated by checking the state from diOutput.
"stopped" and "terminated" were given their own waiter functions to match "running"

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -400,6 +400,8 @@ func (obj *AwsEc2Res) longpollWatch() error {
 		defer obj.wg.Done()
 		defer close(obj.awsChan)
 		for {
+			// get the instance that matches the name specified in
+			// the definition. Ignore terminated instances.
 			diInput := &ec2.DescribeInstancesInput{
 				Filters: []*ec2.Filter{
 					{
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -483,36 +464,15 @@ func (obj *AwsEc2Res) longpollWatch() error {
 						event: event,
 					}:
 					case <-obj.closeChan:
+						return
```
jonathangold: We don't want to return in case of an event, just on closeChan. I think I was looking at the wrong select when I added this. Either way, I should have thought about it more. :\

resources/aws_ec2.go
```diff
@@ -483,36 +464,15 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
 					}
-					return
```
jonathangold: We don't want to return in case of an event, just on closeChan. I think I was looking at the wrong select when I added this. Either way, I should have thought about it more. :\

PR #280, opened 2017-12-07
cli: Add license flag

jonathangold:
This patch adds the option to print the license with a cli flag. It
uses go-bindata to store the license file. The file is generated by
running `make bindata` and the result is stored in the bindata
directory.


jonathangold:
Would you prefer I split this into two commits? (1 for bindata and 1 for the cli flag)

Makefile
```diff
@@ -101,6 +101,10 @@ run:
 race:
 	find . -maxdepth 1 -type f -name '*.go' -not -name '*_test.go' | xargs go run -race -ldflags "-X main.program=$(PROGRAM) -X main.version=$(SVERSION)"
 
+# generate go files from non-go source
+bindata:
```
purpleidea: Good, but we also need to change the build line to be:

`build: bindata $(PROGRAM)`

bindata/Makefile
```diff
@@ -0,0 +1,27 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# The bindata target generates go files from any source defined below. To use
+# the files, import "github.com/purpleidea/mgmt/bindata" and use
+# `bytes, err := bindata.Asset("FILEPATH")` where FILEPATH is the path of the
+# original input file relative to `bindata/`. To make the target you will need
+# to install go-bindata using `go get -u github.com/jteeuwen/go-bindata/...`.
```
purpleidea: why the `...` ?

bindata/Makefile
```diff
@@ -0,0 +1,27 @@
...
+default:
```
purpleidea: can we instead have:

1) A .PHONY target called `build` or something
2) A line with `default: build`
3) the name of the file as the target so it refreshes properly, eg:
`COPYING.go: ../COPYING`

4) `build: COPYING.go` so the dep works (users can add other deps here in the future)

bindata/Makefile
```diff
@@ -0,0 +1,27 @@
...
+	# go-bindata --pkg bindata -o {OUTPUT} {INPUT}
+	go-bindata --pkg bindata -o ./COPYING.go ../COPYING
+	# gofmt the output files
+	gofmt -s -w *.go
```
purpleidea: since we'll have one of these targets per output file, have this be specific file. You can use these magic vars throughout :)

https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html

test/test-golint.sh
```diff
@@ -24,7 +24,7 @@ if [ "$COMMITS" != "" ] && [ "$COMMITS" -gt "1" ]; then
 	HACK="yes"
 fi
 
-LINT=`find . -maxdepth 3 -iname '*.go' -not -path './old/*' -not -path './tmp/*' -exec golint {} \;`	# current golint output
```
purpleidea: Good, but you also need the change on the `LINT1` line.

bindata/Makefile
```diff
@@ -0,0 +1,32 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# The bindata target generates go files from any source defined below. To use
+# the files, import "github.com/purpleidea/mgmt/bindata" and use
```
purpleidea: sorry, last minute nitpick, can you change this to say `import the "bindata" package` so that we don't store more absolute urls than necessary?

misc/make-deps.sh
```diff
@@ -78,7 +78,8 @@ ret=$?
 if [[ $ret != 0 ]]; then
 	go get golang.org/x/tools/cmd/vet      # add in `go vet` for travis
 fi
-go get golang.org/x/tools/cmd/stringer	# for automatic stringer-ing
-go get github.com/golang/lint/golint	# for `golint`-ing
+go get golang.org/x/tools/cmd/stringer			# for automatic stringer-ing
+go get github.com/jteeuwen/go-bindata/go-bindata	# for compiling in non golang files
```
purpleidea: Do you have one extra go-bindata here?

purpleidea:
W00t! Merged in modified commit as discussed in : 62ca12608d9120270a181d04dc2301951271b558

PR #281, opened 2017-12-09
resources: aws: ec2: Store public ip for send/recv

jonathangold:
This patch adds a PublicIP field to AwsEc2Res so other resources can
receive the value.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -156,6 +156,15 @@ type AwsEc2Res struct {
 	// for documantation and examples.
 	UserData string `yaml:"userdata"`
 
+	// PublicIPv4 is reserved for the instance's public IPv4 address and is
+	// updated by CheckApply() whenever it is run. This is a read-only
+	// parameter.
+	PublicIPv4 string
+	// PrivateIPv4 is reserved for the instance's private IPv4 address and
+	// is updated by CheckApply() whenever it is run. This is a read-only
+	// parameter.
+	PrivateIPv4 string
+
 	client *ec2.EC2 // client session for AWS API calls
 
 	snsClient *sns.SNS // client for AWS SNS API calls
@@ -657,6 +666,12 @@ func (obj *AwsEc2Res) snsWatch() error {
 func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 	log.Printf("%s: CheckApply(%t)", obj, apply)
 
+	var instanceID *string
+
+	// clear any IP addresses we've stored for send/recv.
+	obj.PublicIPv4 = ""
+	obj.PrivateIPv4 = ""
+
 	diInput := ec2.DescribeInstancesInput{
 		Filters: []*ec2.Filter{
 			{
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -820,9 +830,37 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
...
 }
 
+// getPublicIPv4 gets the instance's public ip address. If the instance has a
+// public ip address it is returned (as a string), and if it does not, the
+// method returns an empty string.
+func getPublicIP(c *ec2.EC2, instanceID *string) (string, error) {
+	var ip string
+	// get instance info
+	diInput := &ec2.DescribeInstancesInput{
+		InstanceIds: []*string{instanceID},
```
jonathangold: I have tested this with nil, and it does not panic. ;)

resources/aws_ec2.go
```diff
@@ -156,6 +156,11 @@ type AwsEc2Res struct {
 	// for documantation and examples.
 	UserData string `yaml:"userdata"`
 
+	// PublicIP is reserved for the instance's public IPv4 address and is
+	// updated by CheckApply() whenever it is run. This is not a parameter
```
purpleidea: Maybe instead of "This is not a parameter that should be defined by the user." how about "This is a read-only parameter." or similar?

Can we call this PublicIPv4 instead ?
I'm assuming there's always a PrivateIPv4 as well that we could get easily? (If so want to add that as well?)

resources/aws_ec2.go
```diff
@@ -820,9 +830,37 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 		}
 		log.Printf("%s: instance running", obj)
 	}
+	// get the instance's public ip
+	ip, err := getPublicIP(obj.client, instanceID)
+	if err != nil {
+		return false, errwrap.Wrapf(err, "error getting public ip")
+	}
+	// store it in the struct for other resources to receive
```
purpleidea: Don't be fooled at how small the amount of code is for send/recv, as this is incorrect! How come? Well you might imagine that mgmt starts up and checks the state, and the ec2 instance is already in the correct state and we return from CheckApply with (true, nil) much earlier in the function! In this situation, the IPAddress was never set!

So you need to make sure that it is always set anytime before the function returns, whenever it could return and we'd have wanted to update it. (If the machine's IP disappears, we probably want to set it to empty!)

resources/aws_ec2.go
```diff
@@ -820,9 +830,37 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 		}
 		log.Printf("%s: instance running", obj)
 	}
+	// get the instance's public ip
+	ip, err := getPublicIP(obj.client, instanceID)
```
purpleidea: getPublicIPv4 ?

resources/aws_ec2.go
```diff
@@ -820,9 +849,60 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
...
+		return false, errwrap.Wrapf(err, "error getting public ip")
+	}
+
 	return false, nil
 }
 
+// getPublicIPv4 gets the instance's public ip address. If the instance has a
+// public ip address it is returned (as a string), and if it does not, the
+// method returns an empty string.
+func getPublicIPv4(c *ec2.EC2, instanceID *string) (string, error) {
```
purpleidea: Since we're basically making the same call for this and `getPrivateIPv4`, why not just make one generic getInstance call, and then process the output? (Or if you really want to be specific, return the instanceInfo struct (or whatever it's called)

resources/aws_ec2.go
```diff
@@ -680,10 +691,25 @@ func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 	}
 
 	if len(diOutput.Reservations) < 1 && obj.State == "terminated" {
+		// clear any IP addresses we've stored for send/recv.
```
purpleidea: There is a bug in the code (or this patch) because if `len(diOutput.Reservations) > 1 && !apply, then you'll return from CheckApply without ever having touched this value!

So where's the bug? Either it's in this patch since you forgot about that case, OR having a `len(diOutput.Reservations) > 1` is unexpected and unhandled... Meaning either it should be an error or dealt with somehow, etc...

jonathangold:
see #285

PR #282, opened 2017-12-10
resources: aws: ec2: Refactor CheckApply

jonathangold:
@purpleidea I refactored CheckApply by creating some new functions to help tidy up the method. While the start/stop/terminate/launch functions will only be used once, keeping them in-line would have a negative impact on readability. The describeInstanceBy... functions, on the other hand, will be useful throughout, and will help clean up other parts of the code in the future.

I realize this is a big patch, but it didn't make sense to me to do it bit by bit. That said, I really want to make sure this code is A+, so if you want to review the functions one at a time, that might help both of us avoid getting burned out and/or merging something that's not 100%.

Also, since the diff of the method itself is kind of messy, it would probably be easier to review that part sans diff.

LMK what you think. :D
PS. I'm happy to do a live review if you'd prefer. Whatever's less of a burden for you. ;)
PPS. No rush to review. Let's take our time on this one, and make sure it's perfect!

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -656,170 +656,66 @@ func (obj *AwsEc2Res) snsWatch() error {
 // CheckApply method for AwsEc2 resource.
 func (obj *AwsEc2Res) CheckApply(apply bool) (checkOK bool, err error) {
 	log.Printf("%s: CheckApply(%t)", obj, apply)
+	defer func() {
+		log.Printf("%s: CheckApply Done", obj)
+	}()
 
-	diInput := ec2.DescribeInstancesInput{
-		Filters: []*ec2.Filter{
-			{
-				Name:   aws.String("tag:Name"),
-				Values: []*string{aws.String(obj.prependName())},
-			},
-			{
-				Name: aws.String("instance-state-name"),
-				Values: []*string{
-					aws.String("running"),
-					aws.String("pending"),
-					aws.String("stopped"),
-					aws.String("stopping"),
-				},
-			},
-		},
-	}
-	diOutput, err := obj.client.DescribeInstances(&diInput)
+	// find the instance we need to check.
+	instance, err := describeInstanceByName(obj.client, obj.prependName())
 	if err != nil {
-		return false, errwrap.Wrapf(err, "error describing instances")
+		return false, errwrap.Wrapf(err, "error describing instance")
 	}
 
-	if len(diOutput.Reservations) < 1 && obj.State == "terminated" {
+	// If instance is nil, that means it's in a transitional state (pending,
+	// stopping, shutting-down), because an action was performed externally.
+	// We need to wait until the instance converges into a running, stopped,
+	// or terminated state, at which point we will receive an event.
+	if instance == nil {
+		log.Printf("%s: Could not apply, waiting for state to converge.", obj)
+		return false, nil
+	}
+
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -912,6 +800,84 @@ func (obj *AwsEc2Res) prependName() string {
 	return AwsPrefix + obj.GetName()
 }
 
+// describeInstanceByName takes an ec2 client session and an instance name, and
+// returns an ec2Instance or an error.
+func describeInstanceByName(c *ec2.EC2, name string) (ec2Instance, error) {
```
purpleidea: the biggest unknown to me, is why is this a special type we made (maybe it's useful, you have to convince me) instead of just plainly using https://godoc.org/github.com/aws/aws-sdk-go/service/ec2#Instance ? (or possibly a pointer to that actually)

jonathangold: I could do it with ec2.Instance, but here's why I chose to do this:
- I think it's cleaner to use instance.ID and instance.State vs. instance.InstanceId and instance.State.Name
- ec2Instance only contains information we care about, and when we add things like IP addresses, it will be more clear what's going on, IMO.
We talked about wrapping their api in our own (saner) api, and I think this is part of it. If you disagree, I'm happy to change it. :)

purpleidea: > I think it's cleaner to use instance.ID and instance.State vs. instance.InstanceId and instance.State.Name

Why? You've arguably just copied that information, left out a bunch of potentially useful things, and removed one dot. If the argument was that we added additional information to our ec2Instance struct that needed to be cached or stored, I'd be convinced, but a straight copy doesn't do it for me.

In that second case where we need to store additional things, I'd probably even recommend having out ec2Instance be a:

```
type ec2Instance struct {
        *ec2.Instance // anonymous struct field, look it up

        someSpecialField string
        someOtherField uint32
}
```

> it will be more clear what's going on, IMO.

You haven't said _why_.

> We talked about wrapping their api in our own (saner) api

Yes, our _external_ API. The struct fields and behaviours that we expose to the end user is what should be sane, even if it differs from what AWS's public API does. What we're discussing here is our internal helper functions.

resources/aws_ec2.go
```diff
@@ -656,170 +656,66 @@ func (obj *AwsEc2Res) snsWatch() error {
...
+	// We need to wait until the instance converges into a running, stopped,
+	// or terminated state, at which point we will receive an event.
+	if instance == nil {
+		log.Printf("%s: Could not apply, waiting for state to converge.", obj)
+		return false, nil
 	}
-	if len(diOutput.Reservations) == 1 && *diOutput.Reservations[0].Instances[0].State.Name == obj.State {
+
+	// if the instance is in the correct state, we're done.
+	if aws.StringValue(instance.State.Name) == obj.State {
```
jonathangold: FYI, `aws.StringValue()` does the inverse of `aws.String()`. It dereferences the pointer "safely" by returning `""` if the pointer is `nil`

jonathangold:
Closing this. will submit the fixup to the last checkapply patch when I'm sure it's ready. I need to get better at judging when things are ready for review, so I don't give you more work than I need to, and I get better, faster. :)

jonathangold:
@purpleidea: for review once #284 is merged. I included a second commit with send/recv since it's pretty straightforward (now that checkapply has been reorganized)

PR #283, opened 2017-12-12

jonathangold:
Complete rewrite of longpollWatch for correctness and maintanability.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -173,6 +173,7 @@ type AwsEc2Res struct {
 // chanStruct defines the type for a channel used to pass events and errors to watch.
 type chanStruct struct {
 	event awsEc2Event
+	state string
 	err   error
 }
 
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -912,143 +822,149 @@ func (obj *AwsEc2Res) prependName() string {
...
+		},
+	}
+	// call on the appropriate waiter based on current state.
+	switch aws.StringValue(instance.State.Name) {
+	case ec2.InstanceStateNameRunning, ec2.InstanceStateNameStopping:
+		err = c.WaitUntilInstanceStoppedWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameStopped, ec2.InstanceStateNamePending:
+		err = c.WaitUntilInstanceRunningWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameTerminated:
+		err = c.WaitUntilInstanceExistsWithContext(ctx, waitInput)
```
jonathangold: fwiw, there's a race where this returns some milliseconds before the instance is available, so we get 4 or 5 extra "terminated" events. I know sleeping for a second isn't a great fix, but it would work in this case.

resources/aws_ec2.go
```diff
@@ -173,6 +173,7 @@ type AwsEc2Res struct {
 // chanStruct defines the type for a channel used to pass events and errors to watch.
 type chanStruct struct {
 	event awsEc2Event
+	state string
```
jonathangold: Added this, because, now that we have instance state consts to use, we no longer need the awsEc2Event type. The next patch will refactor CheckApply which will include switching to the ec2 pkg consts. Once I do the same for snsWatch, we can get rid of the iota consts and the event field.

resources/aws_ec2.go
```diff
@@ -912,143 +822,149 @@ func (obj *AwsEc2Res) prependName() string {
...
-				}
-			}
-		}
 
-		select {
-		case errChan <- errwrap.Wrapf(err, "unknown error waiting for instance to stop"):
-		case <-closeChan:
+	// get the instance name.
+	for _, tag := range instance.Tags {
+		if aws.StringValue(tag.Key) == "Name" {
```
jonathangold: Should I make this a const?

resources/aws_ec2.go
```diff
@@ -912,143 +822,149 @@ func (obj *AwsEc2Res) prependName() string {
...
-					err = nil
-				}
-			}
+	// instance.InstanceId might be nil in case of terminated instances, which
+	// the api will treat as if it isn't included.
+	waitInput := &ec2.DescribeInstancesInput{
+		InstanceIds: []*string{instance.InstanceId},
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag:Name"),
```
jonathangold: and build this one from it?

jonathangold:
see #284

PR #284, opened 2017-12-12

jonathangold:
Complete rewrite of Watch() for correctness and maintanability.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -90,11 +90,19 @@ const (
 	CweTargetID = "sns"
 	// CweTargetJSON is the json field that cloudwatch will send to our endpoint so we don't get more than we need.
 	CweTargetJSON = "$.detail"
-	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
-	waitTimeout = 400
+	// AwsErrExceededWaitAttempts is the awserr.Message() that gets sent with
+	// the ResourceStateNotReady awserr.Code() when the waiters time out.
+	AwsErrExceededWaitAttempts = "exceeded wait attempts"
 	// AwsErrIncorrectInstanceState is the error returned when an action
 	// cannot be completed due to the current instance state.
 	AwsErrIncorrectInstanceState = "IncorrectInstanceState"
+	// waitTimeout is the duration in seconds of the timeout context in CheckApply.
+	waitTimeout = 400
+	// nameKey is the name of the tag key that stores the instance name in ec2.Instance.
+	// in ec2.Instance
+	nameKey = "Name"
+	// nameTag is used to define the name tag.
+	nameTag = "tag:" + nameKey
 )
 
 //go:generate stringer -type=awsEc2Event -output=awsec2event_stringer.go
@@ -173,6 +181,7 @@ type AwsEc2Res struct {
 // chanStruct defines the type for a channel used to pass events and errors to watch.
 type chanStruct struct {
 	event awsEc2Event
+	state string
 	err   error
 }
 
@@ -381,12 +390,21 @@ func (obj *AwsEc2Res) Watch() error {
 func (obj *AwsEc2Res) longpollWatch() error {
 	send := false
 	var exit *error
+
+	// We tell the engine that we're running right away. This is not correct,
+	// but the api doesn't have a way to signal when the waiters are ready.
 	if err := obj.Running(); err != nil {
 		return err
 	}
+
+	// cancellable context used for exiting cleanly
+	ctx, cancel := context.WithCancel(context.TODO())
+
+	// clean up when we're done
 	defer obj.wg.Wait()
 	defer close(obj.closeChan)
-	ctx, cancel := context.WithCancel(context.TODO())
+
+	// cancel our context if obj.closeChan closes
 	obj.wg.Add(1)
 	go func() {
 		defer obj.wg.Done()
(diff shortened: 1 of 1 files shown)


resources/aws_ec2.go
```diff
@@ -503,71 +280,67 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+		},
+	}
+	// call on the appropriate waiter based on current state.
+	switch aws.StringValue(instance.State.Name) {
+	case ec2.InstanceStateNameRunning, ec2.InstanceStateNameStopping:
+		err = c.WaitUntilInstanceStoppedWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameStopped, ec2.InstanceStateNamePending:
+		err = c.WaitUntilInstanceRunningWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameTerminated:
+		err = c.WaitUntilInstanceExistsWithContext(ctx, waitInput)
```
jonathangold: There's a race where this returns some milliseconds before the instance is available (according to my describeInstanceByName function,) so we get 4 or 5 extra "terminated" events. I know sleeping for a second isn't a great fix, but it _would_ work in this case. Better ideas?

jonathangold:
https://github.com/purpleidea/mgmt/pull/284/commits/61e5392b01a3c45d9da102f8c0abc9071fb22cc8 is just there so you can look at the additions on their own in https://github.com/purpleidea/mgmt/pull/284/commits/79cabc7f3cbcfa2bc4aced762289f98c2125c452 since this is a complete rewrite. https://github.com/purpleidea/mgmt/pull/284/commits/61e5392b01a3c45d9da102f8c0abc9071fb22cc8 contains all the deletions (+ 3 lines to make it "correct"). Now I'm going to bed :P

jonathangold:
@purpleidea Got rid of the deletions.

resources/aws_ec2.go
```diff
@@ -555,6 +491,73 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+		if !ok {
+			return "", errwrap.Wrapf(err, "error casting awserr")
+		}
+		// ignore these errors
+		if aerr.Code() != request.CanceledErrorCode && aerr.Code() != request.WaiterResourceNotReadyErrorCode {
+			return "", errwrap.Wrapf(err, "internal waiter error")
+		}
+		// if the waiter times out, start again recursively.
+		if aerr.Message() == AwsErrExceededWaitAttempts {
+			return watchWaiter(ctx, instance, c)
```
purpleidea: While this might logically make sense, we shouldn't push code that depends on an infinitely deep recursion call stack, since *eventually* we would run out of memory. Not that it's necessarily likely to happen (I didn't measure) but it's logically inappropriate. So we need a different mechanism please.

resources/aws_ec2.go
```diff
@@ -379,174 +388,101 @@ func (obj *AwsEc2Res) Watch() error {
 
 // longpollWatch uses the ec2 api's built in methods to watch ec2 resource state.
 func (obj *AwsEc2Res) longpollWatch() error {
-	send := false
+	var send bool
 	var exit *error
+
+	// we tell the engine we're running right away. Because watchWaiter
+	// checks the instance state directly so we don't miss any events.
 	if err := obj.Running(); err != nil {
```
jonathangold: I'd like to discuss this further, because I imagine when it comes time for send/recv this might no longer be race-free (if it is now.) Also, typo in the comment /s/,

resources/aws_ec2.go
```diff
@@ -379,174 +388,101 @@ func (obj *AwsEc2Res) Watch() error {
...
 		return err
 	}
-	defer obj.wg.Wait()
-	defer close(obj.closeChan)
+
+	// channels and waitgroup for goroutines.
```
jonathangold: I forgot to change these back to the struct channels and wg. I can change it back after you review.

resources/aws_ec2.go
jonathangold: OK. Good thing we didn't go with HASKELL :P 
Do you want me to fix this before you review?

jonathangold:
@purpleidea fixed the recursion.

resources/aws_ec2.go
```diff
@@ -555,6 +486,73 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	if err != nil {
+		aerr, ok := err.(awserr.Error)
+		if !ok {
+			return "", errwrap.Wrapf(err, "error casting awserr")
+		}
+		// ignore these errors
+		if aerr.Code() != request.CanceledErrorCode && aerr.Code() != request.WaiterResourceNotReadyErrorCode {
+			return "", errwrap.Wrapf(err, "internal waiter error")
+		}
+		// if the waiter times out, start again recursively.
```
jonathangold: Forgot to change the comment. Will do after you review.

resources/aws_ec2.go
jonathangold: This waiter doesn't work the way I thought it did. Troubleshooting it now.

jonathangold: Figured it out, I was waiting for any instance with the right name to exist, including terminated ones. ಠ_ಠ
I think I made a change that stopped it waiting for terminated ones in the original version, which might be why waituntilinstancestopped wasn't behaving how I thought.

resources/aws_ec2.go
```diff
@@ -374,19 +383,28 @@ func (obj *AwsEc2Res) Watch() error {
...
+	var send bool
 	var exit *error
+
+	// we tell the engine we're running right away. Because watchWaiter
+	// checks the instance state directly so we don't miss any events.
 	if err := obj.Running(); err != nil {
 		return err
 	}
+
+	// canellable context used for exiting cleanly.
```
purpleidea: typo. one liners are either full sentences with capitalization and period or have neither same above and in a bunch of other comments.

resources/aws_ec2.go
```diff
@@ -374,19 +383,28 @@ func (obj *AwsEc2Res) Watch() error {
...
-	return obj.longpollWatch()
+	return obj.waiterWatch()
 }
 
-// longpollWatch uses the ec2 api's built in methods to watch ec2 resource state.
-func (obj *AwsEc2Res) longpollWatch() error {
-	send := false
+// waiterWatch uses the ec2 api's built in methods to watch ec2 resource state.
+func (obj *AwsEc2Res) waiterWatch() error {
+	var send bool
```
purpleidea: why are we renaming this and also we are we making these sort of unrelated changes in the same commit?

`send := false`
is functionally the same as
`var send bool`

resources/aws_ec2.go
```diff
@@ -555,6 +486,82 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	}
+	// call on the appropriate waiter based on current state.
+	switch aws.StringValue(instance.State.Name) {
+	case ec2.InstanceStateNameRunning, ec2.InstanceStateNameStopping:
+		err = c.WaitUntilInstanceStoppedWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameStopped, ec2.InstanceStateNamePending:
+		err = c.WaitUntilInstanceRunningWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameTerminated:
+		waitInput.Filters = append(waitInput.Filters, existWaiterFilter)
+		err = c.WaitUntilInstanceExistsWithContext(ctx, waitInput)
```
purpleidea: default state should be an error or a panic or ???

resources/aws_ec2.go
```diff
@@ -374,19 +383,28 @@ func (obj *AwsEc2Res) Watch() error {
...
 
-// longpollWatch uses the ec2 api's built in methods to watch ec2 resource state.
-func (obj *AwsEc2Res) longpollWatch() error {
-	send := false
+// waiterWatch uses the ec2 api's built in methods to watch ec2 resource state.
+func (obj *AwsEc2Res) waiterWatch() error {
+	var send bool
 	var exit *error
+
+	// we tell the engine we're running right away. Because watchWaiter
```
purpleidea: So the fact that we previously did this right away was wrong, and now we've got a comment saying it's right. It was wrong before, but it was a bug. Now the comment incorrectly re-enforces the bug. :/

obj.Running _tells_ the engine that we're now watching! This is a lie!! It's limited by their long poll API not exposing when they're ready, but still.

resources/aws_ec2.go
```diff
@@ -555,6 +486,82 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+		if aws.StringValue(tag.Key) == nameKey {
+			name = aws.StringValue(tag.Value)
+		}
+	}
+	// error if we didn't find one.
+	if name == "" {
+		return "", fmt.Errorf("name not found")
+	}
+
+	// build the input for the waiters.
```
purpleidea: why are the following 40-50 lines correct??

resources/aws_ec2.go
jonathangold: I can leave it as longpollWatch, but IMO, longpolling and using the waiters aren't exactly the same, and I think it's causing some confusion. I can also change it back to `send := false`, (I just like `var send bool` better style wise and thought it was okay since this is a refactor)

resources/aws_ec2.go
jonathangold: I guess I picked this up as a bad habit somewhere along the way. Will correct all of them.

resources/aws_ec2.go
jonathangold: I will add some comments explaining the logic and the reasoning behind it. Sorry if it's unclear.

resources/aws_ec2.go
jonathangold: oops. I'll add an error.

resources/aws_ec2.go
jonathangold: I don't think I understand the potential consequences of this race. What exactly does this _mean_ to the engine and what could go wrong if we tell the engine we're watching a second or two before we poll the state and start the waiters?

jonathangold: IOW, is this _just_ about potentially missing events, or is there more to this race on the engine side?

jonathangold:
@purpleidea I made the changes. I am confident in this code. So if you ack you can merge (and squash with the last commit if you don't mind) I still want to discuss the obj.Running() race so I have a better handle on how it affects this resource and future ones, but I updated the comment to reflect the fact that it isn't correct.

resources/aws_ec2.go
```diff
@@ -555,6 +489,93 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	switch aws.StringValue(instance.State.Name) {
+	case ec2.InstanceStateNameRunning, ec2.InstanceStateNameStopping:
+		err = c.WaitUntilInstanceStoppedWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameStopped, ec2.InstanceStateNamePending:
+		err = c.WaitUntilInstanceRunningWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameTerminated:
+		waitInput.Filters = append(waitInput.Filters, existWaiterFilter)
+		err = c.WaitUntilInstanceExistsWithContext(ctx, waitInput)
+	default:
+		return "", errwrap.Wrapf(err, "unrecognized instance state")
```
purpleidea: what is the err var set to here?

resources/aws_ec2.go
```diff
@@ -381,12 +390,24 @@ func (obj *AwsEc2Res) Watch() error {
...
+	// We tell the engine that we're running right away. This is not correct,
+	// but the api doesn't have a way to signal when the waiters are ready.
+	// Because watchWaiter checks the state directly, and the waiters will
+	// return as expected even if they miss the state change, we shouldn't miss
+	// any events.
 	if err := obj.Running(); err != nil {
 		return err
 	}
+
+	// canellable context used for exiting cleanly
```
purpleidea: typo, cannelloni

resources/aws_ec2.go
```diff
@@ -381,12 +390,24 @@ func (obj *AwsEc2Res) Watch() error {
 func (obj *AwsEc2Res) longpollWatch() error {
 	send := false
 	var exit *error
+
+	// We tell the engine that we're running right away. This is not correct,
+	// but the api doesn't have a way to signal when the waiters are ready.
+	// Because watchWaiter checks the state directly, and the waiters will
```
purpleidea: i think these last 3 lines are incorrect and we should drop them

resources/aws_ec2.go
```diff
@@ -555,6 +489,93 @@ func (obj *AwsEc2Res) longpollWatch() error {
...
+	switch aws.StringValue(instance.State.Name) {
+	case ec2.InstanceStateNameRunning, ec2.InstanceStateNameStopping:
+		err = c.WaitUntilInstanceStoppedWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameStopped, ec2.InstanceStateNamePending:
+		err = c.WaitUntilInstanceRunningWithContext(ctx, waitInput)
+	case ec2.InstanceStateNameTerminated:
+		waitInput.Filters = append(waitInput.Filters, existWaiterFilter)
+		err = c.WaitUntilInstanceExistsWithContext(ctx, waitInput)
+	default:
+		return "", fmt.Errorf("unrecognized instance state")
```
purpleidea: it's customary to sometimes say which one, with a %s at the end, eg:

fmt.Errorf("unrecognized instance state: %s", blah)

or %+v if it's a weird value that's not a string

jonathangold:
@purpleidea Fixed. :D

purpleidea:
This is now YOLO merged :)

(I didn't review it 100% but I'll leave it to you!)

Thanks!

Issue #245, opened 2017-11-10
Does aws:ec2 always close properly?


purpleidea:
I think I found a bug in current git master aws:ec2 code. This line:

https://github.com/purpleidea/mgmt/blob/c8ddbeaa5cb894909e22379eaca326b7d8336aa8/resources/aws_ec2.go#L436

with `close(obj.closeChan)` -- I don't know why I didn't notice earlier, or maybe I'm sleepy and it doesn't matter, but there are situations where we exit `Watch` without calling this. So I think this is a bug which could cause a deadlock.

What do you think @jonathangold ?

purpleidea:
https://github.com/purpleidea/mgmt/pull/246

jonathangold:
#288

purpleidea:
Stale, closing.

PR #288, opened 2017-12-21
resources: aws: ec2: Fix closed channel error in snsWatch and longpollWatch

jonathangold:
Found another mistake.

### DIFF
--- a/resources/aws_ec2.go
+++ b/resources/aws_ec2.go
@@ -469,7 +469,7 @@ func (obj *AwsEc2Res) longpollWatch() error {
 			}
 		case msg, ok := <-obj.awsChan:
 			if !ok {
-				return fmt.Errorf("channel closed unexpectedly")
+				return nil
 			}
 			if err := msg.err; err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


jonathangold:
#245

resources/aws_ec2.go
```diff
@@ -650,7 +650,7 @@ func (obj *AwsEc2Res) snsWatch() error {
 			}
 		case msg, ok := <-obj.awsChan:
 			if !ok {
-				return *exit
```
purpleidea: it's true that this is a bug, but why is it an error instead of a nil return?

same question for the longPollWatch too...

jonathangold: I could ignore the channel close, but because the channel is only supposed to be closed when the sender returns, this case should never happen. IOW, this will only come up if something went wrong and the channel was closed early, possibly leading to missed events. Is that not a good enough reason to return an error?

jonathangold: @purpleidea I now see the error of my ways. :P Fixed now.

purpleidea: Errors are ok. I make many of them.

Issue #28, opened 2016-05-05
create a network resource


purpleidea:
Managing a network interface is a fundamental part of managing any system. We should provide a native interface to do so. Some ideas:

* It might be worth basing this on systemd-networkd. (It looks like a safe, stable base!)
* Since mgmt works in a decentralized manner, so transient network "blips" shouldn't disrupt it.
* During CheckApply() if we don't transition to an acceptable state, we could revert and error.
* With the powerful Resource (Res) API, other resources will be able to add [automatic edges](https://ttboj.wordpress.com/2016/03/14/automatic-edges-in-mgmt/) to this network resource in case they require connectivity. Yet another reason this is an important primitive to write.
* I'd focus on getting simple networking working initially, with support for bridges or other systemd-networkd features as an exciting second step.

Patches welcome! Please feel free to make yourself known if you'd like to work on this

mikefaille:
Maybe, we can simply build plugin system as Docker's plugins or Ansible module. The goal is to provide way to implement many solutions. 

At least, we can enforce common api for all systems. And, network management solutions can exist in many ways like : 
- systemd-networkd
- network-manager (used for desktop)
- systems without network manager activated for one specific interface. In this case, we can directly implement netlink's kernel interface used by libcontainer --> https://github.com/mikefaille/tenus/blob/master/macvtap_linux.go
- etc

purpleidea:
@mikefaille I don't think we need a plugin system, at least not until we have more core base resources.

> systemd-networkd

Indeed this is mentioned in the original description.

> systems without network manager activated for one specific interface

Does this code support the interfaces to generating events when something changes? I expect it will be best to wrap networkd for now.

These issue tickets are open to attract someone to want to work on the code. Since you're commenting, are you volunteering? :)

LrsK:
As I was prodded by @purpleidea, I'm taking a stab at this! So far these are my thoughts on how the resource should work.

The resource controls the systemd-networkd configuration, which takes place via configuration files that are placed in the directory /etc/systemd/network. Networks are configured in files with extension ".network", and virtual network devices are configured in files with extension ".netdev". So the mgmt resource will describe and enforce one or more network configurations. Each configuration ends up in a file e.g.  `/etc/systemd/network/10-static_ip.network`

[...]

PR #289, opened 2017-12-22
resources: net: Add net resource

jonathangold:
#28


jonathangold:
@purpleidea Still needs tests (but the code was written with testing in mind,) and I want to chat with you about whether or not to pull out the dbus properties eavesdropping code into a helper function it the signature `func dbusEavesdrop(dbusIface, path string) (chan *dbus.Signal, error) {`

resources/net.go
```diff
@@ -0,0 +1,665 @@
...
+	// check if the interface is up
+	ifaceUp := strings.Contains(iface.Flags.String(), "up")
+
+	// if the interface's state is correct, we're done
+	if obj.State == "down" && !ifaceUp {
+		return true, nil
+	}
+	if obj.State == "up" && ifaceUp {
+		// error if we lose the connection and softfail is false
+		if !obj.SoftFail && obj.oState == dbus.MakeVariant("no-carrier") {
```
jonathangold: This is a bug. If we haven't yet received an event from the NIC, obj.oState will be nil. I'll figure out an alternative.

resources/net.go
```diff
@@ -0,0 +1,671 @@
...
+	if obj.State == "down" && !ifaceUp {
+		return true, nil
+	}
+	// check iface addrs
+	addrs, err := iface.Addrs()
+	if err != nil {
+		return false, errwrap.Wrapf(err, "error getting addrs")
+	}
+	if obj.State == "up" && ifaceUp {
+		// error if we lose the connection and softfail is false
```
jonathangold: I'd like to discuss this with you at some point, maybe before you review. This doesn't really work like it should. Instead I should be getting the `/org/freedesktop/network1/link/$foo` `OperationalState` property so I can detect when it's `no-carrier` to know when we've lost connectivity. But, I can't figure out how to get the property with the dbus api (need to input a systemd unit, nothing I've tried works, but I can get it in d-feet)

resources/net.go
```diff
@@ -0,0 +1,671 @@
...
+		return errwrap.Wrapf(err, "error getting iface dbus path")
+	}
+
+	// create a private message bus for watching the network device
+	bus, err := util.SystemBusPrivateUsable()
+	if err != nil {
+		return errwrap.Wrapf(err, "failed to connect to bus")
+	}
+	defer bus.Close()
+	// Add a DBus rule to watch the "PropertiesChanged" signal and get messages
```
jonathangold: I tried to put the eavesdropping stuff in a function. but I was having trouble handling the bus.Close() and associated operations without closing stuff too early/not closing at all. I'd like to try again, for the sake of unit testing and style, but I need to re-evaluate my approach.

resources/net.go
jonathangold: screenshot from 2017-12-23 08-48-33

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+	if obj.State == IfaceUp {
+		if obj.OpState == dbus.MakeVariant(OpStateRoutable) {
+			return true, nil
+		}
+		// If the connection is lost and timeout is set, retry once a second
+		// until the timeout expires. If the connection comes back, continue
+		// as if nothing happened. Error if the timeout expires (or isn't set.)
+		if obj.OpState == dbus.MakeVariant(OpStateNoCarrier) {
+			for i := 0; i < int(obj.Timeout); i++ {
```
jonathangold: I need some mentoring on making this exit cleanly. I'll try to figure it out in the meantime.

examples/net1.yaml
```diff
@@ -0,0 +1,9 @@
+---
+graph: mygraph
```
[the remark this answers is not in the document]
purpleidea: Yeah :P -- There is some good news:

The consistently named devices *are* actually useful for automation when you have more than one network card. But annoying when you just have *one*. The good news: we could easily add a function in the language that returns the eth device name if there's only one :) -- More on that in February :)

For this patch, let's just name them all `eth0` so users aren't confused about what they are. This will also work best in vms.

examples/net1.yaml
```diff
@@ -0,0 +1,10 @@
...
+resources:
+  net:
+  - name: enp4s0
+    state: up
+    ipv4: 192.168.2.55/24
```
purpleidea: nit: how about: 192.168.42.13/24
and 192.168.42.1 please? ;)

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	// NetworkdUnitFileDir is the location of networkd unit files.
+	NetworkdUnitFileDir = "/etc/systemd/network/"
+	// NetworkdUnitFileExt is the file extension for networkd unit files.
+	NetworkdUnitFileExt = ".network"
+	// NetworkdService is the name of the networkd service.
+	NetworkdService = "systemd-networkd.service"
+
+	// ActiveState is the systemd-networkd service's active state.
+	ActiveState = "ActiveState"
+	// ActiveStateActive is the active service state.
```
purpleidea: These and a few other comments on these consts are in the form: `foo is the foo of the whatever`. While I've certainly been guilt of this kind of thing before, if you could update a few of them so that they're a little bit more self descriptive for those that don't know what the internal networkd API is all about. Remember, these comments get displayed in godoc.

The fact that you have all these const's is probably good! Let's just make them useful in the docs if someone would ever need them, or private if they should never be used externally.

Why should they be public or private? If a third party wants to predict what we might be doing, knowing the prefix (eg `mgmt-`) is something we need public. Also if a consumer of this resource when using mgmt as a lib wants to know something that it should be able to gleam from this, that's useful too.

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+)
+
+func init() {
+	RegisterResource("net", func() Res { return &NetRes{} })
+}
+
+// NetRes is an systemd-networkd resource.
+type NetRes struct {
+	BaseRes `yaml:",inline"`
+	State   string        `yaml:"state"`   // up or down
```
purpleidea: Should we allow this to be empty so that it doesn't change the state? (Eg: maybe it only send's the IPV4 address if one exists?)

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	IPv4    string        `yaml:"ipv4"`    // ipv4 address in cidr format
+	IPv6    string        `yaml:"ipv6"`    // ipv6 address in cidr format
+	Gateway string        `yaml:"gateway"` // gateway address
+	Timeout time.Duration `yaml:"timeout"` // connection timeout in seconds
```
purpleidea: Could you better document what this does for the godoc string please? A multi-line above it might be appropriate.

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+
+	// read-only parameters
+	ActualIPv4 string // the interface's actual ipv4 address
```
purpleidea: Are these values that get populated to `send` ? If so, there's no problem re-using the IPv4 param. You either set it, or it gets set :)

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	ActualIPv6 string // the interface's actual ipv6 address
+
+	OpState dbus.Variant // the network link's operational state
```
purpleidea: What are these three, and why are they part of our public "API" ? Maybe docstrings are needed.

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *NetRes) Default() Res {
+	return &NetRes{
+		BaseRes: BaseRes{
+			MetaParams: DefaultMetaParams, // force a default
+		},
+		State:   IfaceUp,
+		Timeout: time.Duration(10),
```
purpleidea: Where did this timeout come from? If it's arbitrary it should be a const :/

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+
+// Validate if the params passed in are valid data.
+func (obj *NetRes) Validate() error {
+	// validate state
+	if obj.State != IfaceUp && obj.State != IfaceDown {
+		return fmt.Errorf("invalid state: %s", obj.State)
+	}
+
+	// validate network address input
+	if obj.IPv4 != "" {
```
purpleidea: I just realized that it seems the `net` package doesn't have a strong distinction between IPv4 and IPv6 (it just has an `IP` type). Do we need to add a distinction? Can we just have `obj.IP` ?

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+
+func init() {
+	RegisterResource("net", func() Res { return &NetRes{} })
+}
+
+// NetRes is an systemd-networkd resource.
+type NetRes struct {
+	BaseRes `yaml:",inline"`
+	State   string        `yaml:"state"`   // up or down
+	IPv4    string        `yaml:"ipv4"`    // ipv4 address in cidr format
```
purpleidea: Network interfaces can have a list of IP addresses. Don't change this right away, but let's discuss. Shouldn't we accept a list instead? In particular, what do we do if we only wanted mgmt to manage one IP address, but leave the others alone? Or do we only let it manage the whole thing or nothing?

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+
+	*obj = NetRes(raw) // restore from indirection with type conversion!
+	return nil
+}
+
+// Close cleans up some resources so we can exit cleanly.
+func (obj *NetRes) Close() error {
+	// close the dbus connection
+	obj.sdbus.Close()
```
purpleidea: What about these errors that are returned...

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	if obj.Timeout != res.Timeout {
+		return false
+	}
+
+	return true
+}
+
+// UnmarshalYAML is the custom unmarshal handler for this struct.
+// It is primarily useful for setting the defaults.
+func (obj *NetRes) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
purpleidea: Please move this to be after the Close() method.

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	// close the message bus
+	obj.gdbus.Close()
+
+	return obj.BaseRes.Close()
+}
+
+// unitFileContents builds the unit file contents from the input
+func unitFileContents(name, ipv4, ipv6, gateway string) []byte {
+	// build the unit file contents
+	u := "[Match]\n"
```
purpleidea: Instead of adding the `\n` everywhere, how about we build up a list and join with \n?

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+
+	// Add a DBus rule to watch the "PropertiesChanged" signal and get messages
+	// from the network link's dbus path.
+	args := fmt.Sprintf(DBusEavesdropSprint, DBusPropertiesInterface, obj.IfacePath)
+	call := obj.gdbus.BusObject().Call(DBusAddMatch, 0, args)
+	if err := call.Err; err != nil {
+		return errwrap.Wrapf(err, "error creating dbus call")
+	}
+
+	// TODO: verify that implementation doesn't deadlock if there are unread
```
purpleidea: This is a big TODO. If it's a serious thing, change this to a `XXX:` or a `FIXME:` if it's less serious.

Or we can just fix it :)

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+
+	var send = false
+	var exit *error
+
+	for {
+		select {
+		case event, ok := <-ifaceChan:
+			if !ok {
+				return fmt.Errorf("channel closed unexpectedly")
```
purpleidea: In general it's always nice if we can cause the whole "house" to come down, by asking the sender of this to shutdown, we process all it's events until it closes, and then when we get that close we shutdown too.

When we're receiving events from two sources (file and dbus) we ask them *both* to close, and when each closes, we do this:

```
if !ok {
    if someFlag {
        return nil
    }
    someFlag = true
    continue
}
```

This works for two senders. When the 2nd of the two shuts down, we exit :)

(For N senders, you need a `map` and you have to check to see if everyone's done each time someone closes. Remember this is the number of senders which we have a way to actually ask to exit! And if you're not expecting one to close ever then you error of course!)

resources/net.go
purpleidea: Happy to-- let's discuss why we need a timeout at all... I think explaining this in the struct will be the best starting point, although yes, I agree having a blocking `time.Sleep` in the code is not ideal.

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+
+	if !apply {
+		return false, nil
+	}
+
+	log.Printf("%s: ifaceCheckApply(%t)", obj, apply)
+
+	// enable or disable the interface
+	cmd := exec.Command("ip", "link", "set", obj.GetName(), obj.State)
```
purpleidea: Is this the only way to do this?

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+
+	// make sure we run the command as root
```
purpleidea: If mgmt is not running as root, does this error appropriately? Is it possible that something like CAP_SYS_ADMIN or similar would give us this without root?

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+func (obj *NetRes) fileCheckApply(apply bool) (checkOK bool, _ error) {
+	// check if the unit file exists
+	_, err := os.Stat(obj.UnitFilePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return false, errwrap.Wrapf(err, "error checking file")
+		}
+	}
+	// build the unit file contents from the definition
+	contents := unitFileContents(obj.GetName(), obj.IPv4, obj.IPv6, obj.Gateway)
```
purpleidea: It's probably okay to just grab these values off the struct directly and have `unitFileContents` be defined as part of the struct, so you don't have to pass them in.

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	log.Printf("%s: fileCheckApply(%t)", obj, apply)
+
+	// write the file
+	if err := ioutil.WriteFile(obj.UnitFilePath, contents, 0644); err != nil {
+		return false, errwrap.Wrapf(err, "error writing configuration file")
+	}
+	return false, nil
+}
+
+// ipCheckApply checks the network device's ip addresses and performs the
```
purpleidea: What?

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	// get the networkd service state
+	svcState, err := obj.sdbus.GetUnitProperty(NetworkdService, ActiveState)
+	if err != nil {
+		return false, errwrap.Wrapf(err, "failed to get svc unit properties")
+	}
+	// if the service is restarting, there's nothing to do
+	if svcState.Value != dbus.MakeVariant(ActiveStateReloading) {
+		return false, nil
+	}
+	// restart networkd to re-initialize the interface
```
purpleidea: Is this what needs doing? Is there a `reload` instead?

resources/net.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+
+	// check the interface's ip addresses
+	addrs, err := iface.Addrs()
+	if err != nil {
+		return "", "", errwrap.Wrapf(err, "error getting addrs")
+	}
+	for _, addr := range addrs {
+		// ipv4
+		if strings.Contains(addr.String(), ".") {
```
purpleidea: Cute-- Is this the way to check for ipv4 vs. v6?

In any case, as discussed if IP is an array, this just becomes easier, and we don't need to distinguish :)
If need be someone could write an `isIPv4` func for the language.

jonathangold:
@purpleidea No worries re: number of comments. :) Maybe we can hangout this weekend to nail down the unknowns. My pace is a little slower than usual this week, so no rush.

resources/net.go
jonathangold: My first pass used this. https://golang.org/pkg/net/#IP.To4 but ACK on the array (slice?)

purpleidea: Oh! Cool--- Why not use that?

purpleidea:
@jonathangold I happened to notice that you updated this and it passed... Is it ready for another review and possible merge, or are you still hacking on it?

jonathangold:
@purpleidea still a few things to iron out. I should be able to finish it tomorrow.

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+func init() {
+	RegisterResource("net", func() Res { return &NetRes{} })
+}
+
+// NetRes is an systemd-networkd resource.
+type NetRes struct {
+	BaseRes `yaml:",inline"`
+	State   string   `yaml:"state"`   // up or down or empty
+	DHCP    bool     `yaml:"dhcp"`    // enable dhcp
+	Addrs   []string `yaml:"addrs"`   // list of addresses in cidr format
```
purpleidea: Public API in struct looks perfect.

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+
+	return obj.BaseRes.Validate()
+}
+
+// Init runs some startup code for this resource.
+func (obj *NetRes) Init() error {
+	var err error
+	// get the network link's dbus path
+	obj.ifacePath, err = dbusIfacePath(obj.GetName())
+	if err != nil {
```
purpleidea: nit, but fwiw when you're doing an `=` instead of an `:=` you can prefix this, eg:

```
if obj.ifacePath, err = dbusIfacePath(obj.GetName()); err != nil {
 ...
}

```

In general I like to do those as much as possible, unless the line is extremely long.

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+
+	// create a dbus subscription to watch the networkd service unit
+	set := obj.sdbus.NewSubscriptionSet()
+	svcChan, svcErr := set.Subscribe() // svc event channel
+	set.Add(networkdService)
+
+	// Add a DBus rule to watch the "PropertiesChanged" signal and get messages
+	// from the network link's dbus path.
+	args := fmt.Sprintf(dBusEavesdropSprint, dBusPropertiesInterface, obj.ifacePath)
+	call := obj.gdbus.BusObject().Call(dBusAddMatch, 0, args)
```
purpleidea: same as above if call is not used elsewhere. inline with `if`

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+
+	// create a dbus subscription to watch the networkd service unit
+	set := obj.sdbus.NewSubscriptionSet()
+	svcChan, svcErr := set.Subscribe() // svc event channel
+	set.Add(networkdService)
+
+	// Add a DBus rule to watch the "PropertiesChanged" signal and get messages
+	// from the network link's dbus path.
+	args := fmt.Sprintf(dBusEavesdropSprint, dBusPropertiesInterface, obj.ifacePath)
```
purpleidea: missing format string. Otherwise this is just a + b + c

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+	if err := obj.Running(); err != nil {
+		return err // bubble up a NACK...
+	}
+
+	var send = false
+	var exit *error
+
+	// closeMap keeps track of which event sources have closed their channels
+	closeMap := make(map[int]bool)
+	for i := 0; i < eventSources; i++ {
```
purpleidea: the "last one out switches off the lights" pattern is nice, however the static event sources kills it. If anything, use a map.

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+
+	if !apply {
+		return false, nil
+	}
+
+	log.Printf("%s: ifaceCheckApply(%t)", obj, apply)
+
+	// enable or disable the interface
+	cmd := exec.Command("ip", "link", "set", obj.GetName(), obj.State)
```
purpleidea: Is this the only way to do this?

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+	// capture any error messages
+	slurp, err := ioutil.ReadAll(stderr)
+	if err != nil {
+		return false, errwrap.Wrapf(err, "error slurping error message")
+	}
+	// wait until cmd exits and return error message if any
+	if err := cmd.Wait(); err != nil {
+		return false, errwrap.Wrapf(err, "%s", slurp)
+	}
+	// XXX: wait for state?
```
purpleidea: did you find out about the daemon restart stuff?

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+
+	return false, nil
+}
+
+// fileCheckApply checks and maintains the systemd-networkd unit file contents.
+// TODO: check for conflicting unit files and think about unit priority
+func (obj *NetRes) fileCheckApply(apply bool) (checkOK bool, _ error) {
```
purpleidea: just have the return be (bool, error) instead of an unused arg and a `_`

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+	// check if the unit file exists
+	_, err := os.Stat(obj.unitFilePath)
+	if err != nil {
```
purpleidea: && instead of nesting

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+		if !os.IsNotExist(err) {
+			return false, errwrap.Wrapf(err, "error checking file")
+		}
+	}
+	// build the unit file contents from the definition
+	contents := obj.unitFileContents()
+	// check the file contents
+	if err == nil {
```
purpleidea: I think you can do a err != nil above, and then inline this? please look into it

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+		if !ip.IsLinkLocalUnicast() {
+			addrStrings = append(addrStrings, addr.String())
+		}
+	}
+
+	// compare defined addrs with actual addrs
+	objAddrs := obj.Addrs
+	sort.Strings(objAddrs)
+	sort.Strings(addrStrings)
+	if len(objAddrs) == len(addrStrings) {
```
purpleidea: this sort of deep nesting either needs inline-ing, or made into a helper function please

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+		return false, nil
+	}
+
+	log.Printf("%s: ipCheckApply(%t)", obj, apply)
+
+	// restart networkd to re-initialize the interface
+	if err := restartSystemdUnit(obj.sdbus, networkdService, dBusFailMode); err != nil {
+		return false, errwrap.Wrapf(err, "error restarting networkd service")
+	}
+
```
purpleidea: do we need to wait here? what happens with the restartSystemdUnit stuff?
please include your findings about how it works as a comment :)

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+		return false
+	}
+	if (obj.Addrs == nil) != (res.Addrs == nil) {
+		return false
+	}
+	if obj.Addrs != nil && res.Addrs != nil {
+		if len(obj.Addrs) != len(res.Addrs) {
+			return false
+		}
+		objAddrs := obj.Addrs
```
purpleidea: If we had this compare as a helper func, then it could be re-used here :)

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+		return false
+	}
+
+	return true
+}
+
+// Close cleans up some resources so we can exit cleanly.
+func (obj *NetRes) Close() error {
+	// close the dbus connection
+	obj.sdbus.Close()
```
purpleidea: What about these errors that are returned...

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+	// connect to systemd
+	connChan := make(chan string) // catch result information
+	defer close(connChan)
+
+	// restart unit
+	if _, err := conn.RestartUnit(unit, mode, connChan); err != nil {
+		return errwrap.Wrapf(err, "failed to restart unit")
+	}
+
+	// XXX: will this ever hang?
```
purpleidea: you tell me :)

unfortunately, sometimes these sorts of things are difficult to know and occasionally involve reading the upstream code a little. it's good practice, but don't dive too long. if you can't find it in 20 min, move on or ask someone.

resources/net.go
jonathangold: replaced with netlink

resources/net.go
jonathangold: Yes. It may "reload" the unit file, but the ip doesn't change until the service is restarted. tried `systemctl daemon-reload` and dbus.Conn.Reload() but the ip does not change without restarting the service.

resources/net.go
```diff
@@ -0,0 +1,804 @@
...
+
+	// adminState is the name of the dbus property for the network interface's
+	// "administrative state." It is used and to check if the interface is
+	// configured or in a transitional state.
+	adminState = "AdministrativeState"
+	// adminStateConfiguring is the administrative state of the interface when
+	// it is in the process of configuring.
+	adminStateConfiguring = "configuring"
+
+	// eventSources is the number of event sources in watch.
```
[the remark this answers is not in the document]
jonathangold: @purpleidea I could use some guidance on keeping track of which channels are open/closed with a map. Even just a hint like the types in the map I should be using would be helpful. ;)

jonathangold:
Closing this since #371 is merged. I'll keep the branch in my repo for networkd reference.

PR #290, opened 2017-12-25
puppet: avoid empty parameters to puppet mgmtgraph

ffrank:
This solves an issue first observed with golang 1.8.

Creating an exec.Command with an empty string parameter (when no puppet.conf
file is specified) would lead to an error from Puppet, stating that an
unexpected argument was passed to "puppet mgmtgraph print".

The workaround is to not include *any* positional argument (not even the
empty string) when --puppet-conf is not used.


ffrank:
*cough cough*

On the first day of Christmas
ffrank gave to meeeee
...
a patch fo-hor emm-gee-emm teeeeeee.

Thank you, thank you, I'll be here all week.

purpleidea:
Patch LGTM :)

Thank you! Happy holidays... Your song makes me think you need to write 24? more patches...

ffrank:
I think it's twelve, and the other 11 would be not patches to mgmt. Nice try though ;)

purpleidea:
Merged!

@ffrank is back!

PR #292, opened 2018-01-11
test: Don't fail on tag builds

purpleidea:
This seems to be causing our failures with:

$ git fetch --unshallow
fatal: Couldn't find remote ref refs/heads/0.0.14

Hopefully this doesn't break the other use case we added this patch for!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ sudo: true
 dist: trusty
 before_install:
   - sudo apt update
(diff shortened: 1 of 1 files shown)


purpleidea:
I imagine to test:

1) I now have to add a new tag on top of this.
2) Someone has to make a pull-request against a feature branch

Gah, travis you so difficult sometimes :/

purpleidea:
XXX: btw, do not merge, this patch was made with GH webui for testing purposes and commit contains their sig.

jonathangold:
@purpleidea LGTM. Fetching all refs should be ok. IOW, we'll still get whatever branch we need (in addition to the ones we don't, but any penalty should be negligible.)

PR #293, opened 2018-01-12
misc: Don't install packages that are already installed

joejulian:
pacman needs `--needed` to prevent reinstalling packages that are
already installed. Without it, pacman will reinstall packages that are already installed. With this switch, packages that are already installed won't be reinstalled.

Additionally I added `--asdeps` to allow later cleanup of unneeded dependencies. This marks the installed packages as dependencies so they can be cleaned up via pacman when cleaning up unused dependencies.

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -43,7 +43,7 @@ if [ ! -z "$BREW" ]; then
 fi
 
 if [ ! -z "$PACMAN" ]; then
-	$sudo_command $PACMAN -S --noconfirm libvirt augeas libpcap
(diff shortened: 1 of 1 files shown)


joejulian:
Edited the commit message. Hope that makes it more clear.

purpleidea:
@joejulian Unfortunately you have to actually edit the commit message in git and force push to the same branch. Editing the title doesn't have any effect. Sorry!

purpleidea:
Thank you! rebased and merged in 1c8c0b2915416bd64f256e0aabf29918605787b8

PR #294, opened 2018-01-12
misc: Use bash's `command` instead of `which`

joejulian:
Bash has a built-in command, `command`, that will search the path and
return the full path to a command if it exists (or an exit code of 1 if
it does not), preventing the requirement of the `which` package.

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -9,13 +9,13 @@ if env | grep -q '^TRAVIS=true$'; then
 	travis=1
 fi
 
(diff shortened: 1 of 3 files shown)


joejulian:
I ran across this elsewhere where `which` did not behave consistently between distros.

purpleidea:
Gah! Sorry I did not see your update patch sooner. GH does not notify me when you push a new commit.

In any case, your great patch is now MERGED :)

Thank you! bee1aa00f1d3296eb15a056a463a689427dec801

Issue #303, opened 2018-02-07
macOS: Build error `undefined: syscall.Sysinfo_t` since lang feature


aequitas:
## Versions:

* mgmt version (eg: `mgmt --version`): N/A

* operating system/distribution (eg: `uname -a`): Darwin Firefly.local 16.7.0 Darwin Kernel Version 16.7.0: Mon Nov 13 21:56:25 PST 2017; root:xnu-3789.72.11~1/RELEASE_X86_64 x86_64

* golang version (eg: `go version`): go version go1.9.3 darwin/amd64

## Description:

I don't know if macOS is currently supported for lang but on the latest master (eb33a5a5df614d06d5f409106c767d32f5c37937) I get this during build:

```
~/.g/s/g/p/mgmt (master|✔) $ env GOTAGS='noaugeas novirt' make build
Generating: bindata...
Generating: lang...
/Applications/Xcode.app/Contents/Developer/usr/bin/make --quiet -C lang
Building: mgmt, version: 0.0.14-6-geb33a5a...
time go build -i -ldflags "-X main.program=mgmt -X main.version=0.0.14-6-geb33a5a" -o mgmt -tags 'noaugeas novirt';
# github.com/purpleidea/mgmt/lang/funcs/facts/core
lang/funcs/facts/core/loadfact.go:123:14: undefined: syscall.Sysinfo_t
lang/funcs/facts/core/loadfact.go:124:11: undefined: syscall.Sysinfo

real	0m2.755s
user	0m5.688s
sys	0m0.858s
make: *** [mgmt] Error 2
```

Building an older version (1c8c0b2915416bd64f256e0aabf29918605787b8) does seem to work: 

```
~/.g/s/g/p/mgmt ((1c8c0b29…)|✚1…3) $ env GOTAGS='noaugeas novirt' make build
Building: mgmt, version: 0.0.14-3-g1c8c0b2-dirty...
time go build -i -ldflags "-X main.program=mgmt -X main.version=0.0.14-3-g1c8c0b2-dirty" -o mgmt -tags 'noaugeas novirt';

real	0m11.443s
user	0m8.653s
sys	0m5.068s
```

purpleidea:
Re your question, please see: 
https://github.com/purpleidea/mgmt/blob/master/docs/faq.md#does-this-support-windows-osx-gnu-hurd

Happy to merge your patch, discussion in PR...

aequitas:
[...]

PR #296, opened 2018-01-28
docs: Add docs for docker usage

karmab:
the corresponding docker image lives at docker.io/karmab/mgmt:0.0.14

Issue #33, opened 2016-05-14
File resource on OS X fails


tuxmea:
When managing a file which is not yet available:
- on Linux watcher.Add(current) returns no Error.
- on OS X it returns (*os.PathError)

https://github.com/purpleidea/mgmt/blob/master/file.go#L142

purpleidea:
Assign -> @tuxmea ? ;)

tuxmea:
reported upstream: https://github.com/fsnotify/fsnotify/issues/144

purpleidea:
I think just matching the different error would be fine... I don't know enough about OSX to know if it's a bug.

daenney:
Is this gone? I can't find the relevant code path anymore and I see no references to `watcher.Add` in [file.go](/purpleidea/mgmt/blob/master/resources/file.go). Or does this now apply to the [`Configwatcher.Add()`](/purpleidea/mgmt/blob/master/recwatch/configwatch.go)?

purpleidea:
On Mon, Feb 6, 2017 at 9:13 AM, Daniele Sluijters
<notifications@github.com> wrote:
>
> Is this gone? I can't find the relevant code path anymore and I see no references to watcher.Add in file.go. Or does this now apply to the Configwatcher.Add()?



This code is now in the recwatch/ package. Improvement welcome, but
I'm happy to close this.

purpleidea:
Since nobody is actively stepping up to work on OSX support, and it's not a goal to get us to a minimally viable tool, I'm going to close this for now. Happy to accept OSX patches though.

purpleidea:
One more note: Mgmt would be expected to have this issue on an OSX server (not so common/useful) this shouldn't cause an issue for someone using an OSX machine to talk to their servers.

purpleidea:
Merged in 9544ab2e02418398cb178253b3e05ae44be71374

PR #304, opened 2018-02-07
language: Use universal way to retrieve loadavg.

aequitas:
Sysinfo is not supported on macOS and results in a build error. This PR changes the load fact to use a universal method of retrieving load average that should work cross platform.

I don't know what the policy is on excluding external packages into the project with regards to reducing bloat, etc. I would have no problem researching an alternative approach if needed.

This PR fixes my previously posted issue (https://github.com/purpleidea/mgmt/issues/303) and it seems to be the only blocking issue for building/running mgmt on macOS for me.

Only tested on macOS so far:
```
~/.g/s/g/p/mgmt (gopsutil_load|✔) $ uptime; ./mgmt run --tmp-prefix --converged-timeout=1 --lang examples/lang/load0.mcl 2>&1 | grep Msg:\ load
17:09  up 7 days,  5:42, 4 users, load averages: 1.76 1.69 1.82
17:09:12 print.go:89: print[print1]: Msg: load average: 1.86, 1.72, 1.83
```

### DIFF
--- a/docs/quick-start-guide.md
+++ b/docs/quick-start-guide.md
@@ -14,6 +14,7 @@ Once you're familiar with the general idea, please start hacking...
 * You need golang version 1.8 or greater installed.
 ** To install on rpm style systems: `sudo dnf install golang`
 ** To install on apt style systems: `sudo apt install golang`
+** To install on macOS systems install [Homebrew](https://brew.sh) and run: `brew install go`
 * You can run `go version` to check the golang version.
 * If your distro is tool old, you may need to [download](https://golang.org/dl/) a newer golang version.
 
@@ -97,3 +98,43 @@ At the moment we have:
 * [Arch](https://aur.archlinux.org/packages/mgmt/)
 
 Please contribute more! We'd especially like to see a Debian package!
+
+## OSX/macOS/Darwin development
+Developing and running `mgmt` on macOS is currently not supported (but not discouraged either). Meaning it might work but in the case it doesn't you would have to provide your own patches to fix problems (the project maintainer and community are glad to assist where needed).
+
+There are currently some issues that make `mgmt` less suitable to run for provisioning macOS (eg: https://github.com/purpleidea/mgmt/issues/33). But as a client to provision remote servers it should run fine.
+
+Since the primary supported systems are Linux and these are the environments tested for it is wise to run these suites during macOS development as well. To ease this Docker can be levaraged ((Docker for Mac)[https://docs.docker.com/docker-for-mac/]).
+
+Before running any of the commands below create the development Docker image:
+
+```
+docker/scripts/build-development
+```
+
+This image requires updating every time dependencies (`make-deps.sh`) change.
+
+Then to run the test suite:
+
+```
+docker run --rm -ti \
+  -v $PWD:/go/src/github.com/purpleidea/mgmt/ \
+  -w /go/src/github.com/purpleidea/mgmt/ \
+  purpleidea/mgmt:development \
+  make test
+```
+
+For convenience this command is wrapped in `docker/scripts/exec-development`.
+
+Basically any command can be executed this way. Because the repository source is mounted into the Docker container invocation will be quick and allow rapid testing, example:
+
+```
+docker/scripts/exec-development test/test-shell.sh load0.sh
+```
+
+Other examples:
+
+```
+docker/scripts/exec-development make build
+docker/scripts/exec-development ./mgmt run --tmp-prefix --lang examples/lang/load0.mcl
+```
--- a/lang/funcs/facts/core/load_posix.go
+++ b/lang/funcs/facts/core/load_posix.go
@@ -1,0 +1,45 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// +build !darwin
+
+package core // TODO: should this be in its own individual package?
+
+import (
+	"syscall"
+)
+
+const (
+	// LoadScale factor scales the output from sysinfo to the correct float
+	// value.
+	LoadScale = 65536 // XXX: is this correct or should it be 65535?
+)
+
+// load returns the system load averages for the last minute, five minutes and
+// fifteen minutes. Calling this more often than once every five seconds seems
+// to be unnecessary, since the kernel only updates these values that often.
+// TODO: is the kernel update interval configurable?
+func load() (one, five, fifteen float64, err error) {
+	var sysinfo syscall.Sysinfo_t
+	if err = syscall.Sysinfo(&sysinfo); err != nil {
+		return
+	}
+	one = float64(sysinfo.Loads[0]) / LoadScale
+	five = float64(sysinfo.Loads[1]) / LoadScale
+	fifteen = float64(sysinfo.Loads[2]) / LoadScale
+	return
+}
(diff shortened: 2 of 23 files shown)


purpleidea:
@aequitas Thanks for the patch!

Two questions:

1) Did we lose a lack of precision in the values after this patch?
2) Do you think this dep is too big, and is there an easy way instead? I'm not against pasting a small bit of code in if that's all that's needed.

LMK and thanks!

purpleidea:
ps: commit will need to start with a prefix to pass the CI. See: https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

aequitas:
1. I'll investigate.

2. I'm from the Not-Not-Invented-Here syndrome camp, so I always try to find a easy solution someone else has already created before spending the time myself. I don't know what your opinion is on that, another project I work on has a strict aversion for additional dependencies in the core, so I'm tasting the atmosphere for this project :) 

I didn't look at the size of the dependancy for that matter, it might be trivial. But I like to hear your opinion regarding this project.

aequitas:
I can say for sure now that precision will be lost as the gopsutil uses `/proc/loadavg` as source and it has a 2 digit precision compared to the sysinfo call that seems to have 9.

Pasting in the Darwin version of gopustil loadavg would require a bit more than just a few lines: https://github.com/shirou/gopsutil/blob/master/load/load_darwin.go#L19 But should be doable.

Is there currently any mechanism in place for this repo that supports switching implementations based on OS? I have to investigate how gopsutil does this but they have a file per (group of) OS and mark it's compatible platforms: https://github.com/shirou/gopsutil/blob/master/load/load_darwin.go#L1

I think it should be possible to keep the current implementation for platforms supporting syscall but use a different one for macOS but the two implementations will have different precision. I don't know how you'd feel about that. On a quick search I didn't find any way to retrieve higher precision load values for macOS.

purpleidea:
> Pasting in the Darwin version of gopustil loadavg would require a bit more than just a few lines

If you can pull in the relevant parts, I'd be inclined to prefer that if you don't think we'll need everything else from that lib.

> switching implementations based on OS?

Indeed, use these build tags:

https://golang.org/pkg/go/build/

If you have any other questions, please let me know! Thank you!

aequitas:
TIL: build tags, thanks, I'll get on it.

aequitas:
So it turns out you can call C from Go and `getloadavg` on macOS returns load with high precision. 

I still need to if this works for Linux as well so I convert this back to a single implementation. 

The testsuite regex might also need some adjustment as i've seen multiple precisions being returned by `getloadavg`. I think theoretically it might even return precision with 1 or 0 digits but I'd have to dive into some documentation about that.

I currently pushed my WIP, I'll be back tomorrow to fix things up.

aequitas:
I included a separate commit with macOS build environment and doc improvements. I could put this in a separate PR if you want.

One thing I'm unsure about is this change: https://github.com/purpleidea/mgmt/pull/304/files#diff-d59dc5f4cd8b5a47d20873651c6977d9R28

I don't know what kind of behaviour is expected by applying a repetition operator onto another repetition operator. So I'm curious what the intended effect was suppose to be. On macOS grep it throws a warning. 

I also added a new script to @czerasz Docker setup which allows quicker development by not having to copy the files into the container every time but just using a mount. There is another Docker effort going on (https://github.com/purpleidea/mgmt/pull/296) which has the aim for a release/runtime image as far as I can tell. Let me know if you want some changes to combine the efforts.

aequitas:
There also seems to be a bug in the `GetUID` API which only surfaces in tests on macOS. Might need another test with a explicitly created user (without group with matching ID) to assure against retention on Linux. https://github.com/purpleidea/mgmt/pull/304/commits/c2ae3c0a1ec5cf0ee129001c2e1b80044c0162a2

aequitas:
So this PR is getting out of hand quickly :). I marked some tests that currently fail on macOS due to https://github.com/purpleidea/mgmt/issues/33 to skip. Otherwise all velocity is lost in restoring macOS building. The good news, Travis build target for macOS is green now.

test/test-govet.sh
```diff
@@ -25,7 +25,7 @@ function simplify-gocase() {
 
 function token-coloncheck() {
 	# add quotes to avoid matching three X's
-	if grep -Ei "[\/]+[\/]+[ ]*+(FIXME[^:]|TODO[^:]|X"'X'"X[^:])" "$1"; then
```
purpleidea: What's this about?

misc/make-deps.sh
```diff
@@ -39,7 +39,8 @@ if [ ! -z "$APT" ]; then
 fi
 
 if [ ! -z "$BREW" ]; then
-	$BREW install libvirt || true
+	# coreutils contains gtimeout
+	$BREW install libvirt augeas coreutils|| true
```
purpleidea: i think a space before the `||` would be welcome, but i'd certainly merge without this, since it's very nitpicky

docs/quick-start-guide.md
```diff
@@ -97,3 +98,35 @@ At the moment we have:
...
+Before running any of the commands below create the development Docker image:
+
+    docker/scripts/build-development
+
+This image requires updating every time dependencies (`make-deps.sh`) change.
+
+Then to run the test suite:
+
+    docker run --rm -ti \
+      -v $PWD:/go/src/github.com/purpleidea/mgmt/ \
```
purpleidea: Does it make sense to remove all the space identation and surround this with three backticks on either side?

lang/funcs/facts/core/load_posix.go
```diff
@@ -0,0 +1,39 @@
...
+// load returns the system load averages for the last minute, five minutes and
+// fifteen minutes. Calling this more often than once every five seconds seems
+// to be unnecessary, since the kernel only updates these values that often.
+// TODO: is the kernel update interval configurable?
+func load() (one, five, fifteen float64, err error) {
+	var sysinfo syscall.Sysinfo_t
+	if err = syscall.Sysinfo(&sysinfo); err != nil {
+		return
+	}
+	one = float64(sysinfo.Loads[0]) / LoadScale
```
purpleidea: Shouldn't the `LoadScale` const get moved here too?

resources/util.go
```diff
@@ -248,7 +248,7 @@ func GetUID(username string) (int, error) {
 
 	userObj, err = user.Lookup(username)
 	if err == nil {
-		return strconv.Atoi(userObj.Gid)
+		return strconv.Atoi(userObj.Uid)
```
purpleidea: zomg, great find, thanks!

aequitas:
Thanks, I don't mind nitpicking now and then. I must say the work done by you and others on the project quality of the repo is awesome. Stuff like the `.editorconfig`, test suite, project guidelines and automation. I like working on this codebase.

I'll address the nitpicks.

purpleidea:
@aequitas It's already quite obvious from reviewing your commits that your code quality is excellent.
Thanks for working on our codebase :)

test/test-govet.sh
aequitas: As mentioned in an earlier comment: https://github.com/purpleidea/mgmt/pull/304#issuecomment-364084335

This is a fatal error for grep on OSX (not a warning as I stated above). And I know a lot of regex but don't know what a repetition operator applied to an repetition operator is suppose to do?

misc/make-deps.sh
aequitas: You are absolutely right, a space is more than welcome.

test/test-govet.sh
purpleidea: Yeah, I have no idea either :/
Assuming it still works as expected, then ACK :)

docs/quick-start-guide.md
aequitas: I'll stick with triple quotes as that seems to be the convention for this project.

lang/funcs/facts/core/load_posix.go
aequitas: Correct, but in all the excitement about the other stuff I forgot to do the thing I set out for. I think the method I implemented for darwin also works on other osses, and now with the test suite and docker going I can test it. I'll verify if it works and push a universal solution again if it does.

purpleidea: @aequitas Great! However, please keep the pure-golang version in for Linux, in case we ever hit any reasons why we would need that.

purpleidea:
LGTM :) Just waiting for CI, thanks!

purpleidea:
Patch rebased to master and MERGED :)

Thank you!
https://github.com/purpleidea/mgmt/commit/8fb0373f82ec2e2cd3f81e013e696824a51a18e0

PR #321, opened 2018-02-10
lang: Add edges to lexer and parser

purpleidea:
Here is my WIP branch for adding lexing and parsing for standalone edges and send/recv.
Some of my WIP hacking is present in the source but commented out.

Since I suck at yacc, I get this issue:

```
$ make lang
Generating: lang...
make --quiet -C lang

conflicts: 15 shift/reduce, 1 reduce/reduce
```

I think this *might* be due to the fact that the `foo["hey"] -> bar["blah"]` syntax might be confused for a `["list"]`, but I have no idea and I don't know how to debug yacc shift or reduce issues yet.

Help wanted!
/cc @ffrank

### DIFF
--- a/lang/parser.y
+++ b/lang/parser.y
@@ -65,6 +65,9 @@ func init() {
 
 	resFields []*StmtResField
 	resField  *StmtResField
+
+	edgeHalfList []*StmtEdgeHalf
+	edgeHalf     *StmtEdgeHalf
 }
 
 %token OPEN_CURLY CLOSE_CURLY
@@ -74,10 +77,10 @@ func init() {
 %token STRING BOOL INTEGER FLOAT
 %token EQUALS
 %token COMMA COLON SEMICOLON
-%token ROCKET
+%token ROCKET ARROW DOT
 %token STR_IDENTIFIER BOOL_IDENTIFIER INT_IDENTIFIER FLOAT_IDENTIFIER
 %token STRUCT_IDENTIFIER VARIANT_IDENTIFIER VAR_IDENTIFIER IDENTIFIER
-%token VAR_IDENTIFIER_HX
+%token VAR_IDENTIFIER_HX CAPITALIZED_IDENTIFIER
 %token COMMENT ERROR
 
 // precedence table
@@ -154,6 +157,11 @@ stmt:
 		$$.stmt = $1.stmt
 	}
 |	resource
+	{
+		posLast(yylex, yyDollar) // our pos
+		$$.stmt = $1.stmt
+	}
+|	edge
 	{
 		posLast(yylex, yyDollar) // our pos
 		$$.stmt = $1.stmt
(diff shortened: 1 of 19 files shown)


ffrank:
Starting to dig into the parser definition.

From some thinking I had done offline beforehand: What if we went the Puppet way of writing references capitalized like `File["a"] -> Svc["b"]` rather than `file["a"] -> svc["b"]`.

If we can teach the parser to expect lower case for regular old identifiers, and capitalized strings for identifiers in references, it should be possible to generate a single-pass parser.

I whipped up a PoC, but no dice. Will have to do some bisecting before I go on.

Some quick googling seems to indicate that there are alternative repos for goyacc that offer more debugging options, e.g. https://godoc.org/github.com/cznic/goyacc

Aside: I had been wondering why there was a colon in `aws:ec2`, and sure enough
```
/[a-z][a-z0-9]*/
                {
                        yylex.pos(lval) // our pos
                        lval.str = yylex.Text()
                        return IDENTIFIER
                }
```
...I don't think it's a valid resource type name. We *should* probably allow underscores, however.

purpleidea:
> capitalized like File["a"] -> Svc["b"] rather than file["a"] -> svc["b"]

I was thinking the same thing, and it would add some uniqueness, however I did yet, because I was trying to understand the goyacc output first to see if that's *actually* necessary. TBD.

> We should probably allow underscores

Yeah, I'm definitely open to allowing many of these things; I only kept things minimal for now so that I didn't have to remove a special character if we ended up finding a more important use for one. So definitely will add after the basics have been worked out first.

Does that work for you?

ffrank:
Sure, let's not get sidetracked yet :)

lang/parser.y
```diff
@@ -685,6 +693,91 @@ resource_field:
...
+//				e1,
+//				e2,
+//			},
+//			//Notify: false, // unused here, it is implied (i think)
+//		}
+//	}
+//;
+;
+edge_half_list:
+	/* end of list */
```
ffrank: This is the problem. You cannot do recursion this way. You are saying "a list is either 2) a list with one element appended or 1) NOTHING AT ALL". So the recursion has no anchor.

Changing this line to `edge_half` fixes the errors.

In another place you remark in a comment that it's not worth the hassle to catch "single element edge chains" at the parser level. I agree. We should just let that pass here. The fix should be valid.

ffrank:
I was also thinking that this is both inelegant and brittle. Would it not be better to allow chaining expressions? So that references become regular old values? Of type resource reference?

Static type checking should still work, right?

This probably means that a chain becomes an expression in its own right, and it evaluates to a reference. I figure it would probably be right-associative.

```
File[a] -> File[b] -> File[c]
// same as
File[a] -> ( File[b] -> File[c] )
// same as
$ref = File[b] -> File[c]
File[a] -> $ref
```

Sure, can be lots of extra trouble too (e.g., reporting helpful error messages becomes harder perhaps?), but it seems weird to make `IDENTIFIER OPENBRACKET expr CLOSEBRACKET` be a meaningful construct only in the context of resource chaining. (Outright wrong if we opt to add Puppet like metaparameters like `require` and friends. And we've had tentative requests for those.)

ffrank:
Ah, but I now realize that this would require that a chaining expression can be a statement and an expression at the same time. That's probably an issue?

purpleidea:
@ffrank Yeah, I thought about having the $ref be in the type system, and decided we probably don't need it and it adds a lot of complexity. Keep in mind we can still do:

```
if true { # if statement
    file[$whatever] -> svc[$blah]
}

**Comment by purpleidea** [comment]
ps: @ffrank Can you add a commit to this PR with the edge fix change you described please?

**Comment by ffrank** [comment]
I cannot, it would seem, but I did push it to a copy of your branch in ffrank/mgmt.

**Comment by purpleidea** [comment]
@ffrank With your patch, when running `make lang`, I get:

```
conflicts: 13 shift/reduce
```

Meaning, I've still got something wrong :/

ffrank:
Hah. Turns out I forget my first attempt was in my local branch as well (but not in the repo from which I pushed).

Added it now. It mandates that resource references be capitalized, like in Puppet.

purpleidea:
@ffrank I made some changes and pushed an update. I've still got to test a bunch of things and write tests before this is ready. There's a partial test that's not ready. Want to write a patch? I know you love testing!

purpleidea:
merged in 6370f0cb954ae8d973dc9d5eda58e05e0e375192 \o/

contains send/recv things too, however there is at least one bug with the send/recv stuff that needs fixing...

PR #323, opened 2018-02-10
misc: fix graphviz output for hostnames with dot in them.

aequitas:
title says it all. Solves: 
```
# dot -Tpng out.dot@faalserver > out.png
Error: out.dot@faalserver: syntax error in line 1 near '.'
```


purpleidea:
LGTM :)

PR #326, opened 2018-02-11
build: Unify build/crossbuild, explicit crossbuild targets

aequitas:
Changes:

- allows explicit crossbuild targets (eg: `make mgmt-darwin-amd64`)
- adds darwin/amd64 to default crossbuild targets
- gitignore only build artifacts (eg: not all files starting with `mgmt-`)
- `build` and `crossbuild` target now utilize the same build function (`build` still generates only a `mgmt` binary for the current os/arch)
- test crossbuilding
- allow specifying custom GOOSARCHES envvar to override defaults

Points for discussion:

- the crossbuild build function has `-s -w` linker arguments which discards some debug info afaict, @purpleidea would you like to keep the build target without these options? or maybe we should add a specific debug target?

On my mac crossbuilding won't work unless I disable augeas and libvirt:

```
~/.g/s/g/p/mgmt (build|●1✚8…3) $ make build
Generating: bindata...
Generating: lang...
/Applications/Xcode.app/Contents/Developer/usr/bin/make --quiet -C lang
Building: mgmt, os/arch: darwin-amd64, version: 0.0.14-12-g94c8bc1-dirty...
env GOOS=darwin GOARCH=amd64 time go build -ldflags "-X main.program=mgmt -X main.version=0.0.14-12-g94c8bc1-dirty -s -w" -o mgmt-darwin-amd64 ;
        7.14 real        10.36 user         1.73 sys
mv mgmt-darwin-amd64 mgmt
```

```
~/.g/s/g/p/mgmt (build|●1✚8…3) $ time env GOTAGS='noaugeas novirt' make crossbuild
Generating: bindata...
Generating: lang...
/Applications/Xcode.app/Contents/Developer/usr/bin/make --quiet -C lang
Building: mgmt, os/arch: linux-amd64, version: 0.0.14-12-g94c8bc1-dirty...
env GOOS=linux GOARCH=amd64 time go build -ldflags "-X main.program=mgmt -X main.version=0.0.14-12-g94c8bc1-dirty -s -w" -o mgmt-linux-amd64 -tags 'noaugeas novirt';
       18.48 real        50.02 user         5.83 sys
Building: mgmt, os/arch: linux-ppc64, version: 0.0.14-12-g94c8bc1-dirty...
env GOOS=linux GOARCH=ppc64 time go build -ldflags "-X main.program=mgmt -X main.version=0.0.14-12-g94c8bc1-dirty -s -w" -o mgmt-linux-ppc64 -tags 'noaugeas novirt';
       29.83 real        85.09 user        11.54 sys
Building: mgmt, os/arch: linux-ppc64le, version: 0.0.14-12-g94c8bc1-dirty...
env GOOS=linux GOARCH=ppc64le time go build -ldflags "-X main.program=mgmt -X main.version=0.0.14-12-g94c8bc1-dirty -s -w" -o mgmt-linux-ppc64le -tags 'noaugeas novirt';
       29.74 real        85.84 user        11.76 sys
Building: mgmt, os/arch: linux-arm64, version: 0.0.14-12-g94c8bc1-dirty...
env GOOS=linux GOARCH=arm64 time go build -ldflags "-X main.program=mgmt -X main.version=0.0.14-12-g94c8bc1-dirty -s -w" -o mgmt-linux-arm64 -tags 'noaugeas novirt';
       28.33 real        83.24 user        11.40 sys
Building: mgmt, os/arch: darwin-amd64, version: 0.0.14-12-g94c8bc1-dirty...
env GOOS=darwin GOARCH=amd64 time go build -ldflags "-X main.program=mgmt -X main.version=0.0.14-12-g94c8bc1-dirty -s -w" -o mgmt-darwin-amd64 -tags 'noaugeas novirt';
        7.16 real        10.15 user         1.74 sys
      114.71 real       315.26 user        42.44 sys
```


test/test-crossbuild.sh
```diff
@@ -0,0 +1,15 @@
...
+# test if we can build for all OSes and ARCHes.
+
+make crossbuild
+RET=$?
+if [ ! $RET -eq 0 ]; then
+	echo 'FAIL'
+	exit $RET
+fi
+echo 'PASS'
+
```
purpleidea: extra nl

test/test-crossbuild.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+# test if we can build for all OSes and ARCHes.
+
+make crossbuild
```
purpleidea: I think having a test for this is awesome, although if we ever break OSX, I might have to add an exit 0 here. Fine with that though.

Makefile
```diff
@@ -115,31 +118,38 @@ lang:
...
 	go build -a -installsuffix cgo -tags netgo -ldflags '-extldflags "-static" -X main.program=$(PROGRAM) -X main.version=$(SVERSION) -s -w' -o $(PROGRAM).static $(BUILD_FLAGS);
 
-build: bindata lang $(PROGRAM)
+build: $(PROGRAM)
+
+# pattern rule target for (cross)building, mgmt-OS-ARCH will be expanded to the correct build
+# extract os and arch from target pattern
+GOOS=$(firstword $(subst -, ,$*))
+GOARCH=$(lastword $(subst -, ,$*))
+build/mgmt-%: $(GO_FILES) | bindata lang
```
purpleidea: Is there a `mkdir build/` somewhere that is needed? (Test if it works in a clean repo.)

.gitignore
```diff
@@ -9,7 +9,9 @@ tmp/
 bindata/*.go
 mgmt
 mgmt.static
-mgmt-*
+# crossbuild artifacts
+mgmt-linux-*
```
purpleidea: build/ ?

Makefile
aequitas: go build seems to take care of it, I had a (broken) target for it at first but dropped it

test/test-crossbuild.sh
aequitas: No problem, or just set a GOOSARCHES variable with explicitly supported crossbuilds. As long as i'm arround testing OSX I will do best effort to keep supporting it. If not it should not become a burden on you unless you want to.

purpleidea:
@aequitas Okay, sorry for the delay.
I've rebased the first patch and merged. The 2nd had a conflict because I had previously added: 644e5164b154e0022298be30b1306ab010e47c29

So that 2nd commit isn't needed.

Also, I now have no idea how the Makefile works because it is now really complicated ;)
On a serious note, it actually seems to take a non-zero amount of time to run the Makefile now. Weird. Make performance isn't instant now?

Anyways thanks for the patch!

PR #328, opened 2018-02-11
debian: Add graphviz and packagekit runtime dependencies

aequitas:
Like the RPM, graphviz is not required but nice to have. Packagekit is also not strictly a hard requirement  but maybe that should be turned into a Depends for best functionality on resources?

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,8 @@ Build-Depends:
 
 Package: mgmt
 Architecture: any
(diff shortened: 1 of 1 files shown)


aequitas:
Packagekit is a hard dependency for sure.

purpleidea:
Merged in 41eb850b3d72a3d75ef36001d6b0fcdcd7302cc8

Thank you :)

PR #338, opened 2018-02-16
Be more informative about some failure causes/misc

aequitas:


aequitas:
@purpleidea i pulled https://github.com/purpleidea/mgmt/pull/338/commits/03123c8775de6e3754e63c79d1db55a7ef1248c4 into this branch as it was essential for me to know what error caused the testsuite to fail on Travis as I could not replicate locally (i know why now btw).

I'm gonna work on go tests for the prefix stuff, push those later. Implementation will not change much, maybe move into a seperate method for better testability.

lib/main.go
```diff
@@ -164,31 +164,31 @@ func (obj *Main) Init() error {
 		util.FlattenListWithSplit(obj.Seeds, []string{",", ";", " "}),
 	)
 	if err != nil && len(obj.Seeds) > 0 {
-		return fmt.Errorf("the Seeds didn't parse correctly")
+		return fmt.Errorf("the Seeds didn't parse correctly: %s", err)
```
purpleidea: errwrap.Wrapf(err, "message") instead please

lib/cli.go
```diff
@@ -164,7 +165,7 @@ func run(c *cli.Context) error {
 	}()
 
 	if err := obj.Run(); err != nil {
-		//return cli.NewExitError(err.Error(), 1) // TODO: ?
+		log.Printf("%s: %s", reflect.TypeOf(obj).Elem().String(), err.Error())
```
purpleidea: What is this use of `reflect` about?

aequitas: It gets the name of the object the error originated from

lib/cli.go
```diff
@@ -295,7 +295,7 @@ func CLI(program, version string, flags Flags) error {
 		},
 		cli.BoolFlag{
 			Name:  "no-server",
-			Usage: "do not let other servers peer with me",
+			Usage: "do not start embedded etcd server (only act as client of a cluster)",
```
purpleidea: in brackets could you instead write: `(do not promote from client to peer)` ?

purpleidea: also, clearify/clarify/ in commit message please.

aequitas: done

lib/cli.go
aequitas: i looked in some other places in that file and settled on 'log.Printf("Main: Error: %v", err)'

lib/main.go
aequitas: done

lib/main.go
```diff
@@ -200,6 +200,16 @@ func (obj *Main) Exit(err error) {
 	obj.exit <- err // trigger an exit!
 }
 
+func (obj *Main) createTmpPrefix(prefix *string, hostname string) error {
+	var err error
+	if *prefix, err = ioutil.TempDir("", obj.Program+"-"+hostname+"-"); err != nil {
```
purpleidea: this is a bit awkward, just return (string, error) instead.

Also, (and with admitted guilt) even though I've written bad code previously, if you're moving it around, you also get to fix it :) :) So, `obj.Program+"-"+hostname+"-"` should probably use `fmt.Sprintf` instead.

lib/main.go
```diff
@@ -200,6 +200,16 @@ func (obj *Main) Exit(err error) {
...
+		return err
+	}
+	log.Println("Main: Warning: Working prefix directory is temporary!")
```
purpleidea: logging should be in main, not buried in a helper function please

purpleidea:
Other three commits LGTM, and I will merge shortly.

Thanks again!

lib/main.go
aequitas: I was thinking to pull the entire prefix thing into a separate function for easier testing anyways, functionally the helper is really trivial in that it might be better to be less DRY here.

purpleidea:
3/4 of those commits merged. Let's work on the last one in a new PR, so closing this.

Thanks again!

PR #345, opened 2018-02-20
run: Provide cause of tmp-prefix creation failure

aequitas:
Continuation of: https://github.com/purpleidea/mgmt/pull/338

- unwrap logic for creating prefix/tmp-prefix
- expose errors caused during prefix creation process.
- put prefix logic in separate method for easier testing
- add unit test for prefixes

### DIFF
--- a/lib/main.go
+++ b/lib/main.go
@@ -200,6 +200,38 @@ func (obj *Main) Exit(err error) {
 	obj.exit <- err // trigger an exit!
 }
 
+// createPrefix handles logic to create (tmp) prefix directory.
+func (obj *Main) createPrefix(hostname string) (string, error) {
+	if obj.TmpPrefix {
+		// create temporary prefix
+		prefix, err := ioutil.TempDir("", fmt.Sprintf("%s-%s-", obj.Program, hostname))
+		if err != nil {
+			return prefix, fmt.Errorf("can't create tmp-prefix: %s", err)
+		}
+		return prefix, nil
+	}
+
+	// create/use default prefix
+	prefix := fmt.Sprintf("/var/lib/%s/", obj.Program) // default prefix
+	if p := obj.Prefix; p != nil {
+		prefix = *p
+	}
+	err := os.MkdirAll(prefix, 0770)
+	if err != nil && obj.AllowTmpPrefix {
+		// fallback to temporary prefix
+		prefix, err = ioutil.TempDir("", fmt.Sprintf("%s-%s-", obj.Program, hostname))
+		if err != nil {
+			return prefix, fmt.Errorf("can't create tmp-prefix: %s", err)
+		}
+		return prefix, nil
+	}
+	if err != nil {
+		return prefix, fmt.Errorf("can't create prefix: %s", err)
+	}
+
+	return prefix, nil
+}
+
 // Run is the main execution entrypoint to run mgmt.
 func (obj *Main) Run() error {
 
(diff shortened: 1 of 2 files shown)


lib/main.go
```diff
@@ -200,6 +200,41 @@ func (obj *Main) Exit(err error) {
 	obj.exit <- err // trigger an exit!
 }
 
+// logic to create (tmp) prefix directory
```
purpleidea: comment should be in standard `golint` format -- full sentences (ends with period), and starts with exact function name. The linter doesn't enforce on private functions, but it's a good habit to have.

lib/main.go
```diff
@@ -200,6 +200,41 @@ func (obj *Main) Exit(err error) {
...
+	var prefix string
+	var err error
+	if obj.TmpPrefix {
+		// create temporary prefix
+		prefix, err = ioutil.TempDir("", fmt.Sprintf("%s-%s-", obj.Program, hostname))
+		if err != nil {
+			return prefix, fmt.Errorf("can't create tmp-prefix: %s", err)
+		}
+		log.Println("Main: Warning: Working prefix directory is temporary!")
+	} else {
```
purpleidea: Please try and inline the code flow, eg: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

lib/main.go
```diff
@@ -200,6 +200,41 @@ func (obj *Main) Exit(err error) {
 	obj.exit <- err // trigger an exit!
 }
 
+// logic to create (tmp) prefix directory
+func (obj *Main) createPrefix(hostname string) (string, error) {
+	var prefix string
```
purpleidea: Depending on what the end function looks like, in general it is preferable to avoid declaring variables ahead of time, and instead rely on `:=` https://tour.golang.org/basics/10

If it gets reused in a different scope (eg: an if branch), it will be fine due to shadowing.

lib/main.go
aequitas: makes sense

lib/main.go
```diff
@@ -200,6 +200,40 @@ func (obj *Main) Exit(err error) {
...
+	err := os.MkdirAll(prefix, 0770)
+	if err != nil {
+		if obj.AllowTmpPrefix {
+			// fallback to temporary prefix
+			prefix, err = ioutil.TempDir("", fmt.Sprintf("%s-%s-", obj.Program, hostname))
+			if err != nil {
+				return prefix, fmt.Errorf("can't create tmp-prefix: %s", err)
+			}
+			log.Println("Main: Warning: Working prefix directory is temporary!")
+		} else {
```
purpleidea: I think this can still be inlined more...

lib/main.go
```diff
@@ -200,6 +200,40 @@ func (obj *Main) Exit(err error) {
...
+	}
+	err := os.MkdirAll(prefix, 0770)
+	if err != nil {
+		if obj.AllowTmpPrefix {
+			// fallback to temporary prefix
+			prefix, err = ioutil.TempDir("", fmt.Sprintf("%s-%s-", obj.Program, hostname))
+			if err != nil {
+				return prefix, fmt.Errorf("can't create tmp-prefix: %s", err)
+			}
+			log.Println("Main: Warning: Working prefix directory is temporary!")
```
purpleidea: Perhaps it's best to just extract all of these log messages and keep this to a single on in the main calling function there. Keep log messaging un-buried has some good benefits.

lib/main.go
```diff
@@ -200,6 +200,38 @@ func (obj *Main) Exit(err error) {
...
+	}
+	err := os.MkdirAll(prefix, 0770)
+	if err != nil && obj.AllowTmpPrefix {
+		// fallback to temporary prefix
+		prefix, err = ioutil.TempDir("", fmt.Sprintf("%s-%s-", obj.Program, hostname))
+		if err != nil {
+			return prefix, istmp, fmt.Errorf("can't create tmp-prefix: %s", err)
+		}
+		return prefix, true, nil
+	} else if err != nil {
```
purpleidea: the block right above this *always returns* so you *don't* need an else if...

lib/main.go
```diff
@@ -200,6 +200,38 @@ func (obj *Main) Exit(err error) {
 	obj.exit <- err // trigger an exit!
 }
 
+// createPrefix handles logic to create (tmp) prefix directory.
+func (obj *Main) createPrefix(hostname string) (string, bool, error) {
+	istmp := false
```
purpleidea: vars should be in the form `isTmp`, however I don't think this is needed at all. Just log the prefix we make.

aequitas: Fine by me, I just tried to maintain feature parity :)

aequitas:
@purpleidea updated

aequitas:
@purpleidea this PR has also been lingering for to long, I spent all the effort making the changes you requested and it has been open for more than 2 weeks.

PR #351, opened 2018-02-22
misc: Small fixes for makedeps script

ofrzeta:
- install Go distribution package only if no go binary found
- use new govet home to import govet

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -11,6 +11,7 @@ fi
 
 sudo_command=$(command -v sudo)
 
+GO=`command -v go 2>/dev/null`
 YUM=`command -v yum 2>/dev/null`
 DNF=`command -v dnf 2>/dev/null`
 APT=`command -v apt-get 2>/dev/null`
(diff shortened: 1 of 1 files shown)


misc/make-deps.sh
```diff
@@ -55,16 +56,20 @@ fi
 
 if [ $travis -eq 0 ]; then
 	if [ ! -z "$YUM" ]; then
-		# some go dependencies are stored in mercurial
```
purpleidea: don't remove this comment, move it with the command please :)

misc/make-deps.sh
```diff
@@ -84,9 +89,8 @@ echo "done running 'go get -v -t -d ./...'"
 [ -e "$GOBIN/mgmt" ] && rm -f "$GOBIN/mgmt"	# the `go get` version has no -X
 # vet is built-in in go 1.6 - we check for go vet command
 go vet 1> /dev/null 2>&1
-ret=$?
-if [[ $ret != 0 ]]; then
-	go get golang.org/x/tools/cmd/vet      # add in `go vet` for travis
+if [[ $? != 0 ]]; then
+	go get github.com/golang/go/tree/master/src/cmd/vet      # add in `go vet` for travis
```
purpleidea: why did you make this change?

ofrzeta: Because if I didn't I get the following error message:

```
package golang.org/x/tools/cmd/vet: cannot find package "golang.org/x/tools/cmd/vet" in any of:
	/usr/local/go/src/golang.org/x/tools/cmd/vet (from $GOROOT)
	/home/oliver/gotest/src/golang.org/x/tools/cmd/vet (from $GOPATH)
```

I'll remove it from the PR and let you sort it out :)

purpleidea:
@ofrzeta Great, thank you, merged in 318fe4a5dced214ab3ae1ace4ff0dc207dde2094

PR #349, opened 2018-02-21
Emacs support for language

mavit:
Fixes #348.

PR #352, opened 2018-02-22
lang: Add `getenv` function, fix .editorconfig for mcl files, tool links

aequitas:
title says it all

### DIFF
--- a/lang/funcs/simple/env_func.go
+++ b/lang/funcs/simple/env_func.go
@@ -1,0 +1,84 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package simple // TODO: should this be in its own individual package?
+
+import (
+	"os"
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+func init() {
+	Register("getenv", &types.FuncValue{
+		T: types.NewType("func(str) str"),
+		V: GetEnv,
+	})
+	Register("defaultenv", &types.FuncValue{
+		T: types.NewType("func(str, str) str"),
+		V: DefaultEnv,
+	})
+	Register("hasenv", &types.FuncValue{
+		T: types.NewType("func(str) bool"),
+		V: HasEnv,
+	})
+	Register("env", &types.FuncValue{
+		T: types.NewType("func() {str: str}"),
+		V: Env,
+	})
+}
+
+// GetEnv gets environment variable by name or returns empty string if non existing.
+func GetEnv(input []types.Value) (types.Value, error) {
+	return &types.StrValue{
+		V: os.Getenv(input[0].Str()),
+	}, nil
+}
+
+// DefaultEnv gets environment variable by name or returns default if non existing.
+func DefaultEnv(input []types.Value) (types.Value, error) {
+	value, exists := os.LookupEnv(input[0].Str())
+	if !exists {
+		value = input[1].Str()
+	}
+	return &types.StrValue{
+		V: value,
+	}, nil
+}
+
+// HasEnv returns true if environment variable exists.
+func HasEnv(input []types.Value) (types.Value, error) {
+	_, exists := os.LookupEnv(input[0].Str())
+	return &types.BoolValue{
+		V: exists,
+	}, nil
+}
+
+// Env returns a map of all keys and their values.
+func Env(input []types.Value) (types.Value, error) {
+	environ := make(map[types.Value]types.Value)
+	for _, keyval := range os.Environ() {
+		if i := strings.IndexRune(keyval, '='); i != -1 {
+			environ[&types.StrValue{V: keyval[:i]}] = &types.StrValue{V: keyval[i+1:]}
+		}
+	}
+	return &types.MapValue{
+		T: types.NewType("{str: str}"),
+		V: environ,
+	}, nil
+}
--- a/examples/lang/env0.mcl
+++ b/examples/lang/env0.mcl
@@ -1,0 +1,20 @@
+# read and print environment variable
+# env TEST=123 EMPTY= ./mgmt run --tmp-prefix --lang=examples/lang/env0.mcl --converged-timeout=5
+
+$x = getenv("TEST", "321")
+
+print "print1" {
+	msg => printf("the value of the environment variable TEST is: %s", $x),
+}
+
+$y = getenv("DOESNOTEXIT", "321")
+
+print "print2" {
+	msg => printf("environment variable DOESNOTEXIT does not exist, defaulting to: %s", $y),
+}
+
+$z = getenv("EMPTY", "456")
+
+print "print3" {
+	msg => printf("same goes for epmty variables like EMPTY: %s", $z),
+}
--- a/test/shell/env0.mcl
+++ b/test/shell/env0.mcl
@@ -1,0 +1,19 @@
+$tmpdir = defaultenv("TMPDIR", "/tmp")
+
+$x = getenv("TEST")
+$y = getenv("DOESNOTEXIST")
+$z = getenv("EMPTY")
+
+$a = defaultenv("TEST", "321")
+$b = defaultenv("DOESNOTEXIST", "321")
+$c = defaultenv("EMPTY", "456")
+
+$t = hasenv("TEST")
+$f = hasenv("DOESNOTEXIST")
+
+$env = env()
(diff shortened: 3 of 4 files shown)


aequitas:
References: https://github.com/purpleidea/mgmt/pull/349

lang/funcs/simple/getenv_func.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+func init() {
+	Register("getenv", &types.FuncValue{
+		T: types.NewType("func(str, str) str"),
+		V: func(input []types.Value) (types.Value, error) {
+			value := os.Getenv(input[0].Str())
+			// TODO: should empty/non-existing behaviour be different? ie: empty returns empty value?
```
aequitas: Input on this would be welcome.

lang/funcs/simple/getenv_func.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+func init() {
+	Register("getenv", &types.FuncValue{
```
purpleidea: I think this also needs a `hasenv` function that tells us if an env exists. To differentiate between empty string and non-existant. See: https://golang.org/pkg/os/#LookupEnv

Whether we then use the "default" approach as implemented, or "return empty string" instead, needs discussion.

Maybe both?

lang/funcs/simple/getenv_func.go
```diff
@@ -0,0 +1,43 @@
...
+		T: types.NewType("func(str, str) str"),
+		V: func(input []types.Value) (types.Value, error) {
+			value := os.Getenv(input[0].Str())
+			// TODO: should empty/non-existing behaviour be different? ie: empty returns empty value?
+			if value == "" {
```
purpleidea: This is actually wrong. Value could exist and be empty.

lang/funcs/simple/getenv_func.go
aequitas: I have been using an empty string in environment variables, mostly to prevent it from going to the default option, so that might be an explicit feature needed.

aequitas:
I didn't get the linter yet, I don't run `make-deps.sh` that often ;)

purpleidea:
Adding `*.mcl` to the `.editorconfig` is one of those brilliant little tweaks which I never thought of and which is one of the reasons I love Free Software.

That first commit is merged :)

aequitas:
@purpleidea probably your default editor settings is tabs anyways ;)

purpleidea:
@aequitas I didn't realize anyone would build an $EDITOR that allowed anything different!

lang/funcs/simple/environ_func.go
```diff
@@ -0,0 +1,84 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// functions to look up variables in the mgmt server instance environment
```
purpleidea: Unfortunately we don't have a good space for doc strings yet... You can do one of two things:

1) Just kill comment entirely OR
2) Define each function separately with doc string, and then in init put the handle to the function.

eg:

```
// whatever
func whatever(input []types.Value) (types.Value, error) ...

```

with

```

	Register("getenv", &types.FuncValue{
		T: types.NewType("func(str) str"),
		V: whatever,

...
```

lang/funcs/simple/environ_func.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+func init() {
+	// getenv(str) str # get env or empty string is absent
```
purpleidea: s/is/if/ ?

lang/funcs/simple/environ_func.go
```diff
@@ -0,0 +1,84 @@
...
+	Register("getenv", &types.FuncValue{
+		T: types.NewType("func(str) str"),
+		V: func(input []types.Value) (types.Value, error) {
+			value := os.Getenv(input[0].Str())
+			return &types.StrValue{
+				V: fmt.Sprintf("%s", value),
```
purpleidea: why fmt.Sprintf? it's already a string :)
In fact, it's a one-liner here... No need for the `value` var.

same below...

test/shell/environ.sh
```diff
@@ -0,0 +1,21 @@
...
+set -o pipefail
+
+. ../util.sh
+
+# these values represent environment variable values or defaults set in getenv.mcl
+regex="123,,:123,321,:true,false:123"
+
+tmpdir="$($mktemp --tmpdir -d tmp.XXX)"
+
+env TMPDIR="${tmpdir}" TEST=123 EMPTY="" $timeout -sKILL 60s "$MGMT" run --tmp-prefix --converged-timeout=1 --lang environ.mcl
```
purpleidea: `--converged-timeout=5` please

lang/funcs/simple/environ_func.go
aequitas: copied it from your example ;)

purpleidea: i make many mistakes??

aequitas:
@purpleidea everything should be fixed

lang/funcs/simple/env_func.go
```diff
@@ -0,0 +1,76 @@
...
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+// getenv(str) str # get env or empty string if absent
+func getenv(input []types.Value) (types.Value, error) {
+	value := os.Getenv(input[0].Str())
+	return &types.StrValue{
+		V: fmt.Sprintf("%s", value),
```
purpleidea: why is the fmt.Sprintf needed? this is a one-liner without the value var.
same below...

lang/funcs/simple/env_func.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+// getenv(str) str # get env or empty string if absent
```
purpleidea: docstrings should be proper sentences in the expected form please. You can make them public so that they show up in the docs too :)

lang/funcs/simple/env_func.go
```diff
@@ -0,0 +1,76 @@
...
+			environ[&types.StrValue{V: keyval[:i]}] = &types.StrValue{V: keyval[i+1:]}
+		}
+	}
+	return &types.MapValue{
+		T: types.NewType("{str: str}"),
+		V: environ,
+	}, nil
+}
+
+func init() {
```
purpleidea: even though it's not intuitive, init should be at the top where it was before. This is so you don't accidentally forget it's there and running code when you didn't expect it to :)

lang/funcs/simple/env_func.go
```diff
@@ -0,0 +1,76 @@
...
+	Register("getenv", &types.FuncValue{T: types.NewType("func(str) str"), V: getenv})
```
purpleidea: I think these would be more readable across multiple lines as previously shown.

lang/funcs/simple/env_func.go
```diff
@@ -0,0 +1,87 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// functions to look up variables in the mgmt server instance environment
```
purpleidea: NACK this line please.

lang/funcs/simple/env_func.go
```diff
@@ -0,0 +1,87 @@
...
+	})
+	Register("env", &types.FuncValue{
+		T: types.NewType("func() {str: str}"),
+		V: env,
+	})
+}
+
+// getenv gets environment variable by name or empty string if non existing.
+func getenv(input []types.Value) (types.Value, error) {
+	value := os.Getenv(input[0].Str())
```
purpleidea: Can be written as:

```
	return &types.StrValue{
		V: os.Getenv(input[0].Str()),
	}, nil
```

lang/funcs/simple/env_func.go
```diff
@@ -0,0 +1,87 @@
...
+	return &types.StrValue{
+		V: value,
+	}, nil
+}
+
+// defaultenv gets environment variable by name or return default if non existing.
```
purpleidea: Make all the functions public. eg:

`Defaultenv` (capitalized)

so they'll appear in docs :)

aequitas: I didn't know the capitalisation mattered :) Go is weird sometimes. Consistent but weird.

purpleidea: Yup. And it's capitalization of the actual function ;) The comment just has to match or golint will fail.

aequitas:
> I made these review comments previously. If you don't agree you've got to say so, so we can have the discussion, otherwise it just looks like you're forgetting.

Sometimes I'm just forgetful or disorganised, but if I do disagree I always open discussion :)

purpleidea:
L G T M : )

purpleidea:
Merged :)

PR #353, opened 2018-02-22
Integrationtest in golang

aequitas:
Todo:

- [ ] no sleep, proper waiting on events like convergence or shutdown
- [ ] try and make clustering work reliably
- [x] move to top level package

### DIFF
--- a/integration/instance.go
+++ b/integration/instance.go
@@ -1,0 +1,362 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// notes:
+// https://www.philosophicalhacker.com/post/integration-tests-in-go/
+// https://github.com/golang/go/wiki/TableDrivenTests
+// https://blog.golang.org/subtests
+// https://splice.com/blog/lesser-known-features-go-test/
+// https://coreos.com/blog/testing-distributed-systems-in-go.html
+// https://www.philosophicalhacker.com/post/integration-tests-in-go/
+
+package integration
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
+	errwrap "github.com/pkg/errors"
+)
+
+const (
+	defaultTimeout     = 120 * time.Second // longest time a mgmt instance should exist during tests
+	convergedTimeout   = 5 * time.Second
+	idleTimeout        = 5 * time.Second
+	shutdownTimeout    = 10 * time.Second
+	convergedIndicator = "Converged for 5 seconds, exiting!"
+	exitedIndicator    = "Goodbye!"
+)
+
+// Instance represents a single mgmt instance that can be managed
+type Instance struct {
+	// settings passed into the instance
+	Timeout time.Duration
+	Env     []string
+
+	// output and metadata regarding the mgmt instance
+	Name         string
+	Tmpdir       string
+	Prefix       string
+	Workdir      string
+	DeployOutput string
+	Stdout       bytes.Buffer
+	Stderr       bytes.Buffer
+	Err          error
+	Seeds        string
+
+	command string
+	env     string
+	ctx     context.Context
+	cmd     *exec.Cmd
+	cancel  context.CancelFunc
+}
+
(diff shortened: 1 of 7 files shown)


purpleidea:
Which PR should I be looking at?

aequitas:
Not this one until the [WIP] status is removed :)

test/integration/sanity_test.go
```diff
@@ -0,0 +1,137 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// run simple smoke tests to make sure most basic functionality works
```
purpleidea: doc string or nack please :)

test/integration/sanity_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+)
+
+func TestHelp(t *testing.T) {
+	out, err := exec.Command(mgmt).Output()
+	if err != nil {
+		t.Fatal("Error: ", err)
+	}
+	if !strings.Contains(string(out), "--version, -v  print the version") {
+		log.Printf("Command output: %s", string(out))
```
purpleidea: Use `t.Logf`, not `log.` unless privately for debugging.

test/integration/sanity_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"log"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestHelp(t *testing.T) {
+	out, err := exec.Command(mgmt).Output()
+	if err != nil {
+		t.Fatal("Error: ", err)
```
purpleidea: Fatalf instead, and %v it

test/integration/sanity_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if !strings.Contains(string(out), "--version, -v  print the version") {
```
purpleidea: You sure we want to check this exact string? Is that API stable in the cli lib?

test/integration/sanity_test.go
```diff
@@ -0,0 +1,137 @@
...
+		log.Printf("Command output: %s", string(out))
+		t.Fatal("Expected output not found")
+	}
+}
+
+// TestSmoketest makes sure the most basic functionality works
```
purpleidea: Docstrings should be full sentences...

test/integration/sanity_test.go
aequitas: Is anything stable at this point ;)

The test is easy enough modified if the cli lib API changes, also I don't expect it to change that often.

aequitas:
> My biggest suggestion is to think about what the API should be. Currently all your methods "just run things", but they don't usually either return errors, and/or wait for a channel or waitgroup. I'd consider thinking about those before putting this in master.

Currently all the methods handle the test 'fatality', that was not an intentional design but kinda grew this way. I also need to put the API through some more runs to see what is viable or not in regards of usability and debuggability. If channels or wait offer a clear benefit I might consider using them, otherwise I prefer a clean clear API with not to much of Go's verbosity around it. If you're talking code running in production I'm all for advanced Go features. But since the testsuites are mostly just sequential validation of a black box I don't think add anything to make it more Go fancy.

purpleidea:
I realized I had some discussions with @ffrank about testing a while back. Figured I'd ping him here, in case he's interested and wants to have a look!

aequitas:
Rebase on latest master, made all test function t.Helper functions so they provide a better trace on a t.Fatal.

aequitas:
API questions:

- Should functions like `RunBackground` and `DeployLang` have an additional boolean 'wait' argument so they can be merged with `WaitUntilIdle` and `WaitUntilConverged` as that is probably the most common use case?
- Rename `RunBackground`, `StopBackground` to `Start`, `Stop`?

aequitas:
Further development:

- Currently all integration tests live under `integration/`. Ideally these would be moved closer to the code and use the helper functions/types in the integration package. When running the gotest suite a flag should be provided to enable/disable these tests.
- Currently the API is based on helper functions and assertions. Possible common patterns can be extracted so these can be turned into table based tests or struct based tests: eg:

```golang
tests = []integration.testCase{
  Setup integration.TestSetup
  Code string
  Validate []integration.TestValidation
}{
  {
    Setup: integration.TestSetup{"default"},
    Code: `
    file "/pass" {
      content => "not empty",
    }
    `
    Validate: [
      integration.TestValidation{"pass"}
    ]
  },
}
```

aequitas:
/cc @heyitsanthony as discussed via IRC, I was told by @purpleidea to get your advice regarding this PR. 
Specifically this construct: https://github.com/purpleidea/mgmt/pull/353/files#diff-a07f5e8ba679bba1c058680161ef42fbR40

Looking forward to your input.

heyitsanthony:
@aequitas try it like this: https://github.com/heyitsanthony/mgmt/commit/58fbe8d1f2c9bcd5de92d73c35d738815444a5a2

aequitas:
@heyitsanthony thanks for your response. 

The main goal I had with my setup was to not make these tests as complex as the code it is testing. So to reduce the amount of `if` statements to simple single line assertions that can be grokked at a glance, even if you don't know Go. Your suggestion maintains that idea. 

I'd have to re-add the error wrapping to maintain some context information regarding the errors. And probably also move it into a test util file of sorts so it can be shared amongst all integration test files. I'll try this approach with the other testcases written so far to see if I run into additional problems.

@purpleidea can you also have a look if you think this is the direction you like?

aequitas:

PR #354, opened 2018-02-22
recwatch: Fix watching newly created files on macOS

aequitas:
Fixes: https://github.com/purpleidea/mgmt/issues/33

### DIFF
--- a/docs/quick-start-guide.md
+++ b/docs/quick-start-guide.md
@@ -134,9 +134,8 @@ discouraged either). Meaning it might work but in the case it doesn't you would
 have to provide your own patches to fix problems (the project maintainer and
 community are glad to assist where needed).
 
-There are currently some issues that make `mgmt` less suitable to run for
-provisioning macOS (eg: [https://github.com/purpleidea/mgmt/issues/33](https://github.com/purpleidea/mgmt/issues/33)).
-But as a client to provision remote servers it should run fine.
+There are currently some issues that make `mgmt` less suitable to run for provisioning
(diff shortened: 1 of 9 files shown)


test/util.sh
```diff
@@ -11,9 +11,11 @@ export MGMT="$ROOT/mgmt"
 if [[ $(uname) == "Darwin" ]] ; then
 	export timeout="gtimeout"
 	export mktemp="gmktemp"
+	export STAT="gstat"
```
purpleidea: how come some of these are uppercase, and others are lowercase? what is correct?

docs/quick-start-guide.md
```diff
@@ -134,10 +134,6 @@ discouraged either). Meaning it might work but in the case it doesn't you would
 have to provide your own patches to fix problems (the project maintainer and
 community are glad to assist where needed).
 
-There are currently some issues that make `mgmt` less suitable to run for
```
purpleidea: Why kill this? No systemd support or equivalent on osx...

test/util.sh
aequitas: If it is not capitalised it hits: https://github.com/koalaman/shellcheck/wiki/SC2154

Preferably all shell scripts should be conform shellcheck but thats a branch I still need to finish.

docs/quick-start-guide.md
aequitas: True, I just couldn't find any issue to reference :)

purpleidea:
Cools! Merged!

Issue #347, opened 2018-02-21
`make build` doesn't notice deleted files


purpleidea:
If you run `make build`, and then delete a `*.go` file, and then run `make build` again, it won't notice that the build should be remade. Woops.

Not a big deal, but I thought I'd mention it in case there was an easy fix.

If you know the fix, please send us a patch :)

purpleidea:
This has been merged! \o/

PR #359, opened 2018-02-26
[wip] build: Rebuild when a go files gets deleted

aequitas:
Fixes: https://github.com/purpleidea/mgmt/issues/347

Not tested to full extent but this should fix the problem of deleted files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ SHELL = /usr/bin/env bash
 .PHONY: all art cleanart version program lang path deps run race bindata generate build build-debug crossbuild clean test gofmt yamlfmt format docs rpmbuild mkdirs rpm srpm spec tar upload upload-sources upload-srpms upload-rpms copr
 .SILENT: clean bindata
 
+# list of all .go files used to determine when the binary needs rebuilding
 # a large amount of output from this `find`, can cause `make` to be much slower!
 GO_FILES := $(shell find * -name '*.go' -not -path 'old/*' -not -path 'tmp/*')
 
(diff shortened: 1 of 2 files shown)


aequitas:
@purpleidea other thing i noticed, skipping `vendor/` shaves of a few couple of thousands of nanoseconds from make. Which is valuable for when tab-completion targets need to be calculated. I was thinking maybe adding a flag to allow vendor files to be evaluated, or to just require `touch main.go; make` as working in vendor files is not that common. Whats your thought?

With:
```
$ time make
Generating: bindata...
Generating: lang...
/Applications/Xcode.app/Contents/Developer/usr/bin/make --quiet -C lang
        1.29 real         0.98 user         0.17 sys
```

Without:
```
$ time make
Generating: bindata...
Generating: lang...
/Applications/Xcode.app/Contents/Developer/usr/bin/make --quiet -C lang
        0.49 real         0.23 user         0.14 sys
```

purpleidea:
You are brilliant, but I'm a bit afraid to merge this...

Makefile
```diff
@@ -124,7 +125,13 @@ lang:
 $(PROGRAM): build/mgmt-${GOHOSTOS}-${GOHOSTARCH}
 	cp $< $@
 
-$(PROGRAM).static: $(GO_FILES)
+# keep a list of all go files so we can rebuild if one gets deleted
+.make.gofiles: $(GO_FILES) FORCE
+	@echo $^ > $@.tmp
+	@diff $@.tmp $@ >/dev/null || mv $@.tmp $@
+FORCE: ;
```
purpleidea: Is there a need for this even if we can make clean && make build ?

purpleidea:
Small issue:

When testing this by running `make` I get:

```
make: execvp: /usr/bin/env: Argument list too long
make: *** [Makefile:130: .make.gofiles] Error 127
```

aequitas:
@purpleidea can you test with the latest update?

purpleidea:
I still get the same issue. I think the problem is:

```
	@echo "$^" > $@.tmp 
```

because that list of files is too long for echo. Probably needs to be done in a loop or some trickery so that it doesn't hit the bash limit.

aequitas:
You probably have a wee bit more files in your repo than I have atm :D 

I tried with https://www.gnu.org/software/make/manual/html_node/File-Function.html but somehow that doesn't work on my machine. Maybe due to an older make version. I'll do some more research.

Issue #313, opened 2018-02-10
[LOVE] Add DSL fast type-checking as a sub-command


purpleidea:
#### Welcome:

This is a beginner to intermediate level patch for someone who is new to golang, and is new to mgmt. It is meant to encourage new programmers to get involved with mgmt. It will probably take you a few hours to a day or so to complete.

#### Intro:

Mgmt has a built-in language. To run code, we start up the engine, invoke the many steps in the compilation process, and on error, shutdown.
Mgmt should have a sub flag called `--check` (alternate naming suggestions are welcome) which can run the lexer, parser, compiler, etc without actually running the function engine, and without needing to startup anything.
This should be useful as a light weight way to type check code bases before you run a deploy.
To implement this, you'll need to either recommend an alternate design, or use the one provided below.

#### Design:

You should add a new mgmt flag which runs these checks without the etcd startup or the function engine running.
If it cannot easily work as a --check flag, then it could be a check sub command.

Read the contributing tips:
https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Happy hacking, and ping me for questions!

#### Bonus/Future work:

* Write a few tests.
* Add modifiers to only test the lexer, the parser, etc...

aequitas:
I've started working on this.

I suggest using the `--lint` flag or `lint` subcommand instead of `--check` as it might be confused by ansible users for a dry-run/noop deployment.

purpleidea:
@aequitas Maybe it should even be a pre-step of running deploy?

PR #361, opened 2018-02-26
cli: Add Validate command

aequitas:
https://github.com/purpleidea/mgmt/issues/313

Todo:

- [x] docs
- [x] lint on deploy
- [x] test-lint-mcl test for examples


test/test-examples-lint.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo running "$0"
+
+ROOT=$(dirname "${BASH_SOURCE}")/..
+cd "${ROOT}"
+
+failures=''
+
+# lint .mcl examples
+for file in $(find examples/lang/ -maxdepth 3 -type f -name '*.mcl'); do
```
purpleidea: I'm really pleased you added a test for this, because I was thinking we needed this :)

examples/lang/syntaxerror.mcl.txt
```diff
@@ -0,0 +1,4 @@
+# this file generates a syntax error, used in lib/cli_test.go
+file "/tmp/mgmt" {
+  content => "test"  # missing trailing comma
```
purpleidea: tab for indentation, unless your point was that spaces should be a compile error ;)

lib/cli_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"flag"
+	"testing"
+
+	errwrap "github.com/pkg/errors"
+
+	"github.com/urfave/cli"
+)
+
+func TestLintPass(t *testing.T) {
```
purpleidea: what's going on in these two?

lang/lang.go
```diff
@@ -267,3 +267,9 @@ func (obj *Lang) Close() error {
 	obj.wg.Wait()
 	return err
 }
+
+func (obj *Lang) Lint() error {
+	_, err := LexParse(obj.Input)
```
purpleidea: This is actually the part which needs the most work... You probably want all of these to run:

https://github.com/purpleidea/mgmt/blob/master/docs/language-guide.md#stages

With the exception of actually running the function engine. All of those are needed to ascertain if a compile is successful.

examples/lang/syntaxerror.mcl.txt
aequitas: That would be a good point :)

examples/lang/bad/scopeerror.mcl
```diff
@@ -0,0 +1,3 @@
+# this file generates a scope error, used in lib/cli_test.go
```
purpleidea: all "bad" things should be in test/ dir, not easy to find in examples/

in fact, the *real* tests should have nothing to do with examples/

I have a test case that tests examples just to make sure the examples we submit aren't actually junk, but the real testing should happen in test/

examples/lang/bad/syntaxerror.mcl
```diff
@@ -0,0 +1,4 @@
+# this file generates a syntax error, used in lib/cli_test.go
```
purpleidea: let's not refer to a specific file like this

examples/lang/bad/unificationerror.mcl
```diff
@@ -0,0 +1,2 @@
+# this file generates a unification error, used in lib/cli_test.go
```
purpleidea: okay, however why not write this as a test in unification_test.go instead?

lang/gapi.go
```diff
@@ -67,10 +68,26 @@ func (obj *GAPI) Cli(c *cli.Context, fs resources.Fs) (*gapi.Deploy, error) {
 			return nil, fmt.Errorf("input code is empty")
 		}
 
+		// validate syntax before deploying
+		filecontent, err := ioutil.ReadFile(s)
```
purpleidea: `fileContent`

lang/gapi.go
```diff
@@ -67,10 +68,26 @@ func (obj *GAPI) Cli(c *cli.Context, fs resources.Fs) (*gapi.Deploy, error) {
...
+		if err != nil {
+			return nil, errwrap.Wrapf(err, "can't read code from file `%s`", s)
+		}
+		code := strings.NewReader(string(filecontent))
+		obj := &Lang{
+			Input: code,
+			Logf: func(format string, v ...interface{}) {
+				log.Printf(Name+": funcs: "+format, v...)
```
purpleidea: what? why funcs?

lang/gapi.go
```diff
@@ -67,10 +68,26 @@ func (obj *GAPI) Cli(c *cli.Context, fs resources.Fs) (*gapi.Deploy, error) {
...
+		}
+		err = obj.Validate()
+		if err != nil {
+			return nil, errwrap.Wrapf(err, s)
+		}
+
 		// read through this local path, and store it in our file system
 		// since our deploy should work anywhere in the cluster, let the
 		// engine ensure that this file system is replicated everywhere!
-
```
purpleidea: why is this line missing?
if there's a reason, okay, but it sends the message you don't review your diffs before submitting.
the goal should be for me to see your patch and say "perfect, merged" :)

lib/cli_test.go
```diff
@@ -0,0 +1,60 @@
...
+		t.Fatal(errwrap.Wrapf(err, "example load0.mcl should pass validation"))
+	}
+}
+
+// TestValidateError tests if validation fails on files with invalid syntax
+func TestValidateError(t *testing.T) {
+	files := []struct {
+		path string
+	}{
+		{"../examples/lang/bad/syntaxerror.mcl"},
```
purpleidea: no ../ paths please and no references to examples/ in our code.
we need to think of a better way to get the code in. i'd actually prefer to just inline it for now. see:

https://github.com/purpleidea/mgmt/blob/master/lang/lexparse_test.go#L93

examples/lang/bad/unificationerror.mcl
aequitas: i want to test validate is able to detect a unification error, not that this file generates a unification error per se.

lang/gapi.go
aequitas: probably a copy paste error i overlooked

lang/gapi.go
aequitas: must have slipped past

purpleidea: no worries

aequitas:

Issue #312, opened 2018-02-10
[LOVE] Add an info sub command to mgmt to report information


purpleidea:
#### Welcome:

This is a beginner to intermediate level patch for someone who is new to golang, and is new to mgmt. It is meant to encourage new programmers to get involved with mgmt. It will probably take you a few hours to a day or so to complete.

#### Intro:

Mgmt has a number of golang interfaces which are implemented by different components in mgmt. Most of these are registered at runtime. We'd like to have some information about what's available so that users don't need to dig through the source.
Mgmt should have a sub command called `info` (alternate naming suggestions are welcome) which can return the list of resources, functions, and other useful information.
To implement this, you'll need to either recommend an alternate design, or use the one provided below.

#### Design:

You should add a new mgmt command in `lib/info.go` which implements this command. It should take a sub command for each type of thing, and list the information for those things. Eg:
`mgmt info resources`	# prints a list of registered resources
`mgmt info resources --type`	# prints a list of resources with their types
`mgmt info resources file --type`	# prints info on file resource and it's type
`mgmt info resources file --yaml`	# outputs this information as yaml
`mgmt info functions`	# prints a list of registered functions

Read the contributing tips:
https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Happy hacking, and ping me for questions!

#### Bonus/Future work:

* Write a few tests.
* Recommend some other properties that could be listed.
* Consider listing sub properties like resource field types, or function signatures.

PR #363, opened 2018-02-27
[wip] cli: Add info command

aequitas:
For now just lists resources and functions, more to be added soon.

```
$ ./mgmt info funcs -t
answer			answer() int
contains
datetime		datetime() int
datetime_print		datetime_print(int) str
defaultenv		defaultenv(str, str) str
env			env() {str: str}
example_errorbool	example_errorbool(bool) str
exchange		exchange(str, str) {str: str}
flipflop		flipflop() bool
getenv			getenv(str) str
hasenv			hasenv(str) bool
hostname		hostname() str
int2str			int2str(int) str
kvlookup		kvlookup(str) {str: str}
len
load			load() struct{x1 float; x5 float; x15 float}
maplookup
pow			pow(float, float) float
printf
random1			random1(int) str
schedule
structlookup
template
vumeter			vumeter(str, int, float) str
```
```
$ ./mgmt info resources
augeas
aws:ec2
exec
file
graph
group
hostname
kv
msg
noop
nspawn
password
pkg
print
svc
test
timer
user
virt
```

### DIFF
--- a/lib/info.go
+++ b/lib/info.go
@@ -1,0 +1,127 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package lib
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"reflect"
+	"sort"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/purpleidea/mgmt/lang/funcs"
+	"github.com/purpleidea/mgmt/lang/interfaces"
+	"github.com/purpleidea/mgmt/resources"
+	"github.com/urfave/cli"
+)
+
+const (
+	twMinWidth = 0
+	twTabWidth = 8
+	twPadding  = 2   // ensure columns have at least a space between them
+	twPadChar  = ' ' // using a tab here creates 'jumpy' columns on output
+	twFlags    = 0
+)
+
+// info wraps infocmd to produce output to stdout
+func info(c *cli.Context) error {
+	output, err := infoCmd(c)
+	if err != nil {
+		return err
+	}
+	w := tabwriter.NewWriter(os.Stdout, twMinWidth, twTabWidth, twPadding, twPadChar, twFlags)
+	fmt.Fprint(w, output.String())
+	w.Flush()
+	return nil
+}
+
+// infoCmd takes the cli context and returns the requested output
+func infoCmd(c *cli.Context) (bytes.Buffer, error) {
+	var out bytes.Buffer
+	var names []string
+	descriptions := make(map[string]string)
+
+	switch c.Command.Name {
+	case "resources":
+		for _, name := range resources.RegisteredResourcesNames() {
+			names = append(names, name)
+
+			descriptions[name] = ""
+
+			res, err := resources.Lookup(name)
+			if err != nil {
+				continue
+			}
+
+			s := reflect.ValueOf(res).Elem()
+			typeOfT := s.Type()
+
+			var fields []string
+			for i := 0; i < s.NumField(); i++ {
+				field := typeOfT.Field(i)
+				// skip unexported fields
+				if field.PkgPath != "" {
+					continue
+				}
+				fieldname := field.Name
+				if fieldname == "BaseRes" {
+					continue
+				}
+				f := s.Field(i)
+				fieldtype := f.Type()
+				fields = append(fields, fmt.Sprintf("%s (%s)", strings.ToLower(fieldname), fieldtype))
+			}
+			descriptions[name] = strings.Join(fields, ", ")
+
+		}
+	case "functions":
+		for name := range funcs.RegisteredFuncs {
+			// skip internal functions (history, operations, etc)
+			if strings.HasPrefix(name, "_") {
+				continue
+			}
+			names = append(names, name)
+			descriptions[name] = ""
+			fn, err := funcs.Lookup(name)
+			if err != nil {
+				continue
+			}
+			if _, ok := fn.(interfaces.PolyFunc); !ok {
+				// TODO: skip for now, needs Build before Info
+				continue
+			}
+
+			// set function signature as description
+			descriptions[name] = strings.Replace(fn.Info().Sig.String(), "func", name, 1)
+		}
+	default:
+		return out, fmt.Errorf("invalid command")
+	}
+
+	sort.Strings(names)
+	for _, name := range names {
+		if c.Bool("type") {
+			fmt.Fprintf(&out, "%s\t%s\n", name, descriptions[name])
+		} else {
+			fmt.Fprintln(&out, name)
+		}
+	}
+	return out, nil
+}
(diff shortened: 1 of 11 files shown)


aequitas:
fixes https://github.com/purpleidea/mgmt/issues/312

aequitas:
Now also with a little bit of resource types:

```
augeas    file (string), lens (string), sets ([]resources.AugeasSet)
aws:ec2   state (string), region (string), type (string), imageid (string), watchendpoint (string), watchlistenaddr (string), erroronmalformedpost (bool), userdata (string), publicipv4 (string), privateipv4 (string), instanceid (string)
exec      cmd (string), shell (string), timeout (int), watchcmd (string), watchshell (string), ifcmd (string), ifshell (string), user (string), group (string), output (*string), stdout (*string), stderr (*string)
file      path (string), dirname (string), basename (string), content (*string), source (string), state (string), owner (string), group (string), mode (string), recurse (bool), force (bool)
graph     graph (*pgraph.Graph)
group     state (string), gid (*uint32)
hostname  hostname (string), prettyhostname (string), statichostname (string), transienthostname (string)
kv        key (string), value (*string), skiplessthan (bool), skipcmpstyle (resources.KVResSkipCmpStyle)
msg       body (string), priority (string), fields (map[string]string), journal (bool), syslog (bool)
noop      comment (string)
nspawn    state (string)
password  length (uint16), saved (bool), checkrecovery (bool), password (*string)
pkg       state (string), allowuntrusted (bool), allownonfree (bool), allowunsupported (bool)
print     msg (string)
svc       state (string), startup (string), session (bool)
timer     interval (uint32)
user      state (string), uid (*uint32), gid (*uint32), group (*string), groups ([]string), homedir (*string), allowduplicateuid (bool)
virt      uri (string), state (string), transient (bool), cpus (uint), maxcpus (uint), memory (uint64), osinit (string), boot ([]string), disk ([]resources.diskDevice), cdrom ([]resources.cdRomDevice), network ([]resources.networkDevice), filesystem ([]resources.filesystemDevice), auth (*resources.VirtAuth), hotcpus (bool), restartondiverge (string), restartonrefresh (bool)
```

It's ugly atm, eg: no proper indicator for readonly types.

lib/info.go
```diff
@@ -0,0 +1,124 @@
...
+
+var myPrintln = fmt.Printf
+
+// info wraps infocmd to produce output to stdout
+func info(c *cli.Context) error {
+	output, err := infocmd(c)
+	if err != nil {
+		return err
+	}
+	w := tabwriter.NewWriter(os.Stdout, 0, 8, 2, ' ', 0)
```
purpleidea: Can we have these vars be `const` (at the top), if only to improve readability please?

lib/info.go
```diff
@@ -0,0 +1,124 @@
...
+	"strings"
+	"text/tabwriter"
+
+	"github.com/purpleidea/mgmt/lang/funcs"
+	"github.com/purpleidea/mgmt/lang/interfaces"
+	"github.com/purpleidea/mgmt/resources"
+	"github.com/urfave/cli"
+)
+
+var myPrintln = fmt.Printf
```
purpleidea: what's this?

lib/info.go
```diff
@@ -0,0 +1,124 @@
...
+// infocmd takes the cli context and returns the requested output
+func infocmd(c *cli.Context) (bytes.Buffer, error) {
+	var out bytes.Buffer
+	var names []string
+	descriptions := make(map[string]string)
+
+	switch c.Command.Name {
+	case "resources":
+		for _, name := range resources.RegisteredResourcesNames() {
+			if name == "test" {
```
purpleidea: why skip it? It's real :)

lib/info.go
```diff
@@ -0,0 +1,124 @@
...
+
+			res, err := resources.Lookup(name)
+			if err != nil {
+				continue
+			}
+
+			s := reflect.ValueOf(res).Elem()
+			typeOfT := s.Type()
+
+			var fields []string
```
purpleidea: use, improve, or add to one of the functions in `resources/util.go` please,

if anything wouldn't we want the info sub commands to be separate functions?

lib/info.go
```diff
@@ -0,0 +1,124 @@
...
+			if strings.HasPrefix(name, "_") {
+				continue
+			}
+			names = append(names, name)
+			descriptions[name] = ""
+			fn, err := funcs.Lookup(name)
+			if err != nil {
+				continue
+			}
+			switch fn.(type) {
```
purpleidea: `_, ok := fn.(interfaces.PolyFunc)`

lib/info_test.go
```diff
@@ -0,0 +1,64 @@
...
+	ctx.Command = cli.Command{Name: "functions"}
+
+	out, err := infocmd(ctx)
+	if err != nil {
+		t.Fatal("failed")
+	}
+	assert.Contains(t, out.String(), "load() struct{x1 float; x5 float; x15 float}")
+}
+
+// func TestPolyFunctionsWithTypes(t *testing.T) {
```
purpleidea: add or delete?

resources/resources.go
```diff
@@ -673,3 +673,12 @@ func VtoR(v pgraph.Vertex) Res {
 //	}
 //	return nil
 //}
+
+// Lookup returns a pointer to the resource's struct.
```
purpleidea: Not sure why we need a helper function for this.

lib/info.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return err
+	}
+	w := tabwriter.NewWriter(os.Stdout, 0, 8, 2, ' ', 0)
+	fmt.Fprint(w, output.String())
+	w.Flush()
+	return nil
+}
+
+// infocmd takes the cli context and returns the requested output
```
purpleidea: golang style says `InfoCmd` or `infoCmd` please, but not sure why this is a separate function altogether

lib/info.go
aequitas: sure

lib/info.go
aequitas: leftover from the wrong approach to capture test output

lib/info_test.go
aequitas: scratchpad until i have polyfuncs figured out or for a 2.0 patch of this feature

resources/resources.go
aequitas: Just mirroring funcs Lookup here

lib/info.go
aequitas: to allow unit testing the output with jumping through hoops to capture it (eg: rerouting stdout, etc)

aequitas:
@purpleidea 

1: the resource rending is more a hack than a preferred implementation, if you know of a way to get the right output i'm all ears

2: if there is an easy library which can write this to XML in a open format suitable for this kind of data I will add it in, otherwise next patch.

resources/resources.go
aequitas: also registeredResources is private :)

purpleidea: I didn't realize this was in resources.go :P

purpleidea:
@aequitas Indeed, there is already such an XML lib as seen in yamlgraph2/ being used.

aequitas:
@purpleidea I don't see any xml reference in yamlgraph2? Can you point me to where it is?

purpleidea:
@aequitas Sorry, by brain accidentally some letters. Yaml*:

https://github.com/purpleidea/mgmt/blob/master/yamlgraph2/gconfig.go#L31

aequitas:
@purpleidea so you want xml output or yaml output? :) Why not both? We could maybe use YAML as an intermediate format for XML (and JSON for that matter :D)

purpleidea:
@aequitas Just yaml is sufficient :)

lib/info.go
```diff
@@ -0,0 +1,127 @@
...
+	"strings"
+	"text/tabwriter"
+
+	"github.com/purpleidea/mgmt/lang/funcs"
+	"github.com/purpleidea/mgmt/lang/interfaces"
+	"github.com/purpleidea/mgmt/resources"
+	"github.com/urfave/cli"
+)
+
+const (
```
purpleidea: https://golang.org/doc/effective_go.html#mixed-caps

probably these could be public if you want.

aequitas: i know, i'm following the names set by the tabwriter package, let me know if you prefer proper go names instead

purpleidea: oh gosh. Another inconsistency in their own style guide? Gah. Sorry. Do the right thing. If golangv2 fixes these things, we'll be ahead of the curve.

aequitas: fixed

aequitas:

PR #365, opened 2018-02-27
lang: Add file function for reading files

aequitas:
Todos:

- [x] streaming updates
- [ ] better error handling?
- [x] tests
- [ ] https://github.com/purpleidea/mgmt/pull/365/files#diff-a7d4a5bd50bf5e859cbd3bb62e51b65cR92 pending on fix for: https://github.com/purpleidea/mgmt/pull/367

### DIFF
--- a/lang/funcs/core/file_func.go
+++ b/lang/funcs/core/file_func.go
@@ -1,0 +1,156 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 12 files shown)


aequitas:
- [ ] test

lang/funcs/core/file_func.go
```diff
@@ -0,0 +1,103 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package core // TODO: should this be in its own individual package?
+
+import (
+	"io/ioutil"
+
+	errwrap "github.com/pkg/errors"
```
purpleidea: TestUnification1

aequitas: this is what goimports makes of it, can we put this rule into a config file for goimports? I really don't want to be managing imports (and their sorting) by hand. A computer can do this just fine :)

purpleidea: ¯\_(ツ)_/¯

I thought it did do this. If not, it's a bug, missing feature, or a property not explicitly enforced.
It is the common approach we use here and in is seen in many other projects too.

aequitas:
Rebase on latest master, made all test function `t.Helper` functions so they provide a better trace on a `t.Fatal`.

aequitas:

PR #367, opened 2018-03-05
File resource integrationtest

aequitas:
I spotted some deviant behaviour regarding the file resource when dealing with string vs dynamic inputs. When providing an empty string to the file resource a file is created, but when an function is used that returns an empty string the file is not created.


aequitas:

Issue #322, opened 2018-02-10
Enless loop of watching


aequitas:
## Versions:

* mgmt version (eg: `mgmt --version`): 0.0.14-13-g7e6d013-dirty

* operating system/distribution (eg: `uname -a`): Linux faalserver 3.16.0-4-amd64 #1 SMP Debian 3.16.51-2 (2017-12-03) x86_64 GNU/Linux 

* golang version (eg: `go version`): ?

## Description:

When converting Puppet code to YAML using @ffrank module some configurations can cause a lot of `Watching: <service name>.service` messages to be displayed. Mgmt never converges (not tested until eternity though).

Command used are in log output below.

Puppet source:
```
class base {
  notice("fqdn=${::fqdn}, env=${::env}")

  package { 'resolvconf': ensure => latest}
  ~> service { 'resolvconf': ensure => running, enable => true}
  package { 'dnsmasq': ensure => latest}
  ~> service { 'dnsmasq': ensure => running, enable => true}
}
```

Generate YAML:

```
---
graph: faalserver.faalkaart.test
comment: generated from puppet catalog for faalserver.faalkaart.test
resources:
  exec:
  - name: Package:resolvconf
    cmd: |-
      puppet yamlresource package 'resolvconf' '{"provider": "apt", "ensure": "latest", "configfiles": "keep", "reinstall_on_refresh": "false",
        "loglevel": "notice"}'
    timeout: 30
    shell: "/bin/bash"
    watchshell: "/bin/bash"
    ifshell: "/bin/bash"
    watchcmd: 'while : ; do echo "puppet run interval passed" ; /bin/sleep 1800 ;
      done'
    ifcmd: |-
      puppet yamlresource package 'resolvconf' '{"provider": "apt", "ensure": "latest", "configfiles": "keep", "reinstall_on_refresh": "false",
        "loglevel": "notice"}' --noop --color=false | grep -q ^Notice:
    state: present
    pollint: 0
  - name: Package:dnsmasq
    cmd: |-
      puppet yamlresource package 'dnsmasq' '{"provider": "apt", "ensure": "latest", "configfiles": "keep", "reinstall_on_refresh": "false",
        "loglevel": "notice"}'
    timeout: 30
    shell: "/bin/bash"
    watchshell: "/bin/bash"
    ifshell: "/bin/bash"
    watchcmd: 'while : ; do echo "puppet run interval passed" ; /bin/sleep 1800 ;
      done'
    ifcmd: |-
[...]

PR #368, opened 2018-03-09
test, integration: Add an integration test framework

purpleidea:
This adds an initial implementation of an integration test framework for
writing more complicated tests. In particular this also makes some small
additions to the mgmt core so that testing is easier.

### DIFF
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -1,0 +1,130 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package integration
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"sort"
+	"testing"
+)
+
+func TestInstance0(t *testing.T) {
+	code := `
+	$root = getenv("MGMT_TEST_ROOT")
+
+	file "${root}/mgmt-hello-world" {
+		content => "hello world from @purpleidea\n",
+		state => "exists",
+	}
+	`
+	m := Instance{
+		Hostname: "h1", // arbitrary
+		Preserve: true,
+	}
+	if err := m.SimpleDeployLang(code); err != nil {
+		t.Errorf("failed with: %+v", err)
+		return
+	}
+	d := m.Dir()
+	t.Logf("test ran in:\n%s", d)
+	root := path.Join(d, RootDirectory)
+	file := path.Join(root, "mgmt-hello-world")
+	_, err := os.Stat(file)
+	if err != nil {
+		t.Errorf("could not find file: %+v", err)
+		return
+	}
+}
+
+func TestInstance1(t *testing.T) {
+	type test struct { // an individual test
+		name   string
+		code   string // mcl code
+		fail   bool
+		expect map[string]string
+	}
+	values := []test{}
+
+	{
+		code := Code(`
+		$root = getenv("MGMT_TEST_ROOT")
+
+		file "${root}/mgmt-hello-world" {
+			content => "hello world from @purpleidea\n",
+			state => "exists",
+		}
+		`)
+		values = append(values, test{
+			name: "hello world",
+			code: code,
+			fail: false,
+			expect: map[string]string{
+				"mgmt-hello-world": "hello world from @purpleidea\n",
+			},
+		})
+	}
+
+	for index, test := range values { // run all the tests
+		t.Run(fmt.Sprintf("test #%d (%s)", index, test.name), func(t *testing.T) {
+			code, fail, expect := test.code, test.fail, test.expect
+
+			m := Instance{
+				Hostname: "h1",
+				Preserve: true,
+			}
+			err := m.SimpleDeployLang(code)
+			d := m.Dir()
+			if d != "" {
+				t.Logf("test ran in:\n%s", d)
+			}
+
+			if !fail && err != nil {
+				t.Errorf("failed with: %+v", err)
+				return
+			}
+			if fail && err == nil {
+				t.Errorf("passed, expected fail")
+				return
+			}
+
+			if expect == nil { // test done early
+				return
+			}
+
+			files := []string{}
+			for x := range expect {
+				files = append(files, x)
+			}
+			sort.Strings(files) // loop in a deterministic order
+			for _, f := range files {
+				filename := path.Join(d, RootDirectory, f)
+				b, err := ioutil.ReadFile(filename)
+				if err != nil {
+					t.Errorf("could not read file: `%s`", filename)
+					continue
+				}
+				if expect[f] != string(b) {
+					t.Errorf("file: `%s` did not match expected", f)
+				}
+			}
+		})
+	}
+}
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -128,3 +128,125 @@ func TestInstance1(t *testing.T) {
 		})
 	}
 }
+
+func TestCluster0(t *testing.T) {
+	// TODO: implement a simple test for documentation purposes
+}
+
+func TestCluster1(t *testing.T) {
+	type test struct { // an individual test
+		name   string
+		code   string // mcl code
+		fail   bool
+		hosts  []string
+		expect map[string]map[string]string // hostname, file, contents
+	}
+	values := []test{}
+
+	{
+		code := Code(`
+		$root = getenv("MGMT_TEST_ROOT")
+
+		file "${root}/mgmt-hostname" {
+			content => "i am ${hostname()}\n",
+			state => "exists",
+		}
+		`)
+		values = append(values, test{
+			name:  "simple pair",
+			code:  code,
+			fail:  false,
+			hosts: []string{"h1", "h2"},
+			expect: map[string]map[string]string{
+				"h1": {
+					"mgmt-hostname": "i am h1\n",
+				},
+				"h2": {
+					"mgmt-hostname": "i am h2\n",
+				},
+			},
+		})
+	}
+	{
+		code := Code(`
+		$root = getenv("MGMT_TEST_ROOT")
+
+		file "${root}/mgmt-hostname" {
+			content => "i am ${hostname()}\n",
+			state => "exists",
+		}
+		`)
+		values = append(values, test{
+			name:  "hello world",
+			code:  code,
+			fail:  false,
+			hosts: []string{"h1", "h2", "h3"},
+			expect: map[string]map[string]string{
+				"h1": {
+					"mgmt-hostname": "i am h1\n",
+				},
+				"h2": {
+					"mgmt-hostname": "i am h2\n",
+				},
+				"h3": {
+					"mgmt-hostname": "i am h3\n",
+				},
+			},
+		})
+	}
+
+	for index, test := range values { // run all the tests
+		t.Run(fmt.Sprintf("test #%d (%s)", index, test.name), func(t *testing.T) {
+			code, fail, hosts, expect := test.code, test.fail, test.hosts, test.expect
+
+			c := Cluster{
+				Hostnames: hosts,
+				Preserve:  true,
+			}
+			err := c.SimpleDeployLang(code)
+			if d := c.Dir(); d != "" {
+				t.Logf("test ran in:\n%s", d)
+			}
+
+			if !fail && err != nil {
+				t.Errorf("failed with: %+v", err)
+				return
+			}
+			if fail && err == nil {
+				t.Errorf("passed, expected fail")
+				return
+			}
+
+			if expect == nil { // test done early
+				return
+			}
+
+			instances := c.Instances()
+			for _, h := range hosts {
+				instance := instances[h]
+				d := instance.Dir()
+				hexpect, exists := expect[h]
+				if !exists {
+					continue
+				}
+
+				files := []string{}
+				for x := range hexpect {
+					files = append(files, x)
+				}
+				sort.Strings(files) // loop in a deterministic order
+				for _, f := range files {
+					filename := path.Join(d, RootDirectory, f)
+					b, err := ioutil.ReadFile(filename)
+					if err != nil {
+						t.Errorf("could not read file: `%s`", filename)
+						continue
+					}
+					if hexpect[f] != string(b) {
+						t.Errorf("file: `%s` did not match expected", f)
+					}
+				}
+			}
+		})
+	}
+}
--- a/integration/basic_test.go
+++ b/integration/basic_test.go
@@ -41,6 +41,9 @@ func TestInstance0(t *testing.T) {
 	}
 	if err := m.SimpleDeployLang(code); err != nil {
 		t.Errorf("failed with: %+v", err)
+		if output, err := m.CombinedOutput(); err == nil {
+			t.Errorf("logs from instance:\n\n%s", output)
+		}
 		return
 	}
 	d := m.Dir()
@@ -98,6 +101,9 @@ func TestInstance1(t *testing.T) {
 
 			if !fail && err != nil {
 				t.Errorf("failed with: %+v", err)
+				if output, err := m.CombinedOutput(); err == nil {
+					t.Errorf("logs from instance:\n\n%s", output)
+				}
 				return
 			}
 			if fail && err == nil {
@@ -207,9 +213,15 @@ func TestCluster1(t *testing.T) {
 			if d := c.Dir(); d != "" {
 				t.Logf("test ran in:\n%s", d)
 			}
+			instances := c.Instances()
 
 			if !fail && err != nil {
 				t.Errorf("failed with: %+v", err)
+				for _, h := range hosts {
+					if output, err := instances[h].CombinedOutput(); err == nil {
+						t.Errorf("logs from instance `%s`:\n\n%s", h, output)
+					}
+				}
 				return
 			}
 			if fail && err == nil {
@@ -221,7 +233,6 @@ func TestCluster1(t *testing.T) {
 				return
 			}
 
-			instances := c.Instances()
 			for _, h := range hosts {
 				instance := instances[h]
 				d := instance.Dir()
--- a/integration/cluster.go
+++ b/integration/cluster.go
@@ -1,0 +1,224 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"time"
+
+	multierr "github.com/hashicorp/go-multierror"
+	errwrap "github.com/pkg/errors"
+)
+
+// Cluster represents an mgmt cluster. It uses the instance building blocks to
+// run clustered tests.
+type Cluster struct {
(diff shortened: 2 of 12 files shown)


purpleidea:
@aequitas Hey,

So I took some time to hack out integration testing how I was imagining it. I initially started with a thorough review of your code in #353, but then I changed it around to have what is hopefully an easier API, and one that works with events without time.Sleep anywhere.

I hope you don't mind, and hopefully you like it. Please review if you'd like, and I can merge and you can use this to add your magnificent tests on.

I'm also working on a clustered version of this-- you can see the beginnings of this in the seeds parameter. Lastly, you'll see a few "patterns" in the patterns.go file, which I think we can use to add more general patterns, and then re-use them as much as we want, in particular as seen in basic_test.go

Thanks!

aequitas:
@purpleidea I'm glad you finally got around to this and I have some understanding for the time it took. But I'm really frustrated and disappointed by the point we are at now. All the time and effort I spent on constructing and asking for feedback, only getting vague discussions, iterating over it, you're refusal to give proper feedback/discussion due to principal reasons, having me ask outsiders, incorporating their advice and you still abandoning that.

I admit the initial version of my implementation is not complete nor perfect. But I wanted to have something to work with and iterate on the basic principle before draining time into fully expanding all features like proper waits instead of sleep, structs and common patterns. For example a test like this https://github.com/purpleidea/mgmt/pull/365/files#diff-a7d4a5bd50bf5e859cbd3bb62e51b65cR75 already deviated heavily from the standard pattern. And this might not be just an exception. Causing a to simple API to create a high complexity in the testcases themselves making them useless imho as they overshadow the complexity of the thing they try to test. 

All in all I'm left bitter atm regarding contributing to this project. I'll see if this feeling fades away in the upcoming days and if I feel enticed to leave the comments I have on this PR.

purpleidea:
@aequitas I'm sorry about that. My goal wasn't to make you feel bitter, sorry. I would have liked to spend more time on iterating on your branch, but I realized with the amount of changes to get the converge detection working with mgmt core (as I patched here) was too much time.

I also felt bad because I knew the lack of integration testing was blocking you, and it was something we needed for the project anyways. As a result, I decided to just write it "as an example", rather than leaving many review comments on your PR. I didn't merge it yet, because my goal wasn't to make you feel bad about your PR, but rather have something to look at, compare, and discuss.

If you think there are changes we should make to this before I merge it, please let me know.
Hope to have you back and hacking on things.

Please understand that it's a big challenge to run such a project on my own time, and not all the decisions I take are perfect. Thanks for reading!

purpleidea:
I've update the branch with some additions, including a small cluster executor. It all works, however it does hit some occasional intermittent failures which are due to bugs elsewhere in the code. Those are also sometimes experienced by the shell tests, and I hope to resolve all of these in the future :)

aequitas:
@purpleidea It doesn't feel like an example to me, more like your implementation of the same. Instead of providing feedback on what I've spend time on and commenting on the remarks I left in code and in the PR, it feels to me like you just shove it aside with your version. I already had idea's for convergence checking and I had clustering all worked out, only I removed that last part as I ran into the same issue in that it requires some bugfixing in core before it's stable. I just wanted to know if you agreed with the general layout and API before sinking even more time into this. But you refused to even look at it until I refactored it.

I will put this aside for the weekend and come back on this beginning next week.

purpleidea:
@aequitas Sorry again. I just have a lot on my plate and wish I could have handled it better. Hope to have you happy and hacking again soon.

Thanks,
James

purpleidea:
I haven't heard anything for a while, so I'm going to merge this for now.
Happy to change it down the road if necessary.

integration/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+		instance, exists := obj.instances[h]
+		if !exists {
+			continue
+		}
+		if e := instance.Close(); e != nil {
+			err = multierr.Append(err, e)
+		}
+	}
+	if !obj.Preserve {
+		if obj.dir == "" || obj.dir == "/" {
```
aequitas: We might consider any path not created by TempDir to be dangerous, it's a user error of course, as the directory for this should only be a testrun specific tmpdir anyways. But still?

aequitas: Also why fail here and not during Init when the directory is set?

integration/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+			panic("obj.dir is set to a dangerous path")
+		}
+		if err := os.RemoveAll(obj.dir); err != nil { // dangerous ;)
+			return errwrap.Wrapf(err, "can't remove instance dir")
+		}
+	}
+	return err
+}
+
+// RunLinear starts up each instance linearly, one at a time.
```
aequitas: Good to be explicit about this.

integration/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+// to call wait on each member individually.
+func (obj *Cluster) Wait(ctx context.Context) error {
+	var err error
+	for _, h := range obj.Hostnames {
+		instance, exists := obj.instances[h]
+		if !exists {
+			continue
+		}
+		// TODO: do we want individual waits?
+		//ctx, cancel := context.WithTimeout(context.Background(), longTimeout*time.Second)
```
aequitas: Should this commented code be removed? If not you should explain with a comment why it is still in here, as to me it looks like dead code.

integration/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+// Remember to leave a longer timeout when using a context since this will have
+// to call wait on each member individually.
+func (obj *Cluster) Wait(ctx context.Context) error {
+	var err error
+	for _, h := range obj.Hostnames {
+		instance, exists := obj.instances[h]
+		if !exists {
+			continue
+		}
+		// TODO: do we want individual waits?
```
aequitas: Wait in parallel is preferred here imho.

integration/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+		//ctx, cancel := context.WithTimeout(context.Background(), longTimeout*time.Second)
+		//defer cancel()
+		if e := instance.Wait(ctx); e != nil {
+			err = multierr.Append(err, e)
+		}
+	}
+	return err
+}
+
+// DeployLang deploys some code to the cluster. It arbitrarily picks the first
```
aequitas: Good comment on deterministic behaviour and that there is no reason why the first node is preferred.

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+	"strings"
+	"sync"
+	"syscall"
+
+	"github.com/purpleidea/mgmt/recwatch"
+
+	errwrap "github.com/pkg/errors"
+)
+
+const (
```
aequitas: Good documenting of the constants.

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+
+// Instance represents a single running mgmt instance. It is a building block
+// that can be used to run standalone tests, or combined to run clustered tests.
+type Instance struct {
+	// Hostname is a unique identifier for this instance.
+	Hostname string
+
+	// Preserve prevents the runtime output from being explicitly deleted.
+	// This is helpful for running analysis or tests on the output.
+	Preserve bool
```
aequitas: Like said before I think there is no reason why this should be an Instance attribute. Now you have to pass it around everywhere. Maybe better to always preserve on error and set this as a testsuite constant or a test flag instead so it can be enabled if you really want to see the output of a passed test?

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+		}
+	}
+
+	tmpPrefix := path.Join(obj.dir, PrefixDirectory)
+	if err := os.MkdirAll(tmpPrefix, dirMode); err != nil {
+		return errwrap.Wrapf(err, "can't create prefix directory")
+	}
+	obj.tmpPrefixDirectory = tmpPrefix
+
+	testRootDirectory := path.Join(obj.dir, RootDirectory)
```
aequitas: root is indeed a better name than workdir for this

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+// Run launches the instance. It returns an error if it was unable to launch.
+func (obj *Instance) Run(seeds []*Instance) error {
+	if obj.cmd != nil {
+		return fmt.Errorf("an instance is already running")
+	}
+
+	if len(seeds) == 0 {
+		// set so that Deploy can know where to connect
+		// also set so that we can allow others to find us and connect
+		obj.clientURL = "http://127.0.0.1:2379"
```
aequitas: I've had many times where race conditions causes tests to fail as old mgmt instances where occupying ports that where assumed to be free by the testsuite. A solution I was thinking about is using tcp dial on a 0 port to get a available port from the operating system, like: https://github.com/phayes/freeport/blob/master/freeport.go#L9.

aequitas: Ideally instead of tcp, unix domain sockets would be used during test to completely isolate tests and allow for parallel runs. This of course has to wait for the implementation.

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+		urls := []string{}
+		for _, instance := range seeds {
+			if instance.cmd == nil {
+				return fmt.Errorf("instance `%s` has not started yet", instance.Hostname)
+			}
+			urls = append(urls, instance.clientURL)
+		}
+		s := fmt.Sprintf("--seeds=%s", urls[0])
+		// TODO: we could just add all the seeds instead...
+		//s := fmt.Sprintf("--seeds=%s", strings.Join(urls, ","))
```
aequitas: Since we will be testing the behaviour of how instances in cluster behave given different configurations of seeds either way can be expected to work the same. Is there any (speed) benefit to giving all seeds?

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+func (obj *Instance) Kill() error {
+	if obj.cmd == nil {
+		return nil // already dead
+	}
+	if obj.cmd.Process == nil {
+		return nil // nothing running
+	}
+
+	// cause a stack dump first if we can
+	_ = obj.cmd.Process.Signal(syscall.SIGQUIT)
```
aequitas: nice

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+	obj.cmd = nil
+	return err
+}
+
+// Wait until the first converged state we hit. It is not necessary to use the
+// `--converged-timeout` option with mgmt for this to work. It tracks this via
+// the `--converged-status-file` option which can be used to track the varying
+// convergence status.
+func (obj *Instance) Wait(ctx context.Context) error {
+	//if obj.cmd == nil { // TODO: should we include this?
```
aequitas: yes

integration/instance.go
```diff
@@ -0,0 +1,392 @@
...
+	wg.Wait()
+	obj.cmd = nil
+	return err
+}
+
+// Wait until the first converged state we hit. It is not necessary to use the
+// `--converged-timeout` option with mgmt for this to work. It tracks this via
+// the `--converged-status-file` option which can be used to track the varying
+// convergence status.
+func (obj *Instance) Wait(ctx context.Context) error {
```
aequitas: how does this handle race conditions? where mgmt instance/cluster is (partly) converged but Wait is not called?

integration/instance.go
purpleidea: Agreed! I just did it this way since it's the only way I knew. For now it seems to not conflict, but if we experience issue in the future, we can change it.

integration/instance.go
purpleidea: I don't think there's a performance difference, in general it's preferable to connect with more seeds in case one happens to go down before you get the full list. Once you connect to anyone, you get the full list. I left it this way since when I refactor etcd/etcd.go I want to make sure there's no bug here, and I wanted to remind myself to at least see if there's a need to test both or not.

integration/instance.go
purpleidea: What do you mean?

integration/basic_test.go
```diff
@@ -0,0 +1,263 @@
...
+		fail   bool
+		expect map[string]string
+	}
+	values := []test{}
+
+	{
+		code := Code(`
+		$root = getenv("MGMT_TEST_ROOT")
+
+		file "${root}/mgmt-hello-world" {
```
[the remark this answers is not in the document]
purpleidea: Of course! Your idea of using `getenv` in here was great!!

aequitas:
I'm done with this project for now. 

Clearly this feature was not urgent to merge. But you went ahead and merged it without asking if I would still leave comments as I promised I would come back on. If you don't have the patience to wait at least a few days where I waited weeks and spent effort and time to at least have you look and discuss the general principle. In chat you stated you would like people to review your code but as it turns out you rather just merge your own code as soon as you can. 

You scream for contributors to this project but I have the feeling you only see them as code monkeys having to write the code how you would write it, instead of making them part of a team that together builds on something great. I can bare most of your nitpicking and the principles you adhere to for this project. You probably mean well and you have good ideas, but the way you act as lead for this project just frustrates me and doesn't make it seem the effort for me to contribute any further.

integration/basic_test.go
```diff
@@ -0,0 +1,263 @@
...
+			content => "hello world from @purpleidea\n",
+			state => "exists",
+		}
+		`)
+		values = append(values, test{
+			name: "hello world",
+			code: code,
+			fail: false,
+			expect: map[string]string{
```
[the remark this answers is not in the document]
purpleidea: Actually, this is temporary. I agree that `expect` should be a map of some `Expect` interface. That interface could be fulfilled with file checker (which could test for exists and !exists) and any other sort of thing we want to build. I couldn't think of anything obvious to test at the moment, so I didn't generalize it yet.

integration/basic_test.go
```diff
@@ -0,0 +1,263 @@
...
+			content => "i am ${hostname()}\n",
+			state => "exists",
+		}
+		`)
+		values = append(values, test{
+			name:  "simple pair",
+			code:  code,
+			fail:  false,
+			hosts: []string{"h1", "h2"},
+			expect: map[string]map[string]string{
```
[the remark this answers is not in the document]
purpleidea: Agreed! See reply in https://github.com/purpleidea/mgmt/pull/368#discussion_r174101483
This should be map[string]Expect interface.

integration/basic_test.go
```diff
@@ -0,0 +1,263 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"sort"
+	"testing"
+)
+
+func TestInstance0(t *testing.T) {
```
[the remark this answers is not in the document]
purpleidea: Docs would be nice. Missing docs in tests isn't my no.1 prio at this point though. Sorry!

integration/basic_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+	for index, test := range values { // run all the tests
+		t.Run(fmt.Sprintf("test #%d (%s)", index, test.name), func(t *testing.T) {
+			code, fail, hosts, expect := test.code, test.fail, test.hosts, test.expect
+
+			c := Cluster{
+				Hostnames: hosts,
+				Preserve:  true,
+			}
+			err := c.SimpleDeployLang(code)
```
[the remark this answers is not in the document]
purpleidea: I was actually thinking about this, but couldn't find a benefit to actually doing it. At least not until there's some more general thing that wants to consume that interface. However, in that case, we'd probably just only pass in "clusters", which might have one instance or might not.

integration/basic_test.go
```diff
@@ -0,0 +1,263 @@
...
+		state => "exists",
+	}
+	`
+	m := Instance{
+		Hostname: "h1", // arbitrary
+		Preserve: true,
+	}
+	if err := m.SimpleDeployLang(code); err != nil {
+		t.Errorf("failed with: %+v", err)
+		if output, err := m.CombinedOutput(); err == nil {
```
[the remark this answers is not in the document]
purpleidea: ?? Not sure I understand the point, but please don't be sad. It's just code. This isn't a perfect process either.

integration/cluster.go
purpleidea: Agreed. We could also add that, however this is just a "bonus" safety check because I'm a scaredy cat.

purpleidea:
@aequitas 

> I'm done with this project for now.

Sorry to see you go. You're welcome back anytime. I've just got a lot that I'm trying to balance right now, and sorry it didn't all work out in the perfect way. 

Happy Hacking.

aequitas:
@purpleidea 

Thanks for feedback on some of the comments I made.

I closed all my open PR's as most of them require work to adapt to your implementation or have been lingering for to long without conclusion.

> I've just got a lot that I'm trying to balance right now, and sorry it didn't all work out in the perfect way.

I understand your situation but if I may be brutally honest you should really consider how to best spend the time you allocate to this project. To me it just feels that rewriting an entire implementation as example isn't the most efficient way. I think you are a great coder and the standards set for this project are high, which can be a good thing. But if you don't want to end up being the only one having to write all the code you should really consider being more lenient sometimes. Also, and I know your busy and it's your project so I really shouldn't complain, but it seems to me you adhere more strict rules for contributors then for your own commits. Maybe ask other frequent contributors if they like to chime in on reviews so it's not only your voice that does all the complaining.

I set out contributing to mgmt with the idea to answer the question 'does mgmt run in production' with a confident 'yes, and its everything you can wish for'. Meaning I would tackle every problem, small or big, that I encountered to have it fit what I consider a production ready tool that I confidently can 'sell' to co-workers as I don't have warn for sharp edges. But right now the velocity I need to work on this fulltime can't be supported, which creates friction and frustration. That together with how this PR has been handled have made me decide that, for my own sanity, I won't be contributing for a while.

integration/basic_test.go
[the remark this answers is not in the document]
aequitas: Imho, tests are important, more important than code, as that is just the implementation, the test is the proof that code provides the functionality that is required. So for me tests, their implementation and documentation are equal if not higher priority.

integration/basic_test.go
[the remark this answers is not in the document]
aequitas: I agree, it's just code. I'm just sad that the experience and lessons that are in my PR where lost when you rewrote your implementation only to add them again later on when you've spend time discovering it yourself.

integration/instance.go
aequitas: You need to set the converged-timeout to a high value, making waiting for it kinda useless as a sleep would almost suffice as well. (a low timeout is preferred to have faster tests). But with a to low value you have the risk for a configuration to converge after you deploy but before the moment you start waiting for convergence. 

Ideally you would wait for a specific deploy ID to converge instead of just any sign of convergence.

purpleidea:
> To me it just feels that rewriting an entire implementation as example isn't the most efficient way.

There is technical debt which I introduced when I was less familiar with golang, which is causing bugs and intermittent test failures, along with #322 which much be fixed before the project can be considered stable.

> I think you are a great coder and the standards set for this project are high, which can be a good thing. But if you don't want to end up being the only one having to write all the code you should really consider being more lenient sometimes.

No. I'm not taking on any more technical debt that I can't easily patch out. An integration test framework is critical if a lot of tests are added on top of it, because if it ever needs fixing, I either have to nuke all the tests or spend far too much time trying to fix it.

> Also, and I know your busy and it's your project so I really shouldn't complain, but it seems to me you adhere more strict rules for contributors then for your own commits.

Possibly. But the difference is that I'm ultimately responsible for both my code and contributors, so it has to be right.

> Maybe ask other frequent contributors if they like to chime in on reviews so it's not only your voice that does all the complaining.

Others are welcome and encouraged too if their feedback is correct and productive.

aequitas:
> No. I'm not taking on any more technical debt that I can't easily patch out. An integration test framework is critical if a lot of tests are added on top of it, because if it ever needs fixing, I either have to nuke all the tests or spend far too much time trying to fix it.

I'm not saying you should open the floodgates, thats the other extreme, just to accept that your way might not be the only correct one and trust the loyalty of others to come back and iterate. Which I will admit is a hard thing to do with strangers over the internet that can decide to disconnect from this project at any moment and leave you with the broken pieces. 

I would be glad to help with chores and fixing technical debt not related to my immediate goals just as a service to this project. But for that I would need the feeling to be part of a team, right now I don't feel compelled as I have the feeling it just you and only your opinion micromanaging the project and it reminds me of myself in former projects.

I honestly believe your intentions and values are good and you strive for openness, but I think you could do a more active role in delegating and involving people into collaborating on the project together with you instead of pulling everything towards yourself. The mgmtlove issues are a good start, but if it ends up with just only you reviewing them and a contributor walking away half way there is nothing gained. Ideally there would be a team of core enthousiast that carry the project, share the chores, discuss broader subjects and form opinions. This also makes the project easier to 'sell' as production ready to colleagues as right now it's seems just you and some on/off contributors and that's a mighty low bus factor. I'm not involved that long in the project (just weeks), so my view might be completely distorted, correct me if I'm wrong.

In the end it is your project and your codebase, every decision is yours to make. This is just my opinion and you may take it at the value it is worth to you.

One last note, the reason I'm so emotional about this and keep carrying on is because I've come believe in mgmt and dare I say it have fallen in love with it. But the friction, frustration and realisations of the past days made me sad in that this project might never get the tractions it needs to consider it a solution I can actually can get implemented somewhere, which forces me to take a binary decision for now.

purpleidea:
@aequitas I've already apologized multiples times publicly, in channel and DM. I don't think I can improve upon this anymore.

I'm not a perfect maintainer, and it's very difficult me to review patches which are substantially far away from what I'm expecting in terms of design and code quality. It's a skill I need to improve on. To be clear, it's not because I think you're not bright or a good coder, you're obviously both, but you're very new to golang, and at this early stage, it's my mistake for not immediately saying "please start with easier/smaller patches first".

> One last note, the reason I'm so emotional about this and keep carrying on is because I've come believe in mgmt and dare I say it have fallen in love with it.

Well please come back when you want, but if so, then you need to not get offended so easily here.

Thanks

PR #371, opened 2018-03-11
resources: Add net resource based on netlink

jonathangold:
This is very much a work in progress. Let's discuss before you review anything carefully. I haven't yet combined this with the systemd code, because I really like the netlink based ipCheckApply, and I want to discuss with you which parts to keep from each.


purpleidea:
@jonathangold So when mgmt is running, this is good. The big catch is that if you reboot the system, it won't have any state about what IP to use initially, which could make discovering the rest of the cluster difficult.

As a result, we *also* need some mechanism to save that state in something that's part of the core system. I think that should be in this resource, but I'm happy to discuss. It should at least be possible under a flag. I think that mechanism should be networkd because that's a cross distro way to do it, that's better than `/etc/sysconfig/network-scripts/*` (RPM) and `/etc/network/interfaces` (DEB).

jonathangold:
@purpleidea Agreed. I think that fileCheckApply (and the corresponding recWatcher) would be pretty easy to cherry-pick. I guess we should also make sure the networkd service is enabled (it's not by default on some systems afaik)

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+	RegisterResource("net", func() Res { return &NetRes{} })
+}
+
+const (
+	// ifaceUp is the up (on) interface state.
+	ifaceUp = "up"
+	// ifaceDown is the down (off) interface state.
+	ifaceDown = "down"
+
+	// netlink multicast groups to watch for events
```
purpleidea: could you add a docs link in the comments about where all these magic numbers come from please?
same below for rtProto

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+
+	// ip routing protocols for used for netlink route messages
+	rtProtoKernel = 2 // kernel
+	rtProtoStatic = 4 // static
+
+	// networkdUnitFileDir is the location of networkd unit files which define
+	// the systemd network connections.
+	networkdUnitFileDir = "/etc/systemd/network/"
+	// IfacePrefix is the prefix used to identify unit files for managed links.
+	IfacePrefix = "mgmt-"
```
purpleidea: Put the public one at the top of the const list please.

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+			MetaParams: DefaultMetaParams, // force a default
+		},
+	}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *NetRes) Validate() error {
+	// validate state
+	if obj.State != ifaceUp && obj.State != ifaceDown && obj.State != "" {
+		return fmt.Errorf("state must be up down or empty")
```
purpleidea: up COMMA down

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+// nlChanStruct defines the channel used to send netlink messages and errors
+// to the event processing loop in Watch.
+type nlChanStruct struct {
+	msg []syscall.NetlinkMessage
+	err error
+}
+
+// iface wraps net.Interface to add additional methods.
+type iface struct {
+	*net.Interface
```
purpleidea: What is the reason this is anonymous?
It's okay if there's a useful reason, if not, it's good to be explicit so we know where a method comes from.
For discussion instead of immediate change.

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+	}
+
+	// build the path to the networkd configuration file
+	obj.unitFilePath = networkdUnitFileDir + IfacePrefix + obj.GetName() + networkdUnitFileExt
+
+	return obj.BaseRes.Init()
+}
+
+// Watch listens for events from the specified interface via a netlink socket.
+// XXX: currently gets events from ALL interfaces, would be nice to reject
```
purpleidea: XXX should mean "this is bad, fix ASAP". Is it that bad?? If not maybe a FIXME (good thing we should do) or a TODO (nice to have one day)

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+	}
+	if obj.Gateway != "" {
+		u = append(u, fmt.Sprintf("Gateway=%s", obj.Gateway))
+	}
+	c := strings.Join(u, "\n")
+	return []byte(c)
+}
+
+// state reports the state of the interface as up or down.
+func (i *iface) state() (string, error) {
```
purpleidea: if you're adding methods on a struct, the convention is usually to have them immediately follow the struct. so perhaps move it to down here?

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+	var err error
+	if i.Interface, err = net.InterfaceByName(i.Name); err != nil {
+		return "", errwrap.Wrapf(err, "error updating interface")
+	}
+	if i.Flags&net.FlagUp == 0 {
```
purpleidea: add a small comment please

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+	fdSet.Bits[fd/64] |= 1 << uint(fd) // fd = 3 becomes 100 if we add 5, we get 10100
+}
+
+// strSliceCompare takes two lists of strings and returns whether or not they
+// are identical.
+func strSliceCompare(x, y []string) bool {
+	if len(x) != len(y) {
+		return false
+	}
+	sort.Strings(x)
```
purpleidea: I am stupid and forget, but doesn't this modify the input args? Check it doesn't have a side effect please :)

Also your docs are not exactly correct for this function if you sort the two before compare.

Lastly, please add this function to util/util.go package with a test (at least testing the input modification thing) or two please :) Signature should be: `x, y []string) error` instead.

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+// down as necessary.
+func (obj *NetRes) ifaceCheckApply(apply bool) (bool, error) {
+	// check the interface state
+	state, err := obj.iface.state()
+	if err != nil {
+		return false, errwrap.Wrapf(err, "error checking %s state", obj.GetName())
+	}
+	// if up is unspecified
+	if obj.State == "" {
+		obj.State = state // send the state before returning
```
purpleidea: This is incorrect. Do not send state. If state was unspecified it means "we're not touching it", but now we've said "make sure it's up". Easy one line fix though :)

Logic below will change though.

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+	}
+
+	// end of state checking
+	if !apply {
+		return false, nil
+	}
+	log.Printf("%s: gatewayCheckApply(%t)", obj, apply)
+
+	// delete all but one default route
+	for n, route := range defRoutes {
```
purpleidea: `for i:=1` might work better here...

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+		}
+	}
+
+	if !apply {
+		return false, nil
+	}
+	log.Printf("%s: fileCheckApply(%t)", obj, apply)
+
+	// write the file
+	if err := ioutil.WriteFile(obj.unitFilePath, contents, 0644); err != nil {
```
purpleidea: put the umask in a const please

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+		return err
+	}
+	// close the recwatcher when we're done
+	defer recWatcher.Close()
+
+	// channel to unblock selects in goroutine
+	closeChan := make(chan struct{})
+	defer close(closeChan)
+
+	wg.Add(1)
```
purpleidea: Since the pipe watching code is sufficiently complicated, we might want to refactor it out into a separate self contained struct (with associated methods) so that:

1) This code becomes more readable.
2) It's a standalone piece that we could use elsewhere, in particular in a test :) [future patch]

So backup this code first, but then build a small API for this, which we'll then consume here. You can include this all at the bottom of the file.

resources/net.go
```diff
@@ -0,0 +1,849 @@
...
+	// networkdUnitFileDir is the location of networkd unit files which define
+	// the systemd network connections.
+	networkdUnitFileDir = "/etc/systemd/network/"
+	// IfacePrefix is the prefix used to identify unit files for managed links.
+	IfacePrefix = "mgmt-"
+	// networkdUnitFileExt is the file extension for networkd unit files.
+	networkdUnitFileExt = ".network"
+)
+
+// NetRes is an systemd-networkd resource.
```
purpleidea: You should put a much longer description here please :)

jonathangold:
@purpleidea Thanks for the review. I'll get on the changes tomorrow. :D

resources/net.go
jonathangold: TBH, the struct didn't have the netlink.Link to begin with, and when I added it, I never  gave the *net.Interface a name. I don't have a problem changing it.

resources/net.go
jonathangold: You are correct. We discussed this, and I forgot to change it. Coming from Haskell/Rust this seems like something the language shouldn't do without a pointer reference. But, what do I know?
In what case should this return an error (instead of false?)

purpleidea: if we return bool, we have two possible values. if we return error we have two possible values (nil vs. err) but in the second case we can provide a *reason* which is more cool in case anyone cares.

jonathangold:
@purpleidea Done with the changes.
I named the previously anonymous *net.Interface, but I'm not sure I'm totally sold on the idea. Maybe we can discuss before you merge.

resources/util.go
```diff
@@ -270,3 +270,27 @@ func GetGID(group string) (int, error) {
 
 	return -1, errwrap.Wrapf(err, "group lookup error (%s)", group)
 }
+
+// strSliceCompare takes two lists of strings and returns whether or not they
+// are equivalent. It will return nil if both sets contain the same elements,
+// regardless of order, and an error if they do not.
+func strSliceCompare(a, b []string) error {
```
purpleidea: This should be public please. Do you mind renaming to strSortedSliceCompare to be more clear that it sorts?

resources/net.go
```diff
@@ -0,0 +1,877 @@
...
+	rtmGrpIPv4IfRoute = 0x40  // add delete routes
+
+	// IP routing protocols for used for netlink route messages. For all
+	// protocols see:
+	// https://github.com/torvalds/linux/blob/master/include/uapi/linux/rtnetlink.h
+	rtProtoKernel = 2 // kernel
+	rtProtoStatic = 4 // static
+
+	// IfacePrefix is the prefix used to identify unit files for managed links.
+	IfacePrefix = "mgmt-"
```
purpleidea: Forgotten review item: https://github.com/purpleidea/mgmt/pull/371#discussion_r176905420

resources/net.go
```diff
@@ -0,0 +1,877 @@
...
+			return errwrap.Wrapf(err, "error adding addr: %s to %s", objAddr, i.iface.Name)
+		}
+	}
+	return nil
+}
+
+// sockSet is used to receive events from a socket and shut it down cleanly
+// when asked. It contains a socket for events and a pipe socket to unblock
+// receive on shutdown.
+type sockSet struct {
```
purpleidea: why not `socketSet` instead please?

resources/net.go
```diff
@@ -0,0 +1,877 @@
...
+	return &sockSet{
+		fdEvents: fdEvents,
+		fdPipe:   fdPipe,
+		pipeFile: file,
+	}, nil
+}
+
+// shutdown closes the event file descriptor and unblocks select by sending a
+// message to the pipe file descriptor.
+func (s *sockSet) shutdown() error {
```
purpleidea: name the method receiver variable `obj` please as per style guide: https://github.com/purpleidea/mgmt/blob/master/docs/style-guide.md#method-receiver-naming

resources/net.go
```diff
@@ -0,0 +1,877 @@
...
+}
+
+// iface wraps net.Interface to add additional methods.
+type iface struct {
+	iface *net.Interface
+	link  netlink.Link
+}
+
+// state reports the state of the interface as up or down.
+func (i *iface) state() (string, error) {
```
purpleidea: name the method receiver variable `obj` please as per style guide: https://github.com/purpleidea/mgmt/blob/master/docs/style-guide.md#method-receiver-naming

resources/net.go
```diff
@@ -0,0 +1,877 @@
...
+	// waitgroup for netlink receive goroutine
+	wg := &sync.WaitGroup{}
+	defer wg.Wait()
+
+	// create a netlink socket for receiving network interface events
+	conn, err := newSockSet(rtmGrps, path.Join(obj.tempDir, "pipe.sock"))
+	if err != nil {
+		return errwrap.Wrapf(err, "error creating socket set")
+	}
+	defer conn.shutdown() // close the netlink socket and unblock select
```
purpleidea: correct me if i'm wrong, but should this comment be:

"close the netlink socket and unblock conn.receive()"

? (I'm also hoping that's what this does, which if it does, that's great!

resources/net.go
```diff
@@ -0,0 +1,877 @@
...
+	if err := unix.Close(s.fdEvents); err != nil {
+		return err
+	}
+	// send a message to the pipe to unblock select
+	return unix.Sendto(s.fdPipe, nil, 0, &unix.SockaddrUnix{
+		Name: path.Join(s.pipeFile),
+	})
+}
+
+// close closes the pipe file descriptor. It should only be used after receive
```
purpleidea: I'm assuming that shutdown should not be called *after* this was called. I'm also assuming that you shouldn't call this more than once. Whatever the requirements/restrictions are for this and shutdown() please document them in the docstrings.

resources/net.go
jonathangold: Yes. It unblocks the select *within* conn.receive(), but in this context, your comment makes more sense.

jonathangold:
@purpleidea Done! \o/

purpleidea:
@jonathangold LGTM :)

resources/util_test.go
```diff
@@ -378,3 +378,45 @@ func TestCurrentUserGroupById(t *testing.T) {
 		t.Errorf("gid didn't match current user's: %s vs %s", strconv.Itoa(gid), currentGID)
 	}
 }
+
+func TestStrSortedSliceCompare0(t *testing.T) {
+	slice1 := []string{"foo", "bar", "baz"}
```
jonathangold: @purpleidea Since you haven't merged this yet, I'm going to fix these indices so they all match (slice0, slice1.) Won't make any other changes. :)

purpleidea:
On Wed, Mar 28, 2018 at 10:17 AM, jonathangold <notifications@github.com> wrote:
> @jonathangold commented on this pull request.
>
> ________________________________
>
> In resources/util_test.go:
>
>> @@ -378,3 +378,45 @@ func TestCurrentUserGroupById(t *testing.T) {
>   t.Errorf("gid didn't match current user's: %s vs %s", strconv.Itoa(gid),
> currentGID)
>   }
>  }
> +
> +func TestStrSortedSliceCompare0(t *testing.T) {
> + slice1 := []string{"foo", "bar", "baz"}
>
> @purpleidea Since you haven't merged this yet, I'm going to fix these
> indices so they all match (slice0, slice1.) Won't make any other changes. :)
>

Too late, I already merged in my local branch, just didn't push yet.

New patch.

purpleidea:
Original patch has been merged :)

PR #375, opened 2018-03-29
bugfix: Check that lang != nil before calling lang.Stream()

jonathangold:
When a signal interrupt was caught during init, gapi.LangClose() would
set lang = nil before gapi.Next() called lang.Stream(). This patch checks
that lang != nil before calling the method so it doesn't panic.


purpleidea:
As discussed, this is still a race, even though it theoretically reduces the chances of a panic.

PR #376, opened 2018-03-30
lang: Don't race with a ^C to the obj.lang calls

purpleidea:
If we trigger a close, we must not run the LangClose before we've exited
from the loop, because that loop could race and run code which depends
on LangClose not having run first. So run the loop shutdown, then let
the wait group expire, before shutting down the lang.

### DIFF
--- a/lang/gapi.go
+++ b/lang/gapi.go
@@ -178,7 +178,10 @@ func (obj *GAPI) Next() chan gapi.Next {
 				Err:  fmt.Errorf("%s: GAPI is not initialized", Name),
 				Exit: true, // exit, b/c programming error?
 			}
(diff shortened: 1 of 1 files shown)


purpleidea:
This is meant as an alternative to https://github.com/purpleidea/mgmt/pull/375

Instead of a panic, we'll sometimes see a block waiting on the waitgroup in lib/main. I think that is a different bug which will be fixed by my upcoming rewrite/refactor of the lib/main and resources/* mess.

So I'll leave this open until that's ready to merge.

purpleidea:
Reminder note: this can be caused by running mgmt, and then doing a ^C within the first second or so.

PR #378, opened 2018-04-01
test, resources: Add some tests to the net resource

jonathangold:
This patch adds tests for NetRes.unitFileContents(), socketSet.fdSet(),
and socketSet.nfd() methods.

Also have some tests in the works for the functions _with_ side effects, but they're more suited to shell tests, since they require root.

feedback welcome :)


jonathangold:
@purpleidea Pushed the shell test as well. Once this all is reviewed/merged, I'll work on an additional shell test that tests events, since that's missing from this one.

jonathangold:
`ip link add` seems to be stalling travis. For now, the first commit is ready for review. I'll investigate travis networking some more tomorrow.

jonathangold:
@purpleidea sorry if it wasn't clear that was ready for review because of the Travis failure :P My tests did pass, but the integration tests timed out. I'll rebase, and cross my fingers.

purpleidea:
@jonathangold Do you mind waiting until I merge my engine refactor (coming shortly) before merging this? I'm hoping that will resolve the timeouts. It will require a small change of the code and move of the files to a different folder, but it should be a 5 min fix. LMK and thanks for your patience.

jonathangold:
@purpleidea No problem :D

purpleidea:
Re-write of engine is now in git master. Please rebase and ping when ready for review. Thanks again!

This should hopefully make things a lot easier, and less intermittent failure wise.

It might be easiest to simply copy the files to some temp dir, and then copy them back in and commit fresh rather than trying to actually `git rebase` since the dirs moved around.

Thanks!

engine/resources/net_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+// test cases for NetRes.unitFileContents()
+var unitFileContentsTests = []struct {
+	in  *NetRes
+	out []byte
+}{
+	{
+		&NetRes{
+			BaseRes: BaseRes{
+				Name: "eth0",
```
jonathangold: @purpleidea How do I set the name, now that BaseRes is gone?

engine/resources/net.go
```diff
@@ -883,6 +880,6 @@ func (obj *socketSet) nfd() int {
 func (obj *socketSet) fdSet() *unix.FdSet {
 	fdSet := &unix.FdSet{}
 	fdSet.Bits[obj.fdEvents/64] |= 1 << uint(obj.fdEvents)
-	fdSet.Bits[obj.fdPipe/64] |= 1 << uint(obj.fdPipe) // fd = 3 becomes 100 if we add 5, we get 10100
+	fdSet.Bits[obj.fdPipe/64] |= 1 << uint(obj.fdPipe) // fd = 3 becomes 1000 if we add 5, we get 101000
```
purpleidea: What is this change about? Replying here is fine, although a longer explanation as a comment in the source is of course welcome too.

engine/resources/net.go
```diff
@@ -119,9 +119,6 @@ func (obj *NetRes) Validate() error {
 	}
 
 	// validate network address input
-	if (obj.Addrs == nil) != (obj.Gateway == "") {
```
purpleidea: Is this no longer true?

engine/resources/net_test.go
purpleidea: `engine.NewNamedResource`

https://godoc.org/github.com/purpleidea/mgmt/engine#NewNamedResource

Then if you want to set properties on it you need to cast it to a *NetRes and then set fields as you wish. We should probably add a helper function to do this with `reflect`. Patches welcome.

engine/resources/net.go
jonathangold: I tried to explain it in the commit message, but there are scenarios where you'd want to set one without the other. For example, if you're setting up a secondary interface on a different subnet, you might not want to change the default gateway, as having 2 defaults (on the same routing table) leads to issues. That's why travis was losing track of the build when the shell tests added a gateway.

purpleidea:
@jonathangold I think there was a lot of good code in here... You interested in rebasing to master and getting this ready to merge?

jonathangold:
Yeah, I'll have to open a new PR, since github lost the ref to my repo when I deleted it.

PR #382, opened 2018-04-13
resources: mount: Add a mount resource

jonathangold:


purpleidea:
Hey, my engine refactor (with docs) is almost done. Still needs a bit of patching, to fix a small bug in the integration test framework, but should be good enough that you won't need to change any code by the time it's merged (just a rebase).

If you want to update this patch to be based on that, the code is here: https://github.com/purpleidea/mgmt/pull/383

purpleidea:
Update: re-write of engine is now in git master. Please rebase and ping when ready for review. Thanks again!

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	// diskByLabel is the location of symlinks for partitions by label.
+	diskByPartLabel = devDisk + "by-partlabel/"
+
+	// dbusStoragedPath is the dbus path that watch eavesdrops on for signals.
+	dbusStoragedPath = "/org/freedesktop/UDisks2"
+)
+
+// vfsTypes is a list of all valid filesystem types taken from
+// /proc/filesystems.
+var vfsTypes = []string{
```
purpleidea: Are these in a particular order? If not let's make it alphabetical please. Mention whatever order is being used in the docstring.

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+type MountRes struct {
+	traits.Base
+
+	init *engine.Init
+
+	// State must be exists ot absent. If absent, remaining fields are ignored.
+	State  string            `yaml:"state"`
+	Device string            `yaml:"device"` // location of the device or image
+	Type   string            `yaml:"type"`   // the type of filesystem
+	MntOps map[string]string `yaml:"mntops"` // mount options
```
purpleidea: Since this is a `Mount` resource, naming this MntOps is probably a bit redundant. You can probably name it `Options` instead unless there's a reason for being more specific (eg: some other kind of future options we might want to add).

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	"iso9660",
+}
+
+// defaultMntOps is a map that sets the default mount options for fstab mounts
+var defaultMntOps = map[string]string{"defaults": ""}
+
+// MountRes is a systemd mount resource that adds/removes entries from
+// /etc/fstab, and makes sure the defined device is mounted or unmounted
+// accordingly.
+type MountRes struct {
```
purpleidea: If we're depending on the resource `Name` value for the mount location, mention this in the docstring please.

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
```
purpleidea: I think this import belongs in the third block.

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	Freq   int               `yaml:"freq"`   // dump frequency
+	PassNo int               `yaml:"passno"` // verification order
+
+	mount *fstab.Mount // struct representing the mount
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *MountRes) Default() engine.Res {
+	return &MountRes{
+		MntOps: defaultMntOps,
```
purpleidea: Can you test if appending to this MntOps in one resource would change the value in another?

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	var err error
+
+	// validate state
+	if obj.State != "exists" && obj.State != "absent" {
+		return fmt.Errorf("state must be 'exists', or 'absent'")
+	}
+
+	// validate type
+	var typeOK bool
+	for _, t := range vfsTypes {
```
purpleidea: You can simplify this loop by using https://github.com/purpleidea/mgmt/blob/master/util/util.go#L52 and returning right away if not found with a short circuit on != absent.

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	defer recWatcher.Close()
+
+	// notify engine that we're running
+	if err := obj.init.Running(); err != nil {
+		return err // bubble up a NACK...
+	}
+
+	var send bool
+	var done bool
+
```
purpleidea: nit: kill this nl please

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+
+	if !apply {
+		return false, nil
+	}
+
+	obj.init.Logf("mountCheckApply(%t)", apply)
+
+	if obj.State == "exists" {
+		// reload mounts from /etc/fstab
```
purpleidea: can you add a source comment explaining the logic here please?

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+		return fmt.Errorf("the Type differs")
+	}
+	if obj.PassNo != res.PassNo {
+		return fmt.Errorf("the PassNo differs")
+	}
+
+	return nil
+}
+
+// Close is run by the engine to clean up after the resource is done.
```
purpleidea: Can you reorder the method order in this file to match what is done in the other resources please?

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+		return true, nil
+	}
+
+	if !apply {
+		return false, nil
+	}
+	obj.init.Logf("fstabCheckApply(%t)", apply)
+
+	if obj.State == "exists" {
+		fstabEntryAdd(fstabPath, obj.mount)
```
purpleidea: why aren't we checking this error? Same for Remove below...

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+		}
+	}
+	return false, nil
+}
+
+// fstabEntryAdd adds the given mount to the provided fstab file.
+func fstabEntryAdd(file string, mount *fstab.Mount) error {
+	mounts, err := fstab.ParseFile(file)
+	if err != nil {
+		errwrap.Wrapf(err, "error parsing file")
```
purpleidea: missing something?

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	// mount does not exist so we need to add it
+	mounts = append(mounts, mount)
+	return fstabWrite(file, mounts)
+}
+
+// fstabEntryRemove removes the given mount from the provided fstab file.
+func fstabEntryRemove(file string, mount *fstab.Mount) error {
+	mounts, err := fstab.ParseFile(file)
+	if err != nil {
+		errwrap.Wrapf(err, "error parsing file")
```
purpleidea: same (see above)

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+	return fstabWrite(file, mounts)
+}
+
+// fstabWrite generates an fstab file with the given mounts, and writes them
+// to the provided fstab file.
+func fstabWrite(file string, mounts fstab.Mounts) error {
+	// build the file contents
+	contents := fmt.Sprintf("# Generated by mgmt on ")
+	contents = contents + time.Now().Format(time.ANSIC) + "\n"
```
purpleidea: why ANSIC ?

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+		return errwrap.Wrapf(err, "error writing fstab file")
+	}
+	return nil
+}
+
+// mountExists returns true, if a given mount exists in the given file
+// (typically /proc/mounts.)
+func mountExists(file string, mount *fstab.Mount) (bool, error) {
+	var err error
+	var m fstab.Mount
```
purpleidea: `m :=` instead?

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+
+	// resolve the device's symlink if there is one
+	if m.Spec, err = evalSpec(mount.Spec); err != nil {
+		return false, errwrap.Wrapf(err, "error evaluating spec: %s", mount.Spec)
+	}
+
+	// get all mounts
+	mounts, err := fstab.ParseFile(file)
+	if err != nil {
+		return false, errwrap.Wrapf(err, "error parsing /proc/mounts")
```
purpleidea: If the file name is always `/proc/mounts` then this should be a constant, not a variable. Otherwise the error message should mention the variable instead of the static string. Check for this sort of issue in other places too please.

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+	// check for the defined mount
+	for _, p := range mounts {
+		found, err := mountCompare(&m, p)
+		if err != nil {
+			return false, errwrap.Wrapf(err, "error comparing mounts:\n %+v \n and \n %+v", mount, p)
```
purpleidea: why \n ??

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	if !strMapEq(def.MntOps, defaultMntOps) && !strMapEq(def.MntOps, proc.MntOps) {
+		return false, nil
+	}
+	if def.VfsType != "" && def.VfsType != proc.VfsType {
+		return false, nil
+	}
+	return true, nil
+}
+
+// mountReload perorms a daemon-reload and restarts fs-local.target and
```
purpleidea: typo

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	case fstab.Label:
+		path = diskByLabel + m.SpecValue()
+	case fstab.PartUUID:
+		path = diskByPartUUID + m.SpecValue()
+	case fstab.PartLabel:
+		path = diskByPartLabel + m.SpecValue()
+	case fstab.Path:
+		path = m.SpecValue()
+	default:
+		return "", fmt.Errorf("unexpected spec type")
```
purpleidea: add the value to the error message please

engine/resources/mount.go
```diff
@@ -0,0 +1,712 @@
...
+	Flags          uint32
+	FileName       [64]byte
+	CryptName      [64]byte
+	EncryptKey     [32]byte
+	Init           [2]uint64
+}
+
+// getLoopInfo returns a loopInfo struct containing information about the
+// provided file backed loopback device.
+func getLoopInfo(loop string) (loopInfo, error) {
```
purpleidea: maybe this should return *loopInfo so we don't copy the data?

engine/resources/mount_test.go
```diff
@@ -0,0 +1,274 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	fstab "github.com/deniswernert/go-fstab"
+)
+
+const fstabMock1 = `UUID=ef5726f2-615c-4350-b0ab-f106e5fc90ad / ext4 defaults 1 1
+`
```
purpleidea: is the nl here intentional?

engine/resources/mount.go
jonathangold: I imagine it would. :\ Should I make defaultMntOps a function that returns a map instead?

engine/resources/mount.go
jonathangold: It's easier to read, since its printing two structs. I can take them out if you prefer.

engine/resources/mount_test.go
jonathangold: Yeah, the files end in a newline. I still need to add more test cases, so these may end up being multi-line consts.

engine/resources/mount.go
jonathangold: Looked pretty. Needed to pick one.

engine/resources/mount.go
jonathangold: vim :sort to the rescue!

engine/resources/mount.go
jonathangold: I added the file as an argument in these functions to make them easier to test with mock files. I fixed the error message, but I'm inclined to keep the file as an argument, so the functions are testable without root, and without actually mounting anything.

engine/resources/mount.go
```diff
@@ -0,0 +1,715 @@
...
+	// mount does not exist so we need to add it
+	mounts = append(mounts, mount)
+	return fstabWrite(file, mounts)
+}
+
+// fstabEntryRemove removes the given mount from the provided fstab file.
+func fstabEntryRemove(file string, mount *fstab.Mount) error {
+	mounts, err := fstab.ParseFile(file)
+	if err != nil {
+		errwrap.Wrapf(err, "error parsing file: %s", file)
```
purpleidea: When I wrote: "missing something?" in the last review, i meant something else, although you don't have to undo the change you made.

Look closely. Find the others too ;)

engine/resources/mount.go
purpleidea: How about https://github.com/purpleidea/mgmt/blob/master/lib/hello.go#L29 ?

engine/resources/mount.go
```diff
@@ -0,0 +1,715 @@
...
+		if m.Equals(mount) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// fstabEntryAdd adds the given mount to the provided fstab file.
+func fstabEntryAdd(file string, mount *fstab.Mount) error {
+	mounts, err := fstab.ParseFile(file)
```
purpleidea: each use of fstab.ParseFile has the errors done differently. consistency please :)

engine/resources/mount.go
purpleidea: yeah maybe something that prints nicer might be good. you might want to add a Debug + Logf if you want to print some longer thing like this.

engine/resources/mount.go
```diff
@@ -0,0 +1,715 @@
...
+func getLoopInfo(loop string) (*loopInfo, error) {
+	// open the loop file
+	f, err := os.OpenFile(loop, 0, loopFileUmask)
+	if err != nil {
+		return nil, fmt.Errorf("error opening %s: %s", loop, err)
+	}
+	defer f.Close()
+
+	// deserialize the contents
+	retInfo := loopInfo{}
```
purpleidea: the convention is to just do the pointer upfront. eg: 

`retInfo := &loopInfo{}`

engine/resources/mount_test.go
purpleidea: const foo = "blah" + "\n" or `blah` + "\n" instead please

engine/resources/mount.go
jonathangold: You're right. I was going to figure that out, and then I forgot. :/

engine/resources/mount.go
purpleidea: It panics? I'm guessing you forgot to remove the & when you pass it in via unsafe() ?
But current commit looks like it's all good?

engine/resources/mount_test.go
```diff
@@ -0,0 +1,273 @@
...
+			Freq:    1,
+			PassNo:  1,
+		},
+	},
+}
+
+func TestFstabEntryRemove(t *testing.T) {
+	file, err := ioutil.TempFile("", "fstab")
+	if err != nil {
+		t.Error(err)
```
purpleidea: what would happen if this errored? ... the code would continue running which would be bad. So you need to `return` right after.
If possible replace with Errorf.

engine/resources/mount.go
```diff
@@ -0,0 +1,692 @@
...
+	mounts, err := fstab.ParseFile(file)
+	if err != nil {
+		return false, errwrap.Wrapf(err, "error parsing file: %s", file)
+	}
+	// check for the defined mount
+	for _, p := range mounts {
+		found, err := mountCompare(&m, p)
+		if err != nil {
+			if obj.init.Debug {
+				obj.init.Logf("mounts could not be compared:\n %+v \n and \n %+v", mount, p)
```
jonathangold: I'm not sure it's worth it to make mountExists() a struct method just for this one log. I think I should replace it with `return false, errwrap.Wrapf(err, "mounts could not be compared: %s and %s", mount.String(), p.String())` like I did in the tests. What do you think?

engine/resources/mount.go
```diff
@@ -0,0 +1,687 @@
...
+		return errwrap.Wrapf(err, "error establishing dbus connection")
+	}
+	defer conn.Close()
+
+	// add a dbus rule to watch signals from the mount unit.
+	args := fmt.Sprintf("type='signal', path='%s', arg0='%s'",
+		dbusUnitPath+sdbus.PathBusEscape(unit.UnitNamePathEscape((obj.Name()+".mount"))),
+		dbusMountInterface,
+	)
+	if call := conn.BusObject().Call(dbusAddMatch, 0, args); call.Err != nil {
```
purpleidea: dbusAddMatch should be refactored into engine/util/ package, since it's used in a few places in the resources/ namespace already.

purpleidea: Should we have a corresponding `defer` org.freedesktop.DBus.RemoveMatch ?

engine/resources/mount.go
```diff
@@ -0,0 +1,687 @@
...
+	if obj.State == "exists" {
+		// Reload mounts from /etc/fstab by performing a `daemon-reload` and
+		// restarting `local-fs.target` and `remote-fs.target` units.
+		if err := mountReload(); err != nil {
+			return false, errwrap.Wrapf(err, "error reloading /etc/fstab")
+		}
+		return false, nil // we're done
+	}
+	// unmount the device
+	if err := unix.Unmount(obj.Name(), 0); err != nil {
```
purpleidea: can you add a comment about what this 0 is please?

engine/resources/mount.go
```diff
@@ -0,0 +1,687 @@
...
+	if c, err := obj.mountCheckApply(apply); err != nil {
+		return false, err
+	} else if !c {
+		checkOK = false
+	}
+
+	return checkOK, nil
+}
+
+// MountUID is a unique resource identifier.
```
purpleidea: Can you move these things to be in the same order as the other resources please?

engine/resources/mount.go
```diff
@@ -0,0 +1,687 @@
...
+		}
+	}
+	return fstabWrite(file, mounts)
+}
+
+// fstabWrite generates an fstab file with the given mounts, and writes them
+// to the provided fstab file.
+func fstabWrite(file string, mounts fstab.Mounts) error {
+	// build the file contents
+	contents := fmt.Sprintf("# Generated by mgmt at %d", time.Now().UnixNano()) + "\n"
```
purpleidea: Can you replace the `mgmt` by `%s` and use `obj.init.Program` instead please? It will be `mgmt` if that's what the binary is called.

engine/resources/mount_test.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	defer os.Remove(file.Name())
+
+	for _, test := range fstabEntryExistsTests {
+		if err := ioutil.WriteFile(file.Name(), test.fstabMock, 0644); err != nil {
+			t.Errorf("error writing fstab file: %s: %s", file.Name(), err)
+		}
+		result, err := fstabEntryExists(file.Name(), test.in)
+		if err != nil {
+			t.Errorf("error checking if fstab entry %s exists: %s", test.in.String(), err)
```
purpleidea: make sure to add `return`'s when needed...

also as discussed, nit: use of `err` in Errorf should be changed to %v instead of %s throughout.

engine/resources/hostname.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	"github.com/purpleidea/mgmt/engine"
 	"github.com/purpleidea/mgmt/engine/traits"
+	engUtil "github.com/purpleidea/mgmt/engine/util"
```
purpleidea: can you name this alias engineUtil as seen elsewhere for consistency please?

engine/resources/hostname.go
```diff
@@ -36,7 +37,6 @@ func init() {
 const (
 	hostname1Path  = "/org/freedesktop/hostname1"
 	hostname1Iface = "org.freedesktop.hostname1"
-	dbusAddMatch   = "org.freedesktop.DBus.AddMatch"
```
purpleidea: I think this might also exist in the packagekit.go file to be removed as well.

engine/resources/mount.go
```diff
@@ -0,0 +1,732 @@
...
+	}
+	defer conn.BusObject().Call(engUtil.DBusRemoveMatch, 0, args) // ignore the error
+
+	ch := make(chan *dbus.Signal)
+	defer close(ch)
+
+	conn.Signal(ch)
+	defer conn.RemoveSignal(ch)
+
+	// watch the systemd-networkd configuration file
```
purpleidea: networkd?

engine/resources/mount.go
```diff
@@ -0,0 +1,732 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+	"unsafe"
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/engine/traits"
+	engUtil "github.com/purpleidea/mgmt/engine/util"
```
purpleidea: s/engUtil/engineUtil/ please

purpleidea:
Sweet

PR #384, opened 2018-04-26
engine: resources: nspawn: Remove godbus channel buffer

jonathangold:
https://github.com/godbus/dbus/issues/94 is fixed with
https://github.com/godbus/dbus/pull/105, so the
buffered channel is no longer necessary. The call to
RemoveSignal() was also removed, as Close() handles
closing all open channels.

## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/engine/resources/nspawn.go
+++ b/engine/resources/nspawn.go
@@ -151,6 +151,7 @@ func (obj *NspawnRes) Watch() error {
 	if err != nil {
 		return errwrap.Wrapf(err, "failed to connect to bus")
 	}
+	defer bus.Close()
 
 	// add a match rule to match messages going through the message bus
 	call := bus.BusObject().Call("org.freedesktop.DBus.AddMatch", 0,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: engine/resources/nspawn.go [hunk 1]
```diff
@@ -174,7 +172,6 @@ func (obj *NspawnRes) Watch() error {
 
 	defer close(busChan)
 	defer bus.Close()
-	defer bus.RemoveSignal(busChan)
```
purpleidea: Let's make this clean, and keep this, but put it right after the add Signal command. You can add a comment about why it is or isn't absolutely needed if you want, but it's a nice thing to do for symmetry.

engine/resources/nspawn.go
```diff
@@ -174,7 +172,6 @@ func (obj *NspawnRes) Watch() error {
 
 	defer close(busChan)
```
purpleidea: I think this line should go right after the make() for it. Want to move it there in the same patch please?

jonathangold: Is there any reason I shouldn't also move `bus.Close()` to right under `bus, err :=...`? I don't think the order matters in this case.

purpleidea: Oh you definitely should AFAICT :)

I sometimes need to remind myself that present me is often a better programmer than past me, so as a result, I'm always cleaning up after old mistakes or missed mistakes, and should feel free to change anything that's wrong.

purpleidea:
Merged! Thank you :)

PR #385, opened 2018-04-29
Add Arch mapping 'any' for Arch Linux compatibility

alanjjenkins:
Arch Linux uses the mapping architecture name 'any'. This mapping was
missing from mgmt resulting in an error stating that arch 'any' did not
exist. Adding this mapping allows successful installation of packages
under Arch Linux.

### DIFF
--- a/engine/resources/packagekit/packagekit.go
+++ b/engine/resources/packagekit/packagekit.go
@@ -56,6 +56,7 @@ var (
 		// TODO: add more values
 		// noarch
 		"noarch": "ANY", // special value "ANY" (noarch as seen in Fedora)
+		"any":    "ANY", // special value "ANY" ('any' as seen in ArchLinux)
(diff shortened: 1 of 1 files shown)


purpleidea:
@alanjjenkins Thanks for the patch!

Can you `git commit --amend` with a commit message that conforms to our CI test please?

It should look like:

`engine: resources: packagekit: Whatever the message is`

See here: https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md#tips

Ping when pushed again, and I'll merge!

purpleidea:
@alanjjenkins 

> Please run gofmt and get these fixed in the master branch =).

Can you tell me what version of golang you are using? If it's 1.10 it seems they changed the gofmt rules. It's correctly formatted in 1.9.x. When we move to 1.10 we will fix them.

PR #386, opened 2018-05-05
test: Catch scenarios where we forgot to prefix the error with return

purpleidea:
One of our contributors occasionally made this typo, and since core go
vet didn't (surprisingly) catch it, we should add a test!

/cc @jonathangold (present for you!)


purpleidea:
(waiting for your review/ACK)

jonathangold:
@purpleidea Since `errwrap` is an alias for `errors`, maybe it check for both (e.g. `err[a-z]+\.Wrap`)

PR #387, opened 2018-05-06
[WIP] engine: DBus cleanup

jonathangold:
Need to figure out what's wrong with pkg/packagekit before merging this, unless I'm missing something.

### DIFF
--- a/engine/resources/packagekit/packagekit.go
+++ b/engine/resources/packagekit/packagekit.go
@@ -29,6 +29,7 @@ import (
 	"github.com/purpleidea/mgmt/util"
 
 	"github.com/godbus/dbus"
+	multierr "github.com/hashicorp/go-multierror"
 	errwrap "github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


engine/resources/packagekit/packagekit.go
```diff
@@ -186,34 +186,49 @@ func (obj *Conn) Close() error {
...
-		call = bus.Call(engineUtil.DBusAddMatch, 0, "type='signal',path='"+pathStr+"',interface='"+iface+"'")
+		args := fmt.Sprintf("type='signal', path='%s', interface='%s'", pathStr, iface)
+		argsList = append(argsList, args)
+		call = bus.Call(engineUtil.DBusAddMatch, 0, args)
 	} else {
 		for _, signal := range signals {
-			call = bus.Call(engineUtil.DBusAddMatch, 0, "type='signal',path='"+pathStr+"',interface='"+iface+"',member='"+signal+"'")
-			if call.Err != nil {
+			args := fmt.Sprintf("type='signal', path='%s', interface='%s', member'%s'", pathStr, iface, signal)
+			argsList = append(argsList, args)
```
purpleidea: I'm not sure this logic is entirely correct, b/c we might successfully add the first three, but fail on the fourth. In this scenario, we never run the signal remove...

engine/resources/packagekit/packagekit.go
```diff
@@ -186,34 +186,49 @@ func (obj *Conn) Close() error {
...
+		return nil, call.Err
 	}
+
 	// The caller has to make sure that ch is sufficiently buffered; if a
 	// message arrives when a write to c is not possible, it is discarded!
 	// This can be disastrous if we're waiting for a "Finished" signal!
 	obj.GetBus().Signal(ch)
-	return nil
+	return func() error {
+		for _, args := range argsList {
```
purpleidea: Wouldn't we want to run these in the reverse order for correctness? Also, if one fails, why do we break early and skip doing the rest?

engine/resources/packagekit/packagekit.go
```diff
@@ -611,7 +647,11 @@ func (obj *Conn) GetUpdates(filter uint64) ([]string, error) {
 	}
 
 	var signals = []string{"Package", "ErrorCode", "Finished", "Destroy"} // "ItemProgress" ?
-	obj.matchSignal(ch, interfacePath, PkIfaceTransaction, signals)
+	removeSignals, err := obj.matchSignal(ch, interfacePath, PkIfaceTransaction, signals)
+	if err != nil {
+		return nil, err
+	}
+	defer removeSignals()
```
purpleidea: deferring all the removes is a nice touch!

engine/resources/packagekit/packagekit.go
```diff
@@ -186,34 +187,50 @@ func (obj *Conn) Close() error {
...
+			args := fmt.Sprintf("type='signal', path='%s', interface='%s', member'%s'", pathStr, iface, signal)
+			argsList = append(argsList, args)
+			if call = bus.Call(engineUtil.DBusAddMatch, 0, args); call.Err != nil {
+				break // fail if any one fails
 			}
 		}
 	}
 	if call.Err != nil {
-		return call.Err
+		defer removeSignals() // ignore the error
```
jonathangold: Are you ok with ignoring the error here? i.e. if the 3rd of 4 fail, it will try to remove all 4 signals, even though only 2 were added in the first place. The last two will fail silently, since there's nothing to do.

purpleidea: Yeah, the issue of "we're closing/shutting down" but on close there's an error doesn't really mean something good happened, but it's not something we can often do anything about, so in this case it's fine AFAICT.

engine/resources/packagekit/packagekit.go
```diff
@@ -186,34 +187,50 @@ func (obj *Conn) Close() error {
...
 	}
 	// eg: gdbus monitor --system --dest org.freedesktop.PackageKit --object-path /org/freedesktop/PackageKit | grep <signal>
+	bus := obj.GetBus().BusObject()
 	var call *dbus.Call
+	var argsList []string
+	// cleanup function should be called when done or when AddMatch errors
+	removeSignals := func() error {
+		var errList error
+		for i := len(argsList) - 1; i >= 0; i-- { // last in first out
+			if call = bus.Call(engineUtil.DBusRemoveMatch, 0, argsList[i]); call.Err != nil {
```
purpleidea: why not `call := bus.Call ... ` ?

jonathangold: Done. Also moved the var down.

purpleidea:
Rebased to master and merged as 299080f59088e49c74b3def5444496860d96aa68

Thank you!

PR #394, opened 2018-05-19
Update vagrant vm

phaer:
While checking out your interesting work, I noticed a few things:

- the go 1.8 release in fedora 26 seems to be too old to build mgmt, so I bumped the base image to fedora 28 which seems to work fine.
- the gem package manager seems to be needed during the build process but wasn't installed.
- the path to the example yaml file in motd changed, presumably when mgmt lang has been introduced.


purpleidea:
Merged, thanks!

PR #396, opened 2018-07-13
test: Increasing test coverage for graphsync

ginywiny:
This patch is an addition to graphsync_test.go, while also being used to test git. 

I was looking around the structure of test files from pgraph and added something just as an experiment. 

The test passed and coverage increased from 72.4% to 72.9%.

### DIFF
--- a/examples/lib/exec-send-recv.go
+++ b/examples/lib/exec-send-recv.go
@@ -61,7 +61,7 @@ func (obj *MyGAPI) Cli(c *cli.Context, fs engine.Fs) (*gapi.Deploy, error) {
 			Noop: c.GlobalBool("noop"),
 			Sema: c.GlobalInt("sema"),
 			GAPI: &MyGAPI{
-			// TODO: add properties here...
+				// TODO: add properties here...
 			},
 		}, nil
 	}
--- a/pgraph/graphsync_test.go
+++ b/pgraph/graphsync_test.go
@@ -92,3 +92,41 @@ func TestGraphSync2(t *testing.T) {
 		t.Errorf("%s", s)
 	}
 }
+
+func TestGraphSync3(t *testing.T) {
+	v1 := NV("v1")
+	v2 := NV("v2")
+	v3 := NV("v3")
+
+	e1 := NE("e1")
+	e2 := NE("e2")
+	e3 := NE("e3")
+	e4 := NE("e4")
+
+	// g base graph with 3 edges
+	g := &Graph{}
+	g.AddEdge(v1, v2, e1)
+	g.AddEdge(v2, v3, e2)
+	g.AddEdge(v1, v3, e3)
+
+	// newGraph input with 4 edges
+	newGraph := &Graph{}
+	newGraph.AddEdge(v1, v3, e1)
+	newGraph.AddEdge(v2, v3, e2)
+	newGraph.AddEdge(v1, v2, e3)
+	newGraph.AddEdge(v3, v1, e4)
+
+	if runGraphCmp(t, g, newGraph) == "" {
+		t.Errorf("identical graphs: fail")
+		return
+	}
+
+	if err := g.GraphSync(newGraph, strVertexCmpFn, vertexAddFn, vertexRemoveFn, strEdgeCmpFn); err != nil {
+		t.Errorf("fail: %v", err)
+		return
+	}
+
+	if s := runGraphCmp(t, g, newGraph); s != "" {
+		t.Errorf("%s", s)
+	}
+}
--- a/examples/lib/libmgmt-subgraph0.go
+++ b/examples/lib/libmgmt-subgraph0.go
@@ -66,7 +66,7 @@ func (obj *MyGAPI) Cli(c *cli.Context, fs engine.Fs) (*gapi.Deploy, error) {
 			Noop: c.GlobalBool("noop"),
 			Sema: c.GlobalInt("sema"),
 			GAPI: &MyGAPI{
-			// TODO: add properties here...
+				// TODO: add properties here...
 			},
 		}, nil
 	}
--- a/examples/lib/libmgmt-subgraph1.go
+++ b/examples/lib/libmgmt-subgraph1.go
@@ -61,7 +61,7 @@ func (obj *MyGAPI) Cli(c *cli.Context, fs engine.Fs) (*gapi.Deploy, error) {
 			Noop: c.GlobalBool("noop"),
 			Sema: c.GlobalInt("sema"),
 			GAPI: &MyGAPI{
-			// TODO: add properties here...
+				// TODO: add properties here...
 			},
 		}, nil
 	}
--- a/examples/lib/libmgmt2.go
+++ b/examples/lib/libmgmt2.go
@@ -63,7 +63,7 @@ func (obj *MyGAPI) Cli(c *cli.Context, fs engine.Fs) (*gapi.Deploy, error) {
 			Noop: c.GlobalBool("noop"),
 			Sema: c.GlobalInt("sema"),
 			GAPI: &MyGAPI{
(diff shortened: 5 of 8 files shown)


pgraph/graphsync_test.go
```diff
@@ -92,3 +92,41 @@ func TestGraphSync2(t *testing.T) {
...
+	v1 := NV("v1")
+	v2 := NV("v2")
+	v3 := NV("v3")
+
+	e1 := NE("e1")
+	e2 := NE("e2")
+	e3 := NE("e3")
+	e4 := NE("e4")
+
+	//Base Graph with 3 edges
```
purpleidea: comments should be full sentences which start with a capital and end with a period, or not full sentences and not be capitalized. in both case the syntax is `// hello` not `//hello` (add a space after the `//`)

pgraph/graphsync_test.go
```diff
@@ -92,3 +92,41 @@ func TestGraphSync2(t *testing.T) {
...
+	newGraph.AddEdge(v1, v3, e1)
+	newGraph.AddEdge(v2, v3, e2)
+	newGraph.AddEdge(v1, v2, e3)
+	newGraph.AddEdge(v3, v1, e4)
+
+	if runGraphCmp(t, g, newGraph) == "" {
+		return
+	}
+
+	err := g.GraphSync(newGraph, strVertexCmpFn, vertexAddFn, vertexRemoveFn, strEdgeCmpFn)
```
purpleidea: you should use the short statement form here: https://tour.golang.org/flowcontrol/6

purpleidea:
So, this is an improvement *but* since I haven't merged anything into git master yet, we get to *hide* our mistakes. As a result, you should squash the two commits into a single one. Read the linked file for more details.

purpleidea:
Getting much closer, although you need to review: https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md again... Tell me if you can't figure out what you missed.

Hint: try and read the travis logs and see why your patch is failing....
(Note: there is a small bug that causes the integration tests to occasionally fail, ignore that-- other test(s) fail as well-- that's what you need to fix.)

purpleidea:
@ginywiny Hint:

https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md#tips

Note the commit message format. Commit message body should be full sentences. The former is the first line in the commit message. Then if you want to include a body, you press enter twice, and put the body. Eg:

https://github.com/purpleidea/mgmt/commit/05f6ba7297adb7683afb3005f5438bd118b9d9f6

examples/lib/exec-send-recv.go
```diff
@@ -61,7 +61,7 @@ func (obj *MyGAPI) Cli(c *cli.Context, fs engine.Fs) (*gapi.Deploy, error) {
 			Noop: c.GlobalBool("noop"),
 			Sema: c.GlobalInt("sema"),
 			GAPI: &MyGAPI{
-			// TODO: add properties here...
+				// TODO: add properties here...
```
jonathangold: You can probably fix these with `make gofmt` from the project root. Otherwise you can `git checkout examples/lib/exec-send-recv.go` with the path of each changed file and amend the commit.

pgraph/graphsync_test.go
```diff
@@ -92,3 +92,40 @@ func TestGraphSync2(t *testing.T) {
...
+
+	// newGraph input with 4 edges.
+	newGraph := &Graph{}
+	newGraph.AddEdge(v1, v3, e1)
+	newGraph.AddEdge(v2, v3, e2)
+	newGraph.AddEdge(v1, v2, e3)
+	newGraph.AddEdge(v3, v1, e4)
+
+	if runGraphCmp(t, g, newGraph) == "" {
+		return
```
jonathangold: In this case, `g` and `newGraph` are the same. Should this be a t.Error?
Hint: check what the other tests do.

pgraph/graphsync_test.go
```diff
@@ -92,3 +92,40 @@ func TestGraphSync2(t *testing.T) {
...
+	}
+
+	if err := g.GraphSync(newGraph, strVertexCmpFn, vertexAddFn, vertexRemoveFn, strEdgeCmpFn); err != nil {
+		return
```
jonathangold: Shouldn't we report this error too?

pgraph/graphsync_test.go
```diff
@@ -92,3 +92,40 @@ func TestGraphSync2(t *testing.T) {
...
+	v1 := NV("v1")
+	v2 := NV("v2")
+	v3 := NV("v3")
+
+	e1 := NE("e1")
+	e2 := NE("e2")
+	e3 := NE("e3")
+	e4 := NE("e4")
+
+	// g base graph with 3 edges.
```
jonathangold: Only full sentences need punctuation. Please remove the trailing period here, and on the comment below.

pgraph/graphsync_test.go
```diff
@@ -92,3 +92,40 @@ func TestGraphSync2(t *testing.T) {
...
+	}
+
+	if err := g.GraphSync(newGraph, strVertexCmpFn, vertexAddFn, vertexRemoveFn, strEdgeCmpFn); err != nil {
+		return
+	}
+
+	if s := runGraphCmp(t, g, newGraph); s != "" {
+		t.Errorf("%s", s)
+	}
+
```
jonathangold: Please remove this extra newline.

examples/lib/exec-send-recv.go
ginywiny: Hey Jonathan, thanks a ton for the reply. Sounds a bit embarassing, but running make gofmt tells me "make: *** No rule to make target 'gofmt'.  Stop." and I am not sure how to run the command differently. I've tried gofmt (file name), but I am not sure if that accomplishes the same function.

pgraph/graphsync_test.go
ginywiny: You're right, il look into the other tests as well.

examples/lib/exec-send-recv.go
jonathangold: Did you try running it from `$GOPATH/src/github.com/purpleidea/mgmt`? (It won't work from subdirectories)

ginywiny: This is what i ran:

```[michael@localhost mgmt]$ cd $GOPATH/src/github.com/ginywiny/mgmt
[michael@localhost mgmt]$ pwd
/home/michael/gopath/src/github.com/ginywiny/mgmt
[michael@localhost mgmt]$ make gofmt
# TODO: remove gofmt once goimports has a -s option
find . -maxdepth 6 -type f -name '*.go' -not -path './old/*' -not -path './tmp/*' -not -path './vendor/*' -exec gofmt -s -w {} \;
find . -maxdepth 6 -type f -name '*.go' -not -path './old/*' -not -path './tmp/*' -not -path './vendor/*' -exec goimports -w {} \;
```

Is this correct, or must it be run directly from purpleidea's directory rather than ginywiny's?

jonathangold: As long as you're in the correct branch it should be fine.

Caveat: If `make gofmt` invokes go 1.10's gofmt, it won't fix the issue. It may be easier to checkout each affected file from master. 

Or, make a backup copy of `pgraph/graphsync_test.go` (outside of the project directory,) and reset your branch with `git fetch origin` followed by `git reset --hard origin/master`. Then you can copy your backup back into `pgraph/` then `git add -p` to add the changes and `git commit` followed by `git push -f`. Don't do this without making sure you have backed up the file outside `mgmt/`, or you will lose your work.

purpleidea:
Now that we're on golang 1.10, this applied. So merged!

PR #400, opened 2018-11-19
engine: resource: Add a Docker container resource

jonathangold:
I've put together the resource with what I feel is the minimum viable api. I still have to validate the input, but I wanted to put this up so we can discuss a few things in context.

Do we want to allow users to define stopped/terminated resources (like the aws resource?) 
Is there a better data structure than `map[string]string` for portmapping?

We need to talk about vendoring, because this involves vendoring a ~140mb repo, just to delete one of their vendored folders.

Feel free to read through the resource, but note, there aren't many comments at the moment.
Cheers.

### DIFF
--- a/engine/resources/docker_container.go
+++ b/engine/resources/docker_container.go
@@ -1,0 +1,441 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/engine/traits"
+	"github.com/purpleidea/mgmt/util"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
+)
+
+const (
+	// ContainerRunning is the running container state.
+	ContainerRunning = "running"
+	// ContainerStopped is the stopped container state.
+	ContainerStopped = "stopped"
+	// ContainerRemoved is the removed container state.
+	ContainerRemoved = "removed"
+
+	// validateCtxTimeout is the length of time, in seconds, before requests
+	// are cancelled in Validate.
+	validateCtxTimeout = 20
+	// checkApplyCtxTimeout is the length of time, in seconds, before requests
+	// are cancelled in CheckApply.
+	checkApplyCtxTimeout = 120
+)
+
+func init() {
+	engine.RegisterResource("docker:container", func() engine.Res { return &DockerContainerRes{} })
+}
+
+// DockerContainerRes is a docker container resource.
+type DockerContainerRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Edgeable
+
+	State      string                     `yaml:"state"`      // running, stopped, or removed
+	Image      string                     `yaml:"image"`      // docker image or image:tag
+	Cmd        []string                   `yaml:"cmd"`        // command(s) to run on the container
+	Env        []string                   `yaml:"env"`        // environment variables, e.g. ["VAR"="val",],
+	Ports      map[string]map[int64]int64 `yaml:"ports"`      // port bindings, e.g. {"tcp" => {80 => 8080},},
+	APIVersion string                     `yaml:"apiversion"` // optional api version, e.g. "v1.39"
+
+	Force bool `yaml:"force"` // force destroy and redeploy if image is incorrect
+
+	client *client.Client // docker api client
+
+	init *engine.Init
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *DockerContainerRes) Default() engine.Res {
+	return &DockerContainerRes{}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *DockerContainerRes) Validate() error {
+	ctx, cancel := context.WithTimeout(context.Background(), validateCtxTimeout*time.Second)
+	defer cancel()
+
+	// validate state
+	if obj.State != ContainerRunning && obj.State != ContainerStopped && obj.State != ContainerRemoved {
+		return fmt.Errorf("state must be running, stopped or removed")
+	}
+
+	// TODO: should this stuff happen in validate or do we make sure there's no
+	// whitespace, and let it fail in CheckApply?
+	//
+	// set up temporary docker client to validate image
+	c, err := client.NewClient(client.DefaultDockerHost, obj.APIVersion, nil, nil)
+	if err != nil {
+		return fmt.Errorf("error creating docker client: %s", err)
+	}
+	defer c.Close()
+
+	// validate image
+	resp, err := c.ImageSearch(ctx, obj.Image, types.ImageSearchOptions{Limit: 1})
+	if err != nil {
+		return fmt.Errorf("error searching for image: %s", err)
+	}
+	if len(resp) == 0 {
+		return fmt.Errorf("image: %s not found", obj.Image)
+	}
+
+	// validate env
+	for _, env := range obj.Env {
+		if !strings.Contains(env, "=") || strings.Contains(env, " ") {
+			return fmt.Errorf("invalid environment variable: %s", env)
+		}
+	}
+
+	// validate ports
+	for k, v := range obj.Ports {
+		if k != "tcp" && k != "udp" && k != "sctp" {
+			return fmt.Errorf("ports primary key should be tcp, udp or sctp")
+		}
+		for p, q := range v {
+			if (p < 1 || p > 0xFFFF) || (q < 1 || q > 0xFFFF) {
+				return fmt.Errorf("ports should be between 1 and 65535")
+			}
+		}
+	}
+
+	// validate APIVersion
+	if obj.APIVersion != "" {
+		verOK, err := regexp.MatchString(`^(v)[1-9]\.[0-9]\d*$`, obj.APIVersion)
+		if err != nil {
+			return fmt.Errorf("error matching apiversion string: %s", err)
+		}
+		if !verOK {
+			return fmt.Errorf("invalid apiversion: %s", obj.APIVersion)
+		}
+	}
+
+	return nil
+}
+
+// Init runs some startup code for this resource.
+func (obj *DockerContainerRes) Init(init *engine.Init) error {
+	var err error
+	obj.init = init // save for later
+
+	// Initialize the docker client.
+	obj.client, err = client.NewClient(client.DefaultDockerHost, obj.APIVersion, nil, nil)
+	if err != nil {
+		return fmt.Errorf("error creating docker client: %s", err)
+	}
+
+	return nil
+}
+
+// Close is run by the engine to clean up after the resource is done.
+func (obj *DockerContainerRes) Close() error {
+	return obj.client.Close() // close the docker client
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *DockerContainerRes) Watch() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	eventChan, errChan := obj.client.Events(ctx, types.EventsOptions{})
+
+	// notify engine that we're running
+	if err := obj.init.Running(); err != nil {
+		return err // exit if requested
+	}
+
+	var send = false // send event?
+	for {
+		select {
+		case event, ok := <-eventChan:
+			if !ok { // channel shutdown
+				return nil
+			}
+			if obj.init.Debug {
+				obj.init.Logf("%+v", event)
+			}
+			send = true
+			obj.init.Dirty() // dirty
+		case err, ok := <-errChan:
+			if !ok {
+				return nil
+			}
+			return err
+		case event, ok := <-obj.init.Events:
+			if !ok {
+				return nil
+			}
+			if err := obj.init.Read(event); err != nil {
+				return err
+			}
+		}
+
+		// do all our event sending all together to avoid duplicate msgs
+		if send {
+			send = false
+			if err := obj.init.Event(); err != nil {
+				return err // exit if requested
+			}
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


engine/resources/docker.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// notify engine that we're running
+	if err := obj.init.Running(); err != nil {
+		return err // exit if requested
+	}
+
+	var send = false // send event?
+	for {
+		if obj.init.Debug {
+			obj.init.Logf("Watching: %s", passwdFile) // attempting to watch...
```
jonathangold: Already changed. ;)

engine/resources/docker.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/engine/traits"
+)
+
+func init() {
+	engine.RegisterResource("docker", func() engine.Res { return &DockerRes{} })
+}
+
+// DockerRes is a docker account resource.
+type DockerRes struct {
```
purpleidea: So since docker has turned into a behemoth, I'm wondering if it's best to do like aws and have various sub resources, like aws:ec2, etc... So we'd have docker:image, or whatever...

engine/resources/docker.go
purpleidea: you reviewed your own code, hehe, you can just push a new commit ;)

purpleidea:
> jonathangold force-pushed the jonathangold:docker branch 5 times, most recently from 8efcc5d to 70710ef a day ago 

hehe... is GH getting more rude? ;)
review starting as requested...

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+	"github.com/purpleidea/mgmt/engine/traits"
+	"github.com/purpleidea/mgmt/util"
+)
+
+const (
+	// dockerClientVersion is the api version used by the client. See:
+	// https://docs.docker.com/develop/sdk/#api-version-matrix for more info.
+	dockerClientVersion = "v1.37"
+	// dockerContainerRunning is the running container state.
+	containerRunning = "running"
```
purpleidea: I think the states should probably be public in case we end up pulling them in from the lang somehow, but I think we'll eventually need a clean solution for these types of constants. This is sort of nitpicky, but I mentioned it to get you thinking about what we might eventually do on the language side.

engine/resources/docker.go
jonathangold: It makes sense to create new resources to implement different docker features. `docker:network` comes to mind. If we want to create a `docker:image` resource, it would be a matter of pulling out the image search and image pull requests (about six lines of code,) and plumbing in some auto-edges to the new resource. I think it makes sense to get this merged first with the image stuff, and pull it out later if we need to.

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+	if obj.State != containerRunning && obj.State != containerStopped && obj.State != containerRemoved {
+		return fmt.Errorf("state must be running stopped or removed")
+	}
+
+	// validate image
+	c, err := client.NewClient(client.DefaultDockerHost, dockerClientVersion, nil, nil)
+	if err != nil {
+		return fmt.Errorf("error creating docker client: %s", err)
+	}
+	resp, err := c.ImageSearch(ctx, obj.Image, types.ImageSearchOptions{Limit: 1})
```
purpleidea: You've brought up an interesting point here. My first thought was that these sorts of things should not happen in Validate... But maybe they should. Want to start a short discussion on the mailing list about why or why not we should run longer/network tasks in validate??

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+	if err != nil {
+		return fmt.Errorf("error searching for image: %s", err)
+	}
+	if len(resp) == 0 {
+		return fmt.Errorf("image: %s not found", obj.Image)
+	}
+	c.Close()
```
purpleidea: If we care about the return value (probably an err) then use it, otherwise this should be a defer in the right place I think.

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+	// validate env
+	for _, env := range obj.Env {
+		if !strings.Contains(env, "=") || strings.Contains(env, " ") {
+			return fmt.Errorf("invalid environment variable: %s", env)
+		}
+	}
+
+	// validate ports
+	for k, v := range obj.Ports {
+		if k != "tcp" && k != "udp" {
```
purpleidea: does docker support other protocols? sctp?

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating docker client: %s", err)
+	}
+
+	return nil
+}
+
+// Close is run by the engine to clean up after the resource is done.
+func (obj *DockerRes) Close() error {
+	obj.client.Close()
```
purpleidea: just return this instead?

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+	return nil
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *DockerRes) Watch() error {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, os.Interrupt)
```
purpleidea: why is this here?

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+// CheckApply method for Docker resource.
+func (obj *DockerRes) CheckApply(apply bool) (checkOK bool, err error) {
+	obj.init.Logf("CheckApply(%t)", apply)
+
+	var id string
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, os.Interrupt)
```
purpleidea: see above

engine/resources/docker.go
```diff
@@ -0,0 +1,415 @@
...
+			return false, fmt.Errorf("error creating container: %s", err)
+		}
+		id = resp.ID
+	}
+
+	if err := obj.client.ContainerStart(ctx, id, types.ContainerStartOptions{}); err != nil {
+		return false, fmt.Errorf("error starting container: %s", err)
+	}
+
+	// XXX: There is no ContainerWait for the running state, nor do we get any
```
purpleidea: Can we add a brief spin-lock with a timeout then?

examples/lang/docker0.mcl
```diff
@@ -0,0 +1,6 @@
+docker "mgmt-nginx" {
+    state => "running",
+    image => "nginx",
+    cmd => ["nginx", "-g", "daemon off;",],
```
purpleidea: out of curiosity, why is there a semicolon?

engine/resources/docker.go
purpleidea: Just saw this comment now. Sounds good. Shouldn't we name this something different? docker:???

examples/lang/docker0.mcl
jonathangold: Copied directly from the nginx dockerfile. :P
https://github.com/nginxinc/docker-nginx/blob/f4d30145c60c433966df96f618d78513fee9d322/mainline/stretch/Dockerfile#L99

engine/resources/docker.go
jonathangold: To cancel the context in the event that we want to exit, and there are api requests still outstanding. Should I use a struct context and cancel it in close instead?

engine/resources/docker.go
jonathangold: Yup. Missed sctp. Good catch. :)

engine/resources/docker.go
jonathangold: Yeah. `docker:container` is boring but descriptive. Any better ideas? Also, do colons work with the language yet?

purpleidea: Yes in theory, but not in git master yet, and not tested yet:

https://github.com/purpleidea/mgmt/blob/6405b2371bb7a03a23ab67890fc4530930527f7f/lang/lexer.nex#L362

You can try branch feat/import for new things, but might be broken, eat your dog, get rebased often, etc...

engine/resources/docker.go
```diff
@@ -0,0 +1,414 @@
...
+	// ContainerRunning is the running container state.
+	ContainerRunning = "running"
+	// ContainerStopped is the stopped container state.
+	ContainerStopped = "stopped"
+	// ContainerRemoved is the removed container state.
+	ContainerRemoved = "removed"
+
+	// dockerClientVersion is the api version used by the client. See:
+	// https://docs.docker.com/develop/sdk/#api-version-matrix for more info.
+	dockerClientVersion = "v1.37"
```
purpleidea: So it bugged me that we'd have some constant in here which would potentially rot over time, so I did some source diving, and: api.DefaultVersion exists! woo. But even better, it seems if we use the empty string, then that is taken as the default. Woo, woo!

So I think you should move this to a struct arg that defaults to empty string. Unless it's never something you want to specify in which case just empty string. Also there's this: https://github.com/moby/moby/blob/e6244aa598e9894ead8445ca66329d26eb016e64/client/client.go#L249

What do you think? I'm leaning on struct option + warning to specify a version if desired.

engine/resources/docker.go
```diff
@@ -0,0 +1,414 @@
...
+	eventOpts := types.EventsOptions{
+		Filters: filters.NewArgs(filters.KeyValuePair{"container", id}),
+	}
+	eventCh, errCh := client.Events(ctx, eventOpts)
+	// Start the container.
+	if err := client.ContainerStart(ctx, id, opts); err != nil {
+		return fmt.Errorf("error starting container: %s", err)
+	}
+	// Wait for a message on eventChan that says the container has started.
+	select {
```
purpleidea: Is it possible that either of these never occurs and we block indefinitely?

jonathangold: I don't think so but I'll add a `case <- ctx.Done():` for safety.

engine/resources/docker.go
jonathangold: My only concern is that the way certain methods work, or if they're even available at all, may change depending on the API version, leading to undefined behaviour. I think `api.DefaultVersion` is probably the safest bet, and fits with your "do it with git master" philosophy.

jonathangold: :(
```
18:43:23 main.go:186: main: error validating the new graph: error searching for image: Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.39
```

purpleidea:
Ready for a review?

engine/resources/docker_container.go
```diff
@@ -0,0 +1,441 @@
...
+func init() {
+	engine.RegisterResource("docker:container", func() engine.Res { return &DockerContainerRes{} })
+}
+
+// DockerContainerRes is a docker container resource.
+type DockerContainerRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Edgeable
+
+	State      string                     `yaml:"state"`      // running, stopped, or removed
```
purpleidea: Put these lovely comments in the correct format above each struct field. They'll become docs!

engine/resources/docker_container.go
```diff
@@ -0,0 +1,441 @@
...
+	// validate state
+	if obj.State != ContainerRunning && obj.State != ContainerStopped && obj.State != ContainerRemoved {
+		return fmt.Errorf("state must be running, stopped or removed")
+	}
+
+	// TODO: should this stuff happen in validate or do we make sure there's no
+	// whitespace, and let it fail in CheckApply?
+	//
+	// set up temporary docker client to validate image
+	c, err := client.NewClient(client.DefaultDockerHost, obj.APIVersion, nil, nil)
```
purpleidea: Let's not make any connections in Validate()
Keep it "cheap" so we can casually use Validate() in many places.
Add these checks to Init() or CheckApply. Init seems more likely.
Of course, anything you can do without the connection should be done in Validate.
Why? Well we might consider adding Validate() calls from the lang in order to statically check things that are determined statically as a compiler optimization. Eg: If you asked for state => "wrong", since those are all constants we can catch this at compile time... Only if Validate is cheap =D

engine/resources/docker_container.go
```diff
@@ -0,0 +1,441 @@
...
+		}
+	}
+
+	// validate ports
+	for k, v := range obj.Ports {
+		if k != "tcp" && k != "udp" && k != "sctp" {
+			return fmt.Errorf("ports primary key should be tcp, udp or sctp")
+		}
+		for p, q := range v {
+			if (p < 1 || p > 0xFFFF) || (q < 1 || q > 0xFFFF) {
```
purpleidea: why the 0x instead of just an int?

engine/resources/docker_container.go
```diff
@@ -0,0 +1,441 @@
...
+				return fmt.Errorf("ports should be between 1 and 65535")
```
purpleidea: s/should be/must be/ right?

engine/resources/docker_container.go
```diff
@@ -0,0 +1,441 @@
...
+			}
+		}
+	}
+
+	// validate APIVersion
+	if obj.APIVersion != "" {
+		verOK, err := regexp.MatchString(`^(v)[1-9]\.[0-9]\d*$`, obj.APIVersion)
+		if err != nil {
+			return fmt.Errorf("error matching apiversion string: %s", err)
```
purpleidea: errwrap

engine/resources/docker_container.go
```diff
@@ -0,0 +1,441 @@
...
+		}
+
+		c, err := obj.client.ContainerCreate(ctx, containerConfig, hostConfig, nil, obj.Name())
+		if err != nil {
+			return false, fmt.Errorf("error creating container: %s", err)
+		}
+		id = c.ID
+	}
+
+	return false, containerStart(ctx, obj.client, id, types.ContainerStartOptions{})
```
purpleidea: why do you pass around obj.client instead of just having everything be a private struct method and using it directly??

lang/lexer.nex
```diff
@@ -332,7 +332,7 @@
 			lval.str = strings.ToLower(s) // uncapitalize it
 			return CAPITALIZED_IDENTIFIER
 		}
-/[a-z][a-z0-9]*/
+/[a-z]([a-z0-9:]*[a-z0-9]+)?/
```
purpleidea: Unfortunately please wait for this to land in git master from my branch...
It means you'll just have to change it manually to test the resource, sorry.

engine/resources/docker_container.go
jonathangold: I think 0xFFFF better communicates the reason for the check (max value of uint16). Seems less arbitrary than 65536. I wish go had a simple build-in max and min-value methods on number types. How would you prefer I do it, int const?

engine/resources/docker_container.go
jonathangold: It made the tests a little easier to write, but I don't mind changing it if you don't like it.

engine/resources/docker_container.go
purpleidea: yeah, just put the raw int there. 65535 I think, right? not 65536 ??

engine/resources/docker_container.go
purpleidea: Ah. Interesting point. I think you can just init the tests with &DockerContainerRes{client: whatever,} and then call the methods like that.

jonathangold: Yup, will do.

jonathangold:
@purpleidea I deleted my fork when I was playing with hub, so I have to open a new PR, since it can't find my branch.

PR #403, opened 2018-11-26
engine: resources: docker: Add a docker container resource

jonathangold:
See #400

### DIFF
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -17,6 +17,7 @@ You might want to look at the [generated documentation](https://godoc.org/github
 for more up-to-date information about these resources.
 
 * [Augeas](#Augeas): Manipulate files using augeas.
+* [Docker](#Docker):[Container](#Container) Manage docker containers.
 * [Exec](#Exec): Execute shell commands on the system.
 * [File](#File): Manage files and directories.
 * [Group](#Group): Manage system groups.
@@ -39,6 +40,22 @@ for more up-to-date information about these resources.
 
 The augeas resource uses [augeas](http://augeas.net/) commands to manipulate
 files.
+
+## Docker
+
+### Container
+
+The docker:container resource manages docker containers.
+
+It has the following properties:
+
+* `state`: either `running`, `stopped`, or `removed`
+* `image`: docker `image` or `image:tag`
+* `cmd`: a command or list of commands to run on the container
+* `env`: a list of environment variables, e.g. `["VAR=val",],`
+* `ports`: a map of portmappings, e.g. `{"tcp" => {80 => 8080, 443 => 8443,},},`
+* `apiversion:` override the host's default docker version, e.g. `"v1.35"`
+* `force`: destroy and rebuild the container instead of erroring on wrong image
 
 ## Exec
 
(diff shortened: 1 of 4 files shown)


jonathangold:
Sweet. I'll have a crack at the submodule thing tomorrow, and if I can't figure something out, I'll ping you for help.

jonathangold:
@purpleidea I pushed the commit without the git submodule. [Here's the problem](https://travis-ci.org/purpleidea/mgmt/jobs/460010959#L1339) that I'm trying to solve. As you can see, the [container.Config type](https://godoc.org/github.com/docker/docker/api/types/container#Config) uses a vendored type from https://github.com/docker/go-connections/nat, which is just wrong.

My fix was to vendor https://github.com/docker/docker, delete the vendored `go-connections` from `vendor/github.com/docker/docker/vendor/github.com/docker/`, commit the changes and add my commit as a submodule. I'm really not sure how to solve this without pointing it to my fork.

If you have a few minutes to look at it, I'd really appreciate the help. ;)

docs/resources.md
```diff
@@ -17,6 +17,7 @@ You might want to look at the [generated documentation](https://godoc.org/github
 for more up-to-date information about these resources.
 
 * [Augeas](#Augeas): Manipulate files using augeas.
+* [Docker:Container](#Docker:Container) Manage docker containers.
```
jonathangold: Should these all be lower case?

purpleidea:
@jonathangold Okay, sorry for the delay, I had a look today.

You were quite right. AFAICT, it's not possible with a source change. What they're doing is absurd. The magic keywords you could have mentioned were:

"Dear $DEITY, they're statically copying dependencies into the vendor/ directory add adding them to git."

Obviously this "professional" project doesn't understand professional tools like git submodules, or even in the vendor case to use a tool which stores some sort of yaml file, but at least doesn't commit those changes to git.

Sorry I'm ranty. This is absurd. I guess it's the only way they know to deal with their bloat.

One exception though: I was able to just add git submodules for the two docker dependencies to our git submodules, and just add a patch on top of each that did `rm -rf vendor/`, which I think is preferable to having a `jonathangold/docker/...` or `purpleidea/docker/...` fork. Someone will occasionally need to update the dep and re-apply that sort of patch on top. If you'd prefer a different version of either dep, let me know, I just took git master. Is it better to choose a specific tag?

Lastly, I put this patch under your commit here:
https://github.com/purpleidea/mgmt/tree/feat/docker-lol
I can update and merge after you're ready.

Thanks again!

purpleidea:
Oh wait. One sec...

jonathangold:
Vendoring `github.com/docker/docker` and applying `rm -rf vendor/github.com/docker/go-connections` should do the trick, if you want to patch it.

purpleidea:
I've merged this in https://github.com/purpleidea/mgmt/commit/51ec91dd16097f5fe962042b74675f2fe3c83f40 !
(I rebased in the vendor commit just before this, and added a build tag right after.)

Woo!
Any changes we need to make we can patch on top. I didn't consider all of the docker logic exceptionally close, IOW, I'm not personally going to support it (if someone is using it in production, I would recommend more tests be added here!), but happy to have it in, and thanks for your hard work!

PR #404, opened 2018-11-27
make: Release pipeline

jonathangold:
This commit adds new make targets for rpm, deb, and pacman packages.
It also adds a phony target that uploads tarballs of the packages,
along with their signed (and unsigned) checksums to the github release
page. Once the current commit is tagged as a release, run `make release`
to build the packages and upload them to github.

See https://github.com/jonathangold/mgmt/releases/tag/0.0.16 for a preview of the output,


Makefile
```diff
@@ -324,16 +328,76 @@ copr: upload-srpms ## build in copr
...
+	@echo -e "Archiving deb package..."
+	tar -zcf $(DEB_TAR) -C release/deb .
+
+release/deb/sha256sum.asc: release/deb/sha256sum
+	@echo "Signing sha256 sum..."
+	gpg2 --yes --clearsign release/deb/sha256sum
+
+release/deb/sha256sum: release/deb/mgmt_$(VERSION)_amd64.deb
+	@echo "Generating sha256 sum..."
+	sha256sum release/deb/*.deb > release/deb/sha256sum
```
purpleidea: standard output file name is SHA256SUMS IIRC.

Makefile
```diff
@@ -324,16 +328,76 @@ copr: upload-srpms ## build in copr
...
+
+release/deb/changelog:
+	@echo "Generating deb changelog..."
+	./misc/make-deb-changelog.sh
+
+$(RPM_TAR): release/rpm/sha256sum.asc
+	@echo -e "Archiving rpm package..."
+	tar -zcf $(RPM_TAR) -C release/rpm .
+
+release/rpm/sha256sum.asc: release/rpm/sha256sum
```
purpleidea: standard signed SHA256SUMS is named SHA256SUMS.asc IIRC.

Makefile
```diff
@@ -324,16 +328,76 @@ copr: upload-srpms ## build in copr
...
+	@echo "Creating github release..."
+	hub release create \
+		-F <( echo -e "$(VERSION)\n";echo "License: GPLv3" ) \
+		-a $(DEB_TAR) \
+		-a $(RPM_TAR) \
+		-a $(PACMAN_TAR) \
+		$(VERSION)
+
+release/mkdirs:
+	mkdir -p release/{deb,rpm,pacman}
```
purpleidea: One nit pick (sorry!) Could you keep the target named release (awesome) but name the output folder `releases` please?

misc/fpm-pack.sh
```diff
@@ -0,0 +1,65 @@
...
+# maintainer email
+MAINTAINER="mgmt@noreply.github.com"
+# project url
+URL="https://github.com/purpleidea/mgmt"
+# project description
+DESCRIPTION="Next generation distributed, event-driven, parallel config management!"
+# project license
+LICENSE="GPLv3"
+# location to install the binary
+PREFIX="/usr/bin"
```
purpleidea: Is this correct or should it be `/usr/local/bin/` ?

misc/make-deps.sh
```diff
@@ -103,4 +105,5 @@ go get golang.org/x/lint/golint				# for `golint`-ing
 go get github.com/tmthrgd/go-bindata/go-bindata	# for compiling in non golang files
 go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
 command -v mdl &>/dev/null || gem install mdl || true	# for linting markdown files
+command -v fpm &>/dev/null || gem install fpm || true	# for cross distro packaging
```
purpleidea: Hmm. Do we want to use the distro package instead? Is there one everywhere??

.gitignore
```diff
@@ -13,4 +13,4 @@ mgmt.static
 build/mgmt-*
 mgmt.iml
 rpmbuild/
-*.deb
```
purpleidea: Isn't this still used if we do debian/ building normally??

misc/make-deps.sh
purpleidea: Yeah good call. I misread something :/

misc/fpm-pack.sh
purpleidea: Hmm... Interesting... I guess I always thought custom stuff like fpm packages (not stock distro packages) belonged in `local` or `/opt`. I'm okay leaving it as is, and if someone wants to change this later we can do it then.

jonathangold:
@purpleidea. Done. \o/

purpleidea:
\o/

purpleidea:
I've merged this in https://github.com/purpleidea/mgmt/commit/51ec91dd16097f5fe962042b74675f2fe3c83f40 !
(I rebased in the vendor commit just before this, and added a build tag right after.)

Woo!
Any changes we need to make we can patch on top. I didn't consider all of the docker logic exceptionally close, IOW, I'm not personally going to support it (if someone is using it in production, I would recommend more tests be added here!), but happy to have it in, and thanks for your hard work!

purpleidea:
Found one bug...
Please test after running an `rm -rf releases/`
I think you need a dependency on the mkdirs...
```
diff --git a/Makefile b/Makefile
index c6c76d3..b171956 100644
--- a/Makefile
+++ b/Makefile
@@ -359,7 +359,7 @@ releases/deb/mgmt_$(VERSION)_amd64.deb: releases/deb/changelog
        @echo "Building deb package..."
        ./misc/fpm-pack.sh deb libvirt-dev libaugeas-dev
 
-releases/deb/changelog: $(PROGRAM)
+releases/deb/changelog: $(PROGRAM) release/mkdirs
        @echo "Generating deb changelog..."
        ./misc/make-deb-changelog.sh
 
@@ -379,7 +379,7 @@ releases/rpm/mgmt-$(VERSION)-1.x86_64.rpm: releases/rpm/changelog
        @echo "Building rpm package..."
        ./misc/fpm-pack.sh rpm libvirt-devel augeas-devel
 
-releases/rpm/changelog: $(PROGRAM)
+releases/rpm/changelog: $(PROGRAM) release/mkdirs
        @echo "Generating rpm changelog..."
        ./misc/make-rpm-changelog.sh

```

After that, make a fake tag and run `make release` twice. First time should work, second time should do nothing.

Cheers!

jonathangold:
You're right about the dependency, but the `$(PROGRAM)` and `release/mkdirs` targets are causing it to rebuild every time. I'll revisit it a little later.

purpleidea:
On Thu, Nov 29, 2018 at 9:12 AM Jonathan Gold <notifications@github.com> wrote:
>
> You're right about the dependency, but the $(PROGRAM) and release/mkdirs targets are causing it to rebuild every time. I'll revisit it a little later.

I think it's probably because since mkdirs is a fake target it re-runs
each time, which updates dir timestamps, which invalidates the
targets!

The solution is perhaps to just add a README file or similar in the
releases/ dir so that it's added to git, and then make sure your build
changelog script makes the dir it needs if it doesn't exist.

HTH

jonathangold:
So I solved the `release/mkdirs` issue by adding it as an order-only prerequesite, but the release target is still trying to run, since it's phony too. It errors because the github release already exists, so it doesn't actually do anything. Here's the output of the second run: (note the prerequesites for $(PROGRAM) are running too. Not a huge deal, but probably a bug.)

```
jon@localhost ~/go/src/github.com/jonathangold/mgmt 🗙 (fpm) $ make release
Generating: bindata...
Generating: lang...
make --quiet -C lang
mkdir -p releases/{deb,rpm,pacman}
Creating github release...
hub release create \
	-F <( echo -e "0.0.16\n";echo "License: GPLv3" ) \
	-a releases/mgmt_0.0.16_deb_amd64.tar.gz \
	-a releases/mgmt_0.0.16_rpm_x86_64.tar.gz \
	-a releases/mgmt_0.0.16_pacman_x86_64.tar.gz \
	0.0.16
Error creating release: Unprocessable Entity (HTTP 422)
Duplicate value for "tag_name"
make: *** [Makefile:336: release] Error 1
```

purpleidea:
> order-only prerequesite

What's that?

> the release target is still trying to run, since it's phony too

If you run `make release` since it's phony, it will run again, but it's dependencies don't do anything since they're not phony, so it should be okay.

> It errors because the github release already exists

This can be easily prevented by adding a target that either runs a command looking for a version existing (like git tag) or by looking if a file with a certain version in it exists. Right?

> note the prerequesites for $(PROGRAM) are running too. Not a huge deal, but probably a bug.

On git master if I run `make build` twice, only the first run does stuff. Is that not what you get on git master??

jonathangold:
> What's that?

https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

> If you run make release since it's phony, it will run again, but it's dependencies don't do anything since they're not phony, so it should be okay.

Yup.

> On git master if I run make build twice, only the first run does stuff. Is that not what you get on git master??

```
jon@localhost ~/go/src/github.com/jonathangold/mgmt $ make build
Generating: bindata...
Generating: lang...
make --quiet -C lang
jon@localhost ~/go/src/github.com/jonathangold/mgmt $ 
```

jonathangold:
Vis-a-vis checking if the release exists, there's `hub release show <tag>`. The output isn't terribly parse-able, but it's probably the right thing to check.

purpleidea:
RE:

```
jon@localhost ~/go/src/github.com/jonathangold/mgmt $ make build
Generating: bindata...
Generating: lang...
make --quiet -C lang
jon@localhost ~/go/src/github.com/jonathangold/mgmt $ 
```

This is exactly fine... The messages are because it's calling the targets, but it doesn't actually do any work. Technically we could move the messages to non-phony targets so that they don't make noise, and then have our linking call those. That's a welcome 5min patch if you're in the mood.

purpleidea:
> there's hub release show <tag>. The output isn't terribly parse-able, but it's probably the right thing to check.

Are you sure??
Ideally we don't want a network call made every time I tab complete from make.

jonathangold:
I used the output of hub release to write the release url to a file target, and made the phony release target depend on that. Works as expected now.

Makefile
```diff
@@ -324,16 +328,79 @@ copr: upload-srpms ## build in copr
...
+
+releases/mgmt-$(VERSION)-release.url: $(DEB_TAR) $(RPM_TAR) $(PACMAN_TAR)
+	@echo "Creating github release..."
+	hub release create \
+		-F <( echo -e "$(VERSION)\n";echo "License: GPLv3" ) \
+		-a $(DEB_TAR) \
+		-a $(RPM_TAR) \
+		-a $(PACMAN_TAR) \
+		$(VERSION) \
+		> releases/mgmt-$(VERSION)-release.url
```
purpleidea: I think if you're going this method, you need an `|| rm -f releases/mgmt-$(VERSION)-release.url` at the end so the file is deleted if there's an error. (Or something similar)

And if the `hub release create` command echos anything useful, then you could tack on a `&& cat releases/mgmt-$(VERSION)-release.url` before the `||` so that we print it out to screen too. =D

jonathangold: Done and tested. :D

Makefile
```diff
@@ -324,16 +328,81 @@ copr: upload-srpms ## build in copr
...
+		-F <( echo -e "$(VERSION)\n";echo "License: GPLv3" ) \
+		-a $(DEB_TAR) \
+		-a $(RPM_TAR) \
+		-a $(PACMAN_TAR) \
+		$(VERSION) \
+		> releases/mgmt-$(VERSION)-release.url \
+		&& cat releases/mgmt-$(VERSION)-release.url \
+		|| rm -f releases/mgmt-$(VERSION)-release.url
+
+release/mkdirs:
```
jonathangold: should this be `releases/mkdirs`?

purpleidea:
On Fri, Nov 30, 2018 at 7:34 AM Jonathan Gold <notifications@github.com> wrote:
> should this be releases/mkdirs?

nope

purpleidea:
LGTM

purpleidea:
Hmmm... Just did a soft release... Shouldn't make release upload the rpm/deb/etc instead of the .tar ??

https://github.com/purpleidea/mgmt/blob/master/Makefile#L340

purpleidea:
Okay fixed in git master.

PR #405, opened 2018-12-03
net tests

jonathangold:
See #378 for previous feedback


jonathangold:
I responded to your last set of comments in #378 a while back, but I think it got lost in the shuffle.

We should pick up where we left off. The main change we still needed to discuss, iirc, was removing the requirement to set the address and default gateway together.

If you recall, when I submitted #378, travis would lose track of the tests because I was adding a new default gateway to my dummy interface, which was messing with the actual interface's routes (multiple gateways on the same routing table.) So, allowing the resource to set one without the other should be supported, and in the case of travis, may be necessary.

Ultimately, I think the correct solution is to extend the net resource to support additional routing tables, since multiple default routes on the same routing table can lead to undefined behaviour (like travis losing track of the build.) But that's beyond the scope of this patch.

engine/resources/net.go
```diff
@@ -880,6 +880,6 @@ func (obj *socketSet) nfd() int {
 func (obj *socketSet) fdSet() *unix.FdSet {
 	fdSet := &unix.FdSet{}
 	fdSet.Bits[obj.fdEvents/64] |= 1 << uint(obj.fdEvents)
-	fdSet.Bits[obj.fdPipe/64] |= 1 << uint(obj.fdPipe) // fd = 3 becomes 100 if we add 5, we get 10100
+	fdSet.Bits[obj.fdPipe/64] |= 1 << uint(obj.fdPipe) // fd = 3 becomes 1000 if we add 5, we get 101000
```
purpleidea: Your reply about this was:

> I was off by one when I wrote the initial comment. The right-most bit represents 0, not 1, so 1 => 10, 2 => 100, 3 => 1000. The way the initial comment was written, 1,2 would give you 11, but it actually gives you 110. Does that make sense?

Just add a comment above these shifts please.

engine/resources/net.go
```diff
@@ -119,9 +119,6 @@ func (obj *NetRes) Validate() error {
 	}
 
 	// validate network address input
-	if (obj.Addrs == nil) != (obj.Gateway == "") {
```
purpleidea: Previously you wrote:

> 

I tried to explain it in the commit message, but there are scenarios where you'd want to set one without the other. For example, if you're setting up a secondary interface on a different subnet, you might not want to change the default gateway, as having 2 defaults (on the same routing table) leads to issues. That's why travis was losing track of the build when the shell tests added a gateway.

I'm not 100% groking this, but if it's important somewhere, I can merge. I'd prefer to merge it with an example of it being used perhaps? Eg add to this commit something in examples/lang/ perhaps?

purpleidea:
And for some reason travis keeps failing in the first two min running apt... Hmm... Is this travis just being picky again?

jonathangold:
@purpleidea I added the comment in fdSet(). 

re travis: It looks like one of the repos has a non-existent or expired gpg key again.

I'll respond to your other comment a bit later, but in the meantime, [here's a decent writeup](https://kindlund.wordpress.com/2007/11/19/configuring-multiple-default-routes-in-linux/) on what it takes to set up multiple default routes properly (including what can go wrong if you have multiple default routes on the same table.)

purpleidea:
@jonathangold Okay, I cherry-picked another commit and merged!

Can you rebase this patch to master and push again? I'd like to get the shell test to pass in travis before I merge it. The addrs patch I need to think about a bit first. It still needs a concrete example. It might be useful, but there might be a better way *or* it might be too abstract for our needs at this time. So as long as we have a use for it with a _real_ example, then I'll merge it. is that fair? Thanks!!

jonathangold:
@purpleidea the shell test WON'T pass in travis without the change, afaict, since it would require setting a second default route on travis. That'll cause the build to get lost. It works locally, but travis loses the console when the route gets added, and the build times out.

purpleidea:
@jonathangold Why would the test interfere with travis if it's a separate interface that we're touching that's unrelated to the eth0 or whatever it uses?

jonathangold:
Because we're adding a second default route to the default routing table. The only supported config with more than one default gateway involves setting each one up on a dedicated routing table, afaik.

Since the default routes are just routes whose destination is 0.0.0.0, linux doesn't know which one to pick, and packets that should be going out (to travis-ci) on eth0 gets sent out on the dummy interface.

IOW, having multiple default routes means that, if no explicit route exists, one of the default routes will be picked at random.

purpleidea:
> Because we're adding a second default route to the default routing table. The only supported config with more than one default gateway involves setting each one up on a dedicated routing table, afaik.

Oh yeah. I've been kind of slow about this, sorry. I'll wait for travis to work and then we can merge. Shouldn't you rebase so that the change removing this happens before the shell patch so that it passes??

jonathangold:
It does, but github is showing them in the commit date order. Since I found the bug after writing the test, its timestamp is after the shell test even though it's in the right order in my git log.

```
jon@localhost ~/mgmt ● (testnet) $ git log
commit 12f1bbae2d3960cbdb62d8d700cdeaeee77de167 (HEAD -> testnet, jgold/testnet)
Author: Jonathan Gold <jonathan.e.gold@gmail.com>
Date:   Mon Apr 2 00:15:39 2018 -0400

    engine: resources: net: test: Add shell test for net resource
    
    This patch adds a shell test for net, which creates a dummy interface
    and runs mgmt to bring it up and assign it with an address. It then
    checks if the state was applied correctly. Finally, it runs mgmt again
    to bring the interface down, and tests that it comes down and stays
    down.

commit d13f6a7223844d980d5321acb3c2a32e493b9807
Author: Jonathan Gold <jonathan.e.gold@gmail.com>
Date:   Thu Apr 5 08:24:10 2018 -0400

    engine: resources: net: Allow addr without gateway
    
    In some scenarios it is desirable to set the addrs and gateway
    independently, i.e. if a default gateway is already set on
    the machine. This patch removes the requirement to set them
    together.

```

purpleidea:
@jonathangold Oh! Weird. That's a GH bug! LGTM then... Waiting for travis, thanks!

purpleidea:
@jonathangold Do you mind rebasing to new current master and pushing again so that the tests run? ( patched the travis breakage.)

PR #407, opened 2018-12-04
make: remove spurious dependency package 'rubygems' for Debian-like systems

ffrank:
On Ubuntu, the apt-get install call to ruby, ruby-devel, and rubygems will
fail because there is no "rubygems" package in Ubuntu.

In Debian, this package is virtual only. In both cases, the ruby package
is sufficient. (See also https://packages.debian.org/jessie/rubygems)

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -39,7 +39,7 @@ fi
 if [ ! -z "$APT" ]; then
 	$sudo_command $APT install -y libvirt-dev || true
 	$sudo_command $APT install -y libaugeas-dev || true
-	$sudo_command $APT install -y ruby ruby-dev rubygems || true
+	$sudo_command $APT install -y ruby ruby-dev || true
 	$sudo_command $APT install -y libpcap0.8-dev || true
 	# dependencies for building packages with fpm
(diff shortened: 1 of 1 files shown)


purpleidea:
@jonathangold Mind doing a review here please? You're touching more deb/ubuntu systems than I am lately. Thanks!

purpleidea:
On Tue, Dec 4, 2018 at 1:21 AM Jonathan Gold <notifications@github.com> wrote:
> Otherwise LGTM

Does anything in make-deps.sh need `gem` from installing it the current way?

jonathangold:
@purpleidea the `ruby-devel` rpm and the `ruby` deb both include the gem command. I uninstalled `rubygems` on my machine and reinstalled `ruby-devel` to make sure.

ffrank:
FWIW as mentioned in IRC, I had this happen: https://gist.github.com/ffrank/3da501a10eb71c281d4ebc472078a4a2

The Ubuntu 14 machine ended up not getting ruby-dev at all, and then the gem installation fails.

I did not try to reproduce with other Ubuntus. But I'm quite sure it won't break anything.

purpleidea:
Rebased to master (to include the travis fix) and merged! Thanks!

PR #408, opened 2018-12-04
make: Speed up the build by skipping gem docs

ffrank:
Per default, the Ruby gems renerate documentation in two distinct formats
during installation. By passing --no-ri and --no-rdoc, gem is instructed
to skip this step for both formats.

If the user needs documentation for any of the gems after all, they can
manually generate the docs themselves.


purpleidea:
LGTM :) Thanks!

purpleidea:
Rebased to master (to include the travis fix) and merged! Thanks!

PR #409, opened 2018-12-04
travis: Build on Xenial

jonathangold:
Builds were failing on Trusty due to broken GPG keys, and upgrading
the build environment to Xenial Xerus solves the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ go:
   - tip
 go_import_path: github.com/purpleidea/mgmt
 sudo: true
-dist: trusty
+dist: xenial
 # travis requires that you update manually, and provides this key to trigger it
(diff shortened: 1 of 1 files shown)


purpleidea:
heh, I already pushed a fix, but maybe we should do this too? Let's see if it passes...

jonathangold:
Lol, that was quick. It's probably worth upgrading at some point anyway. :P

purpleidea:
Yeah agreed... Let's see what travis says. In the meantime: https://github.com/purpleidea/mgmt/pull/405#issuecomment-444274637 (to git master, not to this)

jonathangold:
Done!

purpleidea:
Merged, thanks!

PR #411, opened 2018-12-06
engine: resources: file: Fix bug preventing use of source

jonathangold:
This patch fixes a previously undiscovered bug which prevented
the use of the source field in the file resource. CheckApply was
returning early if obj.Content was nil. It is also necessary to
check that obj.Source is empty before returning, otherwise
syncCheckApply never runs.

### DIFF
--- a/engine/resources/file.go
+++ b/engine/resources/file.go
@@ -608,7 +608,7 @@ func (obj *FileRes) contentCheckApply(apply bool) (checkOK bool, _ error) {
 	}
 
 	// content is not defined, leave it alone...
-	if obj.Content == nil {
+	if obj.Content == nil && obj.Source == "" {
 		return true, nil
 	}
 
--- a/test/shell/file-source.mcl
+++ b/test/shell/file-source.mcl
@@ -1,0 +1,4 @@
+file "/tmp/mgmt/file-source.txt" {
+	source => "file-source.txt",
+	state => "exists",
+}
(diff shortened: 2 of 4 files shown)


jonathangold:
I'll whip up a shell test in a subsequent pull request, to make sure that obj.Source is working.

purpleidea:
Cool-- are we sure about this one?
TBH file resource (being literally the first thing I wrote in mgmt) needs some cleanups! Eg: https://github.com/purpleidea/mgmt/blob/b2d726051bd192b32a1308cc3ce8ad4951623773/engine/resources/file.go#L181
I'm waiting on the upcoming filesystem changes in my big module patch to add new functions that can read content on disk and return strings, eg:

```
# normal
file "/tmp/foo" {
        content => "hello\n",
}

# coming soon!
file "/tmp/bar" {
        content => file("/etc/passwd"), # copy that file from disk here
}

# also coming soon!
file "/tmp/bar" {
        content => files("static/passwd"), # copy that file from our deploy filesystem
}

# or perhaps fancier:
file "/tmp/bar" {
        content => template("templates/passwd.tmpl"), # template something
}

# or directories?? (uncertain)
file "/tmp/dir/" {
        content => "/var/foo/", # or do we use source =>
        # what about copying a dir *from* the deploy filesystem???
}
```

jonathangold:
Yeah. Can't get the source feature working without this change.

purpleidea:
TBH, I'm still a little unsure about this, but I'm willing to merge it for now until I've got time to write more tests cases and language functions. So merged! =D

(This is because, previously up until now, I've defined the behaviour I want in my head, but it needs to be transferred into tests. So my bad for not having that yet.)

PR #414, opened 2018-12-09
WiP: initial implementation of the langpuppet GAPI

ffrank:
Not for merging yet. Just for easier discussion.

### DIFF
--- a/langpuppet/gapi.go
+++ b/langpuppet/gapi.go
@@ -1,0 +1,324 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package langpuppet
+
+import (
+	"flag"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/gapi"
+	"github.com/purpleidea/mgmt/lang"
+	"github.com/purpleidea/mgmt/pgraph"
+	"github.com/purpleidea/mgmt/puppet"
+
+	multierr "github.com/hashicorp/go-multierror"
+	errwrap "github.com/pkg/errors"
+	"github.com/urfave/cli"
+)
+
+const (
+	// Name is the name of this frontend.
+	Name = "langpuppet"
+	// FlagPrefix gets prepended to each flag of both the puppet and lang GAPI.
+	FlagPrefix = "lp-"
+)
+
+func init() {
+	gapi.Register(Name, func() gapi.GAPI { return &GAPI{} }) // register
+}
+
+// GAPI implements the main langpuppet GAPI interface.
+// It wraps the Puppet and Lang GAPIs and receives graphs from both.
+// It then runs a merging algorithm that mainly just makes a union
+// of both the sets of vertices and edges. Some vertices are merged
+// using a naming convention. Details can be found in the
+// langpuppet.mergeGraphs function.
+type GAPI struct {
+	langGAPI   gapi.GAPI // the wrapped lang entrypoint
+	puppetGAPI gapi.GAPI // the wrapped puppet entrypoint
+
+	currentLangGraph   *pgraph.Graph // the most recent graph received from lang
+	currentPuppetGraph *pgraph.Graph // the most recent graph received from puppet
+
+	langGraphReady   bool // flag to indicate that a new graph from lang is ready
+	puppetGraphReady bool // flag to indicate that a new graph from puppet is ready
+	graphFlagMutex   *sync.Mutex
+
+	data        gapi.Data
+	initialized bool
+	closeChan   chan struct{}
+	wg          sync.WaitGroup // sync group for tunnel go routines
+}
+
+// Cli takes a cli.Context, and returns our GAPI if activated. All arguments
+// should take the prefix of the registered name. On activation, if there are
+// any validation problems, you should return an error. If this was not
+// activated, then you should return a nil GAPI and a nil error.
+func (obj *GAPI) Cli(c *cli.Context, fs engine.Fs) (*gapi.Deploy, error) {
+	if !c.IsSet(FlagPrefix+lang.Name) && !c.IsSet(FlagPrefix+puppet.Name) {
+		return nil, nil
+	}
+
+	if !c.IsSet(FlagPrefix+lang.Name) || c.String(FlagPrefix+lang.Name) == "" {
+		return nil, fmt.Errorf("%s input is empty", FlagPrefix+lang.Name)
+	}
+	if !c.IsSet(FlagPrefix+puppet.Name) || c.String(FlagPrefix+puppet.Name) == "" {
+		return nil, fmt.Errorf("%s input is empty", FlagPrefix+puppet.Name)
+	}
+
+	flagSet := flag.NewFlagSet(Name, flag.ContinueOnError)
+
+	for _, flag := range c.FlagNames() {
+		if !c.IsSet(flag) {
+			continue
+		}
+		childFlagName := strings.TrimPrefix(flag, FlagPrefix)
+		flagSet.String(childFlagName, "", "no usage string needed here")
+		flagSet.Set(childFlagName, c.String(flag))
+	}
+
+	var langDeploy *gapi.Deploy
+	var puppetDeploy *gapi.Deploy
+	var err error
+
+	// we don't really need the deploy object from the child GAPIs
+	if langDeploy, err = (&lang.GAPI{}).Cli(cli.NewContext(c.App, flagSet, nil), fs); err != nil {
+		return nil, err
+	}
+	if puppetDeploy, err = (&puppet.GAPI{}).Cli(cli.NewContext(c.App, flagSet, nil), fs); err != nil {
+		return nil, err
+	}
+
+	return &gapi.Deploy{
+		Name: Name,
+		Noop: c.GlobalBool("noop"),
+		Sema: c.GlobalInt("sema"),
+		GAPI: &GAPI{
+			langGAPI:   langDeploy.GAPI,
+			puppetGAPI: puppetDeploy.GAPI,
+		},
+	}, nil
+}
+
+// CliFlags returns a list of flags used by this deploy subcommand.
+// It consists of all flags accepted by lang and puppet mode,
+// with a respective "lp-" prefix.
+func (obj *GAPI) CliFlags() []cli.Flag {
+	langFlags := (&lang.GAPI{}).CliFlags()
+	puppetFlags := (&puppet.GAPI{}).CliFlags()
+
+	var childFlags []cli.Flag
+	for _, flag := range append(langFlags, puppetFlags...) {
+		childFlags = append(childFlags, &cli.StringFlag{
+			Name:  FlagPrefix + strings.Split(flag.GetName(), ",")[0],
+			Value: "",
+			Usage: fmt.Sprintf("equivalent for '%s' when using the lang/puppet entrypoint", flag.GetName()),
+		})
+	}
+
+	return childFlags
+}
+
+// Init initializes the langpuppet GAPI struct.
+func (obj *GAPI) Init(data gapi.Data) error {
+	if obj.initialized {
+		return fmt.Errorf("already initialized")
+	}
+	obj.data = data // store for later
+	obj.graphFlagMutex = &sync.Mutex{}
+
+	dataLang := gapi.Data{
+		Program:       obj.data.Program,
+		Hostname:      obj.data.Hostname,
+		World:         obj.data.World,
+		Noop:          obj.data.Noop,
+		NoConfigWatch: obj.data.NoConfigWatch,
+		NoStreamWatch: obj.data.NoStreamWatch,
+		Debug:         obj.data.Debug,
+		Logf: func(format string, v ...interface{}) {
+			obj.data.Logf(lang.Name+": "+format, v...)
+		},
+	}
+	dataPuppet := gapi.Data{
+		Program:       obj.data.Program,
+		Hostname:      obj.data.Hostname,
+		World:         obj.data.World,
+		Noop:          obj.data.Noop,
+		NoConfigWatch: obj.data.NoConfigWatch,
+		NoStreamWatch: obj.data.NoStreamWatch,
+		Debug:         obj.data.Debug,
+		Logf: func(format string, v ...interface{}) {
+			obj.data.Logf(puppet.Name+": "+format, v...)
+		},
+	}
+
+	if err := obj.langGAPI.Init(dataLang); err != nil {
+		return err
+	}
+	if err := obj.puppetGAPI.Init(dataPuppet); err != nil {
+		return err
+	}
+
+	obj.closeChan = make(chan struct{})
+	obj.initialized = true
+	return nil
+}
+
+// Graph returns a current Graph.
+func (obj *GAPI) Graph() (*pgraph.Graph, error) {
+	if !obj.initialized {
+		return nil, fmt.Errorf("%s: GAPI is not initialized", Name)
+	}
+
+	var err error
+	obj.graphFlagMutex.Lock()
+	if obj.langGraphReady {
+		obj.langGraphReady = false
+		obj.graphFlagMutex.Unlock()
+		obj.currentLangGraph, err = obj.langGAPI.Graph()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		obj.graphFlagMutex.Unlock()
+	}
+
+	obj.graphFlagMutex.Lock()
+	if obj.puppetGraphReady {
+		obj.puppetGraphReady = false
+		obj.graphFlagMutex.Unlock()
+		obj.currentPuppetGraph, err = obj.puppetGAPI.Graph()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		obj.graphFlagMutex.Unlock()
+	}
+
+	g, err := mergeGraphs(obj.currentLangGraph, obj.currentPuppetGraph)
+
+	if obj.data.Debug {
+		obj.currentLangGraph.Logf(func(format string, v ...interface{}) {
+			obj.data.Logf("graph: "+lang.Name+": "+format, v...)
+		})
+		obj.currentPuppetGraph.Logf(func(format string, v ...interface{}) {
+			obj.data.Logf("graph: "+puppet.Name+": "+format, v...)
+		})
+		if err == nil {
+			g.Logf(func(format string, v ...interface{}) {
+				obj.data.Logf("graph: "+Name+": "+format, v...)
+			})
+		}
+	}
+
+	return g, err
+}
+
+// Next returns nil errors every time there could be a new graph.
+func (obj *GAPI) Next() chan gapi.Next {
+	ch := make(chan gapi.Next)
+	obj.wg.Add(1)
+	go func() {
+		defer obj.wg.Done()
+		defer close(ch) // this will run before the obj.wg.Done()
+		if !obj.initialized {
+			next := gapi.Next{
+				Err:  fmt.Errorf("%s: GAPI is not initialized", Name),
+				Exit: true, // exit, b/c programming error?
+			}
+			ch <- next
+			return
+		}
+		nextLang := obj.langGAPI.Next()
+		nextPuppet := obj.puppetGAPI.Next()
+
+		firstLang := false
+		firstPuppet := false
+
+		for {
+			var err error
+			exit := false
+			select {
+			case nextChild := <-nextLang:
+				if nextChild.Err != nil {
+					err = nextChild.Err
+					exit = nextChild.Exit
+				} else {
+					obj.graphFlagMutex.Lock()
+					obj.langGraphReady = true
+					obj.graphFlagMutex.Unlock()
+					firstLang = true
+				}
+			case nextChild := <-nextPuppet:
+				if nextChild.Err != nil {
+					err = nextChild.Err
+					exit = nextChild.Exit
+				} else {
+					obj.graphFlagMutex.Lock()
+					obj.puppetGraphReady = true
+					obj.graphFlagMutex.Unlock()
+					firstPuppet = true
+				}
+			case <-obj.closeChan:
+				return
+			}
+
+			if (!firstLang || !firstPuppet) && err == nil {
+				continue
+			}
+
+			if err == nil {
+				obj.data.Logf("generating new composite graph...")
+			}
+			next := gapi.Next{
+				Exit: exit,
+				Err:  err,
+			}
+
+			select {
+			case ch <- next: // trigger a run (send a msg)
+			// unblock if we exit while waiting to send!
+			case <-obj.closeChan:
+				return
+			}
+		}
+	}()
+	return ch
+}
+
+// Close shuts down the Puppet GAPI.
+func (obj *GAPI) Close() error {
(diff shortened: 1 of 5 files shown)


langpuppet/gapi.go
```diff
@@ -0,0 +1,290 @@
...
+	closeChan   chan struct{}
+	wg          sync.WaitGroup // sync group for tunnel go routines
+}
+
+// Cli takes a cli.Context, and returns our GAPI if activated. All arguments
+// should take the prefix of the registered name. On activation, if there are
+// any validation problems, you should return an error. If this was not
+// activated, then you should return a nil GAPI and a nil error.
+func (obj *GAPI) Cli(c *cli.Context, fs engine.Fs) (*gapi.Deploy, error) {
+	if s := c.String(Name); c.IsSet(Name) {
```
purpleidea: You can inline this test's body. I didn't in other code and I forget why (probably didn't notice) but I fixed that in an upcoming feature branch.

langpuppet/gapi.go
```diff
@@ -0,0 +1,290 @@
...
+		if s == "" {
+			return nil, fmt.Errorf("%s input is empty", Name)
+		}
+
+		myFlags := All(c.FlagNames(), func(name string) bool {
```
purpleidea: All? Map? Is this pseudo code??

langpuppet/gapi.go
```diff
@@ -0,0 +1,290 @@
...
+		var currentPuppetGraph *pgraph.Graph
+
+		for {
+			var graph_err error
+
+			select {
+			case nextChild <- nextLang:
+				graph_err = nextChild.Err
+				if nextChild.Err == nil {
+					obj.currentLangGraph, graph_err = obj.langGAPI.Graph()
```
purpleidea: Hmm... I'm not 100% sure this is the right approach. IMO (I'd have to think about it.) Next shouldn't call Graph. Keep them separate unless you think I'm wrong.

langpuppet/gapi.go
```diff
@@ -0,0 +1,290 @@
...
+	return ch
+}
+
+// Close shuts down the Puppet GAPI.
+func (obj *GAPI) Close() error {
+	if !obj.initialized {
+		return fmt.Errorf("%s: GAPI is not initialized", Name)
+	}
+
+	var closerGroup sync.WaitGroup
```
purpleidea: Not sure you need goroutines in close. It can just be linear/sync. Not sure there's much of an added benefit, and it makes error passing harder which you're not currently doing. (Not that the errors in Close are very important.)

langpuppet/merge.go
```diff
@@ -0,0 +1,135 @@
...
+// associated with the wrapping GAPI.
+// Vertices are merged if they adhere to the following rules (for any
+// given value of POSTFIX):
+//  1. The graph from lang contains a noop vertex named puppet_POSTFIX
+//  2. The graph from Puppet contains an empty class mgmt_POSTFIX
+//  3. The resulting graph will contain one noop vertex named POSTFIX
+//     that replaces all nodes mentioned in (1) and (2)
+// All edges connecting to any of the vertices merged this way
+// will be present in the merged graph.
+func (obj *GAPI) mergeChildGraphs() (*pgraph.Graph, error) {
```
purpleidea: I'll leave all this to you. =D Unless you have questions. However... If you don't plan on using any of the GAPI stuff in here, I'd instead implement this as a simple merge disconnected to obj.

Eg:

```
func (g1, g2 *pgraph.Graph) (*pgraph.Graph, error)
```

And in the rare chance that it makes sense, you could always do:
```
func (graphs ...*pgraph.Graph) (*pgraph.Graph, error)
```

;)

ffrank:
Actually, at this point it will work without puppet support because of whatever the bug is ^^'

langpuppet/merge.go
ffrank: I took my cue from the puppet stuff, where the functions in puppet.go are actual methods, although they probably don't need to be. (Didn't look into this too deeply, though, and agree that it's not required here at all.)

I *would* be grateful for feedback regarding my use of the Pgraph API, and if my handling of vertices and edges is sound. Look for variables `result`, `lg`, and `pg`.

langpuppet/merge.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	merge_targets := make(map[string]pgraph.Vertex)
+
+	// first add all vertices from the lang graph
+	for _, vertex := range lg.Vertices() {
+		if strings.Index(vertex.String(), "noop[puppet_") == 0 {
+			resource, ok := vertex.(engine.NamedRes)
```
purpleidea: This looks suspicious... Maybe it's okay though. Is it what you want or really need? Why not just `Res` ?

purpleidea:
> However, the GAPI is not selected/activated. The empty GAPI runs instead.

The reason is because this line:

>  	if s := c.String(Name); c.IsSet(Name) { 

It's looking for your thing to say --puppetlang otherwise it won't do anything.
Instead you want to do something like:

```
if... c.IsSet(prefix + puppet.Name) && c.IsSet(prefix + lang.Name) ...
```

Get it?

Apart from that, my earlier review comments stand, especially about the inling :) I didn't verify the prefix/trim logic, but if it works, then it works!

You should be able to run it with the above tip. Let me know how it goes! Other than that, put the lp prefix in a const somewhere.

Thanks!

langpuppet/merge.go
ffrank: For the vertex merging, I need to inspect the `Name` of all `noop` resources.

ffrank:
Just pushed the first revision that actually runs.

1. Check out this branch and `make build`.
2. Make sure `puppet mgmtgraph` is installed and works.
3. Run `./mgmt run --tmp-prefix --lp-lang examples/lang/template0.mcl --lp-puppet 'notify { "hi": }'`

langpuppet/gapi.go
ffrank: I think this has become a fair bit nicer in the current revision. Closing.

langpuppet/merge.go
ffrank: ...which I now realize would work with `Res` just as well, so yes, will probably do it.

ffrank: (unless you end up making the `named` aspect optional, as a code comment suggests ;)

langpuppet/gapi.go
ffrank: Fixed.

langpuppet/merge.go
ffrank: Made the merge func stand alone.

langpuppet/gapi.go
ffrank: Done.

langpuppet/merge.go
ffrank: Anyway, fixed.

langpuppet/gapi.go
```diff
@@ -0,0 +1,284 @@
...
+
+// Graph returns a current Graph.
+func (obj *GAPI) Graph() (*pgraph.Graph, error) {
+	if !obj.initialized {
+		return nil, fmt.Errorf("%s: GAPI is not initialized", Name)
+	}
+
+	var graph_err error
+
+	if obj.currentLangGraph == nil {
```
purpleidea: I think the logic here should actually be something like this:

In Next(), whenever the corresponding lang or puppet child Next() call sends an event, set a "dirty" flag for that frontend. In the Graph() you run the Graph() function when that flag is true, and then unset it. (You might want to consider a lock.)

And you'll never send a message on Next() until you've gotten an initial message from each Next, and then you send on each subsequent message.

So this is a one line change here, and in Next for each. Let me know if this isn't clear enough.

langpuppet/gapi.go
```diff
@@ -0,0 +1,284 @@
...
+				Err:  fmt.Errorf("%s: GAPI is not initialized", Name),
+				Exit: true, // exit, b/c programming error?
+			}
+			ch <- next
+			return
+		}
+		nextLang := obj.langGAPI.Next()
+		nextPuppet := obj.puppetGAPI.Next()
+
+		langReady := false
```
purpleidea: The logic in Next() looks perfect... Except I would expect these "ready" flags to be used in the Graph method to decide which Graph sub method to call... I'm not sure if you'd need a lock or not. I'd need to think about it. (Safer, but can we deadlock??)

langpuppet/gapi.go
```diff
@@ -0,0 +1,284 @@
...
+	return ch
+}
+
+// Close shuts down the Puppet GAPI.
+func (obj *GAPI) Close() error {
+	if !obj.initialized {
+		return fmt.Errorf("%s: GAPI is not initialized", Name)
+	}
+
+	obj.langGAPI.Close()
```
purpleidea: If something returns an error, and you don't acknowledge it or pass it on, you fail. :(

langpuppet/merge.go
```diff
@@ -0,0 +1,167 @@
...
+// Vertices are merged if they adhere to the following rules (for any
+// given value of POSTFIX):
+//  1. The graph from lang contains a noop vertex named puppet_POSTFIX
+//  2. The graph from Puppet contains an empty class mgmt_POSTFIX
+//  3. The resulting graph will contain one noop vertex named POSTFIX
+//     that replaces all nodes mentioned in (1) and (2)
+// All edges connecting to any of the vertices merged this way
+// will be present in the merged graph.
+func mergeGraphs(graphFromLang, graphFromPuppet *pgraph.Graph) (*pgraph.Graph, error) {
+	lg := graphFromLang
```
purpleidea: why the unnecessary copy?

lib/cli.go
```diff
@@ -81,6 +81,7 @@ func run(c *cli.Context) error {
 
 	for _, name := range names {
 		fn := gapi.RegisteredGAPIs[name]
+
```
purpleidea: what's this, and the one below for?

langpuppet/merge.go
ffrank: It was easier to type the code out =D

Undone.

lib/cli.go
ffrank: Oops sorry, didn't properly clean up after having some debug output here.

Fixed.

langpuppet/gapi.go
ffrank: Fixed

langpuppet/gapi.go
ffrank: Raciness is an interesting aspect, but I don't see the benefit of carrying another flag, versus "pointer is now nil" bearing the same semantics as a "dirty" flag.

purpleidea: It's *very* different, because *after* the _first_ run of both, then both of those "flags" are not nil (they contain graphs) and they NEVER go nil again. So a "dirty" operation never actually does anything. Get it?

As a result for langGAPI to return on Next() and you call graph on both, then you've just panic or blocked the system because Puppet wasn't ready for you to call Graph. Get it?

langpuppet/gapi.go
```diff
@@ -0,0 +1,290 @@
...
+		for {
+			select {
+			case nextChild := <-nextLang:
+				if nextChild.Err == nil {
+					obj.currentLangGraph = nil
+					langReady = true
+				}
+			case nextChild := <-nextPuppet:
+				if nextChild.Err == nil {
+					obj.currentPuppetGraph = nil
```
purpleidea: And by erasing this everytime, you can't merge the first graph with the second graph, if one changes often and the other changes rarely...

langpuppet/gapi.go
ffrank: Done.

langpuppet/gapi.go
ffrank: Doing it with the ready flag now.

langpuppet/gapi.go
```diff
@@ -274,8 +276,12 @@ func (obj *GAPI) Close() error {
 		return fmt.Errorf("%s: GAPI is not initialized", Name)
 	}
 
-	obj.langGAPI.Close()
-	obj.puppetGAPI.Close()
+	if err := obj.langGAPI.Close(); err != nil {
+		return errors.Wrap(err, "closing lang GAPI failed")
```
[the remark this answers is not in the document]
ffrank: Done.

langpuppet/gapi.go
```diff
@@ -274,8 +276,12 @@ func (obj *GAPI) Close() error {
...
+	}
+	if err := obj.puppetGAPI.Close(); err != nil {
```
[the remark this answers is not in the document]
ffrank: Changed it to make sure to call both, but if they both err out, we can still only return one of those errors.

Also now makes sure to close the closeChan regardless.

langpuppet/gapi.go
ffrank: I don't follow, but we did resolve this in conversation, so closing this thread.

ffrank:
Oddly, I cannot comment on or resolve your comment towards the bottom. It's not an issue anymore though.

There is one other unresolved item, because I'm not sure what to do in case of two errors.

Please take another look at Next() because I'm not sure about the error handling there, either.

langpuppet/gapi.go
```diff
@@ -0,0 +1,323 @@
...
+	g, err := mergeGraphs(obj.currentLangGraph, obj.currentPuppetGraph)
+
+	// FIXME: Quite noisy, we can drop this soon, or hide it behind a debug flag
+	obj.data.Logf("current graph from lang:")
+	obj.currentLangGraph.Logf(obj.data.Logf)
+	obj.data.Logf("current graph from puppet:")
+	obj.currentPuppetGraph.Logf(obj.data.Logf)
+	if err == nil {
+		obj.data.Logf("merge result:")
+		g.Logf(obj.data.Logf)
```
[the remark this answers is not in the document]
ffrank: Done.

langpuppet/gapi.go
```diff
@@ -0,0 +1,323 @@
...
+		if graph_err != nil {
+			return nil, graph_err
+		}
+	} else {
+		obj.graphFlagMutex.Unlock()
+	}
+
+	g, err := mergeGraphs(obj.currentLangGraph, obj.currentPuppetGraph)
+
+	// FIXME: Quite noisy, we can drop this soon, or hide it behind a debug flag
```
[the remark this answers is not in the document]
ffrank: Cool, thanks. Fixed.

langpuppet/gapi.go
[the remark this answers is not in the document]
ffrank: Changed this to use multierror.

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+
+package langpuppet
+
+import (
+	"flag"
+	"fmt"
+	"strings"
+	"sync"
+
+	multierr "github.com/hashicorp/go-multierror"
```
purpleidea: imports in wrong format (as discussed in IRC)

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/purpleidea/mgmt/lang"
+	"github.com/purpleidea/mgmt/puppet"
+
+	"github.com/urfave/cli"
+)
+
+const (
+	// Name is the name of this frontend.
+	Name       = "langpuppet"
+	FlagPrefix = "lp-"
```
purpleidea: This public var also needs a short comment for lint, or it needs to be private. Up to you.

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+)
+
+func init() {
+	gapi.Register(Name, func() gapi.GAPI { return &GAPI{} }) // register
+}
+
+// GAPI implements the main langpuppet GAPI interface.
```
purpleidea: If you could extend this comment with a short description explaining the magic, that would be great.

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+type GAPI struct {
+	InputURI string
```
purpleidea: Struct comment please.

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+
+// CliFlags returns a list of flags used by this deploy subcommand.
+// It consists of all flags accepted by lang and puppet mode,
+// with a respective "lp-" prefix.
+func (obj *GAPI) CliFlags() []cli.Flag {
+	langFlags := (&lang.GAPI{}).CliFlags()
+	puppetFlags := (&puppet.GAPI{}).CliFlags()
+
+	var childFlags []cli.Flag
+
```
purpleidea: nit: you can nuke this unnecessary nl (i try to add them for readability or logical breaks)

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+	return childFlags
+}
+
+// Init initializes the langpuppet GAPI struct.
+func (obj *GAPI) Init(data gapi.Data) error {
+	if obj.initialized {
+		return fmt.Errorf("already initialized")
+	}
+	obj.data = data // store for later
+
```
purpleidea: nit: nuke nl

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+		World:         obj.data.World,
+		Noop:          obj.data.Noop,
+		NoConfigWatch: obj.data.NoConfigWatch,
+		NoStreamWatch: obj.data.NoStreamWatch,
+		Debug:         obj.data.Debug,
+		Logf: func(format string, v ...interface{}) {
+			obj.data.Logf(lang.Name+": "+format, v...)
+		},
+	}
+
```
purpleidea: nit: nuke nl

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+	return nil
+}
+
+// Graph returns a current Graph.
+func (obj *GAPI) Graph() (*pgraph.Graph, error) {
+	if !obj.initialized {
+		return nil, fmt.Errorf("%s: GAPI is not initialized", Name)
+	}
+
+	var graphErr error
```
purpleidea: just use the standard name `err` since there's only one in this scope.

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+		nextPuppet := obj.puppetGAPI.Next()
+
+		firstLang := false
+		firstPuppet := false
+
+		for {
+			select {
+			case nextChild := <-nextLang:
+				if nextChild.Err != nil {
+					ch <- nextChild
```
purpleidea: deadlock! (same below)

because you can't unblock from here if ch doesn't want to read!

instead, save the err and send it below in one spot at the bottom.
You'll probably also have to do a `(!firstLang || !firstPuppet) && err == nil` below instead of the `!firstLang || !firstPuppet` to pass through.

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+			}
+
+			if !firstLang || !firstPuppet {
+				continue
+			}
+
+			obj.data.Logf("generating new composite graph...")
+			next := gapi.Next{
+				Exit: false,
+				Err:  nil,
```
purpleidea: err goes here :)
you might want to do this too:

https://github.com/purpleidea/mgmt/blob/28f343ac502971754a5be586e22f7c2656a17748/lang/gapi.go#L268

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+// Close shuts down the Puppet GAPI.
+func (obj *GAPI) Close() error {
+	if !obj.initialized {
+		return fmt.Errorf("%s: GAPI is not initialized", Name)
+	}
+
+	var err error
+
```
purpleidea: nit: kill nl

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+	if err1 := obj.langGAPI.Close(); err1 != nil {
```
purpleidea: use var `e` instead of err1 please. err1 implies an err2.

langpuppet/gapi.go
```diff
@@ -0,0 +1,327 @@
...
+		err = multierr.Append(err, errwrap.Wrapf(err1, "closing lang GAPI failed"))
+	}
+	if err1 := obj.puppetGAPI.Close(); err1 != nil {
+		err = multierr.Append(err, errwrap.Wrapf(err1, "closing Puppet GAPI failed"))
+	}
+
+	close(obj.closeChan)
+
```
purpleidea: nit: kill nl

langpuppet/merge.go
```diff
@@ -0,0 +1,165 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/pgraph"
+)
+
+// This method returns the merged graph containing all vertices
```
purpleidea: typically s/This method/mergeGraphs/

langpuppet/merge.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// first add all vertices from the lang graph
+	for _, vertex := range graphFromLang.Vertices() {
+		if strings.Index(vertex.String(), "noop[puppet_") == 0 {
+			resource, ok := vertex.(engine.Res)
+			if !ok {
+				return nil, fmt.Errorf("vertex %s is not a named resource", vertex.String())
+			}
+			basename := strings.TrimPrefix(resource.Name(), "puppet_")
+			// FIXME: will this change the original vertex?
```
purpleidea: YES, AIUI.

langpuppet/merge.go
```diff
@@ -0,0 +1,165 @@
...
+		// strip either prefix (plus the closing bracket)
+		basename := strings.TrimSuffix(
+			strings.TrimPrefix(
+				strings.TrimPrefix(resource.Name(),
+					"admissible_Class[Mgmt_"),
+				"completed_Class[Mgmt_"),
+			"]")
+
+		if _, found := mergeTargets[basename]; !found {
+			// FIXME: should be a warning not an error?
```
purpleidea: That question can always be answered by a flag if you want. But I think error. Really it's up to you.

langpuppet/merge.go
```diff
@@ -0,0 +1,165 @@
...
+			return nil, fmt.Errorf("Puppet graph has unmatched class mgmt_%s", basename)
```
purpleidea: errors are not sentences, so they don't start with a capital
i think the linter catches this. i forget.

langpuppet/merge.go
```diff
@@ -0,0 +1,165 @@
...
+
+		// does this edge not lead to the class end?
+		next := graphFromPuppet.OutgoingGraphVertices(vertex)[0]
+		if next.String() != "noop[completed_Class[Mgmt_"+basename+"]]" {
+			return nil, fmt.Errorf("class mgmt_%s is not empty, start is followed by %s", basename, next.String())
+		}
+	}
+
+	result.AddVertex(anchor)
+
```
purpleidea: so many extra nl's

langpuppet/merge.go
```diff
@@ -0,0 +1,165 @@
...
+	// subject to a merge, so it will be easier do generate the new edges
+	// in the final pass
+	for _, vertex := range graphFromPuppet.Vertices() {
+		if vertex.String() == "noop[admissible_Stage[main]]" {
+			// we can start a depth first search here
+			anchor = vertex
+			continue
+		}
+		// at this stage we don't distinguis between class start and end
+		if strings.Index(vertex.String(), "noop[admissible_Class[Mgmt_") != 0 &&
```
purpleidea: Can you have the `Mgmt_` prefix stuff be a const please? Same for puppet.

langpuppet/merge.go
```diff
@@ -0,0 +1,169 @@
...
+			// if we reach here, this neighbor is a regular vertex
+			result.AddVertex(neighbor)
+			result.AddEdge(source, neighbor, graphFromPuppet.FindEdge(vertex, neighbor))
+		}
+	}
+
+	for _, vertex := range mergeTargets {
+		if !merged[vertex] {
+			// FIXME: should be a warning not an error?
+			return nil, fmt.Errorf("Lang graph has unmatched %s", vertex.String())
```
purpleidea: same as other review comment... errors aren't capitalized.

langpuppet/gapi.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// Graph returns a current Graph.
+func (obj *GAPI) Graph() (*pgraph.Graph, error) {
+	if !obj.initialized {
+		return nil, fmt.Errorf("%s: GAPI is not initialized", Name)
+	}
+
+	var err error
+
```
purpleidea: nit: kill nl

ffrank:
All fixed I believe.

purpleidea:
\o/

purpleidea:
Waiting for travis!

PR #416, opened 2018-12-11
etcd: Use mgmt backend for fs tests

jonathangold:


etcd/fs/fs_test.go
```diff
@@ -39,6 +43,31 @@ const (
 	superblock = "/some/superblock" // TODO: generate randomly per test?
 )
 
+func TestMain(m *testing.M) {
```
purpleidea: Sweet!

Can you add a small comment about what TestMain is, and also why we're calling ourself of course? Thanks!!

etcd/fs/fs_test.go
```diff
@@ -39,6 +43,31 @@ const (
...
+	// cleanup when we're done
+	var exit int
+	defer func() {
+		cmd.Process.Signal(syscall.SIGQUIT)
+		cmd.Process.Kill()
+		os.Exit(exit)
+	}()
+
+	// run the tests
+	exit = m.Run()
```
purpleidea: AFAIK setting exit this way (to be used in defer) is a bit of a dirty hack ;)

jonathangold: There was more code in there before :) It's not neccessary as long as we ignore the errors.

purpleidea: Oh, btw, can this patch also fix the FsTests that don't t.Errorf properly =D

jonathangold:
@purpleidea done

etcd/fs/fs_test.go
```diff
@@ -55,25 +91,15 @@ func TestFs1(t *testing.T) {
 		Metadata:   superblock,
 		DataPrefix: etcdfs.DefaultDataPrefix,
 	}
-	//var etcdFs afero.Fs = NewEtcdFs()
```
purpleidea: Why'd you nuke this and below?

etcd/fs/fs_test.go
```diff
@@ -39,13 +43,45 @@ const (
...
+		fmt.Printf("error starting command %v: %s", cmd, err)
+		os.Exit(1)
+	}
+
+	// run the tests
+	exit := m.Run()
+
+	// cleanup when we're done
+	if err := cmd.Process.Signal(syscall.SIGQUIT); err != nil {
+		fmt.Printf("error sending quit signal: %s", err)
```
purpleidea: printf needs a \n and err is usually %v IIRC
Too bad there's no proper log method for `m`

etcd/fs/fs_test.go
```diff
@@ -39,13 +43,45 @@ const (
...
+		exit += 1
+	}
+	if err := cmd.Process.Kill(); err != nil {
+		fmt.Printf("error killing process: %s", err)
+		exit += 1
```
purpleidea: heh, okay I guess :/

etcd/fs/fs_test.go
jonathangold: I assumed that var was stale, I can put it back, but `mkdir /` doesn't make much sense and will always error. What's the point?

etcd/fs/fs_test.go
jonathangold: hm, maybe not. I'll come back to this later.

jonathangold:
> Does TestMain run once for each test or are the tests not clean?

I guess it's more correct to have a setup() function that returns a cleanup() to call from each test in that case.

jonathangold:
TestFs2 makes no sense, all it does is run `util.FsTree("/")` and `util.FsTree("/tmp")` but it never makes /tmp. It'll fail if etcd is fresh for each test. Can I kill it?

jonathangold:
@purpleidea I went ahead and ditched TestFs2 (see above,) put `//var etcdFs afero.Fs = NewEtcdFs()` back, and replaced TestMain with a function that get's called in each test.

etcd/fs/fs_test.go
```diff
@@ -57,23 +92,14 @@ func TestFs1(t *testing.T) {
 	}
 	//var etcdFs afero.Fs = NewEtcdFs()
 
-	if err := etcdFs.Mkdir("/", umask); err != nil {
```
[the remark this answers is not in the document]
jonathangold: I nuked:

```
if err := etcdFs.Mkdir("/", umask); err != nil {
	t.Logf("error: %+v", err)
	if err != etcdfs.ErrExist {
		return
	}
}
```
because it makes no sense to `mkdir /` and always results in a File Exists error.

I got rid of FsTest2 because all it did was check if `/` and `/tmp` existed, but it never made `/tmp` during the test, so would always fail with a clean etcd node.

purpleidea:
On Tue, Dec 11, 2018 at 3:56 PM Jonathan Gold <notifications@github.com> wrote:
> because it makes no sense to mkdir / and always results in a File Exists error.
>
> I got rid of FsTest2 because all it did was check if / and /tmp existed, but it never made /tmp during the test, so would always fail with a clean etcd node.

Yeah, the fact that it always passed (and seemed redundant) was sort
of the idea behind the test. We're ensuring that a new FS gets the "/"
inode by default!! Put it back unless something is broken please.

jonathangold:
@purpleidea 
>  Put it back

done :)

etcd/fs/fs_test.go
```diff
@@ -245,9 +259,8 @@ func TestFs4(t *testing.T) {
 		DataPrefix: etcdfs.DefaultDataPrefix,
 	}
 
-	etcdFs.Mkdir("/tmp", umask)
```
purpleidea: why did you -1 on depth here?

etcd/fs/fs_test.go
```diff
@@ -82,112 +117,85 @@ func TestFs1(t *testing.T) {
...
-	}
-
-	tree, err := util.FsTree(etcdFs, "/")
+	dir, err = etcdFs.Open("/")
 	if err != nil {
-		t.Errorf("tree error: %+v", err)
+		t.Errorf("error: %+v", err)
 		return
 	}
-	t.Logf("tree: \n%s", tree)
```
purpleidea: why did you nuke the tree?

etcd/fs/fs_test.go
```diff
@@ -82,112 +117,85 @@ func TestFs1(t *testing.T) {
...
+		t.Errorf("rename error: %+v", err)
 		return
 	}
 
-	//f2, err := etcdFs.Create("/tmp/bar")
-	//if err != nil {
-	//	t.Logf("error: %+v", err)
-	//	return
-	//}
+	f2, err := etcdFs.Create("/tmp/bar")
```
purpleidea: awesome :)

etcd/fs/fs_test.go
```diff
@@ -82,112 +117,85 @@ func TestFs1(t *testing.T) {
...
 	if err != nil {
-		t.Logf("error: %+v", err)
+		t.Errorf("error: %+v", err)
 		return
 	}
 	t.Logf("wrote: %d", i)
 
 	if err := etcdFs.Mkdir("/tmp/d1", umask); err != nil {
-		t.Logf("error: %+v", err)
-		if err != etcdfs.ErrExist {
```
purpleidea: Isn't this wrong!? Don't we want to check it's NOT that error before we fail??

etcd/fs/fs_test.go
jonathangold: There's no need for `/tmp`, but it's not a bad idea to test it with a deeper copy.

jonathangold:
@purpleidea I'm happy with it now. Removing TestFs2 messed up the diff a little (see https://github.com/purpleidea/mgmt/pull/416#discussion_r240797786.) I also fixed the error messages, so they're consistent in style and enumeration.

PR #417, opened 2018-12-12
Gitignore vim

ffrank:


ffrank:
Added a comment. The leading dot is because it should only match dotfiles.

This particular expression is quite ubiquitous in gitignore files, so I would not break it up into two lines. But then I don't really care, either. Up to you.

purpleidea:
Merged!

PR #418, opened 2018-12-13
[WIP] engine: resources: timer: Add a systemd-timer resource

jonathangold:
For early feedback. Still have to ~~add autoedges to svc and~~ validate the time.

I didn't initially realize there was already a timer resource :P so I can rename this to something else, but is https://github.com/purpleidea/mgmt/blob/master/engine/resources/timer.go still useful given language functions?

Either way, I think it makes sense to call this resource 'timer' or 'systemd:timer' rather than cron.

Edit: also tests

### DIFF
--- a/engine/resources/cron.go
+++ b/engine/resources/cron.go
@@ -21,7 +21,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"os"
+	"os/user"
+	"path"
 	"strings"
 	"time"
 
@@ -141,15 +142,13 @@ func (obj *CronRes) Default() engine.Res {
 // validate and initialize the nested file resource and to apply the file state
 // in CheckApply.
 func (obj *CronRes) makeComposite() (*FileRes, error) {
-	// root timer
-	path := fmt.Sprintf("/etc/systemd/system/%s.timer", obj.Name())
-	if obj.Session {
-		// user timer
-		path = fmt.Sprintf("%s/.config/systemd/user/%s.timer", os.Getenv("HOME"), obj.Name())
-	}
-	res, err := engine.NewNamedResource("file", path)
+	p, err := obj.UnitFilePath()
 	if err != nil {
-		return nil, err
+		return nil, errwrap.Wrapf(err, "error generating unit file path")
+	}
+	res, err := engine.NewNamedResource("file", p)
+	if err != nil {
+		return nil, errwrap.Wrapf(err, "error creating nested file resource")
 	}
 	file, ok := res.(*FileRes)
 	if !ok {
@@ -260,14 +259,12 @@ func (obj *CronRes) Watch() error {
 	bus.Signal(dbusChan)
 	defer bus.RemoveSignal(dbusChan) // not needed here, but nice for symmetry
 
-	// root timer
-	path := fmt.Sprintf("/etc/systemd/system/%s.timer", obj.Name())
-	if obj.Session {
-		// user timer
-		path = fmt.Sprintf("%s/.config/systemd/user/%s.timer", os.Getenv("HOME"), obj.Name())
+	p, err := obj.UnitFilePath()
+	if err != nil {
+		return errwrap.Wrapf(err, "error generating unit file path")
 	}
 	// recwatcher for the systemd-timer unit file
-	obj.recWatcher, err = recwatch.NewRecWatcher(path, false)
+	obj.recWatcher, err = recwatch.NewRecWatcher(p, false)
 	if err != nil {
 		return err
 	}
@@ -453,7 +450,8 @@ type CronUID struct {
 	// used in the IFF function, is what you see in the struct fields here.
 	engine.BaseUID
 
-	name string // the machine name
+	unit    string // name of target unit
+	session bool   // user session
 }
 
 // IFF aka if and only if they are equivalent, return true. If not, false.
@@ -462,17 +460,38 @@ func (obj *CronUID) IFF(uid engine.ResUID) bool {
 	if !ok {
 		return false
 	}
-	return obj.name == res.name
+	if obj.unit != res.unit {
+		return false
+	}
+	if obj.session != res.session {
+		return false
+	}
+	return true
+}
+
+// AutoEdges returns the AutoEdge interface.
+func (obj *CronRes) AutoEdges() (engine.AutoEdge, error) {
+	return nil, nil
 }
 
 // UIDs includes all params to make a unique identification of this object.
 // Most resources only return one although some resources can return multiple.
 func (obj *CronRes) UIDs() []engine.ResUID {
-	x := &CronUID{
-		BaseUID: engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
-		name:    obj.Name(), // svc name
+	unit := fmt.Sprintf("%s.service", obj.Name())
+	if obj.Unit != "" {
+		unit = obj.Unit
 	}
-	return append([]engine.ResUID{x}, obj.file.UIDs()...)
+	uids := []engine.ResUID{
+		&CronUID{
+			BaseUID: engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
+			unit:    unit,        // name of target unit
+			session: obj.Session, // user session
+		},
+	}
+	if file, err := obj.makeComposite(); err == nil {
+		uids = append(uids, file.UIDs()...) // add the file uid if we can
+	}
+	return uids
 }
 
 // UnmarshalYAML is the custom unmarshal handler for this struct.
@@ -493,6 +512,23 @@ func (obj *CronRes) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 	*obj = CronRes(raw) // restore from indirection with type conversion!
 	return nil
+}
+
+// UnitFilePath returns the path to the systemd-timer unit file.
+func (obj *CronRes) UnitFilePath() (string, error) {
+	// root timer
+	if !obj.Session {
+		return fmt.Sprintf("/etc/systemd/system/%s.timer", obj.Name()), nil
+	}
+	// user timer
+	u, err := user.Current()
+	if err != nil {
+		return "", errwrap.Wrapf(err, "error getting current user")
+	}
+	if u.HomeDir == "" {
+		return "", fmt.Errorf("user has no home directory")
+	}
+	return path.Join(u.HomeDir, "/.config/systemd/user/", fmt.Sprintf("%s.timer", obj.Name())), nil
 }
 
 // unitFileContents returns the contents of the unit file representing the
--- a/engine/resources/svc.go
+++ b/engine/resources/svc.go
@@ -21,9 +21,12 @@ package resources
 
 import (
 	"fmt"
+	"os/user"
+	"path"
 
 	"github.com/purpleidea/mgmt/engine"
 	"github.com/purpleidea/mgmt/engine/traits"
+	engineUtil "github.com/purpleidea/mgmt/engine/util"
 	"github.com/purpleidea/mgmt/util"
 
 	systemd "github.com/coreos/go-systemd/dbus" // change namespace
@@ -380,7 +383,8 @@ type SvcUID struct {
 	// information about the resource we're matching. That data which is
 	// used in the IFF function, is what you see in the struct fields here.
 	engine.BaseUID
-	name string // the svc name
+	name    string // the svc name
+	session bool   // user session
 }
 
 // IFF aka if and only if they are equivalent, return true. If not, false.
@@ -389,7 +393,13 @@ func (obj *SvcUID) IFF(uid engine.ResUID) bool {
 	if !ok {
 		return false
 	}
-	return obj.name == res.name
+	if obj.name != res.name {
+		return false
+	}
+	if obj.session != res.session {
+		return false
+	}
+	return true
 }
 
 // SvcResAutoEdges holds the state of the auto edge generator.
@@ -431,12 +441,55 @@ func (obj *SvcResAutoEdges) Test(input []bool) bool {
 	return true // keep going
 }
 
-// AutoEdges returns the AutoEdge interface. In this case the systemd units.
+// SvcResAutoEdgesCron holds the state of the svc -> cron auto edge generator.
+type SvcResAutoEdgesCron struct {
+	unit    string // target unit
+	session bool   // user session
+}
+
+// Next returns the next automatic edge.
+func (obj *SvcResAutoEdgesCron) Next() []engine.ResUID {
+	// XXX: should this be true if SvcRes State == "stopped"?
+	reversed := false
+	value := &CronUID{
+		BaseUID: engine.BaseUID{
+			Kind:     "CronRes",
+			Reversed: &reversed,
+		},
+		unit:    obj.unit,    // target unit
+		session: obj.session, // user session
+	}
+	return []engine.ResUID{value} // we return one, even though api supports N
+}
+
+// Test takes the output of the last call to Next() and outputs true if we
+// should continue.
+func (obj *SvcResAutoEdgesCron) Test([]bool) bool {
+	return false // only get one svc -> cron edge
+}
+
+// AutoEdges returns the AutoEdge interface. In this case, systemd unit file
+// resources and cron (systemd-timer) resources.
 func (obj *SvcRes) AutoEdges() (engine.AutoEdge, error) {
 	var data []engine.ResUID
-	svcFiles := []string{
+	var svcFiles []string
+	svcFiles = []string{
+		// root svc
 		fmt.Sprintf("/etc/systemd/system/%s.service", obj.Name()),     // takes precedence
 		fmt.Sprintf("/usr/lib/systemd/system/%s.service", obj.Name()), // pkg default
+	}
+	if obj.Session {
+		// user svc
+		u, err := user.Current()
+		if err != nil {
+			return nil, errwrap.Wrapf(err, "error getting current user")
+		}
+		if u.HomeDir == "" {
+			return nil, fmt.Errorf("user has no home directory")
+		}
+		svcFiles = []string{
+			path.Join(u.HomeDir, "/.config/systemd/user/", fmt.Sprintf("%s.service", obj.Name())),
+		}
 	}
 	for _, x := range svcFiles {
 		var reversed = true
@@ -449,11 +502,18 @@ func (obj *SvcRes) AutoEdges() (engine.AutoEdge, error) {
 			path: x, // what matters
 		})
 	}
-	return &FileResAutoEdges{
+
+	fileEdge := &FileResAutoEdges{
 		data:    data,
 		pointer: 0,
 		found:   false,
-	}, nil
+	}
+	cronEdge := &SvcResAutoEdgesCron{
+		session: obj.Session,
+		unit:    fmt.Sprintf("%s.service", obj.Name()),
+	}
+
+	return engineUtil.AutoEdgeCombiner(fileEdge, cronEdge)
 }
 
 // UIDs includes all params to make a unique identification of this object.
@@ -461,7 +521,8 @@ func (obj *SvcRes) AutoEdges() (engine.AutoEdge, error) {
 func (obj *SvcRes) UIDs() []engine.ResUID {
 	x := &SvcUID{
 		BaseUID: engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
-		name:    obj.Name(), // svc name
+		name:    obj.Name(),  // svc name
+		session: obj.Session, // user session
 	}
 	return []engine.ResUID{x}
 }
--- a/examples/lang/cron4.mcl
+++ b/examples/lang/cron4.mcl
@@ -1,0 +1,17 @@
+$home = getenv("HOME")
+
+cron "purpleidea-oneshot" {
+    state => "absent",
+    session => true,
+    trigger => "OnCalendar",
+    time => "*:*:0",
+}
+
+svc "purpleidea-oneshot" {
+    state => "stopped",
+    session => true,
+}
+
+file printf("%s/.config/systemd/user/purpleidea-oneshot.service", $home) {
+    state => "absent",
+}


purpleidea:
> I didn't initially realize there was already a timer resource :P 

Surprise!

> so I can rename this to something else, but is https://github.com/purpleidea/mgmt/blob/master/engine/resources/timer.go still useful given language functions?

That's a good question, and I think actually yes. But I didn't think hard about it. But I'd like to keep the existing `timer` Res in for now anyways.

> Either way, I think it makes sense to call this resource 'timer' or 'systemd:timer' rather than cron.

I thought about this a lot, and it was discussed somewhere, but I think actually naming this cron does make sense. That's historically how it's always been named, and systemd:timer is effectively the `cron` replacement, and theoretically we could have our `cron` resource support actual cron for certain cases.

So... Care to rename to cron and submit a clean patch for review? =D

jonathangold:
@purpleidea 

> So... Care to rename to cron and submit a clean patch for review? =D

done

engine/resources/cron.go
```diff
@@ -0,0 +1,491 @@
...
+				return err // exit if requested
+			}
+		}
+	}
+}
+
+// CheckApply is run to check the state and, if apply is true, to apply the
+// necessary changes to reach the desired state. This is run before Watch and
+// again if Watch finds a change occurring to the state.
+func (obj *CronRes) CheckApply(apply bool) (checkOK bool, err error) {
```
purpleidea: if we can avoid named return variables, and it doesn't make the code worse, I'd do it, even if we leave these names in, but just don't use them as such in the code. Just a preference, it's not required.

engine/resources/cron.go
```diff
@@ -0,0 +1,491 @@
...
+
+// makeComposite creates a pointer to a FileRes. The pointer is used to
+// validate and initialize the nested file resource and to apply the file state
+// in CheckApply.
+func (obj *CronRes) makeComposite() (*FileRes, error) {
+	res, err := engine.NewNamedResource("file", fmt.Sprintf("/etc/systemd/system/%s.timer", obj.Name()))
+	if err != nil {
+		return nil, err
+	}
+	file := res.(*FileRes)
```
purpleidea: unsafe technically (test!)

engine/resources/cron.go
```diff
@@ -0,0 +1,491 @@
...
+		u = append(u, &unit.UnitOption{Section: "Timer", Name: "RemainAfterElapse", Value: "false"})
+	}
+	// [Install]
+	u = append(u, &unit.UnitOption{Section: "Install", Name: "WantedBy", Value: "timers.target"})
+
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(unit.Serialize(u))
+	s := buf.String()
+
+	return &s
```
purpleidea: no way!

string or []byte ... if you want it so you can do you `obj.Content = ...` the real solution is:

```
s := unitFileContents()
obj.Content = &s
```

engine/util/util.go
```diff
@@ -259,3 +275,59 @@ func GetGID(group string) (int, error) {
...
+// units. It waits for the requested job to be complete before it returns.
+func unitStateAction(ctx context.Context, conn *dbus.Conn, unit, action string) error {
+	// Add a dbus rule to watch the systemd1 JobRemoved signal, used to wait
+	// until the job completes.
+	args := []string{
+		"type='signal'",
+		fmt.Sprintf("path='%s'", DBusSystemd1Path),
+		fmt.Sprintf("interface='%s'", DBusSystemd1ManagerIface),
+		fmt.Sprintf("member='%s'", DBusSignalJobRemoved),
+		fmt.Sprintf("arg2='%s'", unit),
```
purpleidea: arg2? really?

jonathangold: Yeah, haha. It goes up to arg63!

engine/resources/cron.go
```diff
@@ -0,0 +1,491 @@
...
+	traits.Base
+	traits.Edgeable
+	traits.Recvable
+
+	init *engine.Init
+
+	// State must be 'exists' or 'absent'.
+	State string `yaml:"state"`
+
+	// Trigger is the type of timer. Valid types are 'OnCalendar',
```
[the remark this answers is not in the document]
jonathangold: Let's discuss when the rest is closer to merge.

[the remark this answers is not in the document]
purpleidea: It is already.

jonathangold:
> Is there any async operation that doesn't block where it should? (Didn't we have something like this for nspawn or something?)

The unitStateAction function waits for the job to be complete before returning. I came up with that solution in mount. Not sure what else would need to block.

engine/resources/cron.go
```diff
@@ -0,0 +1,491 @@
...
+	// 'man systemd-time'.
+	OnUnitInactiveSec = "OnUnitInactiveSec"
+)
+
+func init() {
+	engine.RegisterResource("cron", func() engine.Res { return &CronRes{} })
+}
+
+// CronRes is a systemd-timer cron resource.
+type CronRes struct {
```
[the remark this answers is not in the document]
jonathangold: I'll do this in the next revisiion

engine/resources/cron.go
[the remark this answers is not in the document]
jonathangold: @purpleidea I think that unless we can *greatly* reduce the complexity, then mirroring the systemd-timer interface reduces cognitive overhead compared to learning a new API. Each trigger is unique, so I'm not sure how we could map them all to one or two options, and the formats are consistent and well defined.

engine/resources/cron.go
```diff
@@ -0,0 +1,493 @@
...
+	return false, engineUtil.RestartUnit(ctx, godbusConn, fmt.Sprintf("%s.timer", obj.Name()))
+}
+
+// Cmp compares two resources and returns an error if they are not equivalent.
+func (obj *CronRes) Cmp(r engine.Res) error {
+	res, ok := r.(*CronRes)
+	if !ok {
+		return fmt.Errorf("not a timer res")
+	}
+
```
purpleidea: Should we also have obj.file.Cmp here too?

engine/resources/cron.go
```diff
@@ -208,13 +220,22 @@ func (obj *CronRes) Close() error {
...
 
 	// create a private message bus
-	bus, err := util.SystemBusPrivateUsable()
+	if obj.Session {
+		bus, err = util.SystemBusPrivateUsable()
+	} else {
+		bus, err = dbus.SessionBusPrivate()
+	}
+
+	fmt.Printf("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!1111111111111111")
```
purpleidea: !!!! Please remember to look at your diffs before submitting =D
My personal trick is that whenever I add debugging stuff that I don't want to commit I use XXX and then when I look at the diffs, I search for XXX and go through all of those. You could do that or use !!! or whatever you like!

jonathangold:
@purpleidea The only thing left to do is validate time, but I don't want to commit the cycles unless we're settled on the api. Discuss in IRC?

purpleidea:
I'm mostly not sure about the autoedges... What's that about? Thanks for finding the bug! Awesome.

And if you're in the mood we should really think about an efficient pure golang way to add tests.

purpleidea:
I merged the "Usable" patch.

jonathangold:
@purpleidea I think I've sorted it out.

Here's how it works now (for example with two edges in `obj.ae`):
- `Next()` returns the first edge then increments `obj.ptr`
- `Test()` returns `true` because `(obj.ptr < len(obj.ae))` is true (1 < 2)
- `Next()` returns the second edge and increments `obj.ptr`
- `Test()` returns `false` because `!(obj.ptr < len(obj.ae)` is false (2 == 2)

Still may need some bounds checking for extra safety, but I think it's right this time.

engine/util/util.go
```diff
@@ -57,7 +57,7 @@ const (
 	DBusRestartUnit = DBusSystemd1ManagerIface + ".RestartUnit"
 	// DBusStopUnit is the dbus method for stopping systemd units.
 	DBusStopUnit = DBusSystemd1ManagerIface + ".StopUnit"
-	// DBbusSignalJobRemoved is the name of the dbus signal that produces a
+	// DBusSignalJobRemoved is the name of the dbus signal that produces a
```
purpleidea: This stuff is in the wrong commit

engine/util/util.go
```diff
@@ -331,3 +331,40 @@ func unitStateAction(ctx context.Context, conn *dbus.Conn, unit, action string)
...
+	if len(obj.ae) <= obj.ptr {
+		panic("shouldn't be called anymore!")
+	}
+	return obj.ae[obj.ptr].Next() // return the next edge
+}
+
+// Test takes the output of the last call to Next() and outputs true if we
+// should continue.
+func (obj *edgeCombiner) Test(input []bool) bool {
+	if len(input) == 0 {
```
purpleidea: Is this needed? If anything, you'd *maybe* want something as well in Next() if you had this in case someone ran an empty struct.

jonathangold:
@purpleidea I think I got it this time! If it's still terrible (i.e. needs more than a little fix here and there,) I'll let you take over, but I think I finally wrapped my head around the api. :D

engine/util/util.go
```diff
@@ -331,3 +331,40 @@ func unitStateAction(ctx context.Context, conn *dbus.Conn, unit, action string)
...
+		return false // no more edges
+	}
+	if !obj.ae[obj.ptr].Test(input) {
```
purpleidea: I think this logic is right! But the code could be rewritten to be inlined and nicer please :)

engine/util/util.go
```diff
@@ -331,3 +331,40 @@ func unitStateAction(ctx context.Context, conn *dbus.Conn, unit, action string)
...
+		if len(obj.ae) <= obj.ptr {
+			return false // no more edges
+		}
+	}
+	return true // continue
+}
+
+// EdgeCombiner takes any number of AutoEdges and combines them to be processed
+// together.
+func EdgeCombiner(ae ...engine.AutoEdge) (engine.AutoEdge, error) {
```
purpleidea: Can we name this and the struct AutoEdgeCombiner and autoEdgeCombiner instead?

engine/util/util.go
jonathangold: I guess not. If you run an empty struct you just don't get any auto edges. If you run a nil struct it panics anyway, so I don't have to check that either.

jonathangold:
@purpleidea done. See https://github.com/purpleidea/mgmt/pull/418#discussion_r241906802

engine/util/util.go
```diff
@@ -331,3 +331,37 @@ func unitStateAction(ctx context.Context, conn *dbus.Conn, unit, action string)
...
+// Test takes the output of the last call to Next() and outputs true if we
+// should continue.
+func (obj *autoEdgeCombiner) Test(input []bool) bool {
+	if !obj.ae[obj.ptr].Test(input) {
+		obj.ptr++ // match found, on to the next
+	}
+	return len(obj.ae) > obj.ptr // are there any auto edges left?
+}
+
+// AutoEdgeCombiner takes any number of AutoEdges and combines them to be
```
purpleidea: Use this comment instead please (wrapped to 80):

AutoEdgeCombiner takes any number of AutoEdge structs, and combines them into a single one, so that the logic from each one can be built separately, and then combined using this utility. This makes implementing different AutoEdge generators much easier. This respects the Next() and Test() API, and ratchets through each AutoEdge entry until they have all run their course.

engine/resources/cron.go
```diff
@@ -0,0 +1,493 @@
...
+	"github.com/purpleidea/mgmt/util"
+
+	sdbus "github.com/coreos/go-systemd/dbus"
+	"github.com/coreos/go-systemd/unit"
+	systemdUtil "github.com/coreos/go-systemd/util"
+	"github.com/godbus/dbus"
+	errwrap "github.com/pkg/errors"
+)
+
+const ctxTimeout = 30
```
purpleidea: put all the const in the same block. You can put the private ones at the bottom.

engine/resources/cron.go
```diff
@@ -0,0 +1,493 @@
...
+	AccuracySec string `yaml:"accuracysec"`
+	// RandomizedDelaySec delays the timer by a randomly selected, evenly
+	// distributed amount of time between 0 and the specified time value. The
+	// value must be a valid systemd-time time span.
+	RandomizedDelaySec string `yaml:"randomizeddelaysec"`
+
+	// Unit is the name of the systemd service unit. It is only necessary to
+	// set if you want to specify a service with a different name than the
+	// resource.
+	Unit string `yaml:"unit"`
```
purpleidea: I thought we were moving this to the top?

engine/resources/cron.go
```diff
@@ -0,0 +1,493 @@
...
+	return file, nil
+}
+
+// Validate if the params passed in are valid data.
+func (obj *CronRes) Validate() error {
+	// validate state
+	if obj.State != "absent" && obj.State != "exists" {
+		return fmt.Errorf("state must be 'absent' or 'exists'")
+	}
+	if obj.State == "absent" { // that's all we need to know
```
purpleidea: While true, we shouldn't not do cheap extra validation since the user might have that value come from the bool and the rest be statically defined. So catch the extra things early instead of only when this is present and then it's too late.

Validate the other parts that we can.

purpleidea:
You missed at least one review item, and also the commit adding autoedges: f2766a2 -- something is up. There's way too much churn in that patch, can you take a look and make sure it's intentional??

engine/resources/cron.go
```diff
@@ -80,9 +81,18 @@ type CronRes struct {
 
 	init *engine.Init
 
+	// Unit is the name of the systemd service unit. It is only necessary to
```
purpleidea: You mixed up your commits. This change shouldn't be in this commit, etc...

jonathangold: I meant to rebase that change to the initial commit, but by the time I'd finished breaking up the session commit, it slipped my mind. Now it's fixed.

purpleidea:
Okay, I've merged the first three commits (fast forward) and I cherry-picked the last one.

This just leaves the autoedge commit. That one has more churn in it than I think we need. Eg: why are we changing the *old* svc stuff? The lovely util function we merged means we don't need to change it. Right?

purpleidea:
(When you're ready, please rebase the one commit to master, and review.)

purpleidea:
btw! we still get to celebrate you adding another resource! \o/
now we just have to make sure we discourage most sane infrastructures from using it ;)

jonathangold:
> This just leaves the autoedge commit. That one has more churn in it than I think we need. Eg: why are we changing the _old_ svc stuff? The lovely util function we merged means we don't need to change it. Right?

Nothing used SvcAutoEdges, but I can put it back if you want. I guess it might be useful in the future.

See IRC discussion about API. If we change it, I'll write some tests after we do.

jonathangold:
@purpleidea I added back the old `SvcResAutoEdges` stuff in case anyone needs it in the future.

engine/resources/cron.go
```diff
@@ -462,17 +463,33 @@ func (obj *CronUID) IFF(uid engine.ResUID) bool {
...
+	if obj.Unit != "" {
+		unit = obj.Unit
+	}
 	x := &CronUID{
 		BaseUID: engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
-		name:    obj.Name(), // svc name
+		unit:    unit,        // name of target unit
+		session: obj.Session, // user session
 	}
-	return append([]engine.ResUID{x}, obj.file.UIDs()...)
```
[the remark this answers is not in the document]
jonathangold:
```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1412f3a]

goroutine 246 [running]:
github.com/purpleidea/mgmt/engine/traits.(*Named).Name(...)
	/home/jon/go/src/github.com/purpleidea/mgmt/engine/traits/named.go:33
github.com/purpleidea/mgmt/engine/resources.(*FileRes).UIDs(0x0, 0xc0004c3480, 0xc00002b8e0, 0x1)
	/home/jon/go/src/github.com/purpleidea/mgmt/engine/resources/file.go:909 +0x2a
github.com/purpleidea/mgmt/engine/resources.(*CronRes).UIDs(0xc000164b40, 0xc000268f2d, 0xc000164b40, 0x7ff7c8e83388)
	/home/jon/go/src/github.com/purpleidea/mgmt/engine/resources/cron.go:500 +0x1e2
github.com/purpleidea/mgmt/engine/graph/autoedge.addEdgesByMatchingUIDS(0x7ff7c8e83420, 0xc00003d080, 0xc00061cdd0, 0x1, 0x1, 0xc0004c51c0, 0x0, 0xc00002bd50, 0x18, 0x17cbb40, ...)
	/home/jon/go/src/github.com/purpleidea/mgmt/engine/graph/autoedge/autoedge.go:125 +0x32c
github.com/purpleidea/mgmt/engine/graph/autoedge.AutoEdge(0xc0004c51c0, 0xc000574d00, 0xc00002bd50, 0xfc6f64, 0xc0004c51c0)
	/home/jon/go/src/github.com/purpleidea/mgmt/engine/graph/autoedge/autoedge.go:86 +0x45f
github.com/purpleidea/mgmt/engine/graph.(*Engine).AutoEdge(0xc000436140, 0xc00002beb8, 0x0)
	/home/jon/go/src/github.com/purpleidea/mgmt/engine/graph/autoedge.go:29 +0x5a
github.com/purpleidea/mgmt/lib.(*Main).Run.func10(0x19c92b8, 0xc000392b24, 0xc0001c4480, 0xc0004d4c80, 0xc0001c4488, 0xc0000bd200, 0xc0001ea380, 0x6, 0xc000592480, 0x1bece00, ...)
	/home/jon/go/src/github.com/purpleidea/mgmt/lib/main.go:593 +0x7c5
created by github.com/purpleidea/mgmt/lib.(*Main).Run
	/home/jon/go/src/github.com/purpleidea/mgmt/lib/main.go:423 +0xd7c

shell returned 2
```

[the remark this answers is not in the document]
jonathangold: also, obj.file will be empty in certain cases

engine/resources/svc.go
```diff
@@ -425,12 +434,46 @@ func (obj *SvcResAutoEdges) Test(input []bool) bool {
...
-// AutoEdges returns the AutoEdge interface. In this case the systemd units.
+// SvcResAutoEdgesCron holds the state of the svc -> cron auto edge generator.
+type SvcResAutoEdgesCron struct {
+	unit    string // target unit
+	session bool   // user session
+}
+
+// Next returns the next automatic edge.
+func (obj *SvcResAutoEdgesCron) Next() []engine.ResUID {
+	// XXX: should this be true if SvcRes State == "stopped"?
```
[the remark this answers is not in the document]
jonathangold: It definitely works as is for all the cases I can think of.

I've thought about it for a bit, and I'm pretty sure it never needs to be reversed. The only case I wasn't sure about is if pkg uninstalls a svc backing a timer, but that seems fine too. E.g.

```
Install edges:

pkg - svc - cron
```
```
Uninstall edges:

       cron
svc <
       pkg
```

But tests are good, and I'll write a bunch anyway in my next pr.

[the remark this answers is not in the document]
jonathangold: I'll whip up an integration test with this kind of graph soon.

engine/resources/svc.go
```diff
@@ -425,12 +434,46 @@ func (obj *SvcResAutoEdges) Test(input []bool) bool {
...
+// AutoEdges returns the AutoEdge interface. In this case, systemd unit file
+// resources and cron (systemd-timer) resources.
 func (obj *SvcRes) AutoEdges() (engine.AutoEdge, error) {
 	var data []engine.ResUID
-	svcFiles := []string{
-		fmt.Sprintf("/etc/systemd/system/%s.service", obj.Name()),     // takes precedence
-		fmt.Sprintf("/usr/lib/systemd/system/%s.service", obj.Name()), // pkg default
+	var svcFiles []string
+	svcFiles = []string{
+		fmt.Sprintf("%s/.config/systemd/user/%s.service", os.Getenv("HOME"), obj.Name()), // user svc
```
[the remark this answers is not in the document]
jonathangold: do you want me to pull that out into a separate commit?

engine/resources/cron.go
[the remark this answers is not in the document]
purpleidea: Do we want to include it when it's not nil?

engine/resources/svc.go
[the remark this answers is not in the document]
purpleidea: drbd?

Actually ask Felix before you do this. I think he has a `mock` thing planned that might help. Not sure.

But ideally we need to write some test cases in pure golang that loop through a resource and the matching effects it would have so that we can define resource behaviour to field properties somewhere...

engine/resources/cron.go
```diff
@@ -145,11 +146,18 @@ func (obj *CronRes) makeComposite() (*FileRes, error) {
...
 		// user timer
-		path = fmt.Sprintf("%s/.config/systemd/user/%s.timer", os.Getenv("HOME"), obj.Name())
+		u, err := user.Current()
+		if err != nil {
+			return nil, errwrap.Wrapf(err, "error getting current user")
+		}
+		if u.HomeDir == "" {
+			return nil, fmt.Errorf("user has no home directory")
+		}
+		path = fmt.Sprintf("%s/.config/systemd/user/%s.timer", u.HomeDir, obj.Name())
```
purpleidea: In this case, keep the sprintf for the suffix, but use path.Join for the homedir+prefix to avoid double slashes and stuff.

engine/resources/svc.go
```diff
@@ -425,13 +434,55 @@ func (obj *SvcResAutoEdges) Test(input []bool) bool {
...
+		// user timer
+		u, err := user.Current()
+		if err != nil {
+			return nil, errwrap.Wrapf(err, "error getting current user")
+		}
+		if u.HomeDir == "" {
+			return nil, fmt.Errorf("user has no home directory")
+		}
+		svcFiles = []string{
+			fmt.Sprintf("%s/.config/systemd/user/%s.timer", u.HomeDir, obj.Name()), // user svc
```
purpleidea: same as above

engine/resources/cron.go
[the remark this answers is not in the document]
jonathangold: This panic was not when it was nil. I think it's because obj.file isn't a fully implemented resource unto itself. Maybe you can tell me what I'm missing. ;)

purpleidea:
Trace it? Init() ?

engine/resources/cron.go
```diff
@@ -462,17 +460,34 @@ func (obj *CronUID) IFF(uid engine.ResUID) bool {
...
+		unit = obj.Unit
+	}
+	cronUID := &CronUID{
 		BaseUID: engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
-		name:    obj.Name(), // svc name
+		unit:    unit,        // name of target unit
+		session: obj.Session, // user session
 	}
-	return append([]engine.ResUID{x}, obj.file.UIDs()...)
+	file, _ := obj.makeComposite() // error will be caught in validate
```
purpleidea: To be safe, check the error, and instead of erroring, just *don't* append it... You need a *really* good excuse to not do something with an error, and this isn't it =D

engine/resources/cron.go
```diff
@@ -495,6 +510,23 @@ func (obj *CronRes) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+// UnitFilePath returns the path to the systemd-timer unit file.
+func (obj *CronRes) UnitFilePath() (string, error) {
```
purpleidea: nice consolidation

engine/resources/cron.go
```diff
@@ -462,17 +460,37 @@ func (obj *CronUID) IFF(uid engine.ResUID) bool {
...
+	if obj.Unit != "" {
+		unit = obj.Unit
+	}
+	cronUID := &CronUID{
 		BaseUID: engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
-		name:    obj.Name(), // svc name
+		unit:    unit,        // name of target unit
+		session: obj.Session, // user session
+	}
+	file, err := obj.makeComposite()
```
purpleidea: I think there's a more elegant way to write this...
uids := the one that's always there..
if f, err != nil {
uids=append(uids, f.uids())
return uids

purpleidea:
Merged \o/ thanks!

PR #420, opened 2018-12-16
docs: Explain the langpuppet interface and function

ffrank:


docs/puppet-guide.md
```diff
@@ -164,3 +164,148 @@ language features.
...
+
+`mgmt` will derive a graph that contains all edges and vertices from
+both inputs. You essentially get two unrelated subgraphs that run in
+parallel. To form edges between these subgraphs, you have to define
+special vertices that will be merged. This works through a hard-coded
+naming scheme.
+
+### Mixed graph example 1 - No merges
+
+```
```
purpleidea: nit: you can use `mcl` as the language here if you want.

docs/puppet-guide.md
```diff
@@ -164,3 +164,148 @@ language features.
...
+# lang
+file "/tmp/mgmt_dir/" { state => present }
```
purpleidea: I think this is not correct mcl code. state is a string ;)

ffrank:
Makes sense. All fixed, I believe.

purpleidea:
Merged, thanks!

PR #421, opened 2018-12-17
util: Add PathSlice type that satisfies sort.Interface

jonathangold:
This commit adds a []string{} type alias named PathSlice, and the
Len(), Swap(), and Less() methods required to satisfy sort.Interface.
Now you can do `sort.Sort(util.PathSlice(foo))` where foo is a slice
of paths. It will be sorted by depth in alphabetical order.


purpleidea:
(This is getting used in a feature branch I'm working on.)

PR #413, opened 2018-12-08
etcd: fs: Use source filepath base in CopyFs destination path

jonathangold:
This patch corrects the destination path in CopyFs to use the source's
base filepath, instead of the entire source path. Now copying /foo/bar
to /baz results in /baz/bar instead of /baz/foo/bar. This commit also adds
a test to verify this behaviour.

PR #422, opened 2018-12-17
util: WIP: Add CopyFs testing

purpleidea:
This is currently broken, but it includes a WIP test runner. I get the following output:

```
$ go test -tags nodocker github.com/purpleidea/mgmt/util -run TestCopyDiskToFs1 -v
=== RUN   TestCopyDiskToFs1
--- FAIL: TestCopyDiskToFs1 (0.00s)
	afero_test.go:33: tests directory is: /home/james/code/gopath/src/github.com/purpleidea/mgmt/util/afero_test/TestCopyDiskToFs1/
	afero_test.go:61: testing: simple1.tree
	afero_test.go:78: tree:
		.
		├── 1
		├── 2
		├── 3
		├── a
		├── b
		├── bar/
		├── baz/
		├── c
		├── foo/
		├── simple1/
		└── z
	afero_test.go:81: trees differ for: simple1.tree
FAIL
FAIL	github.com/purpleidea/mgmt/util	0.005s
```

Running an actual `tree` command on the folder produces:

```
.
├── 1
├── 2
├── 3
└── foo
    └── bar
        ├── baz
        │   ├── a
        │   ├── b
        │   └── c
        └── z
```

So it's clear there is a bug and somehow this is getting flattened. If someone could fix this, and send me a patch, I'll merge that patch, and then merge this and then you can send an additional commit with more tests in this format.

### DIFF
--- a/util/afero_test.go
+++ b/util/afero_test.go
@@ -1,0 +1,85 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// +build !root
+
+package util
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/spf13/afero"
+)
+
+func TestCopyDiskToFs1(t *testing.T) {
+	const testDirectory = "afero_test/" + "TestCopyDiskToFs1/"
+	dir, err := TestDir(testDirectory)
+	t.Logf("tests directory is: %s", dir)
+	if err != nil {
+		t.Errorf("could not get tests directory: %+v", err)
+		return
+	}
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		t.Errorf("could not read through tests directory: %+v", err)
+		return
+	}
+	for _, f := range files {
+		if !f.IsDir() {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


jonathangold:
I'll have a closer look when I get home, but how are you providing the etcd
instance/cluster to back the filesystem?

purpleidea:
Just testing with a simple memfs here... This just tests the copying logic, not my etcdfs implementation.

jonathangold:
@purpleidea if you use afero.go from before I "fixed" it for etcd (the tests still don't pass, but they're a lot closer.) I think the changes I made in #413 fixed CopyFs for your implementation of etcdfs, but I'm thinking there was actually a bug in etcdfs, not CopyFs.

```
=== RUN   TestCopyDiskToFs1
--- FAIL: TestCopyDiskToFs1 (0.00s)
    afero_test.go:33: tests directory is: /home/jon/go/src/github.com/purpleidea/mgmt/util/afero_test/TestCopyDiskToFs1/
    afero_test.go:61: testing: simple1.tree
    afero_test.go:78: tree:
        .
        └── home/
            └── jon/
                └── go/
                    └── src/
                        └── github.com/
                            └── purpleidea/
                                └── mgmt/
                                    └── util/
                                        └── afero_test/
                                            └── TestCopyDiskToFs1/
                                                └── simple1/
                                                    ├── 1
                                                    ├── 2
                                                    ├── 3
                                                    └── foo/
                                                        └── bar/
                                                            ├── baz/
                                                            │   ├── a
                                                            │   ├── b
                                                            │   └── c
                                                            └── z
    afero_test.go:81: trees differ for: simple1.tree
FAIL
FAIL	github.com/purpleidea/mgmt/util	0.003s

shell returned 1
```

So it's working, with the old afero.go, but you need to prepend the directory in your test for it to pass.

jonathangold:
As for the bug in etcdfs, I can take a shot at tracing it a little later.

purpleidea:
@jonathangold I'm confused, why would it change whether we're using etcdfs or not? Maybe there is *also* a bug in etcdfs, but there's at least a bug in the code as seen in this test case. (CopyFs is supposed to be fs agnostic, so since it's using the real filesystem, it's probably CopyFs that's broken.) We should fix that before fixing etcdfs which will be trickier.

jonathangold:
@purpleidea 
> it's probably CopyFs that's broken
Yeah. I broke it in #413. The fix for CopyFs is changing it back to how it was before.
Then the etcdfs tests fail, but that's because of the bug in etcdfs, which we're now aware of.

jonathangold:
see #423

PR #423, opened 2018-12-18
util: afero: Use dst + base path for mkdir in walk

jonathangold:
It seems that for etcdfs, the name parameter of walk is fed with
the absolute path instead of the base path. This change does not
solve that problem, but rather strips the parent path from name
and appends it to the destination path. For cases where the name
is already correct, we end up producing the same string again,
but where name has an untrimmed parent path, it will be removed
before appending name to dst. As a result, etcdfs now behaves
like any other filesystem for CopyFs.

### DIFF
--- a/util/afero.go
+++ b/util/afero.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"os"
 	"path"
-	"path/filepath"
 
 	"github.com/spf13/afero"
 )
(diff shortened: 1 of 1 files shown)


jonathangold:
@purpleidea try this with your tests. Like the commit message says, this mitigates the bug, rather than solving it, but it's a start.

jonathangold:
cc: @keur

purpleidea:
Thanks for the patch! ... I think I'm going to hold off on merging this until we figure out more definitively where the issue is. I thought we had decided it was in etcdfs itself? I didn't dig or trace this yet, because I think you two have a handle on it... So let's see where it goes. I'll leave this open for now.

jonathangold:
@purpleidea Yeah, I didn't expect you to merge it, tbh. But, it does show where WalkFn is going wrong (due to the etcdfs bug, I think.) I suspect afero.Walk() is not recognizing path separators correctly for etcdfs, so it's treating /foo/bar/baz as a single file, rather than a file nested in two directories.
cc: @keur

jonathangold:
Closing since @keur's solution in #424 is closer.

PR #424, opened 2018-12-18
Copyfs patch

keur:
CopyFs patch and added tests. Hopefully the tests reflect the desired behavior of this function.

### DIFF
--- a/util/afero.go
+++ b/util/afero.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"os"
 	"path"
-	"path/filepath"
 
 	"github.com/spf13/afero"
 )
@@ -89,19 +88,20 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
 // currently undefined.
 // TODO: this should be made more rsync like and robust!
 func CopyFs(srcFs, dstFs afero.Fs, src, dst string, force bool) error {
-	if src == "" {
-		src = "/"
-	}
-	if dst == "" {
-		dst = "/"
-	}
+	src = path.Join("/", src)
+	dst = path.Join("/", dst)
+
+	// TODO: clean this up with function that gets parent dir?
+	src = path.Clean(src)
+	parentDir, _ := path.Split(src)
+	srcFsLen := len(parentDir)
+
 	walkFn := func(name string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
-		//perm := info.Perm()
-		perm := info.Mode() // TODO: is this correct?
-		p := path.Join(dst, filepath.Base(name))
+		perm := info.Mode() // get file permissions
+		p := path.Join(dst, name[srcFsLen:])
 		if info.IsDir() {
 			err := dstFs.Mkdir(p, perm)
 			if os.IsExist(err) && (name == "/" || force) {
--- a/util/afero_test.go
+++ b/util/afero_test.go
@@ -20,6 +20,7 @@
 package util
 
 import (
+	"bytes"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -27,10 +28,96 @@ import (
 	"github.com/spf13/afero"
 )
 
-func TestCopyDiskToFs1(t *testing.T) {
-	if true {
-		return // XXX: remove me once this test passes
+var dirInputs = []struct {
+	srcDirFull  string
+	srcCopyRoot string
+	dstCopyRoot string
+	dstExpected string
+	force       bool
+}{
+	{"/tmp/foo/bar/baz/", "/tmp/foo/", "/", "/foo/bar/baz", false},
+	{"/tmp/zoo/zar/zaz/", "/tmp/zoo/zar", "/start/dir", "/start/dir/zar/zaz", false},
+	{"/foo", "/foo", "/", "/foo", false},
+	{"/foo", "/foo", "/", "/foo", true},
+}
+
+func TestCopyFs1(t *testing.T) {
+	for _, tt := range dirInputs {
+		src := afero.NewMemMapFs()
+		dst := afero.NewMemMapFs()
+
+		t.Run(tt.srcDirFull, func(t *testing.T) {
+			err := src.MkdirAll(tt.srcDirFull, 0700)
+			if err != nil {
+				t.Errorf("could not MkdirAll %+v", err)
+				return
+			}
+			err = CopyFs(src, dst, tt.srcCopyRoot, tt.dstCopyRoot, tt.force)
+			if err != nil {
+				t.Errorf("error copying source %s to dest %s", tt.srcCopyRoot, tt.dstCopyRoot)
+				return
+			}
+
+			isDir, err := afero.IsDir(dst, tt.dstExpected)
+			if err != nil {
+				t.Errorf("could not check IsDir: %+v", err)
+				return
+			}
+			if !isDir {
+				t.Errorf("expected directory tree %s to exist in dest", tt.dstExpected)
+				return
+			}
+		})
 	}
+}
+
+func TestCopyFs2(t *testing.T) {
+	tree := "/foo/bar/baz/"
+	var files = []struct {
+		path    string
+		content []byte
+	}{
+		{"/foo/foo.txt", []byte("foo")},
+		{"/foo/bar/bar.txt", []byte("bar")},
+		{"/foo/bar/baz/baz.txt", []byte("baz")},
+	}
+
+	src := afero.NewMemMapFs()
+	dst := afero.NewMemMapFs()
+
+	err := src.MkdirAll(tree, 0700)
+	if err != nil {
+		t.Errorf("could not MkdirAll: %+v", err)
+		return
+	}
+
+	for _, f := range files {
+		err = afero.WriteFile(src, f.path, f.content, 0600)
+		if err != nil {
+			t.Errorf("could not WriteFile: %+v", err)
+			return
+		}
+	}
+
+	if err = CopyFs(src, dst, "", "", false); err != nil {
+		t.Errorf("could not CopyFs: %+v", err)
+		return
+	}
+
+	for _, f := range files {
+		content, err := afero.ReadFile(dst, f.path)
+		if err != nil {
+			t.Errorf("could not ReadFile: %+v", err)
+			return
+		}
+		if !bytes.Equal(content, f.content) {
+			t.Errorf("expected: %s, actual: %s, for file %s", string(f.content), string(content), f.path)
+			return
+		}
+	}
+}
+
+func TestCopyDiskToFs1(t *testing.T) {
 	dir, err := TestDirFull()
 	if err != nil {
 		t.Errorf("could not get tests directory: %+v", err)


jonathangold:
@keur does `go test -v github.com/purpleidea/mgmt/etcd/fs -run TestCopyFs3` pass with your changes?

etcd/fs/fs_test.go
```diff
@@ -294,10 +294,6 @@ func TestFs3(t *testing.T) {
 	}
 	t.Logf("tree2: \n%s", tree2)
 
-	if _, err := memFs.Stat("/bar"); err != nil {
```
jonathangold: This test is desired. `util.CopyFs(etcdFs, memFs, "/tmp/foo/bar", "/", false)` is copying `bar` to `foo` so if the test fails on this line, then CopyFs is wrong.

util/afero.go
```diff
@@ -88,20 +88,30 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
 // dst has a different set of contents in the same location, the behaviour is
 // currently undefined.
 // TODO: this should be made more rsync like and robust!
-func CopyFs(srcFs, dstFs afero.Fs, src, dst string, force bool) error {
-	if src == "" {
+func CopyFs(srcFs afero.Fs, dstFs afero.Fs, src string, dst string, force bool) error {
```
jonathangold: This func signature is functionally identical to the one you replaced, but longer, and harder to read IMO.

util/afero.go
```diff
@@ -88,20 +88,30 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
...
+	} else {
+		// todo: clean this up with function that gets parent dir?
+		src = strings.TrimSuffix(src, "/")
+		dirs := strings.Split(src, "/")
+		out := strings.Join(dirs[:len(dirs)-1], "/")
+		srcFsLen = len(out)
 	}
-	if dst == "" {
-		dst = "/"
+	if dst == "" || string(dst[len(dst)-1]) != "/" {
```
jonathangold: All this string manipulation is kind of complicated. Can you try with the version of `afero.go` in #423 to see if your tests pass with that version?

We should settle on the simplest solution that works for all our tests (and is logically correct.) If you want to discuss the bug, ping me on IRC, or here.

util/afero.go
```diff
@@ -88,20 +88,30 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
...
 	}
+
 	walkFn := func(name string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
-		//perm := info.Perm()
-		perm := info.Mode() // TODO: is this correct?
-		p := path.Join(dst, filepath.Base(name))
+		perm := info.Mode() // get file permissions
```
jonathangold: Is this correct?
I'm not suggesting it isn't, just wondering why you got rid of the TODO.

util/util_test.go
```diff
@@ -903,5 +905,74 @@ func TestPathSliceSort(t *testing.T) {
...
+		t.Errorf("Expected /foo/bar/bar.txt to contain %s", barData)
+	}
+
+	contains, _ = afero.FileContainsBytes(dst, "/foo/bar/baz/baz.txt", bazData)
+	if !contains {
+		t.Errorf("Expected /foo/bar/baz/baz.txt to contain %s", bazData)
+	}
+}
+
+func TestCopyFs2(t *testing.T) {
```
jonathangold: Since TestCopyFs2, 3 and 4 are the same, but with different inputs and outputs, you might want to consider combining them into a table driven test and loop through the parameters, like in https://github.com/purpleidea/mgmt/blob/bd80ced9b238016271c55d93e052eaf25344f1c9/util/util_test.go#L861.

Not strictly necessary, but useful if we want to quickly add more test cases down the line.

util/util_test.go
```diff
@@ -903,5 +905,74 @@ func TestPathSliceSort(t *testing.T) {
...
+
+func TestCopyFs1(t *testing.T) {
+	src := afero.NewMemMapFs()
+	dst := afero.NewMemMapFs()
+
+	fooData := []byte("foo")
+	barData := []byte("bar")
+	bazData := []byte("baz")
+
+	src.MkdirAll("/foo/bar/baz/", 0700)
```
jonathangold: file umask(s) should probably be consts.

etcd/fs/fs_test.go
keur: yeah, gonna remove the commit where i deleted this test. my patch passes this test, anyway.

util/afero.go
keur: i believe it is correct, which is why i removed the TODO.

util/util_test.go
keur: Sure. This is a good idea.

keur:
@jonathangold `go test -v github.com/purpleidea/mgmt/etcd/fs -run TestFs3` does pass with these changes. I had rebased wrong before I went to sleep. CI should pass now.

util/util_test.go
```diff
@@ -905,3 +907,74 @@ func TestPathSliceSort(t *testing.T) {
...
+		t.Errorf("Expected /foo/bar/baz/baz.txt to contain %s", bazData)
+	}
+}
+
+func TestCopyFs2(t *testing.T) {
+	src := afero.NewMemMapFs()
+	dst := afero.NewMemMapFs()
+
+	src.MkdirAll("/tmp/foo/bar/baz/", 0700)
+	CopyFs(src, dst, "/tmp/foo/", "/", false)
```
jonathangold: you should check the error returned by CopyFs, and fail `if err != nil`

util/afero.go
```diff
@@ -88,20 +88,30 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
...
+func CopyFs(srcFs afero.Fs, dstFs afero.Fs, src string, dst string, force bool) error {
+	var srcFsLen int
+
+	if len(src) <= 1 {
+		// root directory
 		src = "/"
+		srcFsLen = 1
+	} else {
+		// todo: clean this up with function that gets parent dir?
+		src = strings.TrimSuffix(src, "/")
```
jonathangold: This is looking a lot cleaner! Instead of methods from strings, you should use path.Clean(), path.Split() and path.Join() since we're working with paths.

keur:
@jonathangold I tried to address most of the comments. Thanks for the comments! Really helped clean this PR up.

util/afero.go
```diff
@@ -88,20 +87,27 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
 // dst has a different set of contents in the same location, the behaviour is
 // currently undefined.
 // TODO: this should be made more rsync like and robust!
-func CopyFs(srcFs, dstFs afero.Fs, src, dst string, force bool) error {
-	if src == "" {
+func CopyFs(srcFs afero.Fs, dstFs afero.Fs, src string, dst string, force bool) error {
```
purpleidea: Please leave the sig alone, as jonathan said. (They're equivalent.)

util/afero.go
```diff
@@ -88,20 +87,27 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
...
+	var srcFsLen int
+
+	if len(src) <= 1 {
+		// root directory
 		src = "/"
+		srcFsLen = 1
+	} else {
+		// todo: clean this up with function that gets parent dir?
```
purpleidea: TODO's should be capitalized so that editors catch them, eg: `// TODO: please do some thing`
Same with FIXME, and XXX

util/util_test.go
```diff
@@ -905,3 +907,66 @@ func TestPathSliceSort(t *testing.T) {
 		}
 	}
 }
+
+func TestCopyFs1(t *testing.T) {
```
purpleidea: Since we're testing a function in afero.go, please add the tests in afero_test.go

util/util_test.go
```diff
@@ -905,3 +907,66 @@ func TestPathSliceSort(t *testing.T) {
...
+	src := afero.NewMemMapFs()
+	dst := afero.NewMemMapFs()
+
+	fooData := []byte("foo")
+	barData := []byte("bar")
+	bazData := []byte("baz")
+
+	src.MkdirAll("/foo/bar/baz/", 0700)
```
purpleidea: *any time* we have something that returns an error, we need to explicitly handle it, or explain why we can't in a comment... there are a bunch of these in here... Sometimes when you see an excessive amount of error handling, it's occasionally a hint that it could be re-written differently. See my main comment at the end.

util/util_test.go
```diff
@@ -905,3 +907,66 @@ func TestPathSliceSort(t *testing.T) {
...
+func TestCopyFsDirTree(t *testing.T) {
+	for _, tt := range dirInputs {
+		src := afero.NewMemMapFs()
+		dst := afero.NewMemMapFs()
+
+		t.Run(tt.srcDirFull, func(t *testing.T) {
+			src.MkdirAll(tt.srcDirFull, 0700)
+			err := CopyFs(src, dst, tt.srcCopyRoot, tt.dstCopyRoot, tt.force)
+			if err != nil {
+				t.Errorf("Error copying source %s to dest %s", tt.srcCopyRoot, tt.dstCopyRoot)
```
purpleidea: Errors are not full sentences, so they aren't capitalized. `golint` usually warns about this.

keur:
@purpleidea this good to go?

purpleidea:
@keur Reviewing now...

util/afero_test.go
```diff
@@ -20,17 +20,102 @@
...
+	CopyFs(src, dst, "", "", false)
+
+	for _, f := range files {
+		content, err := afero.ReadFile(dst, f.path)
+		if err != nil {
+			t.Errorf("could not ReadFile: %+v", err)
+			return
+		}
+		if !bytes.Equal(content, f.content) {
+			t.Errorf("Expected: %s, Actual: %s, for file %s", string(f.content), string(content), f.path)
```
purpleidea: nit: expected/actual lower case please

util/afero_test.go
```diff
@@ -20,17 +20,102 @@
...
+
+	for _, f := range files {
+		err = afero.WriteFile(src, f.path, f.content, 0600)
+		if err != nil {
+			t.Errorf("could not WriteFile: %+v", err)
+			return
+		}
+	}
+
+	CopyFs(src, dst, "", "", false)
```
purpleidea: Needs err checking

util/afero_test.go
```diff
@@ -20,17 +20,102 @@
...
+			if !isDir {
+				t.Errorf("expected directory tree %s to exist in dest", tt.dstExpected)
+				return
+			}
+		})
+	}
+}
+
+func TestCopyFs2(t *testing.T) {
+
```
purpleidea: nit: remove nl please

util/afero.go
```diff
@@ -89,19 +88,20 @@ func stringify(fs afero.Fs, name string, indent []bool) (string, error) {
 // currently undefined.
 // TODO: this should be made more rsync like and robust!
 func CopyFs(srcFs, dstFs afero.Fs, src, dst string, force bool) error {
-	if src == "" {
-		src = "/"
-	}
-	if dst == "" {
-		dst = "/"
-	}
+	src = path.Join("/", src)
```
purpleidea: This is slightly different from the earlier version, but maybe it's better! Will test shortly :)

util/afero_test.go
```diff
@@ -20,17 +20,104 @@
...
+	for _, f := range files {
+		err = afero.WriteFile(src, f.path, f.content, 0600)
+		if err != nil {
+			t.Errorf("could not WriteFile: %+v", err)
+			return
+		}
+	}
+
+	err = CopyFs(src, dst, "", "", false)
```
purpleidea: you can use the golang form:

if err := CopyFS(...); err != nil {

and you need a return in there to end the test early.

purpleidea:
Merged, thanks!

Welcome to the club @keur \o/

PR #425, opened 2018-12-19
etcd: fs: Combine etcd FsCopy tests and add some new test cases

jonathangold:
This is for merge only after #424 as one test will definitely fail until that's fixed. Happy to add some more test cases if anyone has any ideas.
/cc @keur

### DIFF
--- a/etcd/fs/fs_test.go
+++ b/etcd/fs/fs_test.go
@@ -306,3 +306,77 @@ func TestFs3(t *testing.T) {
 		return
 	}
 }
+
+func TestEtcdCopyFs0(t *testing.T) {
+	tests := []struct {
+		mkdir, cpsrc, cpdst, check string
+		force                      bool
+	}{
+		{
+			mkdir: "/tmp/foo/bar",
+			cpsrc: "/tmp/foo",
+			cpdst: "/baz/",
+			check: "/baz/foo/bar",
+			force: false,
+		},
+		{
+			mkdir: "/tmp/this/is/a/really/deep/directory/to/make/sure/we/can/handle/deep/copies",
+			cpsrc: "/tmp/this/is/a",
+			cpdst: "/that/was/",
+			check: "/that/was/a/really/deep/directory/to/make/sure/we/can/handle/deep/copies",
+			force: false,
+		},
+	}
+
+	for _, tt := range tests {
+		stopEtcd, err := runEtcd()
+		if err != nil {
+			t.Errorf("setup error: %+v", err)
+			return
+		}
+		defer stopEtcd() // ignore the error
+
+		etcdClient := &etcd.ClientEtcd{
+			Seeds: []string{"localhost:2379"}, // endpoints
+		}
+
+		if err := etcdClient.Connect(); err != nil {
+			t.Errorf("client connection error: %+v", err)
+			return
+		}
+		defer etcdClient.Destroy()
+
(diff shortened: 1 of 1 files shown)


jonathangold:
@purpleidea should this commit message be failing the test? It looks fine to me. Bug in regex?

etcd/fs/fs_test.go
```diff
@@ -183,123 +183,81 @@ func TestFs1(t *testing.T) {
 	}
 }
 
-func TestFs2(t *testing.T) {
```
purpleidea: Why are you nuking these??

jonathangold: The two tests are identical, and we needed more cases, so I combined them into a single test, and added two more test cases. TestEtcdCopyFs, does the same tests as TestFs2 and TestFs3, but adds one case for a recursive copy and one extra deep copy.

purpleidea: I've been to lax on this, and judging by my recent commit, I'm obviously not being a good example, so this is a good time to start: if you'd like to make all of those changes, that's fine, but they need to be each done in separate commits (and reviewed individually).

jonathangold: @purpleidea: I split up the commit, so I add the new test in one, and replace TestFs2 and TestFs3 in the second one.

etcd/fs/fs_test.go
```diff
@@ -306,3 +306,73 @@ func TestFs3(t *testing.T) {
...
+		defer etcdClient.Destroy()
+
+		etcdFs := &etcdfs.Fs{
+			Client:     etcdClient.GetClient(),
+			Metadata:   superblock,
+			DataPrefix: etcdfs.DefaultDataPrefix,
+		}
+
+		if err := etcdFs.MkdirAll(tt.mkdir, umask); err != nil {
+			t.Errorf("mkdir error: %+v", err)
```
purpleidea: missing a return?

etcd/fs/fs_test.go
```diff
@@ -306,3 +306,73 @@ func TestFs3(t *testing.T) {
...
+		}
+		tree, err := util.FsTree(etcdFs, "/")
+		if err != nil {
+			t.Errorf("tree error: %+v", err)
+			return
+		}
+		t.Logf("tree: \n%s", tree)
+
+		var memFs = afero.NewMemMapFs()
+
```
purpleidea: nit: kill nl please

etcd/fs/fs_test.go
```diff
@@ -306,3 +306,73 @@ func TestFs3(t *testing.T) {
 		return
 	}
 }
+
+func TestEtcdCopyFs(t *testing.T) {
```
purpleidea: Let's give this an integer at the end. 0 or 1 or whatever is free :)

purpleidea:
AFAICT in the terrible GitHub UI, the second commit only deletes stuff since you've already added it in the first test, correct? If so, let's just leave those stragglers in. yay, one less commit to write or rebase :)

unless I'm missing something... Thanks for the patch!

jonathangold:
@purpleidea I mixed up my git history. It's all sorted now. Made a few changes, and added back the missing tests.

purpleidea:
LGTM :)

purpleidea:
Now merged, sorry for the delay.

PR #430, opened 2018-12-24
pgraph: Fix GraphCmp function

### LINKED ISSUE #199: [LOVE] Implement graph isomorphism

kkuehlz:
The previous algorithm would fail on the newly added tests due to the
fact that it would check for duplicate vertices. The newer algorithm
doesn't use any fancy isomorphism algorithms. It exploits the fact that
nodes can be expected to have the same label in the two graphs being
compared. This allows us to create a sorted list of vertices based on
the label of a node and the labels of its neighbors.

Once the two graphs are in sorted order, we simply use the vertexCmpFn
on vertices at the same index, and run the edgeCmpFn on each of the
vertices' respective edges.

Resolves #199

### DIFF
--- a/pgraph/pgraph.go
+++ b/pgraph/pgraph.go
@@ -197,16 +197,6 @@ func (g *Graph) FindEdge(v1, v2 Vertex) Edge {
 	return edge
 }
 
-// Vertices returns a randomly sorted slice of all vertices in the graph.
-// The order is random, because the map implementation is intentionally so!
-func (g *Graph) Vertices() []Vertex {
-	var vertices []Vertex
-	for k := range g.adjacency {
-		vertices = append(vertices, k)
-	}
-	return vertices
-}
-
 // Edges returns a randomly sorted slice of all edges in the graph.
 // The order is random, because the map implementation is intentionally so!
 func (g *Graph) Edges() []Edge {
@@ -231,6 +221,25 @@ func (g *Graph) VerticesChan() chan Vertex {
 	return ch
 }
 
+// Vertices returns a randomly sorted slice of all vertices in the graph.
+// The order is random, because the map implementation is intentionally so!
+func (g *Graph) Vertices() []Vertex {
+	var vertices []Vertex
+	for k := range g.adjacency {
+		vertices = append(vertices, k)
+	}
+	return vertices
+}
+
+// EdgeSlice is a linear list of edges. It can be sorted.
+type EdgeSlice []Edge
+
+func (es EdgeSlice) Len() int { return len(es) }
+
+func (es EdgeSlice) Swap(i, j int) { es[i], es[j] = es[j], es[i] }
+
+func (es EdgeSlice) Less(i, j int) bool { return es[i].String() < es[j].String() }
+
 // VertexSlice is a linear list of vertices. It can be sorted.
 type VertexSlice []Vertex
 
@@ -246,14 +255,64 @@ func (vs VertexSlice) Less(i, j int) bool { return vs[i].String() < vs[j].String
 // Sort is a convenience method.
 func (vs VertexSlice) Sort() { sort.Sort(vs) }
 
+// VertexIntenseSlice is a linear list of vertices. It can be sorted.
+type VertexIntenseSlice struct {
+	g        *Graph
+	vertices []Vertex
+}
+
+func (vs VertexIntenseSlice) Len() int {
+	return len(vs.vertices)
+}
+
+func (vs VertexIntenseSlice) Swap(i, j int) {
+	v := vs.vertices
+	v[i], v[j] = v[j], v[i]
+}
+
+func (vs VertexIntenseSlice) Less(i, j int) bool {
+	v := vs.vertices
+	if v[i].String() < v[j].String() {
+		return true
+	} else if v[i].String() > v[j].String() {
+		return false
+	}
+
+	// Labels are equal, time to check neighbor labels
+	neighSet1 := vs.g.GraphVertices(v[i])
+	neighSet2 := vs.g.GraphVertices(v[j])
+
+	if len(neighSet1) < len(neighSet2) {
+		return true
+	}
+
+	sort.Sort(VertexSlice(neighSet1))
+	sort.Sort(VertexSlice(neighSet2))
+
+	// TODO: maybe building a hash here and compare the result is better?
+	for i := 0; i < len(neighSet1); i++ {
+		if neighSet1[i].String() < neighSet2[i].String() {
+			return true
+		}
+	}
+
+	return false
+}
+
 // VerticesSorted returns a sorted slice of all vertices in the graph.
 // The order is sorted by String() to avoid the non-determinism in the map type.
 func (g *Graph) VerticesSorted() []Vertex {
-	var vertices []Vertex
-	for k := range g.adjacency {
-		vertices = append(vertices, k)
-	}
+	vertices := g.Vertices()
 	sort.Sort(VertexSlice(vertices)) // add determinism
+	return vertices
+}
+
+// VerticesSortedIntense returns a sorted slice of all vertices in the
+// graph. Instead of sorting by String() on just the vertex, uses
+// the String() element of all its neighbors.
+func (g *Graph) VerticesSortedIntense() []Vertex {
+	vertices := g.Vertices()
+	sort.Sort(VertexIntenseSlice{g, vertices})
 	return vertices
 }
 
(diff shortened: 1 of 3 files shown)


pgraph/pgraph.go
```diff
@@ -197,16 +197,6 @@ func (g *Graph) FindEdge(v1, v2 Vertex) Edge {
 	return edge
 }
 
-// Vertices returns a randomly sorted slice of all vertices in the graph.
```
purpleidea: Why did you move this function? (Look at the diff)

pgraph/pgraph.go
```diff
@@ -231,6 +221,25 @@ func (g *Graph) VerticesChan() chan Vertex {
...
+	for k := range g.adjacency {
+		vertices = append(vertices, k)
+	}
+	return vertices
+}
+
+// EdgeSlice is a linear list of edges. It can be sorted.
+type EdgeSlice []Edge
+
+func (es EdgeSlice) Len() int { return len(es) }
```
purpleidea: These need the standard comments for golint to pass. You can use the ones from the existing VertexSlice.

pgraph/pgraph.go
```diff
@@ -246,17 +255,67 @@ func (vs VertexSlice) Less(i, j int) bool { return vs[i].String() < vs[j].String
...
+		}
+	}
+
+	return false
+}
+
 // VerticesSorted returns a sorted slice of all vertices in the graph.
 // The order is sorted by String() to avoid the non-determinism in the map type.
 func (g *Graph) VerticesSorted() []Vertex {
-	var vertices []Vertex
```
purpleidea: Not sure if this change is _really_ necessary, but if it is, let's do it in a separate commit, since it has nothing to do with this commit which is not trivial, so we don't want to combine them if there's ever something we learn we need to revert or change in this commit.

pgraph/pgraph.go
```diff
@@ -246,17 +255,67 @@ func (vs VertexSlice) Less(i, j int) bool { return vs[i].String() < vs[j].String
 // Sort is a convenience method.
 func (vs VertexSlice) Sort() { sort.Sort(vs) }
 
+// VertexIntenseSlice is a linear list of vertices. It can be sorted.
+type VertexIntenseSlice struct {
```
purpleidea: Why is this "intense" ?

pgraph/pgraph.go
```diff
@@ -246,17 +255,67 @@ func (vs VertexSlice) Less(i, j int) bool { return vs[i].String() < vs[j].String
...
+	g        *Graph
+	vertices []Vertex
+}
+
+func (vs VertexIntenseSlice) Len() int {
+	return len(vs.vertices)
+}
+
+func (vs VertexIntenseSlice) Swap(i, j int) {
+	v := vs.vertices
```
purpleidea: no need to shorten, if you *really* want to, then rename the struct field (it's private anyways)

pgraph/pgraph.go
```diff
@@ -246,17 +255,67 @@ func (vs VertexSlice) Less(i, j int) bool { return vs[i].String() < vs[j].String
...
+func (vs VertexIntenseSlice) Less(i, j int) bool {
+	v := vs.vertices
+	if v[i].String() < v[j].String() {
+		return true
+	} else if v[i].String() > v[j].String() {
+		return false
+	}
+
+	// Labels are equal, time to check neighbor labels
+	neighSet1 := vs.g.GraphVertices(v[i])
```
purpleidea: If you're doing this, then I'm assuming you have a graph in the "intense" struct, which matches the vertex list stored outside as well? So a redundant copy?? Which makes me think this whole thing should just be a method on Graph instead? Or perhaps I'm misunderstanding the idea here...

pgraph/pgraph.go
```diff
@@ -583,84 +642,62 @@ func (g *Graph) VertexMatchFn(fn func(Vertex) (bool, error)) (Vertex, error) {
 // they're equal. It uses a user defined function to compare topologically
 // equivalent vertices, and edges.
 // FIXME: add more test cases
-func (g *Graph) GraphCmp(graph *Graph, vertexCmpFn func(Vertex, Vertex) (bool, error), edgeCmpFn func(Edge, Edge) (bool, error)) error {
-	if graph == nil || g == nil {
-		if graph != g {
+func (g *Graph) GraphCmp(graphOther *Graph, vertexCmpFn func(Vertex, Vertex) (bool, error), edgeCmpFn func(Edge, Edge) (bool, error)) error {
```
purpleidea: Unnecessary rename, especially if it happens in the same commit.

pgraph/pgraph.go
```diff
@@ -583,84 +642,62 @@ func (g *Graph) VertexMatchFn(fn func(Vertex) (bool, error)) (Vertex, error) {
...
+	n1, n2 := g.NumVertices(), graphOther.NumVertices()
 	if n1 != n2 {
 		return fmt.Errorf("base graph has %d vertices, while input graph has %d", n1, n2)
 	}
-	if e1, e2 := g.NumEdges(), graph.NumEdges(); e1 != e2 {
+	if e1, e2 := g.NumEdges(), graphOther.NumEdges(); e1 != e2 {
 		return fmt.Errorf("base graph has %d edges, while input graph has %d", e1, e2)
 	}
 
-	var m = make(map[Vertex]Vertex) // g to graph vertex correspondence
```
purpleidea: The churn in this function will be easier for me to review, once unrelated changes are split into separate commits (or unnecessary renaming is done in a later patch in the future.)

purpleidea (changes requested):
Very glad to see this, thanks!

Biggest issue is unrelated changes happening in the same commit. I'm generally pretty okay with this (most aren't) but the unofficial rule is that it's for cases when the patch doesn't touch critical or tricky code, or that it's trivial or easy to review. In this case, since there aren't too many tests in this space, and it's not an extremely well-understood part of the code, please split the unrelated changes up :)

Also, VertexSlice might already do some of the things you want. Not sure. This would definitely benefit from an improved doc comment on the complex methods.

Thanks!

pgraph/pgraph.go
keur: Sorry, I couldn't think of a good name. Basically, it's intense because it sorts vertices based on the `String()` value. If two vertices have the same value, then it looks at the labels of neighbors to try and differentiate them. 

Suppose we have two vertices in a graph

```
-------------
|v1 -> v2|
|v1 -> v3|
--------------
```

Then the first v1 gets sorted because it has a different neighbor than the other.

pgraph/pgraph.go
keur: No copies. It iterates over vertices and uses neighbors to compare, as described above. the `neighSet` just has the neighbors of vertex `v[i]`.

pgraph/pgraph.go
keur: The issue with my approach is, if we are to allow duplicates vertex names, what is to say one cannot create a graph entirely of the same vertex labels? Now the ordering imposed by one's vertex label and that of it's neighbors is not unique. So we can't impose any ordering on the two graphs, and this degrades into the traditional graph isomorphism problem.

purpleidea: I need to double check where the vertex name comes from. Assuming I remember correctly that it's a combination of the res type string and the res name, then we're good. Although we might need to double check this in the engine when receiving the graph.

This actually goes back to an early design goal which I don't remember if I enforced with code yet: graphs with two compatible resources in the languages, should be allowed. I forget if this is allowed at the language level or engine level and where it should be taken care of :/

pgraph/pgraph.go
purpleidea: But you have two versions of Vertexes:

1) In the struct as field vs
2) In the struct as field graph.adjacency (which can be accessed by methods if need be)

pgraph/pgraph.go
keur: Okay, so after some researching I realized I implemented a naive version of Weisfeiler-Lehman, checking only the first hop (or a node's immediate neighbors). The good news is these algorithms, while not perfect,  tend to do well in practical circumstances. 

I've included psoudecode below so we can agree on an approach before writing code this time. It will hopefully aid in understanding and make the review process much easier.

```python

def create_hashes(G) ->
  Gi.hash_label <- G.label for all Gi
  
  foreach Gi in G:
    for neigh in Gi:
      neigh.hash_label <- hash(Gi.hash_label + neigh.hash_label)

def isomorphic(G1, G2) ->
  if count(G1.vertices) ≠ count(G2.vertices) then ret false
  
  hG1, hG2 <- create_hashes(G1), create_hashes(G2)  
  sG1, sG2 <- sort_by_hash_labels(hG1), sort_by_hash_labels(hG2)

  foreach (Gi, Gj) in (sG1, -G2):
    if Gi.hash_label ≠ G.hash_label then ret false
  ret true
```

Since every neighbor gets a hash assignment based on its neighbor's labels, we get unique hashes that span the entire graph. My previous implementation only considered one-hop, or the first neighbor. If interested, read more [here](http://blog.smola.org/post/33412570425/the-weisfeiler-lehman-algorithm-and-estimation-on) and [here](https://plus.google.com/114866592715069940152/posts/fmBFhjhQcZF). As the first article mentions, this approach can fail on highly symmetric graphs, but it has the benefit of being much more understandable and quite quick.

purpleidea: > can fail on highly symmetric graphs

This sounds like a blocker right there... :/ If it always worked on DAG's, then that's okay, however the article mentions chains as being problematic too... What do you think?

keur: Unsure. Note that the highly symmetric implies a large majority of the graph consists of duplicate labels. If we are modeling resources, it's not often we'll see a large amount of duplicates, but I may be wrong.

The second article claims infinitesimally small collision likelihood, and the author ran it on over 10k samples. The algorithm is a little more involved than my pseudocode, but still much less complex and _much_ faster than any of the exponential algorithms that guarantee isomorphism in all cases. I am still happy to implement [vf2](https://pdfs.semanticscholar.org/f3e1/0bd7521ec6263a58fdaa4369dfe8ad50888c.pdf). If we could get some edge case graphs of what a resource could _realistically_ look like, I think the algorithm in the second article might work well.

It's also important to note that the runtime of vf2 is horrible since isomorphism is NP.

purpleidea: Okay I did a bunch of research, worked on some tests, and hopefully have some answers. Sorry I wasn't better prepared earlier. One of the downsides of trying to "move fast" with few resources on a new project :P 

1) We can't choose any algorithm that could fail even if that's an infinitesimal probability. It's just not clean.

2) Even if something is NP-complete, doesn't mean that it won't be able to complete the execution in a reasonable amount of time for the size of graphs that we have. Whether our graphs will have 100's, 1000's, 10,000's of vertices, I don't know, but I'm optimistic that we can perform the needed computations fast enough.

3) As you might have already seen, there are multiple DAG's in the code base. (I count 3 atm.) The function graph, and the resource graph are the two main ones that we need to be careful about.

In the resource graph (that the engine runs) we currently allow duplicate resources, but this is a bug. We should consider either error-ing there and/or de-duplicating equivalent resources. This is an issue that can be seen if we use the `yaml` front-end eg:

```
---
graph: mygraph
resources:
  file:
  - name: file0
    path: "/tmp/f0"
    content: |
      i am f0
    state: exists
  - name: file0
    path: "/tmp/f0"
    content: |
      i am f1
    state: exists
edges: []
```

This shouldn't work, but it does :/ Badly. They fight. Oops.

This doesn't happen in `mcl` code, because it's clever enough to de-duplicate output!

The function graph however *can* have identical (same `.String()` values, different pointers of course) vertices. We then ultimately run `interpret` on that AST, and get the resource graph. At this stage we ensure that we don't have duplicate resources.

So it actually seems that `GraphCmp` is very useful for _tests_ (where we compare a function graph with identical vertices) but not actually used anywhere in the production code. The most similar function to `GraphCmp` is `GraphSync`, which fortunately doesn't need to deal with duplicate vertices since this is only used in the engine at the moment where this is supposed to be unique :)

So this patch is useful, but only for tests atm. I'm going to update #199 with some more info. Sorry about being a disorganized maintainer here. If you'd like to update it with the fixes, I'd love to merge it, but it's probably not worth spending days on.

PR #431, opened 2018-12-25
Makefile: make bindata a dependency of deps

keur:


docs/quick-start-guide.md
```diff
@@ -51,7 +51,7 @@ cd $GOPATH/src/github.com/purpleidea/mgmt
 export PATH=$PATH:$GOPATH/bin
 ```
 
-* Run `make deps` to install system and golang dependencies. Take a look at
```
purpleidea: Why would we rename this?

docker/Dockerfile.development
```diff
@@ -25,7 +25,7 @@ COPY . /home/$USER_NAME/mgmt
 WORKDIR /home/$USER_NAME/mgmt
 
 # Install dependencies
-RUN make deps
+RUN make dev
```
purpleidea: Assuming you want to change this because you want `bindata`, we don't, the bindata target is a `build` Makefile dep.

keur:
my mistake, running `make` is sufficient here.

PR #434, opened 2019-01-02
engine: resources: exec: Use WatchShell in Exec resource when needed instead of Shell

VinceMacBuche:


purpleidea:
@VinceMacBuche Great catch, thanks!

One small issue: our CI checks the format of commit messages, so you'll need to format it like:

`engine: resources: Use WatchShell etc...`

(The prefix must be lower case as shown)

Thanks!

VinceMacBuche:
@purpleidea Updated the commit message !!

For future contrib, Does using the (full or part of) path as prefix is ok ?

purpleidea:
All the tests for that are in here: https://github.com/purpleidea/mgmt/blob/master/test/test-commit-message.sh

There is a big regexp for the format, and the prefixes are basically heuristics. Just a weak attempt to try and keep things neater. Sorry about that.

This will also fail because it ends with a period :(

One more try?

VinceMacBuche:
Yeah I ran the test suite, with TRAVIS set to true,  and did not find an error !

purpleidea:
(Sorry I didn't merge yet, my laptop that I do the merging from is currently broken...)

VinceMacBuche:
Haha np ;)

purpleidea:
Merged, thanks again, and sorry for the delay!

VinceMacBuche:
No problem at all! Thank you for all your time and work on mgmt !

PR #435, opened 2019-01-10
make: deps: gometalinter: Use latest stable release

jonathangold:
Gometalinter no longer supports the `--install` flag, and fails
to install some of the linters if it is used. This change employs
the supported installation method to ensure all linters are
successfully installed.

see: https://github.com/alecthomas/gometalinter/issues/552
and: https://github.com/alecthomas/gometalinter#installing

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -103,7 +103,9 @@ go get golang.org/x/tools/cmd/goyacc			# formerly `go tool yacc`
 go get golang.org/x/tools/cmd/stringer			# for automatic stringer-ing
 go get golang.org/x/lint/golint				# for `golint`-ing
 go get github.com/tmthrgd/go-bindata/go-bindata	# for compiling in non golang files
-go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
+if env | grep -q -e '^TRAVIS=true$' -e '^JENKINS_URL=' -e '^BUILD_TAG=jenkins'; then
+	go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
+fi
 command -v mdl &>/dev/null || gem install mdl --no-ri --no-rdoc || true	# for linting markdown files
 command -v fpm &>/dev/null || gem install fpm --no-ri --no-rdoc || true	# for cross distro packaging
 cd "$XPWD" >/dev/null
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -104,7 +104,7 @@ go get golang.org/x/tools/cmd/stringer			# for automatic stringer-ing
 go get golang.org/x/lint/golint				# for `golint`-ing
 go get github.com/tmthrgd/go-bindata/go-bindata	# for compiling in non golang files
 if env | grep -q -e '^TRAVIS=true$' -e '^JENKINS_URL=' -e '^BUILD_TAG=jenkins'; then
-	go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
+	curl -L https://git.io/vp6lP | sh # install latest stable gometalinter release
 fi
 command -v mdl &>/dev/null || gem install mdl --no-ri --no-rdoc || true	# for linting markdown files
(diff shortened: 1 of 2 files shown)


purpleidea:
@jonathangold Thanks for the patch!

While it looks good, this is so horrible (and maybe it was horrible before)

I wonder if there's a better way? Too bad there's no dpkg file in a repo :(

jonathangold:
I'll poke around and see if there's a better way, but I'm not holding my breath. ¯\_(ツ)_/¯

purpleidea:
The apt travis regression is now fixed in master... You can rebase if you like. Although not sure what the best thing to do here is. I don't like advocating a `curl | bash`, particularly when it's hidden inside of a `make deps` that we suggest... Hmmm...

purpleidea:
Here's one idea:

Would you be willing to modify your patch to wrap the install part inside an:

```
if env | grep -q -e '^TRAVIS=true$' -e '^JENKINS_URL=' -e '^BUILD_TAG=jenkins'; then
```
so that we only do these things on travis? (patch 1)

And if you're *really* in the mood we could modify the test case to fail with a helpful reason if gometalinter isn't installed. (patch 3)

That way we don't p0wn the user's laptop if the upstream puts bad code here.

(Ideally this commit could be patch 2, and make the change.)

jonathangold:
> (Ideally this commit could be patch 2, and make the change.)

Can you clarify what you mean. Wrap the existing install command with the env, check and then change the install command in a subsequent commit?

purpleidea:
@jonathangold Sorry, poorly explained on my part. It was a thought about how to break this down, with the actual change to `curl` being in a separate commit so that if we wanted to revert it, we could (since for whatever reason it's not "broken" yet)

jonathangold:
@purpleidea done. Hopefully I understood your comments. :P

misc/make-deps.sh
```diff
@@ -103,7 +103,9 @@ go get golang.org/x/tools/cmd/goyacc			# formerly `go tool yacc`
 go get golang.org/x/tools/cmd/stringer			# for automatic stringer-ing
 go get golang.org/x/lint/golint				# for `golint`-ing
 go get github.com/tmthrgd/go-bindata/go-bindata	# for compiling in non golang files
-go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
+if env | grep -q -e '^TRAVIS=true$' -e '^JENKINS_URL=' -e '^BUILD_TAG=jenkins'; then
+    go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
```
purpleidea: Commit1: I think this is indented with spaces instead of tabs.

Otherwise LGTM :)

purpleidea:
Looks great, just missing the spaces -> tabs fix in the first two commits so that tests pass :)

misc/make-deps.sh
jonathangold: It is. I'm not sure whats wrong with my vimrc... I'll investigate later and ping you when it's fixed.

jonathangold:
@purpleidea I think it's all sorted now. :)

purpleidea:
Patch LGTM, except: https://travis-ci.org/purpleidea/mgmt/jobs/479940643#L1391

"gometalinter not found"

purpleidea:
@jonathangold Yeah, if you look at the old version, I had to do some weird hackery to get it into the right place. But the base installer seemed to get it right, I just needed to rename it... So I don't know what's going on here. I can merge the other two patches, just not the middle one yet.

jonathangold:
@purpleidea Cool. I've got some other stuff to take care of, but I'll debug it later tonight or tomorrow.

purpleidea:
I cherry-picked the two patches that weren't part of the `--install` change into git master. Hopefully this makes it one less thing to worry about when rebasing :)

jonathangold:
@purpleidea I dunno. I made this PR, because the gometalinter install was failing on my dev machine, but the gometalinter project uses `--install` in their own .travis.yml file:

https://github.com/alecthomas/gometalinter/blob/102ac984005d45456a7e3ae6dc94ebcd95c2bb19/.travis.yml#L5

So, unless you want to move away from using the tip of master I think we can close this until it breaks (if it ever does.)

purpleidea:
On Wed, Jan 16, 2019 at 1:35 PM Jonathan Gold <notifications@github.com> wrote:
>
> @purpleidea I dunno. I made this PR, because the gometalinter install was failing on my dev machine, but the gometalinter project uses --install in their own .travis.yml file:
>
> https://github.com/alecthomas/gometalinter/blob/102ac984005d45456a7e3ae6dc94ebcd95c2bb19/.travis.yml#L5
>
> So, unless you want to move away from using the tip of master I think we can close this until it breaks (if it ever does.)

Weird! Well the other two merged commits are useful anyways, so glad
we merged those!

Thanks!

PR #437, opened 2019-01-13
add cpucount fact

keur:
libify socketset api and add cpucount fact.

### DIFF
--- a/engine/resources/net.go
+++ b/engine/resources/net.go
@@ -191,15 +191,19 @@ func (obj *NetRes) Close() error {
 // TODO: currently gets events from ALL interfaces, would be nice to reject
 // events from other interfaces.
 func (obj *NetRes) Watch() error {
-	// waitgroup for netlink receive goroutine
-	wg := &sync.WaitGroup{}
-	defer wg.Wait()
-
 	// create a netlink socket for receiving network interface events
 	conn, err := socketset.NewSocketSet(rtmGrps, obj.socketFile, unix.NETLINK_ROUTE)
 	if err != nil {
 		return errwrap.Wrapf(err, "error creating socket set")
 	}
+
+	// waitgroup for netlink receive goroutine
+	wg := &sync.WaitGroup{}
+	defer conn.Close()
+	// We must wait for the Shutdown() AND the select inside of SocketSet to
+	// complete before we Close, since the unblocking in SocketSet is not a
+	// synchronous operation.
+	defer wg.Wait()
 	defer conn.Shutdown() // close the netlink socket and unblock conn.receive()
 
 	// watch the systemd-networkd configuration file
@@ -220,7 +224,6 @@ func (obj *NetRes) Watch() error {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		defer conn.Close() // close the pipe when we're done with it
 		defer close(nlChan)
 		for {
 			// receive messages from the socket set
--- a/engine/resources/net.go
+++ b/engine/resources/net.go
@@ -34,6 +34,7 @@ import (
 	"github.com/purpleidea/mgmt/engine/traits"
 	"github.com/purpleidea/mgmt/recwatch"
 	"github.com/purpleidea/mgmt/util"
+	"github.com/purpleidea/mgmt/util/socketset"
 
 	multierr "github.com/hashicorp/go-multierror"
 	errwrap "github.com/pkg/errors"
@@ -41,7 +42,6 @@ import (
 	// do not clean up spawned goroutines. Should be replaced when a suitable
 	// alternative is available.
 	"github.com/vishvananda/netlink"
-	"golang.org/x/sys/unix"
 )
 
 func init() {
@@ -195,11 +195,11 @@ func (obj *NetRes) Watch() error {
 	defer wg.Wait()
 
 	// create a netlink socket for receiving network interface events
-	conn, err := newSocketSet(rtmGrps, obj.socketFile)
+	conn, err := socketset.NewSocketSet(rtmGrps, obj.socketFile)
 	if err != nil {
 		return errwrap.Wrapf(err, "error creating socket set")
 	}
-	defer conn.shutdown() // close the netlink socket and unblock conn.receive()
+	defer conn.Shutdown() // close the netlink socket and unblock conn.receive()
 
 	// watch the systemd-networkd configuration file
 	recWatcher, err := recwatch.NewRecWatcher(obj.unitFilePath, false)
@@ -219,11 +219,11 @@ func (obj *NetRes) Watch() error {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		defer conn.close() // close the pipe when we're done with it
+		defer conn.Close() // close the pipe when we're done with it
 		defer close(nlChan)
 		for {
 			// receive messages from the socket set
-			msgs, err := conn.receive()
+			msgs, err := conn.Receive()
 			if err != nil {
 				select {
 				case nlChan <- &nlChanStruct{
@@ -767,123 +767,4 @@ func (obj *iface) addrApplyAdd(objAddrs []string) error {
 		}
 	}
 	return nil
-}
-
-// socketSet is used to receive events from a socket and shut it down cleanly
-// when asked. It contains a socket for events and a pipe socket to unblock
-// receive on shutdown.
-type socketSet struct {
-	fdEvents int
-	fdPipe   int
-	pipeFile string
-}
-
-// newSocketSet returns a socketSet, initialized with the given parameters.
-func newSocketSet(groups uint32, file string) (*socketSet, error) {
-	// make a netlink socket file descriptor
-	fdEvents, err := unix.Socket(unix.AF_NETLINK, unix.SOCK_RAW, unix.NETLINK_ROUTE)
-	if err != nil {
-		return nil, errwrap.Wrapf(err, "error creating netlink socket")
-	}
-	// bind to the socket and add add the netlink groups we need to get events
-	if err := unix.Bind(fdEvents, &unix.SockaddrNetlink{
-		Family: unix.AF_NETLINK,
-		Groups: groups,
-	}); err != nil {
-		return nil, errwrap.Wrapf(err, "error binding netlink socket")
-	}
-
-	// create a pipe socket to unblock unix.Select when we close
-	fdPipe, err := unix.Socket(unix.AF_UNIX, unix.SOCK_RAW, unix.PROT_NONE)
-	if err != nil {
-		return nil, errwrap.Wrapf(err, "error creating pipe socket")
-	}
-	// bind the pipe to a file
-	if err = unix.Bind(fdPipe, &unix.SockaddrUnix{
-		Name: file,
-	}); err != nil {
-		return nil, errwrap.Wrapf(err, "error binding pipe socket")
-	}
-	return &socketSet{
-		fdEvents: fdEvents,
-		fdPipe:   fdPipe,
-		pipeFile: file,
-	}, nil
-}
-
-// shutdown closes the event file descriptor and unblocks receive by sending
-// a message to the pipe file descriptor. It must be called before close, and
-// should only be called once.
-func (obj *socketSet) shutdown() error {
-	// close the event socket so no more events are produced
-	if err := unix.Close(obj.fdEvents); err != nil {
-		return err
-	}
-	// send a message to the pipe to unblock select
-	return unix.Sendto(obj.fdPipe, nil, 0, &unix.SockaddrUnix{
-		Name: path.Join(obj.pipeFile),
-	})
-}
-
-// close closes the pipe file descriptor. It must only be called after
-// shutdown has closed fdEvents, and unblocked receive. It should only be
-// called once.
-func (obj *socketSet) close() error {
-	return unix.Close(obj.fdPipe)
-}
-
-// receive waits for bytes from fdEvents and parses them into a slice of
-// netlink messages. It will block until an event is produced, or shutdown
-// is called.
-func (obj *socketSet) receive() ([]syscall.NetlinkMessage, error) {
-	// Select will return when any fd in fdSet (fdEvents and fdPipe) is ready
-	// to read.
-	_, err := unix.Select(obj.nfd(), obj.fdSet(), nil, nil, nil)
-	if err != nil {
-		// if a system interrupt is caught
-		if err == unix.EINTR { // signal interrupt
-			return nil, nil
-		}
-		return nil, errwrap.Wrapf(err, "error selecting on fd")
-	}
-	// receive the message from the netlink socket into b
-	b := make([]byte, os.Getpagesize())
-	n, _, err := unix.Recvfrom(obj.fdEvents, b, unix.MSG_DONTWAIT) // non-blocking receive
-	if err != nil {
-		// if fdEvents is closed
-		if err == unix.EBADF { // bad file descriptor
-			return nil, nil
-		}
-		return nil, errwrap.Wrapf(err, "error receiving messages")
-	}
-	// if we didn't get enough bytes for a header, something went wrong
-	if n < unix.NLMSG_HDRLEN {
-		return nil, fmt.Errorf("received short header")
-	}
-	b = b[:n] // truncate b to message length
-	// use syscall to parse, as func does not exist in x/sys/unix
-	return syscall.ParseNetlinkMessage(b)
-}
-
-// nfd returns one more than the highest fd value in the struct, for use as as
-// the nfds parameter in select. It represents the file descriptor set maximum
-// size. See man select for more info.
-func (obj *socketSet) nfd() int {
-	if obj.fdEvents > obj.fdPipe {
-		return obj.fdEvents + 1
-	}
-	return obj.fdPipe + 1
-}
-
-// fdSet returns a bitmask representation of the integer values of fdEvents
-// and fdPipe. See man select for more info.
-func (obj *socketSet) fdSet() *unix.FdSet {
-	fdSet := &unix.FdSet{}
-	// Generate the bitmask representing the file descriptors in the socketSet.
-	// The rightmost bit corresponds to file descriptor zero, and each bit to
-	// the left represents the next file descriptor number in the sequence of
-	// all real numbers. E.g. the FdSet containing containing 0 and 4 is 10001.
-	fdSet.Bits[obj.fdEvents/64] |= 1 << uint(obj.fdEvents)
-	fdSet.Bits[obj.fdPipe/64] |= 1 << uint(obj.fdPipe)
-	return fdSet
 }
--- a/engine/resources/net.go
+++ b/engine/resources/net.go
@@ -42,6 +42,7 @@ import (
 	// do not clean up spawned goroutines. Should be replaced when a suitable
 	// alternative is available.
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 )
 
 func init() {
@@ -195,7 +196,7 @@ func (obj *NetRes) Watch() error {
 	defer wg.Wait()
 
 	// create a netlink socket for receiving network interface events
-	conn, err := socketset.NewSocketSet(rtmGrps, obj.socketFile)
+	conn, err := socketset.NewSocketSet(rtmGrps, obj.socketFile, unix.NETLINK_ROUTE)
 	if err != nil {
 		return errwrap.Wrapf(err, "error creating socket set")
 	}
@@ -223,7 +224,7 @@ func (obj *NetRes) Watch() error {
 		defer close(nlChan)
 		for {
 			// receive messages from the socket set
-			msgs, err := conn.Receive()
+			msgs, err := conn.ReceiveNetlinkMessages()
 			if err != nil {
 				select {
 				case nlChan <- &nlChanStruct{
--- a/lang/funcs/core/coresys/cpucount_fact.go
+++ b/lang/funcs/core/coresys/cpucount_fact.go
@@ -76,9 +76,12 @@ func (obj CPUCountFact) Stream() error {
 
 	// waitgroup for netlink receive goroutine
 	wg := &sync.WaitGroup{}
-	defer wg.Wait()
 	defer ss.Close()
-	defer ss.Shutdown()
+	// We must wait for the Shutdown() AND the select inside of SocketSet to
+	// complete before we Close, since the unblocking in SocketSet is not a
+	// synchronous operation.
+	defer wg.Wait()
+	defer ss.Shutdown() // close the netlink socket and unblock conn.receive()
 
 	eventChan := make(chan *nlChanEvent) // updated in goroutine when we receive uevent
 	closeChan := make(chan struct{})     // channel to unblock selects in goroutine
--- a/lang/funcs/core/coresys/cpucount_fact.go
+++ b/lang/funcs/core/coresys/cpucount_fact.go
@@ -1,0 +1,217 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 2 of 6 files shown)


engine/resources/net.go
```diff
@@ -37,6 +37,7 @@ import (
 
 	multierr "github.com/hashicorp/go-multierror"
 	errwrap "github.com/pkg/errors"
+	"github.com/purpleidea/mgmt/lib/socketset"
```
purpleidea: The lib package is meant for general core "mgmt as a lib" stuff (sorry i suck at naming) so how about we put socketset into: `github.com/purpleidea/mgmt/util/socketset` instead?

lib/socketset/socketset.go
```diff
@@ -0,0 +1,218 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Package socketset is in API for creating a select style netlink
+// socket to recieve events from the kernel.
```
purpleidea: receive (typo)

lib/socketset/socketset.go
```diff
@@ -0,0 +1,218 @@
...
+	"os"
+	"path"
+	"strings"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+
+	errwrap "github.com/pkg/errors"
+)
+
```
purpleidea: It's best to re-order the entries in this file. Usually, first we see the `type SocketSet struct`, then it's `New` constructor, and then the methods on that struct in the approximate order they're used in.

And then the dependency struct we use underneath it in the same order.

lib/socketset/socketset.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// use syscall to parse, as func does not exist in x/sys/unix
+	return syscall.ParseNetlinkMessage(msgBytes)
+}
+
+// UEvent struct which has attributes passed from a KOBJECT_NETWORK_UEVENT
+type UEvent struct {
+	// Default keys, as per https://github.com/torvalds/linux/blob/master/lib/kobject_uevent.c
```
purpleidea: nit: short one line comments aren't capitalized, since they aren't sentences.

lib/socketset/socketset.go
```diff
@@ -0,0 +1,218 @@
...
+	Action    string
+	Devpath   string
+	Subsystem string
+
+	// Every other KV pair
+	Data map[string]string
+}
+
+// ReceiveUEvent is a wrapper around ReceiveBytes that returns a UEvent
```
purpleidea: Doc strings and longer comments are full sentences and need to end with a period.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+
+import (
+	"io/ioutil"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/funcs/facts"
+	"github.com/purpleidea/mgmt/lang/types"
+
```
purpleidea: imports go in three groups

first: core imports (this is correct here)
second: internal purpleidea/mgmt imports (squash these together)
third: external imports from other url's

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+	errwrap "github.com/pkg/errors"
+)
+
+const (
+	rtmGrps         = 0x1 // make me a multicast reciever
+	socketFile      = "pipe.sock"
+	cpuDevpathRegex = "/devices/system/cpu/cpu[0-9]"
+)
+
+// CPUCountFact is a fact that returns the current CPU count
```
purpleidea: end with a period... other places as well please.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+type CPUCountFact struct {
+	init      *facts.Init
+	closeChan chan struct{}
+}
+
+func init() {
```
purpleidea: This should be higher up at the top like in the other files. It's so that if a programmer has a quick look, they won't be surprised by code running by itself.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+
+// Init run startup code for this fact. Initializes the closeChan
+// and sets the facts.Init variable
+func (obj *CPUCountFact) Init(init *facts.Init) error {
+	obj.init = init
+	obj.closeChan = make(chan struct{})
+	return nil
+}
+
+type nlChanEvent struct {
```
purpleidea: Order this at the bottom as described in the other review please. You can give it a docstring comment too if you like.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+			count += hi - lo + 1
+		}
+	}
+	return count, nil
+}
+
+// When we receieve a udev event, we filter only those that
+// indicate a CPU is being added or removed.
+// TODO: instead of checking for added/removed CPUS, check for online/offline?
```
purpleidea: Yeah, so initially a cpu is added, and then it's made "online". That's when it becomes useful. So this is probably what we should use instead.

Alternatively, we could instead return a "big struct" of data about the cpus if you think that's useful. To be considered. Maybe better for a future patch.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+func processUdev(event *socketset.UEvent) (int64, bool) {
+	if event.Subsystem != "cpu" {
+		return 0, false
+	}
+	// is this a valid cpu path in sysfs?
+	m, err := regexp.MatchString(cpuDevpathRegex, event.Devpath)
+	if !m || err != nil {
+		// TODO: log error?
```
purpleidea: You can log with obj.init.Logf and if it's better for debugging guard with obj.init.Debug flag.

lang/funcs/core/coresys/cpucount_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// Set up the go function
+	go func() {
+		defer fact.Close()
+		count := 0
+	Loop:
+		for {
+			select {
+			case cpus := <-output:
+				fmt.Printf("CPUS: %d\n", cpus.Int())
```
purpleidea: In tests you'll want `t.Logf` instead of `fmt.Printf`

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+	eventChan := make(chan *nlChanEvent) // updated when we receive uevent
+	defer close(eventChan)
+
+	// Start waiting for kernel to poke us about new
+	// device changes on the system
+	go func() {
+		defer ss.Shutdown()
+		defer ss.Close()
+		for {
+			uevent, err := ss.ReceiveUEvent()
```
purpleidea: Does this block ? How do we unblock it to close down?

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+			default: // don't block
+			}
+		}
+	}()
+
+	startChan := make(chan struct{})
+	close(startChan) // trigger the first event
+	var cpuCount int64
+	select {
+	case <-startChan:
```
purpleidea: structurally you might want this _inside_ the main `for` loop. it will simplify your code substantially I think.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,201 @@
...
+				continue
+			}
+			if event.err != nil {
+				// TODO: log here instead?
+				return errwrap.Wrapf(event.err, "error receiving uevent")
+			}
+			//fmt.Printf("Receiving %s\n", event.Data["SEQNUM"])
+			cpus, cpuEvent := processUdev(event.uevent)
+			if cpuEvent {
+				cpuCount += cpus
```
purpleidea: So it seems you're keeping a running delta AIUI. It might be better to have a function (you already have it) that gets the exact data we want, and then run that every time we get a valid event. Even though the event might produce useful data, we don't necessarily need to use it, unless that sort of information doesn't come from elsewhere.

That makes this more declarative and usually accurate if many events come in all of a sudden or if we loose one somehow and it messes up the count.

keur: yeah, this is a really good idea. thanks.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Start waiting for kernel to poke us about new
+	// device changes on the system
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		defer close(eventChan)
+		defer ss.Close()
+		for {
+			uevent, err := ss.ReceiveUEvent()
```
purpleidea: Please add a comment above this line mentioning what method is called that causes this to unblock. That's super useful information when debugging deadlock/etc issues :)

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/purpleidea/mgmt/lang/funcs/facts"
+	"github.com/purpleidea/mgmt/lang/types"
+	"github.com/purpleidea/mgmt/util/socketset"
+
+	errwrap "github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	rtmGrps         = 0x1 // make me a multicast reciever
```
purpleidea: receiver*

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Info returns static typing info about what the fact returns.
+func (obj *CPUCountFact) Info() *facts.Info {
+	return &facts.Info{
+		Output: types.NewType("int"),
+	}
+}
+
+// Init run startup code for this fact. Initializes the closeChan
```
purpleidea: doc strings should be wrapped to 80 chars please. (Here and elsewhere.)

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+
+// Stream returns the changing values that this fact has over time.
+// It will first poll sysfs to get the initial cpu count, and then
+// receieves UEvents from the kernel as CPUs are added/removed.
+func (obj CPUCountFact) Stream() error {
+	// waitgroup for netlink recieve goroutine
+	wg := &sync.WaitGroup{}
+	defer wg.Wait()
+
+	defer close(obj.init.Output) // signal when we're done
```
purpleidea: This should happen last. IOW should be at the top of this function. We only want to tell the func engine that we're done, when we're done (after the waitgroup Wait returns)

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+	closeChan := make(chan struct{})     // channel to unblock selects in goroutine
+	defer close(closeChan)
+
+	// Start waiting for kernel to poke us about new
+	// device changes on the system
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		defer close(eventChan)
+		defer ss.Close()
```
purpleidea: either this or shutdown() are in the wrong place or you need a waitgroup etc... As per the socketSet docs, Shutdown() which does the unblocking has to happen _before_ close(). So if we start closing, closeChan could close, and then we race to figure out if shutdown or close happens first!

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+			if err != nil {
+				return 0, err
+			}
+			count += hi - lo + 1
+		}
+	}
+	return count, nil
+}
+
+// When we receieve a udev event, we filter only those that
```
purpleidea: typo

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+// indicate a CPU is being added or removed, or being taken
+// online or offline
```
purpleidea: sentence...

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,210 @@
...
+func isCPUEvent(event *socketset.UEvent) bool {
+	if event.Subsystem != "cpu" {
+		return false
+	}
+	// is this a valid cpu path in sysfs?
+	m, err := regexp.MatchString(cpuDevpathRegex, event.Devpath)
+	if !m || err != nil {
+		return false
+	}
+	if event.Action == "add" ||
```
purpleidea: You can put these all on the same line...

lang/funcs/core/coresys/cpucount_test.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var fact *CPUCountFact
+
+func TestSimple(t *testing.T) {
+	fact = &CPUCountFact{}
+
+	output := make(chan types.Value)
+	err := fact.Init(&facts.Init{
+		Hostname: "mgmt",
```
purpleidea: Just leave this out

lang/funcs/core/coresys/cpucount_test.go
```diff
@@ -0,0 +1,66 @@
...
+		Output:   output,
+		Logf:     t.Logf,
```
purpleidea: Wrap the log for tests, eg:
```
		Logf: func(format string, v ...interface{}) {
			t.Logf("test: "+format, v...)
		},
```

lang/funcs/core/coresys/cpucount_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	go func() {
+		defer fact.Close()
+		count := 0
+	Loop:
+		for {
+			select {
+			case cpus := <-output:
+				t.Logf("CPUS: %d\n", cpus.Int())
+				count++
```
purpleidea: What's the logic here?

lang/funcs/core/coresys/cpucount_test.go
```diff
@@ -0,0 +1,66 @@
...
+				// reading from proc
+				if count > 0 {
+					break Loop
+				}
+			}
+		}
+	}()
+
+	// Now start the stream
+	err = fact.Stream()
```
purpleidea: You can combine, eg: `if err := ...`

util/socketset/socketset.go
```diff
@@ -0,0 +1,219 @@
...
+func NewSocketSet(groups uint32, name string, proto int) (*SocketSet, error) {
+	fdEvents, err := unix.Socket(unix.AF_NETLINK, unix.SOCK_RAW, proto)
+	if err != nil {
+		return nil, errwrap.Wrapf(err, "error creating netlink socket")
+	}
+
+	if err := unix.Bind(fdEvents, &unix.SockaddrNetlink{
+		Family: unix.AF_NETLINK,
+		Groups: groups,
+		Pid:    uint32(os.Getpid()),
```
purpleidea: This line and other changes (eg: changed comments, etc) are fine, but should be in a separate commit. So it might be best to have commit (1): move the socketSet without making any changes except the pub/priv stuff (2): edit the socketSet in-place with the changes you wanted to make to it. (3) use the new socket set in commit three with our fact.

Additionally this makes it easier to review because the diff of changes will be separate from the move.

lang/funcs/core/coresys/cpucount_test.go
```diff
@@ -0,0 +1,66 @@
+// Mgmt
```
purpleidea: This file name should be _test.go added to the one we're testing. So we want it to be named cpucount_fact_test.go instead.

purpleidea:
https://travis-ci.org/purpleidea/mgmt/jobs/481207599#L1408

purpleidea:
I think your cpu test deadlocked: https://travis-ci.org/purpleidea/mgmt/jobs/481989442#L1146

util/socketset/socketset.go
```diff
@@ -15,87 +15,70 @@
...
+	fdEvents, err := unix.Socket(unix.AF_NETLINK, unix.SOCK_RAW, proto)
 	if err != nil {
 		return nil, errwrap.Wrapf(err, "error creating netlink socket")
 	}
-	// bind to the socket and add add the netlink groups we need to get events
+
 	if err := unix.Bind(fdEvents, &unix.SockaddrNetlink{
 		Family: unix.AF_NETLINK,
 		Groups: groups,
+		Pid:    uint32(os.Getpid()),
```
purpleidea: Out of curiosity, what's this for? (A small comment on that line wouldn't hurt)

util/socketset/socketset.go
```diff
@@ -15,87 +15,70 @@
...
 // NewSocketSet returns a socketSet, initialized with the given parameters.
-func NewSocketSet(groups uint32, file string) (*SocketSet, error) {
-	// make a netlink socket file descriptor
-	fdEvents, err := unix.Socket(unix.AF_NETLINK, unix.SOCK_RAW, unix.NETLINK_ROUTE)
+func NewSocketSet(groups uint32, name string, proto int) (*SocketSet, error) {
+	fdEvents, err := unix.Socket(unix.AF_NETLINK, unix.SOCK_RAW, proto)
 	if err != nil {
 		return nil, errwrap.Wrapf(err, "error creating netlink socket")
 	}
-	// bind to the socket and add add the netlink groups we need to get events
```
purpleidea: Was there a good reason to nuke this comment?

util/socketset/socketset.go
```diff
@@ -15,87 +15,70 @@
...
+
 	return &SocketSet{
 		fdEvents: fdEvents,
 		fdPipe:   fdPipe,
-		pipeFile: file,
+		pipeFile: name,
 	}, nil
 }
 
-// Shutdown closes the event file descriptor and unblocks receive by sending
```
purpleidea: Just curious: why this big delta/move in the code?

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// Info returns static typing info about what the fact returns.
+func (obj *CPUCountFact) Info() *facts.Info {
+	return &facts.Info{
+		Output: types.NewType("int"),
+	}
+}
+
+// Init run startup code for this fact. Initializes the closeChan and sets the
```
purpleidea: s/run/runs/

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,218 @@
...
+// facts.Init variable.
+func (obj *CPUCountFact) Init(init *facts.Init) error {
+	obj.init = init
+	obj.closeChan = make(chan struct{})
+	return nil
+}
+
+// Stream returns the changing values that this fact has over time. It will
+// first poll sysfs to get the initial cpu count, and then receieves UEvents
```
purpleidea: s/receieves/receives/

lang/funcs/core/coresys/cpucount_fact_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}()
+
+	// Now start the stream
+	err = fact.Stream()
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+var cpulistTests = []struct {
```
purpleidea: Awesome to have such a test! This var can be inside of the test though.
(Eg: inside the scope of TestParseCPUList)

lang/funcs/core/coresys/cpucount_fact_test.go
```diff
@@ -0,0 +1,104 @@
...
+			select {
+			case cpus := <-output:
+				t.Logf("CPUS: %d\n", cpus.Int())
+				break Loop
+			}
+		}
+	}()
+
+	// Now start the stream
+	err = fact.Stream()
```
purpleidea: you can `if err := fact.Stream()`

lang/funcs/core/coresys/cpucount_fact_test.go
```diff
@@ -0,0 +1,104 @@
...
+		for {
+			select {
+			case cpus := <-output:
+				t.Logf("CPUS: %d\n", cpus.Int())
+				break Loop
+			}
+		}
+	}()
+
+	// Now start the stream
```
purpleidea: nit: full sentence or lower case "Now" please

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,218 @@
...
+	startChan := make(chan struct{})
+	close(startChan) // trigger the first event
+	var cpuCount, newCount int64 = 0, -1
+	for {
+		select {
+		case <-startChan:
+			startChan = nil // disable
+			newCount, err = getCPUCount()
+			if err != nil {
+				obj.init.Logf("could not get initial CPU count. Setting to zero.")
```
purpleidea: nit: Is this a sentence or not a sentence? Starts with lowercase `could`.

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+		case <-obj.closeChan:
+			return nil
+		}
+
+		if newCount == cpuCount {
+			continue
+		}
+		cpuCount = newCount
+		st := &types.IntValue{
```
purpleidea: no need to use a var, just put it in directly... and this var name was copied from when it was a struct ;)

lang/funcs/core/coresys/cpucount_fact.go
```diff
@@ -0,0 +1,218 @@
...
+	wg := &sync.WaitGroup{}
+	defer wg.Wait()
+	defer ss.Close()
+	defer ss.Shutdown()
+
+	eventChan := make(chan *nlChanEvent) // updated in goroutine when we receive uevent
+	closeChan := make(chan struct{})     // channel to unblock selects in goroutine
+	defer close(closeChan)
+
+	// Start waiting for kernel to poke us about new device changes on the system
```
purpleidea: nit: lowercase `start` or end with a period.

util/socketset/socketset_test.go
```diff
@@ -104,3 +106,31 @@ func TestNfd(t *testing.T) {
...
+
+// test SocketSet.Shutdown()
+func TestShutdown(t *testing.T) {
+	wg := &sync.WaitGroup{}
+	defer wg.Wait()
+
+	// pass 0 so we create a socket that doesn't receive any events
+	ss, err := NewSocketSet(0, "pipe.sock", 0)
+	if err != nil {
+		t.Errorf("Could not create SocketSet %+v", err)
```
purpleidea: nit: `could` should be lowercase

lang/funcs/core/coresys/cpucount_fact_test.go
```diff
@@ -0,0 +1,103 @@
...
+package coresys
+
+import (
+	"testing"
+
+	"github.com/purpleidea/mgmt/lang/funcs/facts"
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+var fact *CPUCountFact
```
purpleidea: Why is this out here?

util/socketset/socketset.go
```diff
@@ -15,87 +15,71 @@
...
-// a message to the pipe file descriptor. It must be called before close, and
```
purpleidea: See above please, just curious mostly.

util/socketset/socketset_test.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+}
+
+// test SocketSet.Shutdown()
+func TestShutdown(t *testing.T) {
+	// pass 0 so we create a socket that doesn't receive any events
+	ss, err := NewSocketSet(0, "pipe.sock", 0)
+	if err != nil {
+		t.Errorf("could not create SocketSet %+v", err)
```
purpleidea: Super nit: add a colon to the error message like the other one, eg:

`could not create SocketSet: %+v`

util/socketset/socketset.go
keur: Oh, I think this is here because we split the commit into one that just moves + exports the code. Then I split us the methods which pushed this down causing the delta.

purpleidea: Okay cool :)

purpleidea:
LGTM :)

purpleidea:
Well this was fantastic and a lot of fun, thanks!

One quick side question: I noticed we ended up removing the time.Sleep at the end of the `TestShutdown`... Any reason for that? Should we add it back?

Cheers!

PR #454, opened 2019-02-05
langpuppet: Fix Cli method invocations for wrapped GAPIs

ffrank:
Since the langpuppet GAPI creates fresh new CliContext objects,
it has to make sure to provide the original parent context, because
the child GAPIs expect to be able to access its data.


purpleidea:
Awesome, LGTM, Merged =D

Did you want to add a test for this too? ;)

Sorry I broke this in the refactor... I forgot to catch these I guess.

PR #455, opened 2019-02-06
vendor: Add missing purpleidea/go-systemd submodule

roidelapluie:


purpleidea:
I don't think this is needed, because git master of this module is still working AIUI.

PR #458, opened 2019-02-06
make: Use gem --no-document instead of deprecated flags.

landervdb:
Signed-off-by: Lander Van den Bulcke <landervdb@inuits.eu>

The flags `--no-ri` and `--no-rdoc` for `gem install` have been deprecated since version 2.0.0 and have actually been removed in 3.x.x. `--no-document` should be used instead.


purpleidea:
LGTM :)

purpleidea:
Merged, thanks!

PR #460, opened 2019-02-06
lang: Add more string functions, autogenerated

roidelapluie:


Makefile
```diff
@@ -408,4 +410,6 @@ help: ## show this help screen
 		awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
 	@echo ''
 
+funcgen:
```
purpleidea: This target is "PHONY" instead of being a proper make target that is named for the output file and has input file dependencies... That way it will correctly re-run only when necessary.

lang/golang2mgmt/config.go
```diff
@@ -0,0 +1,53 @@
...
+
+// ToMcl prints the arg signature as expected by mcl.
+func (obj *arg) ToMcl() string {
+	if obj.Type == "string" {
+		if obj.Name != "" {
+			return fmt.Sprintf("%s str", obj.Name)
+		}
+		return "str"
+	}
+	log.Fatalf("Can not convert %v to mcl", obj)
```
purpleidea: (Here and elsewhere)

I have to be annoying and enforce cleaner code, even if it's in this small util program-- Please have proper function signatures like `(string, error)`, and propagate the errors upwards... Only at the top-level do we run log.Whatever()

Also errors start with a lower case.

You can use `errwrap.Wrapf` to wrap errors upwards if needed.

Lastly it's "cannot" :)

lang/golang2mgmt/func.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
+	finalName := strings.TrimSuffix(basename, ".tpl")
+	finalPath := filepath.Join(path, finalName)
+	log.Printf("Writing %s\n", finalPath)
+	finalFile, err := os.Create(finalPath)
+	if err != nil {
+		return err
+	}
+	err = t.Execute(finalFile, c)
```
purpleidea: `
if err = whatever(); err != nil {

`

lang/golang2mgmt/func.go
```diff
@@ -0,0 +1,149 @@
...
+	MgmtPackage string
+}
+
+func parseFuncs(c config, path, templates string) error {
+	templateFiles, err := filepath.Glob(templates)
+	if err != nil {
+		return err
+	}
+	for _, tpl := range templateFiles {
+		log.Printf("Generating %s\n", tpl)
```
purpleidea: If the methods aren't on an `obj` which contains a `Logf` method, then you should use the `logf` that you pass down from main. (Here and elsewhere.)

purpleidea:
As an aside, I should have said publicly: this is a great idea, thanks! If you're really keen, we could even use reflection to generate the type signatures. Of course the means our functions need to look exactly like golang's. What do we think about this?

PS: Can this handle lists of strings, etc?

roidelapluie:
Do you want that autogenerated code or we run reflect while loading the language? maybe it is not needed to generate code then.

also, we could use a special prefix:

import "go_strings"

go_string.split("foo")

purpleidea:
@roidelapluie I think it's okay to have it as a generation step. Also, using reflection to call the functions will be much more expensive if it's possible, and very difficult. The current design is a good initial approach. :) We can always break things later if needed, since we're pre-0.1.

roidelapluie:
pushed

Makefile
```diff
@@ -408,4 +410,13 @@ help: ## show this help screen
 		awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
 	@echo ''
 
+funcgen: lang/funcs/core/generated_funcs_tests.go lang/funcs/core/generated_funcs.go
```
purpleidea: Is this the correct Makefile magic? (With the duplicate `funcgen` target?)
I was reading https://www.gnu.org/software/automake/manual/html_node/Multiple-Outputs.html
Does makefile allow the *.go in the input target?

lang/golang2mgmt/func.go
purpleidea: Doing this would still be nice :)

roidelapluie:
@ 
Using: gometalinter --disable-all --enable=goimports --enable=golint --enable=misspell --exclude=lang/lexer.nn.go --exclude=lang/y.go --exclude=bindata/bindata.go


Is it because the file is not added to git?

purpleidea:
@roidelapluie Hmmm Not sure (Like the new tests that run really fast?)
I can check in about one hour from now.

purpleidea:
Confirmed. It's a gofmt issue that you need to fix in the template. You generate some extra newlines in both, and in the test some of the code is missing a space. Do a diff and you'll see.

roidelapluie:
green! \o/

Makefile
roidelapluie: fixed

purpleidea:
You committed the generated files (you should not) and they now format correctly, except the generated_funcs.go has an extra newline at the end.

roidelapluie:
tests are shaky :(

purpleidea:
@roidelapluie I have a feature branch which should hopefully make the tests stable going forward. Just finishing it up as we speak. Sorry about that.

purpleidea:
I want to respectfully state that there are some pretty absurd things in this patch, but I figure if I don't merge it, I'm going to abandon it.

If you'd like to fixup these sorts of hacks, that would be awesome:

https://github.com/purpleidea/mgmt/blob/a0df4829a86d394b4b7efc729a5983693930228d/lang/golang2mgmt/templates/generated_funcs.go.tpl#L37

purpleidea:
(Also the Makefile does not correctly work for the _test.go file.)

PR #466, opened 2019-02-07
Make macOS build work, fix make-deps

aequitas:
This PR restores build for macOS and fixes make-deps for Ubuntu.

### DIFF
--- a/engine/resources/mount_linux_test.go
+++ b/engine/resources/mount_linux_test.go
@@ -1,0 +1,76 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// +build !root !darwin
+
+package resources
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	fstab "github.com/deniswernert/go-fstab"
+)
+
+func TestMountExists(t *testing.T) {
+	const procMock1 = `/tmp/mount0 /mnt/proctest ext4 rw,seclabel,relatime,data=ordered 0 0` + "\n"
+
+	var mountExistsTests = []struct {
+		procMock []byte
+		in       *fstab.Mount
+		out      bool
+	}{
+		{
+			[]byte(procMock1),
+			&fstab.Mount{
+				Spec:    "/tmp/mount0",
+				File:    "/mnt/proctest",
+				VfsType: "ext4",
+				MntOps:  map[string]string{"defaults": ""},
+				Freq:    1,
+				PassNo:  1,
+			},
+			true,
+		},
+	}
+
+	file, err := ioutil.TempFile("", "proc")
+	if err != nil {
+		t.Errorf("error creating temp file: %v", err)
+		return
+	}
+	defer os.Remove(file.Name())
+	for _, test := range mountExistsTests {
+		if err := ioutil.WriteFile(file.Name(), test.procMock, 0664); err != nil {
+			t.Errorf("error writing proc file: %s: %v", file.Name(), err)
+			return
+		}
+		if err := ioutil.WriteFile(test.in.Spec, []byte{}, 0664); err != nil {
+			t.Errorf("error writing fstab file: %s: %v", file.Name(), err)
+			return
+		}
+		result, err := mountExists(file.Name(), test.in)
+		if err != nil {
+			t.Errorf("error checking if fstab entry %s exists: %v", test.in.String(), err)
+			return
+		}
+		if result != test.out {
+			t.Errorf("mountExistsTests test wanted: %t, got: %t", test.out, result)
+		}
+	}
+}
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -102,6 +102,7 @@ go get github.com/blynn/nex				# for lexing
 go get golang.org/x/tools/cmd/goyacc			# formerly `go tool yacc`
 go get golang.org/x/tools/cmd/stringer			# for automatic stringer-ing
 go get golang.org/x/lint/golint				# for `golint`-ing
+go get golang.org/x/tools/cmd/goimports		# for fmt
 go get github.com/tmthrgd/go-bindata/go-bindata	# for compiling in non golang files
 if env | grep -q -e '^TRAVIS=true$' -e '^JENKINS_URL=' -e '^BUILD_TAG=jenkins'; then
 	go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
--- a/test/test-gotest.sh
+++ b/test/test-gotest.sh
@@ -9,20 +9,26 @@ cd "${ROOT}"
 
 # if we want to run this test as root, use build tag -root to ask each test...
 XSUDO=''
-XTAGS=''
+XTAGS=()
 if [[ "$@" = *"--root"* ]]; then
 	if ! timeout 1s sudo -A true; then
 		echo "sudo disabled: can't run as root"
 		exit 1
 	fi
 	XSUDO='sudo -E'
-	XTAGS='-tags root'
+	XTAGS+=('root')
+fi
+
+# As per https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/docker.md
+# Docker is not supported on Travis macOS test instances.
+if [[ "$TRAVIS_OS_NAME" == "osx" ]];then
+	XTAGS+=('nodocker')
 fi
 
 failures=''
 function run-test()
 {
-	$XSUDO $@ || failures=$( [ -n "$failures" ] && echo "$failures\\n$@" || echo "$@" )
+	$XSUDO $@ -tags="${XTAGS[*]}" || failures=$( [ -n "$failures" ] && echo "$failures\\n$@" || echo "$@" )
 }
 
 # NOTE: you can run `go test` with the -tags flag to skip certain tests, eg:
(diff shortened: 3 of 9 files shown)


misc/make-deps.sh
```diff
@@ -102,10 +102,11 @@ go get github.com/blynn/nex				# for lexing
 go get golang.org/x/tools/cmd/goyacc			# formerly `go tool yacc`
 go get golang.org/x/tools/cmd/stringer			# for automatic stringer-ing
 go get golang.org/x/lint/golint				# for `golint`-ing
+go get golang.org/x/tools/cmd/goimports		# for fmt
```
purpleidea: This is needed on OSX? (It works on Linux already)

misc/make-deps.sh
```diff
@@ -102,10 +102,11 @@ go get github.com/blynn/nex				# for lexing
...
 go get github.com/tmthrgd/go-bindata/go-bindata	# for compiling in non golang files
 if env | grep -q -e '^TRAVIS=true$' -e '^JENKINS_URL=' -e '^BUILD_TAG=jenkins'; then
 	go get -u gopkg.in/alecthomas/gometalinter.v1 && mv "$(dirname $(command -v gometalinter.v1))/gometalinter.v1" "$(dirname $(command -v gometalinter.v1))/gometalinter" && gometalinter --install	# bonus
 fi
-command -v mdl &>/dev/null || gem install mdl --no-document || true	# for linting markdown files
-command -v fpm &>/dev/null || gem install fpm --no-document || true	# for cross distro packaging
+command -v mdl &>/dev/null || $sudo_command gem install mdl --no-document || true	# for linting markdown files
+command -v fpm &>/dev/null || $sudo_command gem install fpm --no-document || true	# for cross distro packaging
```
purpleidea: Is root really necessary? For linux it seems it is not. If it is for OSX, can we wrap the sudo variants in an `if OSX` sort of thing please?

engine/resources/mount_linux_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"io/ioutil"
+	"os"
+	"testing"
+
+	fstab "github.com/deniswernert/go-fstab"
+)
+
+const procMock1 = `/tmp/mount0 /mnt/proctest ext4 rw,seclabel,relatime,data=ordered 0 0` + "\n"
+
+var mountExistsTests = []struct {
```
purpleidea: If we're moving these around, can we put them inside of the function. It's great to avoid globals if they're not needed in more than once place :)

engine/resources/net.go
```diff
@@ -38,6 +38,7 @@ import (
 
 	multierr "github.com/hashicorp/go-multierror"
 	errwrap "github.com/pkg/errors"
+
```
purpleidea: unrelated nl?

lib/cli.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/purpleidea/mgmt/bindata"
 	"github.com/purpleidea/mgmt/gapi"
+
```
purpleidea: unrelated nl?

test/test-gotest.sh
```diff
@@ -19,6 +19,12 @@ if [[ "$@" = *"--root"* ]]; then
 	XTAGS='-tags root'
 fi
 
+# as per https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/docker.md
```
purpleidea: nit: Please capitalize "as" since it's a sentence that ends with a period.

misc/make-deps.sh
aequitas: This is for Linux, this is what i get on Ubuntu: 
```
Fetching: mdl-0.5.0.gem (100%)
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /var/lib/gems/2.5.0 directory.
Fetching: fpm-1.11.0.gem (100%)
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /var/lib/gems/2.5.0 directory.
~/g/s/g/p/mgmt (master|✔) $ cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.1 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.1 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```
An alternative is to use `--user` but often this requires modification to `PATH` as most users don't have Ruby setup.

misc/make-deps.sh
aequitas: This if needed for Linux as well. At least on a fresh Ubuntu install I have it was.

misc/make-deps.sh
purpleidea: Interesting! Let's to the --user option if that will work, since the gem stuff is only used for package stuff and making docs, so most users won't need it, and causing less root stuff on their machines is probably a good thing.

engine/resources/net.go
aequitas: `goimports` complains about this otherwise. `goimports -w` added this change and the other one. Both Linux and macOS have this behavior for me. Maybe the version on CI is behind? Afaict CI doesn't complain about missing these newlines.

misc/make-deps.sh
purpleidea: `gofmt` is installed as part as a regular installation for me:

```
$ sudo dnf provides `which gofmt`
gcc-go-8.2.1-6.fc28.x86_64 : Go support
Repo        : updates
Matched from:
Filename    : /usr/bin/gofmt

golang-bin-1.10.7-1.fc28.x86_64 : Golang core compiler tools
Repo        : updates
Matched from:
Filename    : /usr/bin/gofmt

gcc-go-8.0.1-0.20.fc28.x86_64 : Go support
Repo        : fedora
Matched from:
Filename    : /usr/bin/gofmt

golang-bin-1.10.1-1.fc28.x86_64 : Golang core compiler tools
Repo        : fedora
Matched from:
Filename    : /usr/bin/gofmt

```

Is ubuntu missing this? Are you using `gofmt` ?

engine/resources/net.go
purpleidea: Weird... They break `gofmt`/`goimports` formatting from time to time apparently. Let's leave this as is until our tests start failing please. It didn't happen on my machine either, so maybe OSX on your machine has a different interpretation??

misc/make-deps.sh
aequitas: Agree. The commands don't fail the dep install anyways, but can be a red herring for new (or recurring) users.

misc/make-deps.sh
aequitas:
```
~/g/s/g/p/mgmt (macos|✔) $ dpkg -L golang-1.10-go|grep -E 'goimports|gofmt'
/usr/lib/go-1.10/bin/gofmt
```
`gofmt` is but `goimports` isn't and that is what is used in `test/test-gofmt.sh`

engine/resources/net.go
aequitas:
```
~/g/s/g/p/mgmt (master|✔) $ test/test-gofmt.sh 
running test/test-gofmt.sh
FAIL: The following golang files are not properly formatted (goimports -l): engine/resources/net.go
lib/cli.go
~/g/s/g/p/mgmt (master|✔) $ go version 
go version go1.10.4 linux/amd64
```

This is on Linux. But I'll revert the newlines, np.

purpleidea:
```
$ ./test/test-gofmt.sh 
running ./test/test-gofmt.sh
PASS
$ go version
go version go1.10.7 linux/amd64
```

Is it possible your slightly older version has a different way of doing things? This is very strange.

test/test-gotest.sh
```diff
@@ -19,6 +19,12 @@ if [[ "$@" = *"--root"* ]]; then
 	XTAGS='-tags root'
 fi
 
+# As per https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/docker.md
+# Docker is not supported on Travis macOS test instances.
+if [[ "$TRAVIS_OS_NAME" == "osx" ]];then
+	XTAGS="-tags nodocker $XTAGS"
```
aequitas: This breaks if the conditional above is true because only one `-tags` argument may be specified. I'll fix this.

aequitas:
It's possible the osx build on Travis fails due to a timeout like: https://travis-ci.org/aequitas/mgmt/jobs/489979766#L1962

I think this is due to a bug in Travis as the last test hasn't been hanging for 10 minutes, only 3 minutes or so. Someone else seems to hit this as well: https://travis-ci.community/t/no-output-has-been-received-in-the-last-10m0s/2169

purpleidea:
@aequitas I don't know anything about the OSX env, sorry, but maybe the machines are slower and as a result the lang test causes a timeout? (It's the biggest test, but doesn't take even close to 10 min on linux.)

As an aside, you might notice some intermittent test failures in the rest of the ecosystem. They're only in the `.sh` tests and occasionally the integration tests, since they've uncovered a bug in the `engine` core. I've got a rewrite to fix the broken parts there.

Sorry for the noise.

test/test-gotest.sh
```diff
@@ -16,7 +16,13 @@ if [[ "$@" = *"--root"* ]]; then
...
 	fi
 	XSUDO='sudo -E'
-	XTAGS='-tags root'
+	TAGS='root'
+fi
+
+# As per https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/docker.md
+# Docker is not supported on Travis macOS test instances.
+if [[ "$TRAVIS_OS_NAME" == "osx" ]];then
+	TAGS="$XTAGS nodocker"
```
purpleidea: Don't you mean $TAGS ? (Inside)

test/test-gotest.sh
```diff
@@ -25,6 +31,10 @@ function run-test()
 	$XSUDO $@ || failures=$( [ -n "$failures" ] && echo "$failures\\n$@" || echo "$@" )
 }
 
+if [[ "$TAGS" != '' ]];then
+	XTAGS="-tags $TAGS"
```
purpleidea: I think you might want the equivalent of:
```
-tags "tag1 tag2"
```

for this to work correctly with more than one tag.

aequitas: Indeed, stumbled upon another bug as well on a second review. Should both be fixed now.

test/test-gotest.sh
aequitas: Yes, but it is now refactored as an array.

purpleidea:
If travis is happy, then I think this LGTM =D

aequitas:
I noticed the flakiness in the tests, no problem. 

The macOS VM indeed seems slower, but that does not seem the root cause. If you look at time timestamp of the last output and when it complains about the 10 minute mark its only 2-3 minutes between. So my guess is Travis is not seeing the output correctly for the watchdog or some other bug? I'm seeing the same thing on my own Travis runs. I'll see if I can uncover it at a later time and post a fix if needed.

purpleidea:
Unfortunately your change seems to break `mdl` on travis.

purpleidea:
(And I just pushed a small rename... Sorry about that)

aequitas:
I'll push a fix later today.

aequitas:
Should be fixed now.

purpleidea:
LGTM :)

purpleidea:
Merged, thanks! \o/

The OSX users thank you, and glad to see you hacking with us again.

aequitas:
Thanks, good to be working on a nice Golang project again :)

PR #467, opened 2019-02-08
Docker Image Resource

jonathangold:
Would you be opposed to moving `examples/lang/docker_container0.mcl` and `examples/lang/docker_image0.mcl` into a folder, so I could import them in `examples/lang/docker0.mcl` instead of duplicating the code?

### DIFF
--- a/engine/resources/docker_image.go
+++ b/engine/resources/docker_image.go
@@ -1,0 +1,274 @@
+// Mgmt
+// Copyright (C) 2013-2019+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// +build !nodocker
+
+package resources
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/engine/traits"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/client"
+	errwrap "github.com/pkg/errors"
+)
+
+const (
+	// dockerImageInitCtxTimeout is the length of time, in seconds, before requests are
+	// cancelled in Init.
+	dockerImageInitCtxTimeout = 20
+	// dockerImageCheckApplyCtxTimeout is the length of time, in seconds, before requests
+	// are cancelled in CheckApply.
+	dockerImageCheckApplyCtxTimeout = 120
+)
+
+func init() {
+	engine.RegisterResource("docker:image", func() engine.Res { return &DockerImageRes{} })
+}
+
+// DockerImageRes is a docker image resource. The resource's name must be a
+// be a docker image in any supported format (url, image, or image:tag).
+type DockerImageRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Edgeable
+
+	// State of the image must be exists or absent.
+	State string `yaml:"state"`
+	// APIVersion allows you to override the host's default client API version.
+	APIVersion string `yaml:"apiversion"`
+
+	image  string         // full image:tag format
+	client *client.Client // docker api client
+
+	init *engine.Init
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *DockerImageRes) Default() engine.Res {
+	return &DockerImageRes{}
+}
+
+// Validate if the params passed in are valid data.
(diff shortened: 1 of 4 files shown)


jonathangold:
@purpleidea missed a comment. Fixed and re-pushed.

engine/resources/docker_image.go
```diff
@@ -0,0 +1,267 @@
...
+		case event, ok := <-eventChan:
+			if !ok { // channel shutdown
+				return nil
+			}
+			if obj.init.Debug {
+				obj.init.Logf("%+v", event)
+			}
+			send = true
+			obj.init.Dirty() // dirty
+		case err, ok := <-errChan:
```
purpleidea: Looks like we need to handle io.EOF specially... Did I read this correctly?

> the stream has been completely read an io.EOF error will be sent over the error channel.

https://github.com/moby/moby/blob/e6fe7f8f29364345083d466a3dfb21f9d539eda0/client/events.go#L15

engine/resources/docker_image.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// Close is run by the engine to clean up after the resource is done.
+func (obj *DockerImageRes) Close() error {
+	return obj.client.Close() // close the docker client
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *DockerImageRes) Watch() error {
+	ctx, cancel := context.WithCancel(context.Background())
```
purpleidea: What's most troubling is that when we send defer() this looks like it will deadlock:

https://github.com/moby/moby/blob/e6fe7f8f29364345083d466a3dfb21f9d539eda0/client/events.go#L48

```
case <-ctx.Done():
		errs <- ctx.Err()	# wat!
		return
```

I think this is just totally crap on their part, and you should probably report it upstream, but what it means is that we need to "drain" the errChan after close... Which means we need a waitgroup in Watch(), and right after the close happens, we need a defer that has a wg and that starts a go routine to drain the errChan.

```
defer func() { # docker API is WAT.
	wg.Add(1)
	go func() {
		defer wg.Done()
		for {
			_, ok := <- errChan
			if !ok {
				return
			}
		}
	}()
}()
```

Did this not deadlock? Am I reading this all correctly?

engine/resources/docker_image.go
```diff
@@ -0,0 +1,267 @@
...
+	engine.RegisterResource("docker:image", func() engine.Res { return &DockerImageRes{} })
+}
+
+// DockerImageRes is a docker container resource.
+type DockerImageRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Edgeable
+
+	// State of the container must be exists or absent.
+	State string `yaml:"state"`
```
purpleidea: So if we want "nginx:latest" how do we do that? Or if we want a specific version? All in the name? (If so, perhaps you could add a small doc note in the struct main docstring about this please.

examples/lang/docker_image0.mcl
```diff
@@ -0,0 +1,3 @@
+docker:image "nginx" {
+    state => "exists",
```
purpleidea: tabs not spaces

examples/lang/docker0.mcl
```diff
@@ -0,0 +1,10 @@
+docker:container "mgmt-nginx" {
+	state => "running",
+	image => "nginx",
+	cmd => ["nginx", "-g", "daemon off;",],
+	ports => {"tcp" => {80 => 8080,},},
+}
+
+docker:image "nginx" {
+    state => "exists",
```
purpleidea: tabs!

purpleidea:
Thanks for the patches! I didn't understand your question about the examples.

engine/resources/docker_image.go
jonathangold: It doesn't deadlock, because it's a buffered channel: 

https://github.com/moby/moby/blob/e6fe7f8f29364345083d466a3dfb21f9d539eda0/client/events.go#L22

It's a crappy pattern but it's not a blocking issue, afaict.

engine/resources/docker_image.go
jonathangold: It is only completely read when we cancel the context, after we return. It doesn't deadlock because they use a buffered channel (see below.)

engine/resources/docker_image.go
jonathangold: `image:tag` works, but in order to avoid collisions, I think it makes sense to append the `latest` tag if none is specified. E.g. `nginx` becomes `nginx:latest`.

examples/lang/docker_image0.mcl
jonathangold: :P

examples/lang/docker0.mcl
jonathangold: Too much yaml in my life

jonathangold:
Will fix and rebase this weekend. 

See comment about tags above. In that case `nginx` and `nginx:latest` would be the same, but `nginx:1.14-alpine` would not. 

Vis-a-vis the examples. If I make a folder: `examples/lang/docker/`, I could have a docker example in the `examples/lang` that could import resources from there. It's totally unnecessary right now, but could be nice if and when there are more docker resources.

purpleidea:
@jonathangold Heya, not sure if this stalled on my side or yours, but if it's ready for review, could you rebase to master and LMK? Thanks!

jonathangold:
@purpleidea I rebased and fixed Watch() to work with the current engine implementation. I haven't looked at the code in quite a while, but it still works as expected. Pretty sure I addressed all your comments.

purpleidea:
Sorry for the delay in re-reviewing this. I was really sick for a while, and busy a bit :( Anyways, I decided to merge! \o/ Thanks again! There were a few small nitpicks which I added in a new commit after, most line length stuff (I wish we had a test for this!) and some small Cmp bugs.

Thanks again!

PR #343, opened 2018-02-19
[wip] etcd: Connection options (socket file, ipv6)

aequitas:
Changes:

- Allow unix domain socket to be used as client url
- Using ::1 as clienturl should not create default local ipv4 listener: http://[::1]:

Todos:

- [ ] fix gprc warning
- [x] tests

Currently this allows a mgmt instance to be started with no TCP ports opened, eg using:

    ./mgmt run --prefix tmp/ --client-urls=unix://clients.sock:0 --server-urls=unix://servers.sock:0

And run `mgmt deploy` utilising the socket connection (deploy needs to be run from prefix directory).

Only problems I still need to solve is related to gprc: 

    WARNING: 2018/02/19 21:35:35 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: Error while dialing dial tcp: lookup clients.sock: no such host"; Reconnecting to {clients.sock:0 0  <nil>}

I have been debugging for a while now and still unsure why this happens.

PR #468, opened 2019-02-12
 etcd: Connection options (socket file, ipv6)

aequitas:
I know some changes to the etcd code are upcoming. But these changes might be simple enough to already incorporate.

Changes:

* Allow unix domain socket to be used as client url
* Using ::1 as clienturl should not create default local ipv4 listener: http://[::1]:
* Adds `no-network-standalone` option which starts the server without any TCP ports open

The reason behind this is that open TCP ports make the mgmt daemon vulnerable as long as authentication is not implemented. Using `unix://` sockets allows authorization using file system permissions and allows to run a isolated mgmt daemon to run safely. By default the sockets for standalone are created in the prefix directory which should be secured anyway.

This feature was originally posted as: https://github.com/purpleidea/mgmt/pull/343

In that PR there where still issues with grpc. Those have been solved in the upstream.

### DIFF
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -194,6 +194,7 @@ type EmbdEtcd struct { // EMBeddeD etcd
 	advertiseClientURLs etcdtypes.URLs    // client urls to advertise
 	advertiseServerURLs etcdtypes.URLs    // server urls to advertise
 	noServer            bool              // disable all server peering if true
+	noNetwork           bool              // use unix:// sockets instead of TCP for clients/servers
 
 	// local tracked state
 	nominated        etcdtypes.URLsMap // copy of who's nominated to locally track state
@@ -220,7 +221,7 @@ type EmbdEtcd struct { // EMBeddeD etcd
 }
 
 // NewEmbdEtcd creates the top level embedded etcd struct client and server obj.
-func NewEmbdEtcd(hostname string, seeds, clientURLs, serverURLs, advertiseClientURLs, advertiseServerURLs etcdtypes.URLs, noServer bool, idealClusterSize uint16, flags Flags, prefix string, converger converger.Converger) *EmbdEtcd {
+func NewEmbdEtcd(hostname string, seeds, clientURLs, serverURLs, advertiseClientURLs, advertiseServerURLs etcdtypes.URLs, noServer bool, noNetwork bool, idealClusterSize uint16, flags Flags, prefix string, converger converger.Converger) *EmbdEtcd {
 	endpoints := make(etcdtypes.URLsMap)
 	if hostname == seedSentinel { // safety
 		return nil
@@ -229,6 +230,15 @@ func NewEmbdEtcd(hostname string, seeds, clientURLs, serverURLs, advertiseClient
 		log.Printf("Etcd: need at least one seed if running with --no-server!")
 		return nil
 	}
+	if noNetwork {
+		if len(clientURLs) != 0 || len(serverURLs) != 0 || len(seeds) != 0 {
+			log.Printf("--no-network is mutual exclusive with --seeds, --client-urls and --server-urls")
+			return nil
+		}
+		clientURLs, _ = etcdtypes.NewURLs([]string{"unix://clients.sock:0"})
+		serverURLs, _ = etcdtypes.NewURLs([]string{"unix://servers.sock:0"})
+	}
+
 	if len(seeds) > 0 {
 		endpoints[seedSentinel] = seeds
 		idealClusterSize = 0 // unset, get from running cluster
@@ -253,6 +263,7 @@ func NewEmbdEtcd(hostname string, seeds, clientURLs, serverURLs, advertiseClient
 		advertiseClientURLs: advertiseClientURLs,
 		advertiseServerURLs: advertiseServerURLs,
 		noServer:            noServer,
+		noNetwork:           noNetwork,
 
 		idealClusterSize: idealClusterSize,
 		converger:        converger,
@@ -304,7 +315,7 @@ func (obj *EmbdEtcd) GetConfig() etcd.Config {
 	// XXX: filter out any urls which wouldn't resolve here ?
 	for _, eps := range obj.endpoints { // flatten map
 		for _, u := range eps {
-			endpoints = append(endpoints, u.Host) // remove http:// prefix
+			endpoints = append(endpoints, u.String()) // use full url including scheme
 		}
 	}
 	sort.Strings(endpoints) // sort for determinism
@@ -1692,8 +1703,12 @@ func (obj *EmbdEtcd) LocalhostClientURLs() etcdtypes.URLs {
 	// look through obj.clientURLs and return the localhost ones
 	urls := etcdtypes.URLs{}
 	for _, x := range obj.clientURLs {
-		// "localhost" or anything in 127.0.0.0/8 is valid!
-		if s := x.Host; strings.HasPrefix(s, "localhost") || strings.HasPrefix(s, "127.") {
+		// "localhost", ::1 or anything in 127.0.0.0/8 is valid!
+		if s := x.Host; strings.HasPrefix(s, "localhost") || strings.HasPrefix(s, "127.") || strings.HasPrefix(s, "[::1]") {
+			urls = append(urls, x)
+		}
+		// or local unix domain socket
+		if x.Scheme == "unix" {
 			urls = append(urls, x)
 		}
 	}
--- a/lib/cli.go
+++ b/lib/cli.go
@@ -171,7 +171,11 @@ func CLI(program, version string, flags Flags) error {
 			Name:  "no-server",
 			Usage: "do not start embedded etcd server (do not promote from client to peer)",
 		},
-
+		cli.BoolFlag{
+			Name:   "no-network",
+			Usage:  "run single node instance without clustering or opening tcp ports to the outside",
+			EnvVar: "MGMT_NO_NETWORK",
+		},
 		cli.BoolFlag{
 			Name:  "no-pgp",
 			Usage: "don't create pgp keys",
--- a/lib/main.go
+++ b/lib/main.go
@@ -89,6 +89,7 @@ type Main struct {
 	AdvertiseServerURLs []string // list of URLs to advertise for server (peer) traffic
 	IdealClusterSize    int      // ideal number of server peers in cluster; only read by initial server
 	NoServer            bool     // do not let other servers peer with me
+	NoNetwork           bool     // run single node instance without clustering or opening tcp ports to the outside
 
 	seeds               etcdtypes.URLs // processed seeds value
 	clientURLs          etcdtypes.URLs // processed client urls value
@@ -354,6 +355,7 @@ func (obj *Main) Run() error {
 		obj.advertiseClientURLs,
 		obj.advertiseServerURLs,
 		obj.NoServer,
+		obj.NoNetwork,
 		obj.idealClusterSize,
 		etcd.Flags{
 			Debug:   obj.Flags.Debug,
--- a/lib/run.go
+++ b/lib/run.go
@@ -107,6 +107,7 @@ func run(c *cli.Context, name string, gapiObj gapi.GAPI) error {
 	obj.AdvertiseServerURLs = cliContext.StringSlice("advertise-server-urls")
 	obj.IdealClusterSize = cliContext.Int("ideal-cluster-size")
 	obj.NoServer = cliContext.Bool("no-server")
+	obj.NoNetwork = cliContext.Bool("no-network")
 
 	obj.NoPgp = cliContext.Bool("no-pgp")
 
--- a/test/shell/clustersize.sh
+++ b/test/shell/clustersize.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+
+if ! command -v etcdctl >/dev/null; then
+	echo "Missing etcdctl, skipping"
+	exit 0
+fi
+
+. "$(dirname "$0")/../util.sh"
+
+mkdir /tmp/mgmt/{A..E}
+
+# kill servers on error/exit
+trap 'pkill -9 mgmt' EXIT
+
+"$MGMT" run --hostname h1 --tmp-prefix --no-pgp empty &
+"$MGMT" run --hostname h2 --tmp-prefix --no-pgp --seeds http://127.0.0.1:2379 --client-urls http://127.0.0.1:2381 --server-urls http://127.0.0.1:2382 empty &
+"$MGMT" run --hostname h3 --tmp-prefix --no-pgp --seeds http://127.0.0.1:2379 --client-urls http://127.0.0.1:2383 --server-urls http://127.0.0.1:2384 empty &
+
+# wait for everything to converge
+sleep 10
+
+ETCDCTL_API=3 etcdctl --endpoints 127.0.0.1:2379 put /_mgmt/idealClusterSize 3
+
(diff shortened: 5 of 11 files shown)


purpleidea:
Hey, I'm happy to review and merge what I can, I just can't promise which parts of this will get preserved in my upcoming re-write. (I'll try and incorporate as many as possible into the rebase, although my lack of understanding with parts of this might require that you re-submit anything I've missed in the future.)

Having said that, if you're okay with that, review is coming...

etcd/etcd.go
```diff
@@ -304,7 +304,7 @@ func (obj *EmbdEtcd) GetConfig() etcd.Config {
 	// XXX: filter out any urls which wouldn't resolve here ?
 	for _, eps := range obj.endpoints { // flatten map
 		for _, u := range eps {
-			endpoints = append(endpoints, u.Host) // remove http:// prefix
+			endpoints = append(endpoints, u.String()) // use full url including scheme
```
purpleidea: I feel like I took it off for a reason... Doesn't doing this break something elsewhere?

lib/cli.go
```diff
@@ -171,7 +171,11 @@ func CLI(program, version string, flags Flags) error {
 			Name:  "no-server",
 			Usage: "do not start embedded etcd server (do not promote from client to peer)",
 		},
-
+		cli.BoolFlag{
+			Name:   "no-network-standalone",
```
purpleidea: Seems redundant, would `no-network` by itself be sufficient?
Can you elaborate more on the use case for this? Is it just standalone single host usage?

lib/deploy.go
```diff
@@ -99,6 +100,28 @@ func deploy(c *cli.Context, name string, gapiObj gapi.GAPI) error {
...
+		}
+	}
+	if unixsockets {
+		wd, err := os.Getwd()
+		if err != nil {
+			return errwrap.Wrapf(err, "could not get current working directory")
+		}
+		prefix := c.GlobalString("prefix")
+		log.Printf("Deploy: changing to directory %s", prefix)
+		if err := os.Chdir(prefix); err != nil {
```
purpleidea: I'm concerned this might cause a problem for other parts of the code that might be expecting the original working directory. Do you think that's the case? Is there a more isolated way of doing this, or can we do so using absolute paths so we don't need to `os.Chdir` ?

lib/run.go
```diff
@@ -118,6 +118,14 @@ func run(c *cli.Context, name string, gapiObj gapi.GAPI) error {
 		obj.PgpIdentity = &us
 	}
 
+	if cliContext.Bool("no-network-standalone") {
```
purpleidea: In general we try and put this "defaults" logic inside of the etcd/ if possible, as lib/ just copies the values in... Do you think that's possible? Look at the `const` default, for example. HTH

aequitas:
Sound good to me. I'll jump in if you push the rewrite code to adapt where needed for this feature. But I don't expect much issues anyways as nothing fundamental is changed in this PR, only that more URL types (which etcd already supports) are allowed to be passed.

etcd/etcd.go
aequitas: AFAICT no tests are breaking. I tested this extensively last year when I started working on this feature and didn't find any issue. But there might be scenarios I'm not aware of?

lib/cli.go
aequitas: Use case is indeed to have mgmt run (as root) on a single host without any TCP connections on localhost so an unprivileged user is not able to access the mgmt API and elevate to root.

I believe we discussed the name back then on irc and ended up with this. But `no-network` would be fine for me as well.

lib/deploy.go
aequitas: There is a defer to change back to the original directory after the enclosing code block is complete. I will check if etcd/grpc has been updated to support non-relative `unix://` paths in the meantime. Then we can drop this entire logic.

lib/run.go
aequitas: I'll look into that.

etcd/etcd.go
purpleidea: Can you try this one please:

https://github.com/purpleidea/mgmt/blob/90fa83a5cf28394617583578e982eeedd94a5628/examples/lang/exchange0.mcl#L1

lib/deploy.go
purpleidea: That would be excellent! I just honestly don't know if it would affect other goroutines or if the chdir affects the entire program. My guess is the latter, which would potentially be problematic for any case that looks at the working dir.

lib/run.go
purpleidea: Great!

etcd/etcd.go
aequitas: Test added: https://github.com/purpleidea/mgmt/pull/468/files#diff-34b136176f6a9b9c42585fa1ae2a07df

lib/deploy.go
aequitas: The problem still persists in etcd, in that if you specify a `unix://` socket it tries to unpack it into a host and port. So `unix://clients.sock:0` is unpacked to 'host' `clients.sock` and 'port' `0`. It will result in the socket file `clients.sock:0` being created in the current working directory. Adding any path before the file causes an error of it not being a 'host'.

The logic here is only invoked when unix sockets are specified as client or server urls. So unless a user explicitly specifies these in arguments (or runs with `--no-network`) this code won't trigger.

As far as the `Chdir` goes. It is only needed when etcd creates the socket files. After that we can chdir back to the original working directory. I could wrap this logic around the etcd clients creation if you want to make it more explicit. Then the chdir back would happen right after instead of at the end of the function.

purpleidea: So it seems the Chdir thing is a problem, see: https://github.com/golang/go/issues/27658
So can we open a bug upstream wherever that socket is written so that it uses an absolute path?

etcd/etcd.go
purpleidea: Sweet! I was actually hoping that was it, but here's one more...
https://github.com/purpleidea/mgmt/blob/master/etcd/etcd.go#L38
I remember that code in the back of my mind, I just can't remember why it was necessary, but that it was. So it's really my bad for having all this old code... Which is why I'm nuking it shortly... In any case...

lib/deploy.go
aequitas: That might be a problem indeed. I think it would be a slim chance since this is called early on in the program so not many other goroutines relying on the working dir should be active. But I don't know for certain so there might be the odd race condition.

This part of the code is only used if you want to deploy mcl on the host itself to a non-networking instance of mgmt. So functionally it could be replaced with restarting mgmt instance after you change the configuration files locally.

etcd/etcd.go
aequitas: I'm trying to convert this into a testcase as well, but on master this fails with:
```
~/p/m/mgmt (master|…) $ ./mgmt run --hostname h1 --tmp-prefix --no-pgp yaml --yaml examples/yaml/etcd1a.yaml
22:09:06 hello.go:49: this is: mgmt, version: 0.0.17-46-g4c80869
22:09:06 hello.go:50: main: start: 1550005746276466928
22:09:06 main.go:191: main: warning: working prefix directory is temporary!
22:09:06 main.go:191: main: working prefix is: /var/folders/00/wq2lzdvd6sg2w1_5l0092bj00000gn/T/mgmt-h1-776165371
22:09:06 main.go:191: main: etcd: seeds: no seeds specified!
22:09:06 etcd.go:435: Etcd: Bootstrapping...
22:09:06 etcd.go:350: Etcd: Connect: Endpoints: []
22:09:06 etcd.go:364: Etcd: Connect: CtxError...
22:09:06 etcd.go:644: Etcd: CtxError: Reason: CtxDelayErr(1s): No endpoints available yet!
22:09:06 etcd.go:1472: Etcd: Nominated: h1=http://localhost:2380
22:09:06 etcd.go:1487: Etcd: StartServer(newCluster: true): h1=http://localhost:2380
22:09:06 etcd.go:1758: Etcd: StartServer: Starting server...
22:09:06 etcd.go:1765: Etcd: StartServer: Done starting server!
22:09:06 etcd.go:1783: Etcd: StartServer: Server running...
22:09:06 etcd.go:1510: Etcd: Addresses are: http://localhost:2379
22:09:07 etcd.go:348: Etcd: Connect: Endpoints: h1=http://localhost:2379
22:09:07 etcd.go:1830: Etcd: ApplyDeltaEvents: Event(PUT): h1
22:09:07 etcd.go:1472: Etcd: Nominated: h1=http://localhost:2380
22:09:07 etcd.go:455: Etcd: Startup: Volunteering...
22:09:07 etcd.go:1678: Etcd: Ideal cluster size is now: 5
22:09:07 etcd.go:1022: Etcd: Set(/_mgmt/idealClusterSize): &{cluster_id:14841639068965178418 member_id:10276657743932975437 revision:4 raft_term:2  <nil>}
22:09:07 etcd.go:1272: Etcd: Members: List: [h1]
22:09:07 etcd.go:1295: Etcd: Leader: h1
22:09:07 etcd.go:1314: Etcd: Volunteers: [h1]
22:09:07 etcd.go:1318: Etcd: Quitters: []
22:09:07 etcd.go:1330: Etcd: Candidates: []
22:09:07 etcd.go:1272: Etcd: Members: List: [h1]
22:09:07 etcd.go:1295: Etcd: Leader: h1
22:09:07 etcd.go:1314: Etcd: Volunteers: [h1]
22:09:07 etcd.go:1318: Etcd: Quitters: []
22:09:07 etcd.go:1330: Etcd: Candidates: []
22:09:08 main.go:191: main: waiting...
22:09:08 main.go:191: main: running...
22:09:08 main.go:191: main: waiting...
22:09:08 main.go:492: gapi: generating new graph...
22:09:08 resources.go:41: Etcd: Watch: Path: /_mgmt/exported/
22:09:08 main.go:492: gapi: generating new graph...
22:09:08 resources.go:175: Etcd: Get: (Hostname, Kind, Name): (h1, file, filea)
22:09:08 main.go:492: gapi: yaml: collect: file; pattern: /tmp/mgmtA/
22:09:08 main.go:410: engine: autoedge: adding autoedges...
22:09:08 main.go:410: engine: autogroup: algorithm: wrappedGrouper: NonReachabilityGrouper...
22:09:08 main.go:191: main: commit...
22:09:08 main.go:410: engine: graph sync...
22:09:08 main.go:191: main: graph: Vertices(0), Edges(0)
22:09:08 main.go:191: main: waiting...
22:09:08 resources.go:175: Etcd: Get: (Hostname, Kind, Name): (h1, file, filea)
22:09:08 resources.go:109: Etcd: SetResources: Error: Empty kind: 
```

lib/run.go
aequitas: This has been moved into `etcd/etcd.go`

etcd/etcd.go
purpleidea: Oh gah! That is so embarrassing of me, it's still a yaml example ;)
I told you this file was getting nuked =D

Just try the same example with the `yaml --yaml examples/yaml/etcd1a.yaml` replaced with `empty`

It's the cluster growing/shrinking volunteering/etc part that mattered, not the code that ran, sorry.

Thanks!

aequitas: Tests added :) https://github.com/purpleidea/mgmt/pull/468/files#diff-2fb98fd83d458bf7aeff523d04b46264

lib/cli.go
```diff
@@ -171,7 +171,11 @@ func CLI(program, version string, flags Flags) error {
 			Name:  "no-server",
 			Usage: "do not start embedded etcd server (do not promote from client to peer)",
 		},
-
+		cli.BoolFlag{
+			Name:   "no-network",
+			Usage:  "run single node instance without clustering or opening tcp ports to the outside",
+			EnvVar: "MGMT_NO_NETWORK_STANDALONE",
```
purpleidea: Should we rename this var to match the flag name?

lib/main.go
```diff
@@ -225,6 +226,12 @@ func (obj *Main) Run() error {
 		}
 	}
 	Logf("working prefix is: %s", prefix)
+	// change to prefix directory to have a deterministic path where unix:// socket files are created
```
purpleidea: I can't commit this block just yet until we've figured out the `chdir` plan as discussed, but I can commit everything else if you'd like to split this part out into a separate commit.

aequitas:
All chdir is removed. The feature is less convenient now but should still provide the core functionality needed.

aequitas:
The build is currently failing on the ipv6-localhost tests. I guess this is due to TravisCI not having ipv6 enabled, but I'm still investigating.

aequitas:
I disabled the test for hosts that don't have ipv6 localhost.

test/shell/ipv6-localhost.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+
+if ! ifconfig lo | grep 'inet6 ::1' >/dev/null;then
```
purpleidea: nit: Can you add a space after the last semicolon please

test/shell/clustersize.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+
+if ! command -v etcdctl >/dev/null;then
```
purpleidea: nit: Can you add a space after the last semicolon please

test/shell/clustersize.sh
```diff
@@ -0,0 +1,40 @@
...
+sleep 10
+
+test "$(ETCDCTL_API=3 etcdctl --endpoints 127.0.0.1:2379 member list|wc -l)" -eq 3
+
+ETCDCTL_API=3 etcdctl --endpoints 127.0.0.1:2381 put /_mgmt/idealClusterSize 5
+
+# wait for everything to converge
+sleep 5
+
+test "$(ETCDCTL_API=3 etcdctl --endpoints 127.0.0.1:2381 member list|wc -l)" -eq 5
```
purpleidea: Missing nl at the end & spaces around the pipe please

test/shell/clustersize.sh
```diff
@@ -0,0 +1,40 @@
...
+
+ETCDCTL_API=3 etcdctl --endpoints 127.0.0.1:2379 put /_mgmt/idealClusterSize 3
+
+"$MGMT" run --hostname h4 --tmp-prefix --no-pgp --seeds http://127.0.0.1:2379 --client-urls http://127.0.0.1:2385 --server-urls http://127.0.0.1:2386 empty &
+"$MGMT" run --hostname h5 --tmp-prefix --no-pgp --seeds http://127.0.0.1:2379 --client-urls http://127.0.0.1:2387 --server-urls http://127.0.0.1:2388 empty &
+
+# wait for everything to converge
+sleep 10
+
+test "$(ETCDCTL_API=3 etcdctl --endpoints 127.0.0.1:2379 member list|wc -l)" -eq 3
```
purpleidea: spaces around the pipe please

lib/main.go
```diff
@@ -225,7 +226,6 @@ func (obj *Main) Run() error {
 		}
 	}
 	Logf("working prefix is: %s", prefix)
-
```
purpleidea: unnecessary nl deletion

aequitas:
That should be everything I think.

purpleidea:
Looks perfect! Thank you again, Merged :)

aequitas:
Thanks \o/

purpleidea:
Thank you!

(If you feel like elaborating on where the `chdir` was needed-- (I'm guessing in etcd?) then you could open an upstream issue asking for absolute paths perhaps?)

PR #470, opened 2019-02-14
build: Fix macOS build

jdub:
Add pkg-config to fix builds with augeas and libvirt on macOS.

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -52,7 +52,7 @@ fi
 
 if [ ! -z "$BREW" ]; then
 	# coreutils contains gtimeout, gstat, etc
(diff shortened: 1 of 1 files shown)


purpleidea:
@jdub Welcome and thanks for the patch!

It's perfect for you to learn about one of our build quirks:

We have a test for commit message, so they should be in the like so:

https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md#tips

If the body itself could be a properly formatted sentence (capitalized/period at end) that would be appreciated too!

Thanks for the fix, and welcome to the project!

jdub:
@purpleidea Tidied.

purpleidea:
LGTM, thanks!

PR #473, opened 2019-02-24
pippet resource support

ffrank:
The pippet resource implements faster integration of Puppet resources
in mgmt at runtime, by piping synchronization commands to a Puppet
process that keeps running alongside mgmt. This avoids huge overhead
through launching a Puppet process for each operation on a resource
that is delegated to Puppet.

### DIFF
--- a/engine/resources/pippet.go
+++ b/engine/resources/pippet.go
@@ -1,5 +1,5 @@
 // Mgmt
-// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Copyright (C) 2013-2019+ James Shubin and the project contributors
 // Written by James Shubin <james@shubin.ca> and the project contributors
 //
 // This program is free software: you can redistribute it and/or modify
--- a/engine/resources/pippet.go
+++ b/engine/resources/pippet.go
@@ -1,0 +1,350 @@
+// Mgmt
+// Copyright (C) 2013-2018+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os/exec"
+	"sync"
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/engine/traits"
+	"github.com/purpleidea/mgmt/util/errwrap"
+)
+
+var pippetReceiverInstance *pippetReceiver
+var pippetReceiverOnce sync.Once
+
+func init() {
+	engine.RegisterResource("pippet", func() engine.Res { return &PippetRes{} })
+}
+
+// PippetRes is a wrapper resource for puppet.  It implements the functional
+// equivalent of an exec resource that calls "pupept resource <type> <title>
+// <params>", but offers superior performance through a long-running Puppet
+// process that receives resources through a pipe (hence the name).
+type PippetRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Refreshable
+
+	init *engine.Init
+
+	// Type is the exact name of the wrapped Puppet resource type, e.g.
+	// "file", "mount". This needs not be a core type. It can be
+	// a type from a module. The Puppet installation local to the mgmt
+	// agent machine must be able recognize it. It has to be a native
+	// type though, as opposed to defined types from your Puppet manifest
+	// code.
+	Type string `yaml:"type" json:"type"`
+	// Title is used by Puppet as the resource title. Puppet will often
+	// assign special meaning to the title, e.g. use it as the path for a
+	// file resource, or the name of a package.
+	Title string `yaml:"title" json:"title"`
+	// Params is expected to be a hash in YAML format, pairing resource
+	// parameter names with their respective values,
+	// e.g. { ensure: present }
+	Params string `yaml:"params" json:"params"`
+
+	runner *pippetReceiver
+}
+
+// Default returns an example Pippet resource.
+func (obj *PippetRes) Default() engine.Res {
+	return &PippetRes{
+		Params: "{}", // use an empty params hash per default
+	}
+}
+
+// Validate never errors out. We don't know the set of potential types that
+// Puppet supports. Resource names are arbitrary. We cannot really validate the
+// parameter YAML, because we cannot assume that it can be unmarshalled into a
+// map[string]string; Puppet supports complex parameter values.
+func (obj *PippetRes) Validate() error {
+	return nil
+}
+
+// Init makes sure that the PippetReceiver object is initialized.
+func (obj *PippetRes) Init(init *engine.Init) error {
+	obj.init = init // save for later
+
+	var err error
+	obj.runner, err = PippetReceiverInstance()
+	if err != nil {
+		return err
+	}
+
+	if err = obj.runner.Register(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Close is run by the engine to clean up after the resource is done.
+func (obj *PippetRes) Close() error {
+	return obj.runner.Unregister()
+}
+
+// Watch is the primary listener for this resource and it outputs events.
+func (obj *PippetRes) Watch() error {
+	obj.init.Running() // when started, notify engine that we're running
+
+	select {
+	case <-obj.init.Done: // closed by the engine to signal shutdown
+	}
+
+	//obj.init.Event() // notify engine of an event (this can block)
+
+	return nil
+}
+
+// CheckApply synchronizes the resource if required.
+func (obj *PippetRes) CheckApply(apply bool) (bool, error) {
+	var changed bool
+	var err error
+	if changed, err = applyPippetRes(obj.runner, obj); err != nil {
+		return false, fmt.Errorf("pippet: %s[%s]: ERROR - %v", obj.Type, obj.Title, err)
+	}
+	return !changed, nil
+}
+
+// Cmp compares two resources and returns an error if they are not equivalent.
+func (obj *PippetRes) Cmp(r engine.Res) error {
+	res, ok := r.(*PippetRes)
+	if !ok {
+		return fmt.Errorf("not a %s", obj.Kind())
+	}
+
+	if obj.Type != res.Type {
+		return fmt.Errorf("the type params differ")
+	}
+
+	if obj.Title != res.Title {
+		return fmt.Errorf("the resource titles differ")
+	}
+
+	// FIXME: This is a lie. Parameter lists can be equivalent
+	// but not lexically identical (e.g. whitespace differences,
+	// parameter order).
+	// This is difficult to handle because we cannot casually
+	// unmarshall the YAML content.
+	if obj.Params != res.Params {
+		return fmt.Errorf("the parameters differ")
+	}
+
+	return nil
+}
+
+// PippetUID is the UID struct for PippetRes.
+type PippetUID struct {
+	engine.BaseUID
+	resourceType  string
+	resourceTitle string
+}
+
+// UIDs includes all params to make a unique identification of this object.
+// Most resources only return one, although some resources can return multiple.
+func (obj *PippetRes) UIDs() []engine.ResUID {
+	x := &PippetUID{
+		BaseUID:       engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
+		resourceType:  obj.Type,
+		resourceTitle: obj.Title,
+	}
+	return []engine.ResUID{x}
+}
+
+// GroupCmp returns whether two resources can be grouped together or not.
+func (obj *PippetRes) GroupCmp(r engine.GroupableRes) error {
+	return fmt.Errorf("pippet resources are never grouped")
+}
+
+// UnmarshalYAML is the custom unmarshal handler for this struct.
+// It is primarily useful for setting the defaults.
+func (obj *PippetRes) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type rawRes PippetRes // indirection to avoid infinite recursion
+
+	def := obj.Default()        // get the default
+	res, ok := def.(*PippetRes) // put in the right format
+	if !ok {
+		return fmt.Errorf("could not convert to PippetRes")
+	}
+	raw := rawRes(*res) // convert; the defaults go here
+
+	if err := unmarshal(&raw); err != nil {
+		return err
+	}
+
+	*obj = PippetRes(raw) // restore from indirection with type conversion!
+	return nil
+}
+
+// PippetRunner is the interface used to communicate with the PippetReceiver
+// object. Its main purpose is dependency injection.
+type PippetRunner interface {
+	LockApply()
+	UnlockApply()
+	InputStream() io.WriteCloser
+	OutputStream() io.ReadCloser
+}
+
+// PippetResult is the structured return value type for the PippetReceiver's
+// Apply function.
+type PippetResult struct {
+	Error     bool
+	Failed    bool
+	Changed   bool
+	Exception string
+}
+
+// PippetReceiverInstance returns a pointer to the PippetReceiver object.
+// The PippetReceiver is supposed to be a singleton object. The pippet resource
+// code should always use the PippetReceiverInstance function to gain access to
+// the pippetReceiver object. Other objects of type pippetReceiver should not
+// be created.
+func PippetReceiverInstance() (*pippetReceiver, error) {
+	if pippetReceiverInstance == nil {
+		pippetReceiverOnce.Do(func() { pippetReceiverInstance = &pippetReceiver{} })
+	}
+	return pippetReceiverInstance, nil
+}
+
+type pippetReceiver struct {
+	stdin         io.WriteCloser
+	stdout        io.ReadCloser
+	registerMutex sync.Mutex
+	applyMutex    sync.Mutex
+	registered    int
+}
+
+// Init runs the Puppet process that will perform the work of synchronizing
+// resources that are sent to its stdin. The process will keep running until
+// Close is called. Init should not be called directly. It is implicitly called
+// by the Register function.
+func (obj *pippetReceiver) Init() error {
+	cmd := exec.Command("puppet", "yamlresource", "receive", "--color=no")
+	var err error
+	obj.stdin, err = cmd.StdinPipe()
+	if err != nil {
+		return err
+	}
+	obj.stdout, err = cmd.StdoutPipe()
+	if err != nil {
+		err = errwrap.Append(err, obj.stdin.Close())
+		return err
+	}
+	if err = cmd.Start(); err != nil {
+		err = errwrap.Append(err, obj.stdin.Close())
+		return err
+	}
+	buf := make([]byte, 80)
+	if _, err = obj.stdout.Read(buf); err != nil {
+		err = errwrap.Append(err, obj.stdin.Close())
+	}
+	return err
+}
+
+// Register should be called by any user (i.e., any pippet resource) before
+// using the PippetRunner functions on this receiver object. Register implicitly takes
+// care of calling Init if required.
+func (obj *pippetReceiver) Register() error {
+	obj.registerMutex.Lock()
+	defer obj.registerMutex.Unlock()
+	obj.registered = obj.registered + 1
+	if obj.registered > 1 {
+		return nil
+	}
+	// count was increased from 0 to 1, we need to (re-)init
+	var err error
+	if err = obj.Init(); err != nil {
+		obj.registered = 0
+	}
+	return err
+}
+
+// Unregister should be called by any object that registered itself using the
+// Register function, and which no longer needs the receiver. This should
+// typically happen at closing time of the pippet resource that registered
+// itself.  Unregister implicitly calls Close in case all registered resources
+// have unregistered.
+func (obj *pippetReceiver) Unregister() error {
+	obj.registerMutex.Lock()
+	defer obj.registerMutex.Unlock()
+	obj.registered = obj.registered - 1
+	var err error
+	if obj.registered == 0 {
+		err = obj.Close()
+	}
+	if obj.registered < 0 {
+		err = errwrap.Append(err, fmt.Errorf("pippet runner: ERROR: unregistered more resources than were registered"))
+	}
+	return err
+}
+
+// LockApply locks the pippetReceiver's mutex for an "Apply"transaction.
+func (obj *pippetReceiver) LockApply() {
+	obj.applyMutex.Lock()
+}
+
+// UnlockApply unlocks the pippetReceiver's mutex for an "Apply"transaction.
+func (obj *pippetReceiver) UnlockApply() {
+	obj.applyMutex.Unlock()
+}
+
+// InputStream returns the pippetReceiver's pipe writer.
+func (obj *pippetReceiver) InputStream() io.WriteCloser {
+	return obj.stdin
+}
+
+// OutputStream returns the pippetReceiver's pipe reader.
+func (obj *pippetReceiver) OutputStream() io.ReadCloser {
+	return obj.stdout
+}
+
+// Close stops the backend puppet process by closing its stdin handle. It should
+// not be called directly. It is implicitly called by the Unregister function if
+// appropriate.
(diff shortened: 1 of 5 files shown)


ffrank:
Missing context here: Once this is in, the Puppet->mgmt translator can stop making this:
```
exec "whatever" { cmd => "puppet yamlresource resname restype { params ... }", ... }
```

And instead make this
```
pippet "whatever" { name: "resname", type: "restype", params: "{ <yaml stuff> }" }
```

That way, when the translator has to decide to fall back to Puppet, synchronizing the graph will not be as atrociously slow.

purpleidea:
Should I be reviewing or is this still meant as WIP?

ffrank:
@purpleidea this is now ready for its first review.

The PuppetReceiver code (in util) lacks logging, which is unfortunate. Should probably be added before merge, but I'm not yet sure how to approach it.

Another sore spot are missing tests. We'd need to add Puppet to the build requirements. This is necessary anyway, I feel, in order to make sure that any Puppet integration/support works.

purpleidea:
@ffrank Cool, before I review can you move: `util/pippet_run.go` under resources? If you *need* a new package, you can do so below that dir like packagekit does. (Probably not necessary here.) Util is for stuff that's common to more than one resource or package.

ffrank:
I had tried that originally, but failed to get it to work because of import cycles. I'll see if I can resolve this but it might not work out.

ffrank:
Never mind. It's done.

ffrank:
Addressed the warnings I could manage.
I don't understand why the gofmt test fails, it's what I use on my end. Wrong go version? (This is terrible then.)

purpleidea:
> gofmt test fails

Yeah, it's very shitty, but they break compat between versions. At least twice so far I think. Unfortunately we'll have to use 1.10 until Fedora 30 is out, and then I'll move to 1.11.

Reviewing now...

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"fmt"
+	//"sync"
```
purpleidea: rm please

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/engine/traits"
+)
+
+func init() {
+	engine.RegisterResource("pippet", func() engine.Res { return &PippetRes{} })
+}
+
+// PippetRes is a wrapper resource for puppet.
```
purpleidea: Wrap and reflow everything to 80 chars please.

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+// that calls "pupept resource <type> <title> <params>", but offers
+// superior performance through a long-running Puppet process that
+// receives resources through a pipe (hence the name).
+type PippetRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Refreshable
+
+	init *engine.Init
+
+	Type   string `yaml:"type" json:"type"`
```
purpleidea: Why json?
Also, you'll want to add the `lang` tags only if you want anyone to be able to use this from the DSL. (Do we?)

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+	Title  string `yaml:"title" json:"title"`
+	Params string `yaml:"params" json:"params"`
+
+	runner PippetRunner
+}
+
+// Default returns an example Pippet resource.
+func (obj *PippetRes) Default() engine.Res {
+	return &PippetRes{
+		Params: "{}",
```
purpleidea: Could you add a comment on this line explaining what this means please?

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+// superior performance through a long-running Puppet process that
+// receives resources through a pipe (hence the name).
+type PippetRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Refreshable
+
+	init *engine.Init
+
+	Type   string `yaml:"type" json:"type"`
+	Title  string `yaml:"title" json:"title"`
```
purpleidea: Each public field can have a docstring comment, eg: https://github.com/purpleidea/mgmt/blob/master/engine/resources/file.go#L63

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+func (obj *PippetRes) Init(init *engine.Init) error {
+	obj.init = init // save for later
+
+	var err error
+	obj.runner, err = PippetRunnerInstance()
+	if err != nil {
+		return err
+	}
+
+	obj.runner.Register(obj)
```
purpleidea: Should this method have an error returned that we should check, or is it guaranteed to always pass?

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return nil
+}
+
+// Close is run by the engine to clean up after the resource is done.
+func (obj *PippetRes) Close() error {
+	obj.runner.Unregister(obj)
```
purpleidea: if err := ?

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+
+	//obj.init.Event() // notify engine of an event (this can block)
+
+	return nil
+}
+
+// CheckApply synchronizes the resource if required.
+func (obj *PippetRes) CheckApply(apply bool) (checkOK bool, err error) {
+	if obj.init.Refresh() {
+		obj.init.Logf("WARNING - pippet does not support refreshing yet")
```
purpleidea: Then just remove this entirely please...

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Cmp compares two resources and returns an error if they are not equivalent.
+func (obj *PippetRes) Cmp(r engine.Res) error {
+	res, ok := r.(*PippetRes)
+	if !ok {
+		return fmt.Errorf("not a %s", obj.Kind())
+	}
+
+	if obj.Type != res.Type {
+		return fmt.Errorf("the resource types differ")
```
purpleidea: nit: just because the word `type` is so overloaded, perhaps: `the Type param differs` instead?

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	if obj.Title != res.Title {
+		return fmt.Errorf("the resource titles differ")
+	}
+
+	// FIXME: This is a lie. Parameter lists can be equivalent
```
purpleidea: You're allowed to do a more complex Cmp if it's appropriate. Is this what you mean? Eg: compare the sorted lists that you mapped with Trim?

engine/resources/pippet.go
```diff
@@ -0,0 +1,183 @@
...
+	// unmarshall the YAML content.
+	if obj.Params != res.Params {
+		return fmt.Errorf("the parameters differ")
+	}
+
+	return nil
+}
+
+// PippetUID is the UID struct for PippetRes.
+// FIXME: Is it fine to ignore parameters here? S.a. the note in Cmp.
```
purpleidea: S.a ???

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+	"fmt"
+	"io"
+	"os/exec"
+	"sync"
+)
+
+// PippetRunner is the interface to the wrapped "puppet yamlresource receive"
+// process that keeps running alongside mgmt. Individual resources are expected
+// to register themselves, and can use the Apply function from that point on.
+type PippetRunner interface {
```
purpleidea: I'm not sure why you have an interface type. Interfaces are useful (in general) when more than one thing implements them, or when you plan a future second implementation. AIUI there's only one here...

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+	Register(interface{}) error
+	Unregister(interface{}) error
+	Apply(string, string, string) PippetResult
```
purpleidea: Weird return arg, would: (PippetResult, error) be more appropriate?

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+	Close() error
+}
+
+// PippetResult is the structured return value type for the PippetRunner's
+// Apply function.
+type PippetResult struct {
+	Error     bool
```
purpleidea: What's the difference between Error and failed? And Exception?

My guess is you might prefer:

`(PippetResult, error)` (as mentioned above), with:

```type PippetResult struct{
   Changed bool
   AnyOtherNeededParam int
}```

File: engine/resources/pippet_run.go [hunk 15]
```diff
@@ -0,0 +1,160 @@
...
+
+type pippetReceiver struct {
+	stdin         io.WriteCloser
+	stdout        io.ReadCloser
+	registerMutex sync.Mutex
+	applyMutex    sync.Mutex
+	registered    int
+}
+
+var pippetRunnerInstance PippetRunner
```
purpleidea: global vars (if required) should be at the top of the file, right after imports. Basically so they're easily seen, since they can be problematic if you aren't expecting them.

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+	if pippetRunnerInstance == nil {
+		pippetRunnerOnce.Do(func() { pippetRunnerInstance = &pippetReceiver{} })
+	}
+	return pippetRunnerInstance, nil
+}
+
+// Init runs the Puppet process that will perform the work of synchronizing resources
+// that are sent to its stdin. The process will keep running until Close is called.
+// Init should not be called directly. It is implicitly called by the Register function.
+func (obj *pippetReceiver) Init() error {
```
purpleidea: method receivers should follow their struct... Keep them all together.

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+	cmd := exec.Command("puppet", "yamlresource", "receive", "--color=no")
+	var err error
+	obj.stdin, err = cmd.StdinPipe()
+	if err != nil {
+		return err
+	}
+	obj.stdout, err = cmd.StdoutPipe()
+	if err != nil {
+		obj.stdin.Close()
```
purpleidea: err ?

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+		return err
+	}
+	if err := cmd.Start(); err != nil {
+		obj.stdin.Close()
```
purpleidea: err?

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+		return err
+	}
+	return nil
+}
+
+// Register should be called by any user (i.e., any pippet resource) before using
+// the Apply function on this receiver object.
+// Register implicitly takes care of calling Init if required.
+func (obj *pippetReceiver) Register(res interface{}) error {
```
purpleidea: The input isn't even used in this function... ?

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
...
+	if obj.registered > 1 {
+		obj.registerMutex.Unlock()
+		return nil
+	}
+	// count was increased from 0 to 1, we need to (re-)init
+	var err error
+	if err = obj.Init(); err != nil {
+		obj.registered = 0
+	}
+	obj.registerMutex.Unlock()
```
purpleidea: Use defer to Unlock instead.

engine/resources/pippet_run.go
```diff
@@ -0,0 +1,160 @@
+// Mgmt
```
purpleidea: I'm not sure that this needs to be in a separate file. Unless you have a good reason, put the contents at the bottom of the pippet.go resource file.

engine/resources/pippet.go
ffrank: Fixed.

engine/resources/pippet.go
ffrank: Oops, sorry. That was from an earlier test I did. Removed.

I would not expose this resource type to the language until we get a specific feature request for this. It should be considered an integral part of the Puppet Crutch.

engine/resources/pippet.go
ffrank: Done.

engine/resources/pippet.go
ffrank: No. You're right. Fixed.

engine/resources/pippet.go
ffrank: I went for `return obj.runner.Unregister(obj)`. Is that fine?

engine/resources/pippet.go
ffrank: Done.

engine/resources/pippet.go
ffrank: Done.

engine/resources/pippet.go
ffrank: I don't think it's that simple. Given two arbitrary YAML hashes, you will be hard pressed to make a call on their being equivalent without actually parsing the YAML. And Go cannot casually do this.

engine/resources/pippet.go
ffrank: Ended up removing this note.
Upon further thought, it's fine to use type/title as the UID, because these resources are expected to be generated from Puppet code, which will not allow this combination to be duplicated, anyway.

engine/resources/pippet_run.go
ffrank: It will make testing much easier :)

engine/resources/pippet_run.go
ffrank: It's a good point. The result from Puppet's sync operation is unmarshalled from JSON, in PippetResult form, and that *has* to include any errors that happened (we have no other way to get diagnostics from Puppet to mgmt).

Of course, we could go ahead and extract the interesting fields within the Apply function, and then return something that needs not be a PippetResult plus an error. Would you prefer that?

engine/resources/pippet_run.go
ffrank: It's subtle, so I'm not sure we need it.

"Failed" is a flag that puppet can set during the sync operation. E.g., a *mount* resource could not mount the file system.

"Error" is a flag my own module sets to indicate that there was an issue with the input, e.g. a YAML parsing error, or an invalid resource type.

"Exception" is the message that describes what went wrong specifically in case Error==true.

A failed resource does not fill the Exception field. Any diagnostics will be written to stderr by Puppet itself, and we don't bother trying to catch those atm. The result is some red Puppet output mixed into mgmt's own output. This is not future-proof, but I would prefer not to over-engineer this detail right now.

engine/resources/pippet_run.go
ffrank: Fixed.

engine/resources/pippet_run.go
purpleidea: If so, then can you write a small stub test to show how it's useful/easier that way please?

engine/resources/pippet_run.go
purpleidea: I think you want (PippetResult, error) here... Let's see what that looks like.

engine/resources/pippet_run.go
ffrank: FIxed.

engine/resources/pippet_run.go
ffrank: Fixed.

engine/resources/pippet_run.go
ffrank: True, it's just flavor now - the resource registers "itself" even though the runner object uses a simple counter.
I removed it.

engine/resources/pippet_run.go
ffrank: D'oh. So much nicer now. Fixed.

engine/resources/pippet_run.go
ffrank: Done.

engine/resources/pippet.go
ffrank: Done.

engine/resources/pippet.go
ffrank: Done.

purpleidea:
@ffrank Is everything addressed and ready for a new review?

engine/resources/pippet_run.go
ffrank: I reworked the function, and it became apparent that returning (bool, error) is sufficient (I feel).

Let me know what you think. Keeping this in a separate commit for now, so you can review the change itself if you wish.

engine/resources/pippet_run.go
ffrank: I think this is "fine" now because the Apply function was reworked. (?)

ffrank:
I have yet to add a test. The rest is addressed, I believe.

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+
+var pippetRunnerInstance PippetRunner
+var pippetRunnerOnce sync.Once
+
+func init() {
+	engine.RegisterResource("pippet", func() engine.Res { return &PippetRes{} })
+}
+
+// PippetRes is a wrapper resource for puppet.  It implements the functional
+// equivalent of an exec resource that calls "pupept resource <type> <title>
```
purpleidea: pupept

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+func (obj *PippetRes) Init(init *engine.Init) error {
+	obj.init = init // save for later
+
+	var err error
+	obj.runner, err = PippetRunnerInstance()
+	if err != nil {
+		return err
+	}
+
+	if err = obj.runner.Register(); err != nil {
```
purpleidea: `err :=` inside a new scope

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+	case <-obj.init.Done: // closed by the engine to signal shutdown
+	}
+
+	//obj.init.Event() // notify engine of an event (this can block)
+
+	return nil
+}
+
+// CheckApply synchronizes the resource if required.
+func (obj *PippetRes) CheckApply(apply bool) (checkOK bool, err error) {
```
purpleidea: Get rid of the named return values please to be consistent with https://github.com/purpleidea/mgmt/commit/e4d874cc69837784527ae79044d7c0cd0e7ebdbc

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+		BaseUID:       engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
+		resourceType:  obj.Type,
+		resourceTitle: obj.Title,
+	}
+	return []engine.ResUID{x}
+}
+
+// GroupCmp returns whether two resources can be grouped together or not.
+func (obj *PippetRes) GroupCmp(r engine.GroupableRes) error {
+	return fmt.Errorf("pippet resources are never grouped")
```
purpleidea: Just omit the method then.

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+func (obj *pippetReceiver) Init() error {
+	cmd := exec.Command("puppet", "yamlresource", "receive", "--color=no")
+	var err error
+	obj.stdin, err = cmd.StdinPipe()
+	if err != nil {
+		return err
+	}
+	obj.stdout, err = cmd.StdoutPipe()
+	if err != nil {
+		err = errwrap.Append(err, obj.stdin.Close())
```
purpleidea: In this case you can just `return errwrap.Append(err, obj.stdin.Close())`

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+		return err
+	}
+	if err = cmd.Start(); err != nil {
+		err = errwrap.Append(err, obj.stdin.Close())
```
purpleidea: same as above

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+		return err
+	}
+	buf := make([]byte, 80)
+	if _, err = obj.stdout.Read(buf); err != nil {
```
purpleidea: errwrap allows you to inline these...

iow:

```
err = obj.stdout.Read(buf)
err = errwrap.Append(err, obj.stdin.Close())
```

IIRC

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+	return err
+}
+
+// Apply does the actual work of making Puppet synchronize a resource.
+func (obj *pippetReceiver) Apply(resource *PippetRes) (changed bool, err error) {
+	obj.applyMutex.Lock()
+	defer obj.applyMutex.Unlock()
+	if err = json.NewEncoder(obj.stdin).Encode(resource); err != nil {
+		return false, fmt.Errorf("failed to send resource to puppet: %v", err)
```
purpleidea: errwrap.Wrapf instead

engine/resources/pippet.go
```diff
@@ -0,0 +1,325 @@
...
+func (obj *pippetReceiver) Unregister() error {
+	obj.registerMutex.Lock()
+	defer obj.registerMutex.Unlock()
+	obj.registered = obj.registered - 1
+	var err error
+	if obj.registered == 0 {
+		err = obj.Close()
+	}
+	if obj.registered < 0 {
+		err = errwrap.Append(err, fmt.Errorf("pippet runner: ERROR: unregistered more resources than were registered"))
```
purpleidea: s/....

```
e := fmt.Errorf("pippet runner: unregistered more resources than were registered")
err = errwrap.Append(err, e)
```

purpleidea:
(Also, please rebase to master.)

ffrank:
Heyyyy, so, this sat for way too long. I tuned out of it for a while there, mulling around mocks and tests in my head. I think at one point you had challenged me to show why I wanted that interface in there so badly? I don't clearly recall, and cannot find the comment anymore.

So I refactored a bit so that I can now use Dependency Injection for testing. I'm rather pleased with it. Let me know your thoughts.

I also rebased and tried to act on all your comments. There is a separate commit up top that has all the changes regarding the review comments.

Thanks for the patience!

purpleidea:
Ping when the basic tests pass and I'll review. There is one race that occasionally causes the shell tests to fail. You can ignore that. (It's in the middle group.)

ffrank:
@purpleidea this is ready for another round of review. Thanks in advance.

engine/resources/pippet.go
```diff
@@ -0,0 +1,335 @@
...
+)
+
+var pippetReceiverInstance *pippetReceiver
+var pippetReceiverOnce sync.Once
+
+func init() {
+	engine.RegisterResource("pippet", func() engine.Res { return &PippetRes{} })
+}
+
+// PippetRes is a wrapper resource for puppet.  It implements the functional
```
purpleidea: We only use single spaces. Both here and in word processors =D

engine/resources/pippet.go
```diff
@@ -0,0 +1,335 @@
...
+// process that receives resources through a pipe (hence the name).
+type PippetRes struct {
+	traits.Base // add the base methods without re-implementation
+	traits.Refreshable
+
+	init *engine.Init
+
+	// Type is the exact name of the wrapped Puppet resource type, e.g.
+	// "file", "mount". This needs not be a core type. It can be
+	// a type from a module. The Puppet installation local to the mgmt
```
purpleidea: All docstrings *must* be wrapped to 80 chars. This means 80, not less than 80, not more than 80. Please check the entire patch for this. I recommend using an editor that either does this for you and/or adds a line at 80 chars. Sadly we don't have an automatic test for this yet. Please write one =D

ffrank:
I wrapped the docstrings to the best of my ability.
I will squash all the commits once you feel this is ready for merge.

ffrank:
@purpleidea ready for another round whenever you are.

engine/resources/pippet.go
ffrank: I may take that challenge some day.

But not today. :)

purpleidea:
@ffrank Sorry for the delay, as you know I was sick, but I'm back and catching up on life.

If you rebase to git master, the tests should look a lot better now. I fixed the bbolt issue mostly.

There's still one failing thing of importance, but at least the whole thing is failing less and only in that one test mostly.

Also etcd logging changed in the new version so that's super ugly and needs fixing too :( Patches welcome!

ffrank:
@purpleidea great, thanks. Pushed the rebase. Tests fail pretty badly indeed, but it does build 👍 Cheers!

purpleidea:
Okay, sorry I was AFK for a bit, as I mentioned to you in IRC, I was quite sick :( Anyways, I decided to merge this and get it done. I squashed all your commits. I also fixed all the wrapping issues I saw at a quick glance, and added these (and a few tiny fixups) in a separate commit here: https://github.com/purpleidea/mgmt/commit/d34715b4ba5bfe9f53a3c5826de77b994fdbda33

Please have a look at figure out why your editor doesn't see these. To be fair, it's something we should totally have a test for. At least for the docstring parts. That could be pretty easy to write if we use the golang AST to pull out the docstrings and check that they're wrapped correctly. If you're in the mood, that would be a good patch for you to write!

Cheers!
James

PR #474, opened 2019-02-26
docs: Fix mode syntax in example resource

amousset:


purpleidea:
Unfortunately this is a bug in the resource code, that is missing this feature, and not in the docs :(

Patches welcome?? Thanks!!

PR #475, opened 2019-02-26
lang: funcs: Added a function to detect Debian and RedHat like systems

ncharles:
Adding methods to detect if OS is RedHat like or Debian like


purpleidea:
LGTM :)

Merci et Bienvenue a mgmt!

purpleidea:
\o/ Merged, thank you!

PR #476, opened 2019-02-27
resources: print: Add option RefreshOnly for print, to print only when notified

ncharles:
Add a RefreshOnly option on print to trigger the print only on nofity/listen


purpleidea:
LGTM :)

purpleidea:
Thanks for the patch! Merged =D

PR #479, opened 2019-02-28
engine: resources: exec: Add lang for all field

ncharles:


purpleidea:
LGTM :)

purpleidea:
@ncharles Sorry about the delay. I am almost done solving the problem. I did the much needed exec resource cleanups that I promised, and that makes this work properly. (Will make sense shortly.)

Patch is almost done, but I need to do one last thing. You can look at the recent feature branch to see what it's approximately going to look like. Just need to get WatchCmd to fail early if the command isn't present!

purpleidea:
@ncharles Exec has been fixed up considerably, so it should make it much easier for you to rebase your patch to current git master and not have any more issues. Thanks for your patience.

purpleidea:
Since this no longer applies cleanly, and I haven't heard anything recently, I'm going to close for the time being, but please feel free to re-open in a new PR when ready. Thanks!

PR #481, opened 2019-02-28
docs: Slightly clarify parameter defaults

ffrank:


purpleidea:
@ffrank Thanks for the patch! One subtlety change:

>  makes no sense (e.g., the `state` parameter of the `file` resource).

It's not that it the zero value makes no sense, and in fact some resources have a zero value that doesn't Validate for example (eg: `cwd` in exec, so you must specify it) but rather it's that Default produces a struct with the correct values to use when those parameters are omitted. IOW, if you had:

```
launch_missile "missile1" {
    safety => true,
}
```

In this case we want the `safety` to be true most of the time, and if it was omitted, it would be `false` (the zero value) so since it's better to name it `safety` rather than `no_safety`, we keep it as such, and ensure that the Default() function specifies it as `true`. That way if we omit the option, it will be true. We'd have to write `safety => false` if we wanted it off.

HTH

ffrank:
Makes sense.

ffrank:
Added more prose.

docs/resource-guide.md
```diff
@@ -99,6 +99,14 @@ This returns a populated resource struct as a `Res`. It shouldn't populate any
 values which already have the correct default as the golang zero value. In
 general it is preferable if the zero values make for the correct defaults.
 
+It should be noted that the default values are not expected to form a resource
+that is sensible or functional. Only fill values for which the zero value
+makes no sense (e.g., the `state` parameter of the `file` resource).
```
purpleidea: As alluded to in my earlier comment, I don't think this part of this particular sentence is correct.

docs/resource-guide.md
```diff
@@ -99,6 +99,14 @@ This returns a populated resource struct as a `Res`. It shouldn't populate any
...
+Even if the zero value is valid, you should also define a default value here
```
purpleidea: s/you should also/you may/ please

docs/resource-guide.md
```diff
@@ -99,6 +99,14 @@ This returns a populated resource struct as a `Res`. It shouldn't populate any
...
+in situations where this zero value violates the principle of least surprise.
+For example, all boolean flags default to `false`. This may be counter-intuitive
+or dangerous in some contexts. You should add a `true` default then.
```
purpleidea: s/should/could/

docs/resource-guide.md
ffrank: "Generally, you will want to fill values..." ?

purpleidea: I don't want to write it for you... ;) I was hoping my super long comment in https://github.com/purpleidea/mgmt/pull/481#issuecomment-468481155 was helpful enough to give you the knowledge to be able to rephrase into "human" since I suck at docs :(

ffrank:
How about it @purpleidea ?

purpleidea:
@ffrank Sorry, I don't know why I didn't see your earlier notification. Thanks for the ping.

Patch LGTM :)

I'll merge shortly after I finish a big rebase/branch that I don't want to stash atm.
(In other news, it would be awesome if editors could restore your ^Z history and cursor position if you reload a file with the same sha1.)

ffrank:
Hmm fascinating:
https://vim.fandom.com/wiki/Restore_cursor_to_file_position_in_previous_editing_session
https://jovicailic.org/2017/04/vim-persistent-undo/

purpleidea:
Merged, thanks!

PR #492, opened 2019-03-23
uptime_fact_posix.go: Remove deprecated syscall import

mfossen:
According to https://godoc.org/syscall the `syscall` package is deprecated, and points to using https://godoc.org/golang.org/x/sys instead.

I stumbled into looking at the `syscall` package in the first place due to #493. 

I've updated the import in a single file for now to see what your thoughts are on it and if it should be done with the other `syscall` imports or not.

### DIFF
--- a/lang/funcs/core/sys/uptime_fact_posix.go
+++ b/lang/funcs/core/sys/uptime_fact_posix.go
@@ -20,12 +20,12 @@
 package coresys
 
 import (
-	"syscall"
+	"golang.org/x/sys/unix"
 )
 
 func uptime() (int64, error) {
-	var sysinfo syscall.Sysinfo_t
-	if err := syscall.Sysinfo(&sysinfo); err != nil {
+	var sysinfo unix.Sysinfo_t
+	if err := unix.Sysinfo(&sysinfo); err != nil {
 		return 0, err
 	}
 	return sysinfo.Uptime, nil


purpleidea:
@mfossen Thanks for the patch!

I suppose this makes sense, so happy to apply. Patch is failing tests because you need to `gofmt -w` your file that you changed and the commit message needs to be in the official format, see here:

https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Please force push a new version, thanks!

purpleidea:
Rebased to master and merged as 7db3e8556a0cc6215d48f9212a1fc0620c635a19

Thanks and welcome to the club! :)

(Random GitHub specific tip: When you force push a new commit, it doesn't send maintainers a new notification, so leaving a small "pushed" comment or similar will let me re-review your patch quicker, so sorry for the delay!)

PS: We seem to use `"syscall"` in a number of other places. If you'd like to port more of them, and add a test in https://github.com/purpleidea/mgmt/blob/7db3e8556a0cc6215d48f9212a1fc0620c635a19/test/test-govet.sh#L52 that would be awesome!

mfossen:
Perfect, thanks! Also no worries for the delay, that's my fault I got busy and forgot to come back and ping you on it :grimacing: 

Yep I'll definitely go through and switch out the other `syscall` imports and check for them!

Issue #381, opened 2018-04-10
[LOVE] Add a new function to the mgmt language


purpleidea:
This is a beginner to intermediate level patch for someone who is either new to golang or mgmt. It is meant to encourage new programmers to get involved with mgmt. It will probably take you a few hours to a day or so to complete, depending on what function(s) you choose to implement.

## Background

Mgmt has a language. For more on that please read: https://purpleidea.com/blog/2018/02/05/mgmt-configuration-language/

The language has a number of functions in it which can be used. We'd like to implement more. Your patch for this issue should initially propose a design, and then once discussed and accepted should follow up with a patch. Your patch may add a single function, or it may add a family of related functions.

## Technical internals

To implement a function in mgmt, there are a few different ways to do so. You might want to start by reading the https://github.com/purpleidea/mgmt/blob/master/docs/language-guide.md and then the https://github.com/purpleidea/mgmt/blob/master/docs/function-guide.md 

You'll probably want to use either the 
https://github.com/purpleidea/mgmt/blob/master/docs/function-guide.md#simple-function-api
or the:
https://github.com/purpleidea/mgmt/blob/master/docs/function-guide.md#simple-polymorphic-function-api

Examples can be seen in the source at:
https://github.com/purpleidea/mgmt/tree/master/lang/funcs/simple
and at:
https://github.com/purpleidea/mgmt/tree/master/lang/funcs/simplepoly
respectively.

## Next steps

When you're ready to implement a function, open up a new issue describing the function(s) you'd like to implement. After we all agree it makes sense, please send a patch! Please read the contributing tips:
https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

## Ideas?

Here are some ideas for functions that need implementing. (Basically anything you'd expect in a stdlib.)
* [ ] datetime related functions
* [ ] math related functions
* [ ] string manipulation functions
* [ ] add your idea here...

Happy hacking, and ping me for questions!

## Bonus steps

* Include tests for your function with your patch
*

purpleidea:
Functions we'd like to have:

`float`: converts an int to a float. You'll want to add an entry like this in the parser:
```
|	FLOAT_IDENTIFIER OPEN_PAREN expr CLOSE_PAREN
	{
		posLast(yylex, yyDollar) // our pos
		$$.expr = &ExprCall{
			Name: floatFuncName,
			Args: []interfaces.Expr{
				$3.expr,
			},
		}
	}
```
[...]

PR #495, opened 2019-03-25
imports: Use golang.org/x/sys/unix in place of syscall where possible

mfossen:
## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

## Thanks for contributing to mgmt and welcome to the team!

### DIFF
--- a/engine/resources/exec.go
+++ b/engine/resources/exec.go
@@ -33,6 +33,8 @@ import (
 	"github.com/purpleidea/mgmt/engine/traits"
 	engineUtil "github.com/purpleidea/mgmt/engine/util"
 	"github.com/purpleidea/mgmt/util/errwrap"
+
+	"golang.org/x/sys/unix"
 )
 
 func init() {
@@ -182,7 +184,7 @@ func (obj *ExecRes) Watch() error {
 		cmd := exec.CommandContext(ctx, cmdName, cmdArgs...)
 		cmd.Dir = obj.WatchCwd // run program in pwd if ""
 		// ignore signals sent to parent process (we're in our own group)
-		cmd.SysProcAttr = &syscall.SysProcAttr{
+		cmd.SysProcAttr = &unix.SysProcAttr{
 			Setpgid: true,
 			Pgid:    0,
 		}
@@ -217,7 +219,7 @@ func (obj *ExecRes) Watch() error {
 					return errwrap.Wrapf(err, "unknown error")
 				}
 				pStateSys := exitErr.Sys() // (*os.ProcessState) Sys
-				wStatus, ok := pStateSys.(syscall.WaitStatus)
+				wStatus, ok := pStateSys.(unix.WaitStatus)
 				if !ok {
 					return errwrap.Wrapf(err, "error running cmd")
 				}
@@ -278,7 +280,7 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 		cmd := exec.Command(cmdName, cmdArgs...)
 		cmd.Dir = obj.IfCwd // run program in pwd if ""
 		// ignore signals sent to parent process (we're in our own group)
-		cmd.SysProcAttr = &syscall.SysProcAttr{
+		cmd.SysProcAttr = &unix.SysProcAttr{
 			Setpgid: true,
 			Pgid:    0,
 		}
@@ -301,7 +303,7 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 				return false, err
 			}
 			pStateSys := exitErr.Sys() // (*os.ProcessState) Sys
-			wStatus, ok := pStateSys.(syscall.WaitStatus)
+			wStatus, ok := pStateSys.(unix.WaitStatus)
 			if !ok {
 				return false, errwrap.Wrapf(err, "error running cmd")
 			}
@@ -369,7 +371,7 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 	cmd := exec.CommandContext(ctx, cmdName, cmdArgs...)
 	cmd.Dir = obj.Cwd // run program in pwd if ""
 	// ignore signals sent to parent process (we're in our own group)
-	cmd.SysProcAttr = &syscall.SysProcAttr{
+	cmd.SysProcAttr = &unix.SysProcAttr{
 		Setpgid: true,
 		Pgid:    0,
 	}
@@ -423,7 +425,7 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 	exitErr, ok := err.(*exec.ExitError) // embeds an os.ProcessState
 	if err != nil && ok {
 		pStateSys := exitErr.Sys() // (*os.ProcessState) Sys
-		wStatus, ok := pStateSys.(syscall.WaitStatus)
+		wStatus, ok := pStateSys.(unix.WaitStatus)
 		if !ok {
 			return false, errwrap.Wrapf(err, "error running cmd")
 		}
@@ -434,7 +436,7 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 		sig := wStatus.Signal()
 
 		// we get this on timeout, because ctx calls cmd.Process.Kill()
-		if sig == syscall.SIGKILL {
+		if sig == unix.SIGKILL {
 			return false, errwrap.Wrapf(err, "cmd timeout, exit status: %d", exitStatus)
 		}
 
--- a/engine/resources/file.go
+++ b/engine/resources/file.go
@@ -29,7 +29,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"syscall"
 
 	"github.com/purpleidea/mgmt/engine"
 	"github.com/purpleidea/mgmt/engine/traits"
@@ -37,6 +36,8 @@ import (
 	"github.com/purpleidea/mgmt/recwatch"
 	"github.com/purpleidea/mgmt/util"
 	"github.com/purpleidea/mgmt/util/errwrap"
+
+	"golang.org/x/sys/unix"
 )
 
 func init() {
@@ -125,7 +126,7 @@ func (obj *FileRes) Validate() error {
 		if err != nil {
 			return fmt.Errorf("can't stat root to get system information")
 		}
-		_, ok := fileInfo.Sys().(*syscall.Stat_t)
+		_, ok := fileInfo.Sys().(*unix.Stat_t)
 		if !ok {
 			return fmt.Errorf("can't set Owner or Group on this platform")
 		}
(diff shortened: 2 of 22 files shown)

### CODE REVIEW

**Comment by purpleidea** [comment]
Sweet... Before I review, I should mention this will fail CI because of the commit format.

File: test/test-govet.sh [hunk 1]
```diff
@@ -65,6 +65,18 @@ function consistent-imports() {
 	if grep $'\t"github.com/purpleidea/mgmt/engine/util"' "$1"; then	# import as engineUtil
 		return 1
 	fi
+
+	# import as golang.org/x/sys/unix
+	if grep -E "^[[:space:]]+.syscall." "$1"; then
```
mfossen: There are 3 files that use `syscall` features and `unix` doesn't have a drop-in replacement for them. For now, I have it checking for those cases and returning 0 if it finds them.

mfossen:
Is the topic invalid?

purpleidea:
commit message titles must be in the form:

topic: Capitalized message with no trailing period

test/test-govet.sh
purpleidea: Fantastic!

mfossen:
Ah got it, overlooked the lowercase `u`. Fixed and force-pushed. Thanks!

engine/resources/exec.go
```diff
@@ -29,6 +29,8 @@ import (
 	"syscall"
 	"time"
 
+	"golang.org/x/sys/unix"
```
purpleidea: Imports go in three groups: with the middle group being our internal code... This needs to go in the third at the end. Here and elsewhere...

https://github.com/golang/go/wiki/CodeReviewComments#imports

lib/cli.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/purpleidea/mgmt/bindata"
 	"github.com/purpleidea/mgmt/gapi"
+
```
purpleidea: What's this?

test/test-govet.sh
purpleidea: Why is this DOT syscall DOT ? (what's the leading period for?)

Perhaps it's also worth it to match as done in the import example to guard against aliases too?

purpleidea (changes requested):
Thanks for the patch! Here's the review. Very close, thanks again!

lib/cli.go
mfossen: Looks like `gofmt -w lib/cli.go` is fine with no line between that and the comment, but `make gofmt` adds it in.

test/test-govet.sh
mfossen: Looking at this again, it doesn't make a lot of sense. I changed it to match for aliases too, check it out and see if that looks alright.

mfossen:
@purpleidea Made some fixes, let me know how that looks. Thanks for the review!

test/test-govet.sh
mfossen: Also suppressing output of the `grep` unless an actual bad use of `syscall` is found, then print that out with the filename.

lib/cli.go
purpleidea: Not in my version of golang. They've been known to break compatibility before. For now we're on 1.10.x. Please remove this one-line nl addition. If there's a change in `gofmt` we can address that in a separate commit. We'll be moving to 1.11.x soon anyways. (We basically track the version in Fedora stable -1.)

test/test-govet.sh
```diff
@@ -65,6 +65,19 @@ function consistent-imports() {
...
+
+	# import as golang.org/x/sys/unix
+	if grep -q -E $'\t\w*[[:space:]]*"syscall"' "$1"; then
+		if grep -q "syscall.Credential" "$1"; then
+			return 0  # Credential not available in golang.org/x/sys/unix
+		fi
+		if grep -q "syscall.NetlinkMessage"; then
+			return 0 # NetlinkMessage not available in golang.org/x/sys/unix
+		fi
+		grep -Hn -E $'\t\w*[[:space:]]*"syscall"' "$1"
```
purpleidea: What's this line for?

test/test-govet.sh
```diff
@@ -65,6 +65,19 @@ function consistent-imports() {
 	if grep $'\t"github.com/purpleidea/mgmt/engine/util"' "$1"; then	# import as engineUtil
 		return 1
 	fi
+
+	# import as golang.org/x/sys/unix
+	if grep -q -E $'\t\w*[[:space:]]*"syscall"' "$1"; then
```
purpleidea: If we're doing it this way, rather than checking for the actual import, then we'd want `syscall.` (with a trailing explicit period, so maybe `\.` ?)

purpleidea (changes requested):
Patch looks perfect except for a pesky additional nl (as commented.) Also need some slight tweaking or clarification in the test. Thanks again!

test/test-govet.sh
mfossen: Ah, I think I understand it now, updated!

test/test-govet.sh
mfossen: Added a comment for it.

mfossen:
@purpleidea Not sure if you get notified on review comments or not, made some changes so let me know how they look!

purpleidea:
@mfossen Assuming this passes tests, then LGTM =D Thanks!

engine/resources/file.go
```diff
@@ -371,7 +372,7 @@ func (obj *FileRes) fileCheckApply(apply bool, src io.ReadSeeker, dst string, sh
 	}
 
 	// TODO: attempt to reflink with Splice() and int(file.Fd()) as input...
-	// syscall.Splice(rfd int, roff *int64, wfd int, woff *int64, len int, flags int) (n int64, err error)
```
purpleidea: nice! =D

test/test-govet.sh
```diff
@@ -65,6 +65,20 @@ function consistent-imports() {
...
+
+	# import as golang.org/x/sys/unix
+	if grep -q -E $'syscall\.' "$1"; then
+		if grep -q "syscall.Credential" "$1"; then
+			return 0  # Credential not available in golang.org/x/sys/unix
+		fi
+		if grep -q "syscall.NetlinkMessage"; then
+			return 0 # NetlinkMessage not available in golang.org/x/sys/unix
+		fi
+		# Print out the filename and line number of the improper syscall import
```
purpleidea: One nit: comments should be either full sentences (which end with periods) or not capitalized one-liners. I'd recommend removing the capitalization of Print.

purpleidea:
Note your test has caused a timeout. Test it, it won't work. Hint:

>  		if grep -q "syscall.NetlinkMessage"; then

test/test-govet.sh
mfossen: :+1: Made it lowercase.

mfossen:
Ugh, sorry missed that one. Fixed!

purpleidea (approved):

purpleidea:
Just waiting for tests, all looks good, thanks!

purpleidea:
As an aside (and feel free to reply privately over IRC if you prefer) while this is very helpful, and we appreciate your contributions, if you'd be interesting in working on something more interesting and or involved, we're happy to make more suggestions, and mentor you if you require it. I don't know if you're new to golang and looking to do more, or a seasoned programmer contributing on the side. In any case, all are welcome, let me know! =D

mfossen:
@purpleidea For sure! I'm pretty new to golang, I've just been kicking it around with some simple stuff the past two or three weeks and liking it so far. I've been following `mgmt` for awhile and think the idea is awesome and really want to use it for some deployment/ops stuff, so figured I'd try contributing a little if I can.

Any suggestions for where to start? Or I figured just get into the issues backlog and pick up some easier ones to begin.

Also ugh, looks like the basic test timed out again, was that something from my PR?

purpleidea:
@mfossen I think it timed out because of a GH bug... One sec.

purpleidea (dismissed):
Just trying to workaround a GH bug. No changes necessary atm.

purpleidea:
Correct test running at: https://travis-ci.org/purpleidea/mgmt/builds/511307496
I think the bug is that when I clicked "approve" before I saw that hang bug, it refused to look at (and test) new builds. That's the fail when you depend on proprietary sw :/

purpleidea:
> I've been following mgmt for awhile and think the idea is awesome and really want to use it for some deployment/ops stuff, so figured I'd try contributing a little if I can.

Awesome!

> Any suggestions for where to start? Or I figured just get into the issues backlog and pick up some easier ones to begin.

It all depends on your general programming experience and your golang experience. (We don't want to overwhelm new contributors with something too complex, because then it's not fun.)

How familiar would you rate your experience with channels and goroutines?

(There are some good projects listed in https://github.com/purpleidea/mgmt/issues/381 -- we can help match you with one if you're interested.)

mfossen:
Sweet! I'm not super familiar with them, I've recently been reading up and doing the go by example lessons and think I get the basic idea. I have more of a background with python and Java so right now I've more been  wrapping my head around how Go does things compared to those.

purpleidea:
@mfossen I would recommend going through the https://tour.golang.org/ (don't spend more than 5-15 min on any one problem, just get the main gist. Spend a bit more time on the concurrency parts.)

Some bad news:

https://travis-ci.org/purpleidea/mgmt/jobs/511307497#L1013

Something seems to be causing the tests to hang. Is it possible one of these changes is responsible?

purpleidea:
(Note: you can run the test locally to narrow down which test(s) are hanging...) That's useful information.

mfossen:
@purpleidea Will do! Also I used a Docker container set up with `golang:1.11` and mounted my local `mgmt` directory into it, ran `make deps`, ``make build`, `make test` and I'm able to run the tests, and `make govet` runs successfully.

purpleidea:
It's failing in go-test.sh I think.

mfossen:
Looks like `TestResources1/test_#0_(simple_exec)` is the one that's hanging and timing out. I'm trying to track down why that would be.

purpleidea:
> TestResources1/test_#0_(simple_exec)

When you test it locally does it also fail/hang? When you try on git master does it always pass?

LMK, thanks!

purpleidea:
Note: you can run individual tests with go get github.com/purpleidea/mgmt/whateverthepackageisnamed/ -v -r TestResources1/test_#0_

purpleidea:
I rebased this to master which has a lot of stability fixes and it's now testing at: 

https://travis-ci.org/purpleidea/mgmt/builds/519010924

purpleidea:
This is a total mystery! I can confirm that with the updated patch it fails in the travis test and locally! Although I have no idea what it would have changed. Ghosts! Anyone with more expertise in the difference between syscall and unix package, please LMK!

engine/resources/exec.go
```diff
@@ -182,7 +184,7 @@ func (obj *ExecRes) Watch() error {
 		cmd := exec.CommandContext(ctx, cmdName, cmdArgs...)
 		cmd.Dir = obj.WatchCwd // run program in pwd if ""
 		// ignore signals sent to parent process (we're in our own group)
-		cmd.SysProcAttr = &syscall.SysProcAttr{
+		cmd.SysProcAttr = &unix.SysProcAttr{
```
shibumi: shouldn't it be syscall.SysProcAttr here instead of unix.SysProcAttr? At least according to the unix documentation: https://godoc.org/golang.org/x/sys/unix#SysProcAttr

purpleidea: I've tried reverting that part of the change and also the waitstatus parts. It still doesn't work.
So it's at least something *else* in the patch.

purpleidea:
I've merged a simple version of this, mostly to get it done. I don't think there's much left here, and I admit I don't truly understand some of the implications, but thanks for bringing up the topic! If you have new work please rebase and submit! Thanks!

PR #505, opened 2019-04-22
Removed obsolete no-watch-config flag

wdullaer:
I removed the "no-watch-config" flag, since it was only ever written and never read.
Having it around creates the expectation that by default mgmt will put a watch on the config.

The test suite still passes locally.

@purpleidea let me know what you think.

### DIFF
--- a/lib/cli.go
+++ b/lib/cli.go
@@ -73,10 +73,6 @@ func CLI(program, version string, flags Flags) error {
 			Usage: "do not update graph under any switch events",
 		},
 		cli.BoolFlag{
-			Name:  "no-config-watch",
-			Usage: "do not update graph on config switch events",
-		},
-		cli.BoolFlag{
 			Name:  "no-stream-watch",
 			Usage: "do not update graph on stream switch events",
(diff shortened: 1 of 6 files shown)


lib/cli.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/purpleidea/mgmt/bindata"
 	"github.com/purpleidea/mgmt/gapi"
+
```
purpleidea: nack this nl change... this might be due to change in gofmt, we're still on 1.10.x but will be moving to 1.11.x shortly. For now, please don't include this.

wdullaer: amended the commit

purpleidea:
Please read: https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md#tips specifically the part about the commit message.

wdullaer:
The new commit message should be in line with the style guide

purpleidea:
Merged, thank you and welcome to the club! =D

PR #511, opened 2019-04-24
Add uptime implementation for macOS

aequitas:
This fixes build for macOS.

Example taken from: https://github.com/mackerelio/go-osstat/blob/master/uptime/uptime_bsd.go

### DIFF
--- a/lang/funcs/core/sys/uptime_fact_darwin.go
+++ b/lang/funcs/core/sys/uptime_fact_darwin.go
@@ -1,0 +1,47 @@
+// Mgmt
+// Copyright (C) 2013-2019+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// +build darwin
+
+package coresys
+
+import (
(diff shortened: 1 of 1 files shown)


purpleidea:
Patch looks perfect, however as per your comment:

> Example taken from: https://github.com/mackerelio/go-osstat/blob/master/uptime/uptime_bsd.go

This means that code is under a different license, and we can't copy it without attribution. If you can get permission from that original author to allow your patch as-is in mgmt, then this is a +1 from me.

It's just on the borderline of sufficiently small that it shouldn't necessarily be copyrightable, but nevertheless, we need to be safe, sorry :(

purpleidea:
A git blame looks like it's from @itchyny -- do we have your permission to merge this as-is and assign the copyright to @purpleidea ?

aequitas:
@purpleidea there is also a near identical implementation of the same code over here: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/util/boottime_util_darwin.go Also apache licence though. I think this is as generic as this implementation can get, like you say on the border of copyrightable. 

I reduced some of the conversions as we handle it as epoch anyways.

lang/funcs/core/sys/uptime_fact_darwin.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+func uptime() (int64, error) {
+	out, err := unix.SysctlRaw("kern.boottime")
+	if err != nil {
+		return 0, err
+	}
+	var boottime syscall.Timeval
+	if len(out) != int(unsafe.Sizeof(boottime)) {
+		return 0, fmt.Errorf("Failed to get boottime")
```
purpleidea: error messages should not be capitalized. and we should say "uptime" not boottime.

aequitas: well technically its the time-at-boot we failed to fetch. It's only later that it's converted into the uptime by subtracting it from the current time since epoch.

purpleidea:
> Also please change commit to lang: funcs: ... Thanks!

@aequitas

purpleidea:
Okay, since I don't think it's possible to really simplify this new version much more since there's really only one logical way to do this, I'm not sure it's applicable for copyright, especially since it's so small. In any case, I rebased to master and merged. Thanks!

PR #513, opened 2019-05-03
Support official builds on supported LTS Ubuntu/Debian releases

aequitas:
These are the current LTS releases of Debian/Ubunutu (https://wiki.debian.org/LTS/, https://wiki.ubuntu.com/Releases): Debian 8, Debian 9, Ubuntu 16.04, Ubuntu 18.04.

I think it would be nice if mgmt would support these with the official builds. The current build in Github releases does not work on any of these OS releases because it is linked against libraries not available on these systems [0].

This PR adds a test script to verify if a `.deb` release build under `releases/` will install and run on these supported LTS releases.

You can run tests using: 

     git tag -f 0.0.0; make releases/0.0.0/deb/mgmt_0.0.0_amd64.deb; test/test-deb-releases.sh 

There is also a change to the build scripting included which allows to build the `.deb` release in a Debian 8 system which ensures it will install and run in all OS releases mentioned above. 

A branch with only the test scripting is available at: https://github.com/aequitas/mgmt/tree/deb-test

[0]
```
$ wget https://github.com/purpleidea/mgmt/releases/download/0.0.19/mgmt_0.0.19_amd64.deb
$ mv mgmt_0.0.19_amd64.deb releases/0.0.19/deb/ 
$ test/test-deb-releases.sh
Testing: ubuntu:19.04
+ apt-get update
+ dpkg -i /releases/0.0.19/deb/mgmt_0.0.19_amd64.deb
Selecting previously unselected package mgmt.
(Reading database ... 4058 files and directories currently installed.)
Preparing to unpack .../deb/mgmt_0.0.19_amd64.deb ...
Unpacking mgmt (0.0.19) ...
+ apt-get install -yqqf
debconf: delaying package configuration, since apt-utils is not installed
+ mgmt --version
mgmt version 0.0.19
Testing: ubuntu:18.04
+ apt-get update
+ dpkg -i /releases/0.0.19/deb/mgmt_0.0.19_amd64.deb
Selecting previously unselected package mgmt.
(Reading database ... 4038 files and directories currently installed.)
Preparing to unpack .../deb/mgmt_0.0.19_amd64.deb ...
Unpacking mgmt (0.0.19) ...
+ apt-get install -yqqf
debconf: delaying package configuration, since apt-utils is not installed
+ mgmt --version
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_4.1.0' not found (required by mgmt)
Testing: debian:9
+ apt-get update
+ dpkg -i /releases/0.0.19/deb/mgmt_0.0.19_amd64.deb
Selecting previously unselected package mgmt.
(Reading database ... 6498 files and directories currently installed.)
Preparing to unpack .../deb/mgmt_0.0.19_amd64.deb ...
Unpacking mgmt (0.0.19) ...
+ apt-get install -yqqf
debconf: delaying package configuration, since apt-utils is not installed
+ mgmt --version
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_4.1.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.9.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.7.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.1.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by mgmt)
Testing: ubuntu:16.04
+ apt-get update
+ dpkg -i /releases/0.0.19/deb/mgmt_0.0.19_amd64.deb
Selecting previously unselected package mgmt.
(Reading database ... 4768 files and directories currently installed.)
Preparing to unpack .../deb/mgmt_0.0.19_amd64.deb ...
Unpacking mgmt (0.0.19) ...
+ apt-get install -yqqf
debconf: delaying package configuration, since apt-utils is not installed
+ mgmt --version
mgmt: /usr/lib/x86_64-linux-gnu/libvirt-lxc.so.0: version `LIBVIRT_LXC_2.0.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_2.2.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_4.1.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_3.9.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_3.7.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_3.1.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_3.0.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_1.3.3' not found (required by mgmt)
mgmt: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_2.0.0' not found (required by mgmt)
Testing: debian:8
+ apt-get update
+ dpkg -i /releases/0.0.19/deb/mgmt_0.0.19_amd64.deb
Selecting previously unselected package mgmt.
(Reading database ... 7627 files and directories currently installed.)
Preparing to unpack .../deb/mgmt_0.0.19_amd64.deb ...
Unpacking mgmt (0.0.19) ...
+ apt-get install -yqqf
debconf: delaying package configuration, since apt-utils is not installed
+ mgmt --version
mgmt: /usr/lib/libvirt-lxc.so.0: version `LIBVIRT_LXC_2.0.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_2.2.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_1.2.11' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_1.2.12' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_4.1.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.9.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_1.2.16' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.7.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.1.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_1.2.15' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_1.2.19' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_1.2.14' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.0.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_1.3.3' not found (required by mgmt)
mgmt: /usr/lib/libvirt.so.0: version `LIBVIRT_2.0.0' not found (required by mgmt)
FAIL
The following tests failed:
 ubuntu:18.04 debian:9 ubuntu:16.04 debian:8
```

### DIFF
--- a/test/test-deb-releases.sh
+++ b/test/test-deb-releases.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# test if .deb package installs and mgmt runs on supported LTS Debian/Ubuntu
+
+set -o errexit
+set -o pipefail
+
+releases=(ubuntu:18.04 debian:9 ubuntu:16.04 debian:8 )
+
+. "$(dirname "$0")/util.sh"
+
+version=$(git describe --match '[0-9]*\.[0-9]*\.[0-9]*' --tags --abbrev=0)
+release_deb="releases/$version/deb/mgmt_${version}_amd64.deb"
+testcmd="set -x; apt-get update >/dev/null && \
+	dpkg -i /$release_deb 2>/dev/null; \
+	apt-get install -yqqf >/dev/null && \
+	mgmt --version"
+
+# test if release build is available
+if ! test -f "$release_deb"; then
+	fail_test "Release file to test ($release_deb) is not available!"
+fi
+
+failures=""
+
+for release in "${releases[@]}";do
+	echo "Testing: $release"
+	if ! docker run -ti --rm \
+			-v "$PWD/releases/:/releases/" \
+			"$release" /bin/sh -c "$testcmd";
+	then
+		failures="$failures $release"
+	fi
+done
+
+if [[ -n "${failures}" ]]; then
+	echo 'FAIL'
+	echo 'The following tests failed:'
+	echo "${failures}"
+	exit 1
+fi
+echo 'PASS'
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 docker
+releases
--- a/Makefile
+++ b/Makefile
@@ -373,7 +373,7 @@ releases/$(VERSION)/deb/changelog: $(PROGRAM) releases/$(VERSION)/.mkdir
 
 $(DEB_PKG): releases/$(VERSION)/deb/changelog
 	@echo "Building: deb package..."
-	./misc/fpm-pack.sh deb $(VERSION) libvirt-dev libaugeas-dev
+	./misc/fpm-pack.sh deb $(VERSION) libvirt0 libaugeas0
 
 $(PACMAN_PKG): $(PROGRAM) releases/$(VERSION)/.mkdir
 	@echo "Building: pacman package..."
--- a/Makefile
+++ b/Makefile
@@ -367,13 +367,26 @@ $(RPM_PKG): releases/$(VERSION)/rpm/changelog
 	@echo "Building: rpm package..."
 	./misc/fpm-pack.sh rpm $(VERSION) libvirt-devel augeas-devel
 
-releases/$(VERSION)/deb/changelog: $(PROGRAM) releases/$(VERSION)/.mkdir
+# Build deb release in oldest, most common denominator LTS for Debian/Ubuntu
+# to make sure it can be installed and run on every supported LTS.
+ifeq ($(shell grep 'Debian GNU/Linux 8 (jessie)' /etc/os-release),)
+$(DEB_PKG): releases/$(VERSION)/deb/changelog
+	@echo "Building: deb package using Docker environment..."
+	docker build -t mgmt-deb-release -f docker/Dockerfile.deb-release .
+	-docker rm -f mgmt-deb-release &>/dev/null
+	docker run --name mgmt-deb-release -ti --rm \
+		-v $$PWD:/gopath/src/github.com/purpleidea/mgmt/ \
+		mgmt-deb-release /bin/sh -c 'make clean && make $@ && make clean'
+
+# Git on Debian 8 is to old for the formatting arguments used in make-deb-changelog.sh.
+releases/$(VERSION)/deb/changelog: releases/$(VERSION)/.mkdir
 	@echo "Generating: deb changelog..."
 	./misc/make-deb-changelog.sh $(VERSION)
-
-$(DEB_PKG): releases/$(VERSION)/deb/changelog
+else
+$(DEB_PKG): $(PROGRAM) releases/$(VERSION)/deb/changelog
 	@echo "Building: deb package..."
 	./misc/fpm-pack.sh deb $(VERSION) libvirt0 libaugeas0
+endif
 
 $(PACMAN_PKG): $(PROGRAM) releases/$(VERSION)/.mkdir
 	@echo "Building: pacman package..."
--- a/docker/Dockerfile.deb-release
+++ b/docker/Dockerfile.deb-release
@@ -1,0 +1,18 @@
+FROM debian:8
+
+# Create environment to build release .deb for supported LTS Debian/Ubuntu versions
+
+RUN apt-get update
+RUN apt-get install -yqq git make curl
+
+# install at least go1.9
+RUN curl -sSL https://dl.google.com/go/go1.12.4.linux-amd64.tar.gz | tar -C /usr/local -xzf -
+
+ENV GOPATH /gopath
+ENV PATH $PATH:$GOPATH/bin:/usr/local/go/bin
+
+RUN mkdir -p $GOPATH/src/github.com/purpleidea/
+COPY . $GOPATH/src/github.com/purpleidea/mgmt/
+
+WORKDIR $GOPATH/src/github.com/purpleidea/mgmt
+RUN make deps
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -84,8 +84,8 @@ if [ $travis -eq 0 ]; then
 fi
 
 # if golang is too old, we don't want to fail with an obscure error later
-if go version | grep 'go1\.[012345]\.'; then
-	echo "mgmt requires go1.6 or higher."
+if go version | grep 'go1\.[012345678]\.'; then
+	echo "mgmt requires go1.9 or higher."
 	exit 1
 fi
 


purpleidea:
@aequitas First off, really cool, thanks! Any patch that improves my releases is awesome!

Before I do a more in depth review, I'll point out the one big issue:

I don't want to have docker be a build dependency, so here's what I suggest:

1) I'd love it if you could update this to remove docker entirely. I admit I'm not knowledgeable about what packaging people use to build, but I know there was a lot of it before docker came along. Was it just plain chroots? If you could switch this to use whatever the standard thing is, I'd appreciate it.

2) Alternatively, if you really *don't* want to move away from docker, you're welcome to run your own PPA, it just won't be official, and we'd still need to fix this upstream anyways.

Let me know what you think!
Thanks!

test/test-deb-releases.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
purpleidea: Having this test is a great idea btw! We should integrate some form of it! Maybe for the test itself we could switch it to using systemd-nspawn?

aequitas: What would be the best way doing this with nspawn? From what I've seen it will not be as trivial to do this with nspawn as with docker containers.

purpleidea: I'm not 100% sure, but there are many ways to pass an image into nspawn, via debootstrap, a filesystem image, etc...

I'd recommend doing the main part of building things properly and we'll add the test in step 2?

aequitas:
I understand docker as a build dependency issue, I just took it as it was the easiest thing to get this PoC going. Normally you do this on 'build servers' which means provisioning an entire OS just for this. So basically containers is the way to go for this. If you have a preferred method (eg nspawn) which does not add to much overhead to the entire process I can have a go at implementing that. Otherwise putting these builds into a PPA is also something I can do.

purpleidea:
Having a test is fantastic, but I think it's more important that we just build things correctly even without the test initially... So whatever people use to build packages normally would be great for an initial version of this that doesn't use docker.

Maybe @cgwalters could tell us the name of the correct tool to use to build deb packages from my fedora (or any) environment. Apparently fpm isn't doing it properly.

test/test-deb-releases.sh
aequitas: Running an image in nspawn is not really the problem I guess. Getting the image with the desired content on is a bit more complicated, that is why docker is so popular.

Doing the tests first would be preferred for me. Because it acts as the proof things are working.

aequitas:
It's not really the problem of the build tool. The problem is the binary is linked against libraries (libvirt in this case) that are not available on older systems. All the 'fix' does is build it on the oldest system to support (in this case Debian 8). So it is compatible with systems form that point. Maybe there are some other tricks you can use to accomplish this with linking flags, but my knowledge on build systems is lacking here as well.

purpleidea:
@aequitas One reason I'm less excited about the tests is that they're only something that would run _after_ the fact on CI, and that's if we even pushed them there to begin with. But if we pushed them, it would already be because we thought they passed... (So it confirms it worked, but doesn't block something broken.)

Maybe @jordansissel or @jonathangold know what we need to fix with our fpm setup ?

purpleidea:
@cgwalters Can you recommend the classic, customary way to build packages for debian/ubuntu while on a Fedora machine? (Without using docker.) Thanks!

jordansissel:
For linking against older binaries, I don’t have any great options. I’ve
hacked the ELF headers before to point to different locations per platform,
but that’s kind of not a good-feeling solution.

If the library binary is compatible, you can often be successful by setting
the dependency appropriately in the package (-d flag in fpm)

Beyond that, not sure I have context to answer.

cgwalters:
> @cgwalters Can you recommend the classic, customary way to build packages for debian/ubuntu while on a Fedora machine? (Without using docker.) Thanks!

You don't have to use docker the software, there's also [podman](https://github.com/containers/libpod/).  I think using some container like that is going to be by far the easiest, just install the build deps and use `-v $(pwd):/srv/build` and write to `/srv/build` or so.  

But it's certainly possible to install the deb build tools as RPMs if you prefer, I think some of those are packaged for Fedora.  There's also [FPM](https://fpm.readthedocs.io/en/latest/index.html) which I haven't tried.  I know that OpenEmbedded has tooling to do this too (it's a whole host-agnostic cross-compilation system), it abstracts over deb/rpm.

aequitas:
The problem isn't really the packaging. This can be done on any system just fine, even on Fedora using FPM I guess. What I encountered is that if the `mgmt` binary is not built on an older Debian distribution it is linked against newer versions of the `libvirt` library and will not run on the older distribution. So it really is a Go linking question that would need to be solved. How to tell Go to link the binary against older versions of the library.

purpleidea:
@aequitas Right! I suppose we need to build in some sort of chroot so that the right dependencies are seen by golang when it's building... What did people use historically? Chroot??

purpleidea:
@aequitas If you're in the mood, I guess I would relent and accept a patch to run systemd-nspawn with the correct os image inside, and build each package separately in N of those (one for each distro) and then suck them out to upload. Does that make sense?

aequitas:
@purpleidea we only need to build one `mgmt` binary on the oldest distro to support, then copy the binary out and we can just continue the release process like it is currently. I tested it and both newer Debian and Ubuntu distros then support that package.

purpleidea:
@aequitas I suppose that's true. Although if we're building in nspawn for Debian-9, we might as well just have a separate one for each distro we build a package for so that it's all neat and unlikely to cause problems.

purpleidea:
@aequitas (PS: If you do work on this, feel free to discuss the design early in here or in an issue so that you don't go down a long road for nothing. I've done a bit of nspawn stuff so happy to help if I can.)

purpleidea:
(For arch it's like this: https://wiki.archlinux.org/index.php/Systemd-nspawn#Create_and_boot_a_minimal_Arch_Linux_distribution_in_a_container for debian it's debootstrap, and so on...)

purpleidea:
(And sudo dnf --releasever=XX --installroot=/absolute/path/ install systemd passwd dnf fedora-release) for fedora. Maybe it's slightly different now, that was the way for F24.

purpleidea:
@aequitas Thanks again for the patch... We have ubuntu and debian releases now without docker. If your version is missing or there are any issues with the builds, please lmk.

https://github.com/purpleidea/mgmt/releases/

Issue #319, opened 2018-02-10
[LOVE] Add a config file to replace common command line flags


purpleidea:
#### Welcome:

This is a beginner level patch for someone who is new to golang, and is new to mgmt. It is meant to encourage new programmers to get involved with mgmt. It will probably take you a few hours or so to complete.

#### Intro:

Mgmt has a number of flags which are commonly passed in when testing and debugging. They include `--tmp-prefix`, `--no-pgp`, and others.
We'd like a way to specify some of these in a config file so that we can avoid typing them constantly.
To implement this, you'll need to either recommend an alternate design, or use the one provided below.

#### Design:

We need a config file for some of our args. It should probably obey the xdg spec and as such default to `~/.config/mgmt/`.
When running as root it should probably look in `/etc/mgmt/`
The argument parsing library has some built-in notion of looking up defaults in a config. Look into that and see if it's sufficient for our needs.
Lastly, ensure that the config file uses some existing well-known format. Yaml might be an acceptable choice. We don't need to unnecessarily invent new file formats.
Once you've decided on a design, present it here for discussion before you implement things.

Read the contributing tips:
https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Happy hacking, and ping me for questions!

#### Bonus/Future work:

* Add some tests!
* Add an example config to the github repo.

PR #515, opened 2019-05-14
[WIP] Add a config file to replace common command line flags

sebastianrakel:
This PR is currently WIP it should add the new feature in the end

I openend this PR so other people can see the development state of this feature

This is realted to #319

### DIFF
--- a/gapi/empty/empty.go
+++ b/gapi/empty/empty.go
@@ -24,7 +24,7 @@ import (
 	"github.com/purpleidea/mgmt/gapi"
 	"github.com/purpleidea/mgmt/pgraph"
 
-	"github.com/urfave/cli"
+	"gopkg.in/urfave/cli.v1"
 )
 
 const (
--- a/gapi/gapi.go
+++ b/gapi/gapi.go
@@ -25,7 +25,7 @@ import (
 	"github.com/purpleidea/mgmt/engine"
 	"github.com/purpleidea/mgmt/pgraph"
 
-	"github.com/urfave/cli"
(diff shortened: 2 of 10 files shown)


purpleidea:
@sebastianrakel Thanks for the patch! So in general, unless I'm misunderstanding something, this would pin the cli library at an older version, instead of the current git master... Here's an FAQ about this:

https://github.com/purpleidea/mgmt/blob/master/docs/faq.md#why-arent-you-using-glide-or-godep-for-dependency-management

Is there a reason we want to make an exception here?

sebastianrakel:
As i wrote in the commit 

> To add the config option of the cli lib, we need to switch to the
> gopkg.in package, because it's used in the internal development
> of this lib

The altsrc package witch is needed to load a config file, references "gopkg.in/urfave/cli.v1"
like here: https://github.com/urfave/cli/blob/master/altsrc/yaml_file_loader.go

so we have to options, pin our lib to v1 because v2 is unstable and not ready or open a bug upstream to change this behavior

purpleidea:
@sebastianrakel Sorry for my delay in replying. I was a bit buried with recent travel. Back on track now =D

> so we have two options, pin our lib to v1 because v2 is unstable and not ready or open a bug for upstream to change his import statement

Gah, you're right. Thanks for pointing this out. Can you open a bug? Feel free to CC me, and explain that this blocks us from running `git master`. Thanks again.

purpleidea:
@sebastianrakel I don't know if this is an issue anymore. If so please LMK and I'll reopen. If you still want to work on this please lmk! Thanks!

PR #524, opened 2019-06-03
engine: resources: mount: Fix the dbus call for reloading systemd

### LINKED ISSUE #509: Mount resource only updated /etc/fstab, did not mount
## Versions:

* mgmt version (eg: `mgmt --version`): 0.0.18

* operating system/distribution (eg: `uname -a`): Ubuntu 19.04

## Description:

Trying to mount a few local ext4 filesystems updated /etc/fstab correctly, but did not mount the disks after mgmt had converged.

The mcl files was simple

```
file "/mnt/disk1/" {
	state => "exists",
}

mount "/mnt/disk1" {
	state => "exists",
	device => "UUID=8f3b7486-7529-4552-820c-284232c8354d",
	type => "ext4",
	options => {
		"noatime"=>"",
		"defaults"=>"",
	},

	Depend => File["/mnt/disk1/"],
}
```
The log contained statements like this one, which seems to indicate that the mount logic inside of mgmt did get called

```
17:54:48 main.go:496: gapi: lang: running interpret...
17:54:48 main.go:414: engine: mount[/mnt/disk4]: CheckApply(true)
17:54:48 main.go:414: engine: mount[/mnt/disk4]: resource: mountCheckApply(true)
```

ffrank:
The Reload method cannot just be invoked on the administrative DBus
object. Make it work by specifying the proper systemd object name, path, and
interface.

Fixes #509

### DIFF
--- a/engine/resources/mount.go
+++ b/engine/resources/mount.go
@@ -75,6 +75,8 @@ const (
 	// diskByLabel is the location of symlinks for partitions by label.
 	diskByPartLabel = devDisk + "by-partlabel/"
 
+	// dbusSystemdService is the service to connect to systemd itself.
+	dbusSystemd1Service = "org.freedesktop.systemd1"
 	// dbusSystemd1Interface is the base systemd1 path.
 	dbusSystemd1Path = "/org/freedesktop/systemd1"
 	// dbusUnitPath is the dbus path where mount unit files are found.
@@ -88,6 +90,9 @@ const (
 	dbusManagerInterface = dbusSystemd1Interface + ".Manager"
 	// dbusRestartUnit is the dbus method for restarting systemd units.
 	dbusRestartUnit = dbusManagerInterface + ".RestartUnit"
+	// dbusReloadSystemd is the dbus method for reloading systemd settings.
+	// (i.e. systemctl daemon-reload)
+	dbusReloadSystemd = dbusManagerInterface + ".Reload"
 	// restartTimeout is the delay before restartUnit is assumed to have
 	// failed.
 	dbusRestartCtxTimeout = 10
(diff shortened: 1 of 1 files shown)


purpleidea:
@ffrank Cool! Thanks for the patch... Do you think you could follow the convention and use the constants, and add a new one for Reload?

Also, if you could provide a bit more background in the commit message, this would really be helpful. Would love to understand what we did wrong and why this fixes it.

Cheers!

purpleidea:
Also you mentioned:

> ffrank: Trying to unmount using state=>absent makes mgmt crash for me too btw

Does that happen all the time or only with this patch, and could you post the trace in an issue please?

Thanks again!

ffrank:
Done and done. With some extra code clarification work.

engine/resources/mount.go
```diff
@@ -88,6 +90,9 @@ const (
 	dbusManagerInterface = dbusSystemd1Interface + ".Manager"
 	// dbusRestartUnit is the dbus method for restarting systemd units.
 	dbusRestartUnit = dbusManagerInterface + ".RestartUnit"
+	// dbusReloadSystemd is the dbus method for reloading systemd settings
```
purpleidea: nit: add trailing period please

ffrank:
Nits were addressed.

The "panic" really was just the dbus race we've been seeing recently.

purpleidea:
LGTM! Thanks!

FYI, will merge soon, this is queued up, I just haven't merged yet because I don't want to stash my WIP work atm.

Cheers!

purpleidea:
Merged, thank you and sorry for the delay!

PR #529, opened 2019-07-02
Added test for FindEdge() and HasVertex() functions of pgraph.go

adamsigal:
## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/pgraph/pgraph_test.go
+++ b/pgraph/pgraph_test.go
@@ -840,3 +840,47 @@ func TestDeleteEdge2(t *testing.T) {
 		t.Errorf("expected outgoing graph edges for vertex v1: e4, instead of: %s", res)
 	}
 }
+
+func TestFindEdge1(t *testing.T) {
+	g, _ := NewGraph("g")
+
+	v1 := NV("v1")
+	v2 := NV("v2")
+	v3 := NV("v3")
+	v4 := NV("v4")
+	v5 := NV("v5")
+	v6 := NV("v6")
+	v7 := NV("v7")
+
+	e1 := NE("e1")
+	e2 := NE("e2")
+	e3 := NE("e3")
+	e4 := NE("e4")
+	e5 := NE("e5")
+	e6 := NE("e6")
+	e7 := NE("e7")
+	e8 := NE("e8")
+	e9 := NE("e9")
+	e10 := NE("e10")
+
+	g.AddEdge(v1, v2, e1)
+	g.AddEdge(v1, v4, e2)
+	g.AddEdge(v1, v3, e3)
+	g.AddEdge(v2, v3, e4)
+	g.AddEdge(v2, v4, e5)
+	g.AddEdge(v2, v6, e6)
+	g.AddEdge(v2, v5, e7)
+	g.AddEdge(v3, v6, e8)
+	g.AddEdge(v4, v7, e9)
+	g.AddEdge(v5, v7, e10)
+
+	if !(g.HasVertex(v1) && g.HasVertex(v4) && g.HasVertex(v7)) {
+		t.Errorf("graph expected to have vertices v1, v4, and v7")
+	}
+	if g.FindEdge(v1, v4) != e2 {
+		t.Errorf("edge e2 was not returned")
+	}
+	if g.FindEdge(v1, v7) != nil {
+		t.Errorf("an edge was found although it did not exist")
+	}
+}

### CODE REVIEW

File: pgraph/pgraph_test.go [hunk 1]
```diff
@@ -840,3 +840,48 @@ func TestDeleteEdge2(t *testing.T) {
...
+	g.AddEdge(v2, v4, e5)
+	g.AddEdge(v2, v6, e6)
+	g.AddEdge(v2, v5, e7)
+	g.AddEdge(v3, v6, e8)
+	g.AddEdge(v4, v7, e9)
+	g.AddEdge(v5, v7, e10)
+
+	if g.HasVertex(v1) && g.HasVertex(v4) && g.HasVertex(v7) {
+		if g.FindEdge(v1, v4) != e2 {
+			t.Errorf("Edge e2 was not returned")
```
purpleidea: Errorf should start with a lowercase char and not end with punctuation

pgraph/pgraph_test.go
```diff
@@ -840,3 +840,48 @@ func TestDeleteEdge2(t *testing.T) {
...
+		}
+		if g.FindEdge(v1, v10) != nil {
+			t.Errorf("An edge was found although it did not exist")
+		}
+	} else {
```
purpleidea: You can inline this and the rest please.

adamsigal: What do you mean by "inline"?

purpleidea:

pgraph/pgraph_test.go
```diff
@@ -840,3 +840,49 @@ func TestDeleteEdge2(t *testing.T) {
...
+		t.Errorf("graph expected to have vertices v1, v4, and v7")
+		return
+	}
+	if g.FindEdge(v1, v4) != e2 {
+		t.Errorf("edge e2 was not returned")
+	}
+	if g.FindEdge(v1, v7) != nil {
+		t.Errorf("an edge was found although it did not exist")
+	}
+	return
```
purpleidea: This return isn't necessary, since we fall off at the end.

pgraph/pgraph_test.go
```diff
@@ -840,3 +840,49 @@ func TestDeleteEdge2(t *testing.T) {
...
+	g.AddEdge(v2, v4, e5)
+	g.AddEdge(v2, v6, e6)
+	g.AddEdge(v2, v5, e7)
+	g.AddEdge(v3, v6, e8)
+	g.AddEdge(v4, v7, e9)
+	g.AddEdge(v5, v7, e10)
+
+	if !(g.HasVertex(v1) && g.HasVertex(v4) && g.HasVertex(v7)) {
+		t.Errorf("graph expected to have vertices v1, v4, and v7")
+		return
```
purpleidea: We could return here if it would affect future tests, but it won't. So either add a return for the other two cases, or remove this one. I think the latter is best =D

pgraph/pgraph_test.go
```diff
@@ -840,3 +840,49 @@ func TestDeleteEdge2(t *testing.T) {
 		t.Errorf("expected outgoing graph edges for vertex v1: e4, instead of: %s", res)
 	}
 }
+
+func TestFindEdge(t *testing.T) {
```
purpleidea: Maybe name this `TestFindEdge1` please.

pgraph/pgraph_test.go
adamsigal: If this test fails, that means that v1, v4, or v7 doesn't exist, in which case the condition of one of the following if statements will create an error. In that case we don't want to continue to the other cases. Let me know what you think.

purpleidea: Agreed, however it's not necessarily harmful. It's fine to just "return" and stop the test early. But be consistent.

adamsigal: Okay, I'll follow your suggestion in that case!

purpleidea:
@adamsigal One more fix, sorry. Your commit message shouldn't mention a file name. You could say in "pgraph package" or similar if you wanted to.

And the body makes no sense. And it's not a full sentence. It should be something useful, not a log of your development work unless a dead end you learned about needed sharing.

purpleidea:
LGTM, thanks! (Waiting for travis...)

purpleidea:
Merged, thanks! \o/

Welcome to the club of real code committed.

Issue #498, opened 2019-04-05
[LOVE] State directory with xdg dirs


purpleidea:
This is a beginner level golang patch for someone who is interested in getting involved.

We should consider having our default `/var/lib/mgmt/` dir be a combination of https://github.com/systemd/portable-walkthrough-go/blob/5f2b36a0f2ce30325119224956c837d43fa6782e/main.go#L34 (systemd's STATE_DIRECTORY env var) and xdg dirs when that is not specified.

You'll need to see where that initial path comes from, and add the logic to implement the above.

Anyone interested in writing the patch, please let us know here!

hooksie1:
Are you thinking the xdg dir should be XDG_CONFIG_HOME or something different?

purpleidea:
I think the state part should probably go in `~/.cache/` which is XDG_CACHE_HOME. (Better suggestions welcome if I'm interpreting XDG spec wrong.)

Of course this is all secondary to if the systemd STATE_DIRECTORY is not present.

@hooksie1 Want to write a patch?

hooksie1:
Ok yeah I think you're right, that makes more sense. I might give it a shot and see what terrible thing I can come up with.

purpleidea:
@hooksie1 Okay awesome! Join us in #mgmtconfig on Freenode IRC for questions, and feel free to send an early (even broken) PR/patch for early review.

Cheers!

hooksie1:
PR #530 is in.

purpleidea:
Merged!

PR #530, opened 2019-07-05
lib: Support for systemd STATE_DIRECTORY or XDG cache dir

hooksie1:
PR to add support for STATE_DIRECTORY or XDG_CACHE_HOME from Issue #498

### DIFF
--- a/lib/main.go
+++ b/lib/main.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/user"
 	"path"
 	"strings"
 	"sync"
@@ -214,7 +215,35 @@ func (obj *Main) Run() error {
 		return fmt.Errorf("hostname cannot be empty")
 	}
 
+	user, err := user.Current()
+	if err != nil {
+		return errwrap.Wrapf(err, "can't get current user")
+	}
+
+	// Use systemd StateDirectory if set. If not, use XDG_CACHE_DIR unless user
+	// is root, then use /var/lib/mgmt/.
 	var prefix = fmt.Sprintf("/var/lib/%s/", obj.Program) // default prefix
+	stateDir := os.Getenv("STATE_DIRECTORY")
+	// Ensure there is a / at the end of the directory path.
+	if stateDir != "" && !strings.HasSuffix(stateDir, "/") {
+		stateDir = stateDir + "/"
+	}
+
+	xdg := os.Getenv("XDG_CACHE_HOME")
+	// Ensure there is a / at the end of the directory path.
+	if xdg != "" && !strings.HasSuffix(xdg, "/") {
+		xdg = xdg + "/"
+	}
+	if xdg == "" && user.HomeDir != "" {
+		xdg = fmt.Sprintf("%s/.cache/%s/", user.HomeDir, obj.Program)
+	}
+
+	if stateDir != "" {
+		prefix = stateDir
+	} else if user.Uid != "0" {
+		prefix = xdg
+	}
+
 	if p := obj.Prefix; p != nil {
 		prefix = *p
 	}


lib/main.go
```diff
@@ -214,7 +214,20 @@ func (obj *Main) Run() error {
 		return fmt.Errorf("hostname cannot be empty")
 	}
 
-	var prefix = fmt.Sprintf("/var/lib/%s/", obj.Program) // default prefix
```
purpleidea: Wouldn't it be better to leave this first line as-is, and then If state dir, use that, else if xdg is set, use that, and so on?

hooksie1: Oh I guess I misunderstood. I thought you wanted to use XDG_CACHE_HOME instead of /var/lib/mgmt. The reason I did it that way was because if XDG_CACHE_HOME isn't set it's supposed to be $HOME/.cache. I know at least on Fedora (what I'm using) it's not set as an environment variable out of the box so you would need to consciously add it. I can take out the default of $HOME/.cache and leave the /var/lib/%s line at the top.

purpleidea: Maybe we should think about what makes the most sense... Hmmm

purpleidea: How about if root, then that's the default (/var/lib/ ...) otherwise the default is XDG. And either can be overridden by the systemd state var.

What do you think? (Don't necessarily just agree with me, decide yourself =D)

hooksie1: Yeah that sounds like it makes sense. I'm not sure how useful XDG dirs are for root anyway since you normally don't run an X session as root.

purpleidea:
On Fri, Jul 5, 2019 at 6:49 PM John Hooks <notifications@github.com> wrote:
>
> @hooksie1 commented on this pull request.
>
> ________________________________
>
> In lib/main.go:
>
> > @@ -214,7 +214,20 @@ func (obj *Main) Run() error {
>   return fmt.Errorf("hostname cannot be empty")
>   }
>
> - var prefix = fmt.Sprintf("/var/lib/%s/", obj.Program) // default prefix
>
> Yeah that sounds like it makes sense. I'm not sure how useful XDG dirs are for root anyway since you normally don't run an X session as root.

Agreed!

hooksie1:
Not sure if you care about pulling in os/user. I didn't want to rely on doing os.Getenv("USER") because if someone weirdly named a user root for some reason it might not have the affect they expected.

purpleidea:
Have at it, we already pull it in here: https://github.com/purpleidea/mgmt/blob/191a2495a5008ac76c25a6af0e832dff63571d27/engine/util/util.go#L26

lib/main.go
```diff
@@ -214,7 +215,33 @@ func (obj *Main) Run() error {
 		return fmt.Errorf("hostname cannot be empty")
 	}
 
+	// Get current user running mgmt
+	user, _ := user.Current()
```
purpleidea: You always have to handle errors!

lib/main.go
```diff
@@ -214,7 +215,33 @@ func (obj *Main) Run() error {
...
+		}
+	}
+
+	if user.Uid != "0" {
+		if stateDir != "" {
+			prefix = fmt.Sprintf("%s/", stateDir)
+		} else if xdg != "" {
+			prefix = fmt.Sprintf("%s/", xdg)
+		} else {
+			stateDir = os.Getenv("HOME") + "/.cache/"
```
purpleidea: You can get HOME from the os/user User struct.

lib/main.go
```diff
@@ -214,7 +215,33 @@ func (obj *Main) Run() error {
...
+	user, _ := user.Current()
+
+	// Use /var/lib/mgmt if user is root, otherwise
+	// get systemd state directory if set, if not then
+	// use XDG_CACHE_DIR or default for XDG_CACHE_DIR
 	var prefix = fmt.Sprintf("/var/lib/%s/", obj.Program) // default prefix
+	stateDir := os.Getenv("STATE_DIRECTORY")
+	xdg := os.Getenv("XDG_CACHE_HOME")
+
+	if user.Uid == "0" {
```
purpleidea: I think this logic can be refactored to be less nested and more straight forward... Want to give it a try?

lib/main.go
hooksie1: bah my bad. I did that to test, and forgot to check how you were handling errors. I'll fix that.

lib/main.go
hooksie1: Ok, I'm glad you said that. I didn't like it but didn't put enough thought into it. I think this is better, is this more along the lines of what you were looking for? I'll rebase and squash into a single commit once you like everything.

lib/main.go
```diff
@@ -214,7 +215,29 @@ func (obj *Main) Run() error {
 		return fmt.Errorf("hostname cannot be empty")
 	}
 
+	// Get current user running mgmt
+	user, err := user.Current()
+	if err != nil {
+		return errwrap.Wrapf(err, "can't get current user")
+	}
+
+	// Use systemd SateDirectory if set. If not, use
```
purpleidea: typo

lib/main.go
```diff
@@ -214,7 +215,29 @@ func (obj *Main) Run() error {
...
+	// XDG_CACHE_DIR unless user is root, then use
+	// /var/lib/mgmt
```
purpleidea: dir's (even in comments) have a trailing slash please, and full sentences end with a period.

(Make sure comments are wrapped to 80 chars please.)

lib/main.go
```diff
@@ -214,7 +215,29 @@ func (obj *Main) Run() error {
...
 	var prefix = fmt.Sprintf("/var/lib/%s/", obj.Program) // default prefix
+	stateDir := os.Getenv("STATE_DIRECTORY")
+	xdg := os.Getenv("XDG_CACHE_HOME")
+
```
purpleidea: kill nl

lib/main.go
```diff
@@ -214,7 +215,29 @@ func (obj *Main) Run() error {
...
+	if xdg == "" {
+		xdg = fmt.Sprintf("%s/.cache/%s", user.HomeDir, obj.Program)
```
purpleidea: What if HomeDir is "" ?

lib/main.go
```diff
@@ -214,7 +215,29 @@ func (obj *Main) Run() error {
...
+	}
+
+	if stateDir != "" {
+		prefix = stateDir
+	} else if user.Uid != "0" {
+		prefix = fmt.Sprintf("%s", xdg)
```
purpleidea: unnecessary sprintf

lib/main.go
```diff
@@ -214,7 +215,29 @@ func (obj *Main) Run() error {
...
+	}
```
purpleidea: Probably need to add a check to add trailing slash if one is not present. (STATEDIR or XDG might not necessarily provide one.)

lib/main.go
hooksie1: Sorry, I thought I tested it without that and it didn't work. I ran it again and it did. My bad.

lib/main.go
hooksie1: What do you think is an appropriate dir for this? Something in /tmp?

lib/main.go
hooksie1: I tested by both using StateDirectory in the unit file and STATE_DIRECTORY as an environment variable without it. I didn't pass the / to either and it put it under each correctly.

purpleidea: It's likely IIRC that later parts of the code that consume `prefix` check for it and add one, but we should do it anyways in case they don't.

lib/main.go
purpleidea: `if xdg == "" && HomeDir != "" {` ?

hooksie1: Oh I see what you mean. I thought you meant assign one if they didn't have one.

lib/main.go
hooksie1: ok will do

hooksie1: So if I have it add the / suffix, it errors out here:
```
pgpPrefix := fmt.Sprintf("%s/", path.Join(prefix, "pgp"))
                if err := os.MkdirAll(pgpPrefix, 0770); err != nil {
                        return errwrap.Wrapf(err, "can't create pgp prefix")
                }
```

Looks like since the / is in the path there, it's adding a double slash and you get the error:

    21:03:28 run: error: can't create pgp prefix: mkdir /pgp/: permission denied

hooksie1: So I realized it doesn't matter anyway. I need to add `obj.Program` to the path. Because the mgmt  directory should exist in the XDG_CACHE_HOME directory, so the slash will be added to prefix when we assign xdg.

hooksie1: Just ignore me, I'm dumb. It works with adding the trailing slash.

hooksie1:
Also, thanks for being so patient with me. I really appreciate the help, and it's awesome that you've been so kind.

lib/main.go
```diff
@@ -214,7 +215,36 @@ func (obj *Main) Run() error {
 		return fmt.Errorf("hostname cannot be empty")
 	}
 
+	// Get current user running mgmt
```
purpleidea: nit: remove comment b/c it's obvious or add the trailing period please. You can have full sentences (capitalized & period) or lower case mini-ones.

purpleidea:
@hooksie1 

> Also, thanks for being so patient with me. I really appreciate the help, and it's awesome that you've been so kind.

Patch looks perfect now. One small nit to fix and we merge! \o/

Happy to review and help. Obviously it's more work for us initially, than to just write the patch, but happy to mentor code with the hopes that it helps you improve, and as you get better with golang, your future patches will help even more! Keep it up and join us on IRC. We can suggest the next thing to work on if you don't have any ideas.

hooksie1:
Ok, I just removed that comment since it's not needed. I had a crazy rebase issue there so if something is screwed up let me know.

hooksie1:
Not sure what I did wrong with the commit message. Is it just too long?

purpleidea:
@hooksie1 Looks good to me. There's a typo: `unning` but I think it's failing because it doesn't like the `/` in the message. Remove/edit that and I think it should pass =D

Thanks again!

purpleidea:
I think this should be good to go! I'll have a look tomorrow I think!

Thanks again :)

hooksie1:
Awesome! No problem, thanks for the help!

purpleidea:
@hooksie1 Merged, thanks!

Welcome to the project. LMK when you're ready for your next patch!

hooksie1:
Will do. I should be able to soon.

PR #538, opened 2019-09-09
make deps: Add golang version test to fix Ubuntu bug

adamsigal:
Running make-deps for the first time on Ubuntu can erroneously lead to
installing outdated version of golang, even if a newer version is already
installed. A version check is now added to the relevant conditions.

## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/misc/make-deps.sh
+++ b/misc/make-deps.sh
@@ -75,7 +75,7 @@ fi
 
 if [ $travis -eq 0 ]; then
 	if [ ! -z "$YUM" ]; then
-		if [ -z "$GO" ]; then
+		if [ -z "$GO" ] && [ go version | grep -e 'go1\.[0123456789]\.' -e 'go1\.10\.' ]; then
 			$sudo_command $YUM install -y golang golang-googlecode-tools-stringer
 		fi
 		# some go dependencies are stored in mercurial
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: misc/make-deps.sh [hunk 1]
```diff
@@ -75,15 +75,15 @@ fi
 
 if [ $travis -eq 0 ]; then
 	if [ ! -z "$YUM" ]; then
-		if [ -z "$GO" ]; then
+		if [ -z "$GO" ] && [ go version | grep -e 'go1\.[0123456789]\.' -e 'go1\.10\.' ]; then
```
purpleidea: Wouldn't `! go version 2> /dev/null` be more what you want??

adamsigal: If I'm not mistaken, `[ -z "$GO" ]`does the same thing as ! go version `2>/dev/null` in theory since GO=`command -v go 2>/dev/null`. The issue is that this first condition succeeds when it should not in some installation cases. That's why I added this other condition so we are absolutely sure that go must be installed.

purpleidea: I'm just not quite clear what's going on here and why it's needed. Care to discuss in channel or try to explain more please?

PR #542, opened 2019-09-13
Use go mod and remove vendor submodules

joejulian:


purpleidea:
Thanks for the patch! Git submodules actually work *great*, and in particular are far easier to use when you need to pin a particular version and develop in that, so we really don't want to move to git master and add another tool like `go mod`.

Please see:
https://github.com/purpleidea/mgmt/blob/master/docs/faq.md#why-arent-you-using-glide-godep-or-go-mod-for-dependency-management

However, we *would* like to have some sort of explicit dependency listing for releases! If you can help us integrate a go mod workflow for everytime I tag a release, so that we commit the mod file into releases/ and have a way to restore it when you `git checkout <tag>` then that would be great!

I'm going to close this, but LMK if you can help. Thanks!

joejulian:
Actually, they're not working at all, and are making it impossible to build the aur package.

purpleidea:
@joejulian Can you provide some details as to why? I'm sure it was possible to build golang packages before `go mod` was invented.

joejulian:
Some submodule links are broken. You cannot pull the tarball and build the package as it does not have the bits necessary to build the package. Reproducible release builds are broken; I cannot build past releases.

Since go mod is the defacto solution that had been merged into the main golang project and this project is not going to adopt it, I cannot continue to maintain the aur package. I've orphaned it and it is available for adoption if someone else wishes to do so.

purpleidea:
> Some submodule links are broken.

No they're not or the tests would fail. If you know of something broken please LMK or file an issue.

> Reproducible release builds are broken; I cannot build past releases.

Agreed, please see: https://github.com/purpleidea/mgmt/pull/542#issuecomment-531389410

If you'd like to help fix this, please LMK.

> go mod

Please see: https://github.com/purpleidea/mgmt/blob/master/docs/faq.md#why-arent-you-using-glide-godep-or-go-mod-for-dependency-management

joejulian:
> If you'd like to help fix this, please LMK.
I did. 😝

PR #548, opened 2019-10-05
engine: Add Env to exec.go

adamsigal:
Add functionality to specify environment variables in exec.go.

## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/engine/resources/exec.go
+++ b/engine/resources/exec.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os/exec"
 	"os/user"
+	"sort"
 	"strings"
 	"sync"
 	"syscall"
@@ -65,6 +66,9 @@ type ExecRes struct {
 	// running command. If the Kill is received before the process exits,
 	// then this be treated as an error.
 	Timeout uint64 `yaml:"timeout"`
+	// Env allows the user to specify environment variables for script
+	// execution. These are taken using a map of format of VAR_NAME -> value.
+	Env map[string]string `yaml:"env"`
 
 	// Watch is the command to run to detect event changes. Each line of
 	// output from this command is treated as an event.
@@ -138,6 +142,12 @@ func (obj *ExecRes) Validate() error {
 		}
 	}
 
+	// check that environment variables' format is valid
+	for key := range obj.Env {
+		if err := isNameValid(key); err != nil {
+			return errwrap.Wrapf(err, "invalid variable name")
+		}
+	}
 	return nil
 }
 
@@ -371,6 +381,18 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 	defer cancel()
 	cmd := exec.CommandContext(ctx, cmdName, cmdArgs...)
 	cmd.Dir = obj.Cwd // run program in pwd if ""
+
+	envKeys := []string{}
+	for key := range obj.Env {
+		envKeys = append(envKeys, key)
+	}
+	sort.Strings(envKeys)
+	cmdEnv := []string{}
+	for _, k := range envKeys {
+		cmdEnv = append(cmdEnv, k+"="+obj.Env[k])
+	}
+	cmd.Env = cmdEnv
+
 	// ignore signals sent to parent process (we're in our own group)
 	cmd.SysProcAttr = &syscall.SysProcAttr{
 		Setpgid: true,
@@ -803,4 +825,21 @@ func (obj *wrapWriter) Write(p []byte) (int, error) {
 // String returns the contents of the unshared buffer.
 func (obj *wrapWriter) String() string {
 	return obj.Buffer.String()
+}
+
+// isNameValid checks that environment variable name is valid.
+func isNameValid(varName string) error {
+	if varName == "" {
+		return fmt.Errorf("variable name cannot be an empty string")
+	}
+	for i := range varName {
+		c := varName[i]
+		if i == 0 && '0' <= c && c <= '9' {
+			return fmt.Errorf("variable name cannot begin with number")
+		}
+		if !(c == '_' || '0' <= c && c <= '9' || 'a' <= c && c <= 'z' || 'A' <= c && c <= 'Z') {
+			return fmt.Errorf("invalid character in variable name")
+		}
+	}
+	return nil
 }
--- a/engine/resources/resources_test.go
+++ b/engine/resources/resources_test.go
@@ -322,6 +322,42 @@ func TestResources1(t *testing.T) {
 		})
 	}
 	{
+		r := makeRes("exec", "x2")
+		res := r.(*ExecRes) // if this panics, the test will panic
+		res.Env = map[string]string{
+			"boiling": "one hundred",
+		}
+		f := "/tmp/whatever"
+		res.Cmd = fmt.Sprintf("env | grep boiling > %s", f)
+		res.Shell = "/bin/bash"
+		res.IfCmd = "! diff <(cat /tmp/whatever) <(echo boiling=one hundred)"
+		res.IfShell = "/bin/bash"
+		res.WatchCmd = fmt.Sprintf("/usr/bin/inotifywait -e modify -m %s", f)
+		res.WatchShell = "/bin/bash"
+
+		timeline := []Step{
+			NewStartupStep(1000 * 60),              // startup
+			NewChangedStep(1000*60, false),         // did we do something?
+			FileExpect(f, "boiling=one hundred\n"), // check initial state
+			NewClearChangedStep(1000 * 15),         // did we do something?
+			FileWrite(f, "this is stuff!\n"),       // change state
+			NewChangedStep(1000*60, false),         // did we do something?
+			FileExpect(f, "boiling=one hundred\n"), // check again
+			sleep(1),                               // we can sleep too!
+		}
+
+		testCases = append(testCases, test{
+			name:     "exec with env",
+			res:      res,
+			fail:     false,
+			timeline: timeline,
+			expect:   func() error { return nil },
+			// build file for inotifywait
+			startup: func() error { return ioutil.WriteFile(f, []byte("starting...\n"), 0666) },
+			cleanup: func() error { return os.Remove(f) },
+		})
+	}
+	{
 		r := makeRes("file", "r1")
 		res := r.(*FileRes) // if this panics, the test will panic
 		p := "/tmp/emptyfile"

### CODE REVIEW

File: engine/resources/exec.go [hunk 1]
```diff
@@ -138,9 +141,38 @@ func (obj *ExecRes) Validate() error {
...
+	for key := range obj.Env {
+		if !isNameValid(key) {
+			return fmt.Errorf("environment variable name must only contain alphanumeric characters or underscore")
+		}
+	}
+
 	return nil
 }
 
+// TODO: move this function further down in code
```
purpleidea: Why have a TODO in here?

engine/resources/exec.go
```diff
@@ -138,9 +141,38 @@ func (obj *ExecRes) Validate() error {
...
+// checks that environment variable name is valid
+func isNameValid(varName string) bool {
```
purpleidea: change sig to return `error` instead. Instead of `return false` use fmt.Errorf(...reason...) instead.

engine/resources/exec.go
```diff
@@ -138,9 +141,38 @@ func (obj *ExecRes) Validate() error {
 		}
 	}
 
+	// check that environment variables' format is valid
+	for key := range obj.Env {
+		if !isNameValid(key) {
+			return fmt.Errorf("environment variable name must only contain alphanumeric characters or underscore")
```
purpleidea: errwrap.Wrapf please with shorter message now that below sig changes

engine/resources/exec.go
```diff
@@ -138,9 +141,38 @@ func (obj *ExecRes) Validate() error {
...
 	return nil
 }
 
+// TODO: move this function further down in code
+// checks that environment variable name is valid
+func isNameValid(varName string) bool {
+	if word == "" {
+		return false
+	}
+	var i int
```
purpleidea: not needed. `i := 0` instead.

engine/resources/exec.go
```diff
@@ -138,9 +141,38 @@ func (obj *ExecRes) Validate() error {
...
+	for i = 0; i < len(varName); i++ {
+		c := varName[i]
+		// if first character is number, name is invalid
+		if i == 0 && '0' <= c && c <= '9' {
+			return false
+		}
+		// if name does not only contain alphanumeric characters or underscore,
```
purpleidea: one line mini comments or full sentences. Full sentences are capitalized and end with a period.

engine/resources/exec.go
```diff
@@ -277,6 +309,16 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 		}
 		cmd := exec.Command(cmdName, cmdArgs...)
 		cmd.Dir = obj.IfCwd // run program in pwd if ""
+
+		//TODO: verify with James
```
purpleidea: Don't put a comment like this =D Ask in review instead ;)

engine/resources/exec.go
```diff
@@ -277,6 +309,16 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
...
+		cmdEnv := make([]string, len(obj.Env))
```
purpleidea: I dislike this syntax, instead just do: `cmdEnv := []string{}`

engine/resources/exec.go
```diff
@@ -277,6 +309,16 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
...
+		i := 0
+		for key := range obj.Env {
```
purpleidea: Look up the map iteration syntax, don't use a counter like `i`. And we'll want to make this deterministic somehow... I'll let you figure it out. Do you know why it isn't deterministic and why it should be?

engine/resources/exec.go
adamsigal: I thought that this function was out of place in its current position, but I wanted it to be easily visible the first time you were looking at it, since it is related to the neighbouring function.

engine/resources/exec.go
adamsigal: I know that the order of map traversal isn't deterministic in this case. Because of that the end result isn't deterministic because duplicate definitions of a variable aren't assured to be in the same order, possibly leading to an incorrect final definition.
I see the problem, I'll find a solution.

engine/resources/exec.go
adamsigal: I don't quite understand what you mean by "now that below sig changes"

adamsigal: Sorry, I get it now

engine/resources/exec.go
purpleidea: This doesn't matter, I'd see it anyways because `git diff` shows me the changes. But if a future reader of the code (not the patch) would find it easier if they were co-located, then it makes sense to do so. This depends on convention and other things. Eg: do you layout chained functions in standard execution order or dependency order? Or do private util functions go at the end so they don't interleave with the struct methods? (That is what I do.)

engine/resources/exec.go
purpleidea: Wrong, a map (in golang or mcl) can't have duplicate keys, so _that's_ not the problem. But the list we build will be in different orders which can make debugging or other things tricky. IOW a heisenbug! We never want to cause those. (Eg: "i ran the code and it did X, now it did Y... what's going on? *brain explodes*)

purpleidea:
Ping when ready for a new review. Thanks!

engine/resources/exec.go
adamsigal: So we have to make sure that all of the variables are processed in the order they were written, right? For example we couldn't just sort them alphabetically and process them in that order.

adamsigal: I can't think of how we would do that in order just by using a map. I'm assuming that a map is still the best option because you confirmed that to me at one point, but I fail to think of a non-hacky solution

purpleidea: > For example we couldn't just sort them alphabetically and process them in that order.

Why not? Of course we can. The solution is to first pull all the keys out of the map into a new list. Then sort that list. Then loop through that list (now deterministic) and read the values out from the map in that order. TADA =D I'll let you write the code for this. Max 6 lines.

adamsigal: Okay, I believe I've implemented the necessary changes for this part correctly. I'll submit a new PR in a second and let you take a look.

engine/resources/exec.go
```diff
@@ -138,6 +142,10 @@ func (obj *ExecRes) Validate() error {
 		}
 	}
 
+	// check that environment variables' format is valid
+	for key := range obj.Env {
+		isNameValid(key)
```
purpleidea: Missing something here...

engine/resources/exec.go
```diff
@@ -277,6 +285,14 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 		}
 		cmd := exec.Command(cmdName, cmdArgs...)
 		cmd.Dir = obj.IfCwd // run program in pwd if ""
+
+		cmdEnv := []string{}
+		for key := range obj.Env {
+			cmdEnv = append(cmdEnv, key+"="+obj.Env[key])
```
purpleidea: This actually _does_ do what we want, but it's non-standard and ends up sorting more than we need to. Try and implement the above described solution that *just* sorts the map keys. They're unique (it's a map) so we don't need to sort the values alongside them.

adamsigal: Let me make sure I'm understanding correctly. What you're suggesting is
1. Make a list of the keys by iterating through map
2. Sort that list
3. Iterate through sorted list, in each iteration append "<key>=<value>" to cmd.Env

If this is what you mean, isn't it a worse solution to have to use 2 for loops to accomplish the same thing?

engine/resources/exec.go
adamsigal: return, right?

purpleidea: errwrap.Wrapf...

engine/resources/exec.go
purpleidea: Yes and no. Technically there's a loop inside the Sort() that you're using, and you're sorting additional stuff which isn't needed.

purpleidea: Plus it would be good to get you to learn the standard sort approach when the map keys aren't sortable or are a different type.

engine/resources/exec.go
adamsigal: Is it not redundant to have error messages in the `isNameValid` function _and_ in the errwrap?

purpleidea: this currently does nothing

adamsigal: I changed it in the most recent pull request to 
```
for key := range obj.Env {
	return errwrap.Wrapf(isNameValid(key), "invalid variable name")
}
return nil
```

purpleidea:
If you rebase to current master, the builds should be fixed now.

engine/resources/exec.go
```diff
@@ -54,7 +55,7 @@ type ExecRes struct {
 	// only be used when a Shell is *not* specified. The advantage of this
 	// is that you don't have to worry about escape characters.
 	Args []string `yaml:"args"`
-	// Cmd is the dir to run the command in. If empty, then this will use
+	// Cwd is the dir to run the command in. If empty, then this will use
```
purpleidea: This is a good catch, but technically it's unrelated to this commit and so it should be in a separate commit. If the commit as-is was perfect except for this, I'd probably make an exception and just merge, but it's good practice for you to learn how to "split a commit" (google or ddg that phrase) so let's do it here! (And remember to fix commit message!)

engine/resources/exec.go
```diff
@@ -277,6 +285,7 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
 		}
 		cmd := exec.Command(cmdName, cmdArgs...)
 		cmd.Dir = obj.IfCwd // run program in pwd if ""
+
```
purpleidea: Random nl add? This sort of mistake should be caught by you running `git diff` *before* you submit the patch. Please do =D

engine/resources/exec.go
```diff
@@ -368,6 +377,18 @@ func (obj *ExecRes) CheckApply(apply bool) (bool, error) {
...
 	cmd := exec.CommandContext(ctx, cmdName, cmdArgs...)
 	cmd.Dir = obj.Cwd // run program in pwd if ""
+
+	envKeys := []string{}
+	for i := range obj.Env {
+		envKeys = append(envKeys, i)
+	}
+	sort.Strings(envKeys)
+	cmdEnv := []string{}
+	for k := range envKeys {
```
purpleidea: Use the `for _, x := range` form and code will look nicer.

engine/resources/exec.go
```diff
@@ -800,3 +821,21 @@ func (obj *wrapWriter) Write(p []byte) (int, error) {
 func (obj *wrapWriter) String() string {
 	return obj.Buffer.String()
 }
+
+// checks that environment variable name is valid
```
purpleidea: docstring in wrong format for two reasons. Please fix and tell me what the two mistakes are.

engine/resources/exec.go
```diff
@@ -800,3 +821,21 @@ func (obj *wrapWriter) Write(p []byte) (int, error) {
...
+func isNameValid(varName string) error {
+	if varName == "" {
+		return fmt.Errorf("variable name cannot be an empty string")
+	}
+	i := 0
```
purpleidea: This line and the next can be combined. Know how?

engine/resources/exec.go
```diff
@@ -800,3 +821,21 @@ func (obj *wrapWriter) Write(p []byte) (int, error) {
...
+	for i = 0; i < len(varName); i++ {
+		c := varName[i]
+		if i == 0 && '0' <= c && c <= '9' {
```
purpleidea: This is yucky. Just use strings.HasPrefix outside the loop.

engine/resources/exec.go
```diff
@@ -800,3 +821,21 @@ func (obj *wrapWriter) Write(p []byte) (int, error) {
...
+			return fmt.Errorf("variable name cannot begin with number")
+		}
+		if !(c == '_' || '0' <= c && c <= '9' || 'a' <= c && c <= 'z' || 'A' <= c && c <= 'Z') {
```
purpleidea: O.K., but double check there isn't a more succinct or idiomatic alternative in the `strings` package.

engine/resources/resources_test.go
```diff
@@ -286,6 +286,44 @@ func TestResources1(t *testing.T) {
...
+		r := makeRes("exec", "x1")
+		res := r.(*ExecRes) // if this panics, the test will panic
+		s := "hello, world"
+		f := "/tmp/whatever"
+		res.Cmd = fmt.Sprintf("echo '%s' > '%s'", s, f)
+		res.Shell = "/bin/bash"
+		res.IfCmd = "! diff <(cat /tmp/whatever) <(echo hello, world)"
+		res.IfShell = "/bin/bash"
+		res.WatchCmd = fmt.Sprintf("/usr/bin/inotifywait -e modify -m %s", f)
+		res.Env = map[string]string{
```
purpleidea: It would be logical to put this higher next to the part of the resource it goes with, right?

engine/resources/resources_test.go
```diff
@@ -286,6 +286,44 @@ func TestResources1(t *testing.T) {
...
+			"boiling": "one hundred", "freezing": "zero", "room": "twenty-five",
```
purpleidea: I like =D
But put one on each line instead of the awkward split of three and then two.

engine/resources/resources_test.go
```diff
@@ -286,6 +286,44 @@ func TestResources1(t *testing.T) {
...
+			"house": "twenty-two", "canada": "negative thirty",
+		}
+		//res.WatchShell = "/bin/bash"
+
+		timeline := []Step{
```
purpleidea: Do we ever check in here somewhere that `Env` actually did anything? IOW if you took all the `Env` code in CheckApply and other methods away, would this test still pass? If so, then it's missing something.

examples/lang/exec_env.mcl
```diff
@@ -0,0 +1,7 @@
+exec "exec2" {
+	shell => "/bin/bash",
+	ifcmd => "/bin/true",
+	ifshell => "/bin/bash",
+	env => {"BOILING" => "one hundred",},
+    cmd => "/usr/bin/env",
```
purpleidea: wat! (tabs!)

engine/resources/exec.go
adamsigal: do you want me to do this for all of my for loops?

engine/resources/exec.go
adamsigal: No trailing period, and the comment doesn't begin with the function name.

engine/resources/exec.go
adamsigal: I believe lines 830 and 831 can be combined like so:
```
for i := 0; i < len(varName); i++ {
```

engine/resources/exec.go
adamsigal: Maybe I'm missing something, but wouldn't I have to loop through characters 0-9 if I used strings.HasPrefix? That doesn't sound like the simplest solution.

adamsigal: come to think of it I could just use regex

engine/resources/exec.go
adamsigal: Only in the strings package, or should I consider other packages too (like regex)

engine/resources/resources_test.go
adamsigal: I don't know what you're referring to

engine/resources/exec.go
purpleidea: It depends on the case. In this case it pulls out the value automatically so you don't need to have it look more complex. Know what I mean?

engine/resources/exec.go
purpleidea: Bingo =D

engine/resources/exec.go
purpleidea: Just look in strings for now. Regex should be used as a last resort.

engine/resources/exec.go
adamsigal: gotcha

engine/resources/exec.go
adamsigal: Am I wrong that it would be more simple to not add a second loop?

engine/resources/exec.go
adamsigal: I can't find a more succinct way in the strings package.

engine/resources/exec.go
purpleidea: There are more readable ways to write this, but after looking at it further, they're less efficient. You can keep it as is, but at least use the loop `range` construct.

engine/resources/resources_test.go
adamsigal: I just tested it with a print. It does in fact add the correct variables to env like it is supposed to. However, if I take out necessary code, the test still passes... I'm going to keep working on it.

engine/resources/resources_test.go
adamsigal: It's right after the "simple exec" test, it seems well placed to me.

engine/resources/resources_test.go
purpleidea: Nowhere in this test are you actually testing the ENV parts :/

engine/resources/exec.go
purpleidea: Not sure you did this...

purpleidea:
I think this LGTM!

adamsigal:
> I think this LGTM!

Yes!

purpleidea:
Sorry for the delay in actually merging! It was queued up behind my messing working directory!

All in now \o/

PR #549, opened 2019-10-07
misc: Point to the archives to download the box

jcftang:
Fedora has archived fedora-28 and the vagrant cloud redirects are
broken. This fixes the download so new devs won't need to go digging
around for the archive link.

This may not be the right fix as vagrant supports a bunch of different providers for base boxes.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,7 +6,7 @@ Vagrant.configure(2) do |config|
 	config.vm.synced_folder ".", "/vagrant", disabled: true
 
 	config.vm.define "mgmt-dev" do |instance|
-		instance.vm.box = "fedora/28-cloud-base"
+		instance.vm.box = "fedora/30-cloud-base"
 	end
 
 	config.vm.provider "virtualbox" do |v|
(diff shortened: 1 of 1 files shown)


purpleidea:
@jcftang Cool =D

TBH I don't really use the Vagrantfile, and so I think it is a bit bit rotten, but patch is good for now =D

You're more than welcome to play with Vagrant side of things and improve it if you like!

Patch LGTM, will merge shortly. Thanks!

jcftang:
I've one or two minor changes to this if you can wait to merge. The vagrantfile looks like its a bit stale

jcftang:
This pr just makes the vagrant box come up, but it's not in a usable state at all. I think I will open a separate PR to make it actually usable. I'm interested in being able to try things out in a self-contained environment since my work machine is a mac.

purpleidea:
@jcftang Okay that's a good idea, thanks! I shall wait and we can merge a larger "clean up" patch for it instead. Thank you!

jcftang:
@purpleidea it's in a reasonable state for someone to use I think

```
FAIL
The following tests have failed:
./test/test-govet.sh
./test/test-examples.sh
./test/test-gotest.sh
./test/test-gometalinter.sh

You can rerun a single suite like so:

`make test-gofmt` or `make test-shell-<testname>`
```

purpleidea:
@jcftang Patch LGTM, except the commit message doesn't fit anymore.
Also, I don't see the failing tests you quoted...

purpleidea:
@jcftang Hiya, didn't realize you had pushed a new version, sorry. (Please ping when you do because GH doesn't send notifications of updated patches.) -- Why remove hg?

jcftang:
@purpleidea there was another error where if hg was present the install would error out due to a duplication elsewhere, removing it silences an error, its not ideal but it's less noisy for new users imho

purpleidea:
@jcftang Can you paste the error please?

jcftang:
@purpleidea I dont seem to be able to recreate the error, i can undo that change

purpleidea:
@jcftang sounds good!

purpleidea:
Out Vagrantfile points to F31 now, so closing. @jcftang Please ping if you'd like to get involved elsewhere. Thanks!

PR #550, opened 2019-10-08
engine: resources: Support a symbolic mode parameter for the file res (WIP)

dbakong:
This patch would add symbolic filemode support to the file resource. Currently incomplete.

## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/engine/resources/symbolic.go
+++ b/engine/resources/symbolic.go
@@ -1,0 +1,109 @@
+package resources
+
+import (
+	"fmt"
+	"log"
+	"strings"
+)
+
+// symbolicMode calculates the filemodes from the symbolic file input
+func symbolicMode(p string) (mode string, err error) {
+	var octal int64
+	var out int64
+	m := parse(p)
+	for _, v := range m {
+		v = string(v)
+		switch v[0] {
+		default:
+			err = fmt.Errorf("error")
+			return v, err
+		case 'g':
+			switch v[2] {
+			case 'r':
+				octal = groupRead
+			case 'w':
+				octal = groupWrite
+			case 'x':
+				octal = groupExecute
+			}
+		case 'o':
+			switch v[2] {
+			case 'r':
+				octal = otherRead
+			case 'w':
+				octal = otherWrite
+			case 'x':
+				octal = otherExecute
+			}
+		case 'u':
+			switch v[2] {
+
+			case 'r':
+				octal = userRead
+			case 'w':
+				octal = userWrite
+			case 'x':
+				octal = userExecute
+			}
+		}
+		out = out + octal
+		if err != nil {
+			log.Fatal("Invalid perm")
+
+		}
+	}
+
+	mode = fmt.Sprintf("%#04o", out)
+	fmt.Println(mode)
+	return mode, err
+}
+
+// parse returns individual file modes from a string
+func parse(symbolic string) []string {
+	var p []string
+	for _, v := range strings.Split(symbolic, ",") {
+		v = strings.ReplaceAll(v, "a", "ugo")
+		s := strings.Split(v, "=")
+		for _, w := range s[0] {
+			for _, q := range s[1] {
+				c := string(w) + "=" + string(q) + "\n"
+				p = append(p, c)
+			}
+		}
+	}
+	p = unique(p)
+	return p
+}
+
+// unique removes repeating filemodes
+func unique(p []string) []string {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: engine/resources/symbolic.go [hunk 1]
```diff
@@ -0,0 +1,109 @@
...
+			case 'r':
+				octal = otherRead
+			case 'w':
+				octal = otherWrite
+			case 'x':
+				octal = otherExecute
+			}
+		case 'u':
+			switch v[2] {
+
```
purpleidea: extra white space not seen in other switches

engine/resources/symbolic.go
```diff
@@ -0,0 +1,109 @@
...
+				octal = userRead
+			case 'w':
+				octal = userWrite
+			case 'x':
+				octal = userExecute
+			}
+		}
+		out = out + octal
+		if err != nil {
+			log.Fatal("Invalid perm")
```
purpleidea: You should never fatal deep inside the code. If at all, then only at the top-level. This would cause the whole program to die and potentially corrupt any in progress database operation or write. We always return an error and propagate upwards.

engine/resources/symbolic.go
```diff
@@ -0,0 +1,109 @@
...
+				c := string(w) + "=" + string(q) + "\n"
+				p = append(p, c)
+			}
+		}
+	}
+	p = unique(p)
+	return p
+}
+
+// unique removes repeating filemodes
```
purpleidea: Don't we want to either (1) error or (2) take the last one, instead?

engine/resources/symbolic_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+		// currently failing tests
+		// {input: "o=args", err: fmt.Errorf("permission setting failed")},
+		// {input: "o=arq", err: fmt.Errorf("permission setting failed")},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.input, func(t *testing.T) {
+			result, err := symbolicMode(tc.input)
+			if err != tc.err {
+				t.Errorf("error mismatch. Actual: %v, Expected: %v", err, tc.err)
```
purpleidea: typically we'd add a `return` here to end this test.

engine/resources/symbolic_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{input: "u=r,g=w,o=x", want: "0421"},
+		{input: "ug=rw,o=rx", want: "0665"},
+		{input: "g=r,ugo=x,a=w", want: "0373"},
+
+		// currently failing tests
+		// {input: "o=args", err: fmt.Errorf("permission setting failed")},
+		// {input: "o=arq", err: fmt.Errorf("permission setting failed")},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.input, func(t *testing.T) {
```
purpleidea: There's a possible bug here... It's tricky. Learn it and explain it...

https://golang.org/doc/articles/race_detector.html#Race_on_loop_counter

engine/resources/symbolic.go
```diff
@@ -0,0 +1,109 @@
...
+	mode = fmt.Sprintf("%#04o", out)
+	fmt.Println(mode)
+	return mode, err
+}
+
+// parse returns individual file modes from a string
+func parse(symbolic string) []string {
+	var p []string
+	for _, v := range strings.Split(symbolic, ",") {
+		v = strings.ReplaceAll(v, "a", "ugo")
```
purpleidea: This is a bit of a hack ;) Cute though

engine/resources/symbolic_test.go
dbakong: I ran this package's test with the race detector and couldn't find any race condition.
My guess would be `t *testing.T` potentially being an accidentally shared variable between `func(t *testing.T)` and `Test_mode(t *testing.T)`

purpleidea:
@dbakong Now that the other parser is merged, can you rebase this patch and add the tests! Those are quite good, and it would be great to check they work too!

purpleidea:
I think we can close this now. If in error, LMK. Thanks again!

PR #551, opened 2019-10-16
engine, resources: Adds symbolic mode parsing to file resource

dgbuckley:
Adds symbolic mode parsing functionality to the "util" package for parsing in the
file resource.

### DIFF
--- a/engine/resources/file.go
+++ b/engine/resources/file.go
@@ -94,8 +94,7 @@ type FileRes struct {
 	// name, or a string representation of the group integer gid.
 	Group string `lang:"group" yaml:"group"`
 	// Mode is the mode of the file as a string representation of the octal
-	// form.
-	// TODO: add symbolic representations
+	// form or symbolic form.
 	Mode    string `lang:"mode" yaml:"mode"`
 	Recurse bool   `lang:"recurse" yaml:"recurse"`
 	Force   bool   `lang:"force" yaml:"force"`
@@ -137,10 +136,23 @@ func (obj *FileRes) isDir() bool {
 // the case where the mode is not specified. The caller should check obj.Mode is
 // not empty.
 func (obj *FileRes) mode() (os.FileMode, error) {
-	m, err := strconv.ParseInt(obj.Mode, 8, 32)
-	if err != nil {
-		return os.FileMode(0), errwrap.Wrapf(err, "mode should be an octal number (%s)", obj.Mode)
+	if n, err := strconv.ParseInt(obj.Mode, 8, 32); err == nil {
+		return os.FileMode(n), nil
 	}
+
+	// Try parsing symbolic by first getting the files current mode.
+	stat, err := os.Stat(obj.getPath())
+	if err != nil {
+		return os.FileMode(0), errwrap.Wrapf(err, "failed to get the current file mode")
+	}
+	m := stat.Mode()
+
+	modes := strings.Split(obj.Mode, ",")
+	m, err = engineUtil.ParseSymbolicModes(modes, m, false)
+	if err != nil {
+		return os.FileMode(0), errwrap.Wrapf(err, "mode should be an octal number or symbolic mode (%s)", obj.Mode)
+	}
+
 	return os.FileMode(m), nil
 }
 
--- a/engine/util/mode.go
+++ b/engine/util/mode.go
@@ -1,0 +1,307 @@
+// Mgmt
+// Copyright (C) 2013-2019+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package util
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+// Constant bytes for the who (u, g, or o) and the what (r, w, x, s, or t).
+const (
+	ModeUser  uint32 = 64
+	ModeGroup uint32 = 8
+	ModeOther uint32 = 1
+
+	ModeRead  uint32 = 4
+	ModeWrite uint32 = 2
+	ModeExec  uint32 = 1
+
+	ModeSetU   uint32 = 4
+	ModeSetG   uint32 = 2
+	ModeSticky uint32 = 1
+)
+
+// modeIsValidWho checks that only the characters 'u', 'g', 'o' are in the who
+// string. It expects that 'a' was expanded to "ugo" already and will return
+// false if not.
+func modeIsValidWho(who string) bool {
+	for _, w := range []string{"u", "g", "o"} {
+		who = strings.Replace(who, w, "", -1)
+	}
+	return len(who) == 0
+}
+
+// modeIsValidWhat checks that only the valid mode characters are in the what
+// string ('w', 'r', 'x', 's', 't').
+func modeIsValidWhat(what string) bool {
+	for _, w := range []string{"w", "r", "x", "s", "t"} {
+		what = strings.Replace(what, w, "", -1)
+	}
+	return len(what) == 0
+}
+
+// modeAssigned executes an assigment symbolic mode string (u=r). It clears out
+// any bits for every subject in who and then assigns the specified modes in
+// what.
+func modeAssigned(who, what string, from os.FileMode) (os.FileMode, error) {
+	// Clear out any users defined in 'who'.
+	for _, w := range who {
+		switch w {
+		case 'u':
+			from = from &^ os.FileMode(448) // 111000000 in bytes
+			from = from &^ os.ModeSetuid
+		case 'g':
+			from = from &^ os.FileMode(56) // 111000 in bytes
+			from = from &^ os.ModeSetgid
+		case 'o':
+			from = from &^ os.FileMode(7) // 111 in bytes
+		}
+	}
+
+	for _, c := range what {
+		switch c {
+		case 'w':
+			m := modeValueFrom(who, ModeWrite)
+			if from&m == 0 {
+				from = from | m
+			}
+		case 'r':
+			m := modeValueFrom(who, ModeRead)
+			if from&m == 0 {
+				from = from | m
+			}
+		case 'x':
+			m := modeValueFrom(who, ModeExec)
+			if from&m == 0 {
+				from = from | m
+			}
+		case 's':
+			for _, w := range who {
+				switch w {
+				case 'u':
+					from = from | os.ModeSetuid
+				case 'g':
+					from = from | os.ModeSetgid
+				}
+			}
+		case 't':
+			from = from | os.ModeSticky
+		default:
+			return os.FileMode(0), fmt.Errorf("invalid character %q", c)
+		}
+	}
+
+	return from, nil
+}
+
+// modeAdded executes an addition symbolic mode string (u+x) and will add the
+// bits requested in what if not present.
+func modeAdded(who, what string, from os.FileMode) (os.FileMode, error) {
+	for _, c := range what {
+		switch c {
+		case 'w':
+			m := modeValueFrom(who, ModeWrite)
+			if from&m == 0 {
+				from = from | m
+			}
+		case 'r':
+			m := modeValueFrom(who, ModeRead)
+			if from&m == 0 {
+				from = from | m
+			}
+		case 'x':
+			m := modeValueFrom(who, ModeExec)
+			if from&m == 0 {
+				from = from | m
+			}
+		case 's':
+			for _, w := range who {
+				switch w {
+				case 'u':
+					from = from | os.ModeSetuid
+				case 'g':
+					from = from | os.ModeSetgid
+				}
+			}
+		case 't':
+			from = from | os.ModeSticky
+		default:
+			return os.FileMode(0), fmt.Errorf("invalid character %q", c)
+		}
+	}
+
+	return from, nil
+}
+
+// modeSubtracted executes an subtraction symbolic mode string (u+x) and will
+// removethe bits requested in what if present.
+func modeSubtracted(who, what string, from os.FileMode) (os.FileMode, error) {
+	for _, c := range what {
+		switch c {
+		case 'w':
+			m := modeValueFrom(who, ModeWrite)
+			if from&m != 0 {
+				from = from &^ m
+			}
+		case 'r':
+			m := modeValueFrom(who, ModeRead)
+			if from&m != 0 {
+				from = from &^ m
+			}
+		case 'x':
+			m := modeValueFrom(who, ModeExec)
+			if from&m != 0 {
+				from = from &^ m
+			}
+		case 's':
+			for _, w := range who {
+				switch w {
+				case 'u':
+					if from&os.ModeSetuid != 0 {
+						from = from &^ os.ModeSetuid
+					}
+				case 'g':
+					if from&os.ModeSetgid != 0 {
+						from = from &^ os.ModeSetgid
+					}
+				}
+			}
+		case 't':
+			if from&os.ModeSticky != 0 {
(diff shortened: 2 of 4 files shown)


purpleidea:
@dgbuckley Thanks for the patch!

As a funny coincidence, a mentee of the project was also working on this issue as a learning exercise, however this version is much closer to what we're looking for.

There are a few review items which I hope you wouldn't mind fixing, but I was wondering if you would be okay waiting a short while to give my mentee a chance to practice reviewing this first.

@dbakong (currently learning golang) will review as a learning exercise. Please ignore those comments. After his are submitted, I'll do a final review.

Hope that's okay, and thank you for your patch!

util/sym-mode_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/purpleidea/mgmt/util"
+)
+
+func TestSymbolicMode(t *testing.T) {
+	def := os.FileMode(0644) | os.ModeSetgid
+	symModeTests := []struct {
+		input  string
+		expect os.FileMode
+	}{
+		{"ug=rwx", 0774 | os.ModeSetgid},
```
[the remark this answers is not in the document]
purpleidea: Or maybe that should be an error?

dgbuckley:
@purpleidea that sounds great. No rush on this PR, I'm just excited to be contributing to this awesome project. Also sorry about the overlap of work; I'll do better to watch PRs moving forward.

engine/resources/file.go
```diff
@@ -137,9 +137,19 @@ func (obj *FileRes) isDir() bool {
 // the case where the mode is not specified. The caller should check obj.Mode is
 // not empty.
 func (obj *FileRes) mode() (os.FileMode, error) {
-	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	n, err := strconv.ParseInt(obj.Mode, 8, 32)
+	m := os.FileMode(n)
```
purpleidea: Having this between the `xxx, err := ...` and the `if err != nil` is awkward, might be better to inline the code and exit early if `err == nil`.

engine/resources/file.go
```diff
@@ -137,9 +137,19 @@ func (obj *FileRes) isDir() bool {
...
 	if err != nil {
-		return os.FileMode(0), errwrap.Wrapf(err, "mode should be an octal number (%s)", obj.Mode)
+		// Try parsing symbolic
+		modes := strings.Split(obj.Mode, ",")
+		m = os.FileMode(0)
```
purpleidea: If this is the initial condition, don't we want this to be the mode of the current file since we might +/- on it?

util/sym-mode.go
```diff
@@ -0,0 +1,262 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package util
+
+import (
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
```
purpleidea: use our util errwrap package instead please

util/sym-mode.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+const (
```
purpleidea: public const block needs a comment for golint to pass IIRC.

engine/resources/file.go
```diff
@@ -137,9 +137,19 @@ func (obj *FileRes) isDir() bool {
 // the case where the mode is not specified. The caller should check obj.Mode is
 // not empty.
 func (obj *FileRes) mode() (os.FileMode, error) {
-	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	n, err := strconv.ParseInt(obj.Mode, 8, 32)
+	m := os.FileMode(n)
 	if err != nil {
-		return os.FileMode(0), errwrap.Wrapf(err, "mode should be an octal number (%s)", obj.Mode)
+		// Try parsing symbolic
+		modes := strings.Split(obj.Mode, ",")
```
purpleidea: As per @dbakong's comment, I think we should track duplicates so that we don't allow `u=rw,u=r` or `u+rw,u-r` sorts of things. As a result of adding that logic, perhaps this should be in a separate helper function?

util/sym-mode.go
```diff
@@ -0,0 +1,262 @@
...
+	// var i int
+	if strings.Contains(mode, "=") && !strings.ContainsAny(mode, "+-") {
+		return modeAssigned(mode, from)
+	} else if strings.Contains(mode, "+") && !strings.ContainsAny(mode, "=-") {
+		return modeAdded(mode, from)
+	} else if strings.Contains(mode, "-") && !strings.ContainsAny(mode, "=+") {
+		return modeSubtracted(mode, from)
+	}
+
+	return from, errors.Errorf("%s is not a valid symbolic mode", mode)
```
purpleidea: fmt.Errorf, and the return value should be the zero or nil.
Same in other places too.

util/sym-mode.go
```diff
@@ -0,0 +1,262 @@
...
+// applies it to the from FileMode.
+//
+// SymbolicFile mode is expected to be a list of who (user, group, other) then
+// the operation (=, +, -) then the change (read, write, execute, setuid,
+// setgid, sticky).
+//
+// Ex. ug=rw
+func SymbolicMode(mode string, from os.FileMode) (os.FileMode, error) {
+	// parse the who
+	// var i int
```
purpleidea: dead code?

util/sym-mode.go
```diff
@@ -0,0 +1,262 @@
...
+	if strings.Contains(mode, "=") && !strings.ContainsAny(mode, "+-") {
```
purpleidea: Why isn't this `mode == "="` ?

util/sym-mode.go
```diff
@@ -0,0 +1,262 @@
...
+
+func modeAssigned(mode string, from os.FileMode) (os.FileMode, error) {
+	var who, what string
+	if sp := strings.Split(mode, "="); len(sp) != 2 {
+		return from, errors.Errorf("invalid mode: expected only one '=' but got %d", len(sp))
+	} else {
+		who, what = sp[0], sp[1]
+	}
+
+	// Validate input
```
purpleidea: Short comments are either lower case snippets or full sentences that are capitalized and end with a period. Same comment for other places. https://github.com/purpleidea/mgmt/blob/master/docs/style-guide.md#comments

util/sym-mode_test.go
```diff
@@ -0,0 +1,58 @@
...
+		{"o-rwx", 0640 | os.ModeSetgid},
+		{"u-w", 0444 | os.ModeSetgid},
+		{"g-s", 0644},
+		{"u-t", 0644 | os.ModeSetgid},
+	}
+
+	for _, test := range symModeTests {
+		t.Run(test.input, func(t *testing.T) {
+			if got, err := util.SymbolicMode(test.input, def); err != nil {
+				t.Fatalf("Failed to parse the symbolic string: %v", err)
```
purpleidea: Errorf instead of fatal please. You can return or continue as needed.

util/sym-mode_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+
+	"github.com/purpleidea/mgmt/util"
+)
+
+func TestSymbolicMode(t *testing.T) {
+	def := os.FileMode(0644) | os.ModeSetgid
+	symModeTests := []struct {
+		input  string
+		expect os.FileMode
```
purpleidea: Please look at our other tests to add an `experr` sort of variable or similar. The recent pr by @dbakong shows this too.

util/sym-mode_test.go
```diff
@@ -0,0 +1,58 @@
...
+		{"u+s", 0644 | os.ModeSetgid | os.ModeSetuid},
+		{"u+t", 0644 | os.ModeSetgid | os.ModeSticky},
+		{"o-rwx", 0640 | os.ModeSetgid},
+		{"u-w", 0444 | os.ModeSetgid},
+		{"g-s", 0644},
+		{"u-t", 0644 | os.ModeSetgid},
+	}
+
+	for _, test := range symModeTests {
+		t.Run(test.input, func(t *testing.T) {
```
purpleidea: Needs to be race free... See other tests to avoid this. https://golang.org/doc/articles/race_detector.html#Race_on_loop_counter

util/sym-mode.go
```diff
@@ -0,0 +1,262 @@
+// Mgmt
```
purpleidea: Please move this file to engine/util/ and name it mode.go or something without a dash in it please. (Same with test.)

purpleidea:
On Thu, Oct 17, 2019 at 4:01 PM dgbuckley <notifications@github.com> wrote:
>
> @purpleidea that sounds great. No rush on this PR, I'm just excited to be contributing to this awesome project. Also sorry about the overlap of work; I'll do better to watch PRs moving forward.

It's not a problem. We're happy to have you as a new contributor. I
just finished the review. Please feel free to join us in #mgmtconfig
on Freenode IRC or ping me wherever if you're looking for more hacking
with us! Happy to help you find something to work on if you don't
already have something in mind.

engine/resources/file.go
dgbuckley: My initial interpretation of symbolic mode is that it should be idempotent and it shouldn't matter if you gave `u=x,u=r,u=rw` you would just get `u=rw`. Thinking about it further that may allow for some confusing inputs. Should repeated subjects (ex uug=rw) be an error as well?

util/sym-mode.go
dgbuckley: My strategy was to check that the mode string (ex `u=rw`) containes a `=` anywhere and no `+/-` which is the bare minimum check to say it is an assignment.

purpleidea: Oh, okay. I'm okay with this, can you please add a short, clear sentence comment explaining that logic at the top of the `if` please? I might just be out of it today, but since this is sort of untraditional for parsing style, a quick reminder will help! =D

engine/resources/file.go
purpleidea: Honestly, unless the `chmod` "standard" allows such weird things (please have a look) then we shouldn't allow them at all. We want to be less permissive on what we accept so that accidentally incorrect input here causes an error instead of silently allowing wrong behaviour.

As a slight warning: I was planning to write this patch (but you're welcome to include it if you like) but while it's great that you support the `+` and `-`, I was thinking of having those be enabled by a bool const. That way, we can leave support for those off by default. My reasoning: I'm contemplating whether they're really necessary here or not. But I like having the possibility to flip it on if we find a good reason. Cheers!

util/sym-mode_test.go
dgbuckley: In that PR the errors are being compared and even though  the text is the same they are separate error values so it won't pass that way ([playground](https://play.golang.org/p/nEYvEGxgVg3)). Would it be preferred to check for nil then compare output of `Error()` or could error constants and `errors.Is` be used?

util/sym-mode_test.go
dgbuckley: Sorry I thought range with `t.Run` was race free because each run has its own struct from the slice. Does a copy need to be made before `t.Run()`?

engine/resources/file.go
dgbuckley: If there is an error getting the mode from the file should it be raised then or just set to 0 and have that handled elsewhere?

purpleidea: The first one. So if when trying to read the file mode that errors, then we propagate that error up and the engine handles it.

util/sym-mode_test.go
purpleidea: If we made that error compare somewhere, please let me know! In general it's best to compare error constants, or if it's part of a test case and we're checking for a particular error that the user never needs to compare to, then the string compare is okay.

util/sym-mode_test.go
purpleidea: Honestly I have to dig into the code again and check, but at one time I'm fairly sure it wasn't race free. So I did this: https://github.com/purpleidea/mgmt/blob/2ea492c96512a0802c98c9a06984457e88d5f2a4/lang/interpret_test.go#L431

dgbuckley:
@purpleidea I think I covered all of your feedback. I changed the main function to accept a slice of modes and fail safely if there is repeated subjects. I also added the support for your suggestion to make +/- toggle with a bool and the missing 'a' subject I realized I forgot.

engine/resources/file.go
```diff
@@ -137,10 +137,25 @@ func (obj *FileRes) isDir() bool {
 // the case where the mode is not specified. The caller should check obj.Mode is
 // not empty.
 func (obj *FileRes) mode() (os.FileMode, error) {
-	m, err := strconv.ParseInt(obj.Mode, 8, 32)
+	if n, err := strconv.ParseInt(obj.Mode, 8, 32); err == nil {
+		return os.FileMode(n), nil
+	}
+
+	// Try parsing symbolic by first getting the files current mode.
+	m := os.FileMode(0)
```
purpleidea: You can inline this section and not do this awkward initialization. Eg:

```
stat, err := ...
if err != nil {
...
}
m := stat.Mode()
```

Better?

engine/resources/file.go
```diff
@@ -137,10 +137,25 @@ func (obj *FileRes) isDir() bool {
...
+	if stat, err := os.Stat(obj.getPath()); err != nil {
+		return os.FileMode(0), errwrap.Wrapf(err, "failed to get the current file mode")
+	} else {
+		m = stat.Mode()
+	}
+
+	modes := strings.Split(obj.Mode, ",")
+	var err error
```
purpleidea: and then you can nuke this line

engine/util/mode.go
```diff
@@ -0,0 +1,297 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+type ModeError string
+
+func (m ModeError) Error() string { return string("invalid mode: " + m) }
```
purpleidea: fmt.Sprintf instead please.

purpleidea: https://github.com/purpleidea/mgmt/blob/master/docs/style-guide.md#method-receiver-naming please.

purpleidea: Note: I'm not actually sure why you're defining a special error type unless you want to check against it somewhere. Eg:

https://github.com/purpleidea/mgmt/blob/06952c224b734b5146539d1fedc953e7b1a50041/lang/interfaces/error.go#L29

If you're planning to, go ahead, if not you can just use: `fmt.Errorf` in the error returns.

engine/util/mode.go
```diff
@@ -0,0 +1,297 @@
...
+		return os.FileMode(0), ModeError(fmt.Sprintf("%s does not appear to me a symbolic mode", symModes[i].mode))
+	}
+
+	// Validate input and verify the slice of symbolic modes does not contain
+	// redundancy.
+	seen := make(map[rune]struct{}) // To validate all subjects are not dupicated
+	for i, _ := range symModes {
+		if strings.ContainsRune(symModes[i].who, 'a') || symModes[i].who == "" {
+			// If 'a' or empty who (implicit 'a') is called and there are more
+			// symbolic modes in the slice then it must be a repitition.
```
purpleidea: repetition repetition

purpleidea:
PS: This commit should also remove this line and update the docstring above it:

https://github.com/purpleidea/mgmt/blob/06952c224b734b5146539d1fedc953e7b1a50041/engine/resources/file.go#L98

And the docs/resources.md file mode section too please!

dgbuckley:
@purpleidea. I think I've covered all of your feedback. I removed the error type because I only added it for testing and the same can be achieved with checking the error string.

purpleidea:
@dgbuckley Doesn't build: https://travis-ci.org/purpleidea/mgmt/jobs/602558300#L837 and sorry commit will fail unless it's `engine: resources: Add symbolic mode to file resource` (the resources prefix needs an S)

purpleidea:
@dgbuckley I'm planning on merging this shortly, thanks! There are a few nitpicks that I was planning on adding in a subsequent commit. Is this okay with you or would you prefer I enumerate them for you to fix up instead? I only mention because they're possibly easier to understand as a diff.

Thanks again!

dgbuckley:
You can definitely make any changes in a later commit; I look forward to seeing the diff. Thank you for the feedback you have provides.

purpleidea:
This patch has now been rebased to master and merged! Thank you again!
As discussed I made a few cleanups to get this in. This is in 1eba5833d5418b9473b9e961a455c7ae2e9466fe I've also updated the style guide so that I can better provide feedback about things to change. Please have a look at that commit, and let me know if you agree with it all or if you have any questions!

Thank you again for your patch, and hopefully we can persuade you to write something else! -- Adding a Reverse method to an existing resource might be a good one!

Cheers!

PR #557, opened 2019-10-31
engine: resources: consul:kv resource

roidelapluie:

## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/engine/resources/consul_kv.go
+++ b/engine/resources/consul_kv.go
@@ -1,0 +1,221 @@
+// Mgmt
+// Copyright (C) 2013-2019+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package resources
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"sync"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/engine/traits"
+	"github.com/purpleidea/mgmt/util"
+	"github.com/purpleidea/mgmt/util/errwrap"
+)
+
+func init() {
+	engine.RegisterResource("consul:kv", func() engine.Res { return &ConsulKVRes{} })
+}
+
+// ConsulKVRes is a resource that writes a value into a Consul datastore. The
+// name of the resource can either be the key name, or the concatenation of the
+// server address and the key name: http://127.0.0.1/my-key.
+type ConsulKVRes struct {
+	traits.Base
+	init *engine.Init
+
+	// Key is the name of the key. Defaults to the name of the resource.
+	Key string `lang:"key" yaml:"key"`
+
+	// Value is the value for the key.
+	Value string `lang:"value" yaml:"value"`
+
+	// cheme is the URI scheme for the Consul server. Default: http.
+	Scheme string `lang:"scheme" yaml:"scheme"`
+
+	// Address is the address of the Consul server. Default: 127.0.0.1:8500.
+	Address string `lang:"address" yaml:"address"`
+
+	// Token is used to provide an ACL token to use for this resource.
+	Token string `lang:"token" yaml:"token"`
+
+	client *api.Client
+	config *api.Config // needed to close the idle connections
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *ConsulKVRes) Default() engine.Res {
+	return &ConsulKVRes{}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *ConsulKVRes) Validate() error {
+	if obj.Scheme != "" && obj.Scheme != "http" && obj.Scheme != "https" {
+		return fmt.Errorf("unknown Scheme")
+	}
+	return nil
+}
+
+// Init runs some startup code for this resource.
+func (obj *ConsulKVRes) Init(init *engine.Init) error {
+	obj.init = init // save for later
+
+	s, a, k := parseConsulName(obj.Name())
+	if obj.Key == "" {
+		obj.Key = k
+	}
+	if obj.Address == "" {
+		obj.Address = a
+	}
+	if obj.Scheme == "" {
+		obj.Scheme = s
+	}
+
+	obj.config = api.DefaultConfig()
+	if obj.Address != "" {
+		obj.config.Address = obj.Address
+	}
+	if obj.Scheme != "" {
+		obj.config.Scheme = obj.Scheme
+	}
+	if obj.Token != "" {
+		obj.config.Token = obj.Token
+	}
+
+	var err error
+	obj.client, err = api.NewClient(obj.config)
+	return err
+}
+
+// Close is run by the engine to clean up after the resource is done.
+func (obj *ConsulKVRes) Close() error {
+	if obj.config != nil && obj.config.Transport != nil {
+		obj.config.Transport.CloseIdleConnections()
+	}
+	return nil
+}
+
+// Watch is the listener and main loop for this resource and it outputs events.
+func (obj *ConsulKVRes) Watch() error {
+	wg := &sync.WaitGroup{}
+	defer wg.Wait()
+
+	ch := make(chan error)
+	kv := obj.client.KV()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		opts := &api.QueryOptions{RequireConsistent: true}
+		ctx, cancel := util.ContextWithCloser(context.Background(), obj.init.Done)
+		defer cancel()
+		opts = opts.WithContext(ctx)
+
+		for {
+			_, meta, err := kv.Get(obj.Key, opts)
+			select {
+			case ch <- err:
+				if err != nil {
+					return
+				}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by roidelapluie** [comment]
cannot find package "go.uber.org/zap" in any of:


@purpleidea not related to my code..

File: engine/resources/consul_kv.go [hunk 1]
```diff
@@ -0,0 +1,186 @@
...
+func init() {
+	engine.RegisterResource("consul:kv", func() engine.Res { return &ConsulKVRes{} })
+}
+
+// ConsulKVRes is a resource that writes a value into a consul datastore.
+type ConsulKVRes struct {
+	traits.Base
+	init *engine.Init
+
+	// Key is the path of the value.
```
purpleidea: Rephrase, this is a bit confusing.

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+	Key string `lang:"key" yaml:"key"`
+
+	// Key is the value that will be written into consul.
```
purpleidea: Copy-pasta bug

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+	Value string `lang:"value" yaml:"value"` // the consul key value
+
+	// Address is the address of the consul server.
```
purpleidea: Can you add an example too, eg: "http://localhost:999" or whatever the common format is. Use example.com if you need a domain. Should this be named Server instead?

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+	Address string `lang:"address" yaml:"address"` // the consul server
+
+	// Token is the token used to connect to the consul server.
```
purpleidea: Is this a password?

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+
+// Default returns some sensible defaults for this resource.
+func (obj *ConsulKVRes) Default() engine.Res {
+	return &ConsulKVRes{}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *ConsulKVRes) Validate() error {
+	if obj.Key == "" {
+		return errors.New("key is mandatory")
```
purpleidea: fmt.Errorf and use the format used elsewhere have a word before and `Key` is capitalized.

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+	config *api.Config // needed to close the idle connections
+}
+
+// Default returns some sensible defaults for this resource.
+func (obj *ConsulKVRes) Default() engine.Res {
+	return &ConsulKVRes{}
+}
+
+// Validate if the params passed in are valid data.
+func (obj *ConsulKVRes) Validate() error {
```
purpleidea: Shouldn't the unique name be what we use to identify the key, and have the Key and Server params overwrite that if they're defined? Eg: name could be: "http://xxx.com:999/some/key"

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	var err error
+	obj.client, err = api.NewClient(obj.config)
+	return err
+}
+
+// Close is run by the engine to clean up after the resource is done.
+func (obj *ConsulKVRes) Close() error {
+	obj.config.Transport.CloseIdleConnections()
```
purpleidea: You should wrap in `if obj.Config != nil` in case something in Init fails ... Can .Transport be nil?

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+	return nil
+}
+
+// Watch is the listener and main loop for this resource and it outputs events.
+func (obj *ConsulKVRes) Watch() error {
+	wg := &sync.WaitGroup{}
+	defer wg.Wait()
+
+	c := make(chan error)
```
purpleidea: s/c/ch/

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+	}()
+
+	for {
+		select {
+		case err := <-c:
+			if err == nil {
+				obj.init.Event()
+			} else {
+				return err
+			}
```
purpleidea: add a nl after this please

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+				opts.WaitIndex = meta.LastIndex
+				goto POLL
+			}
+		case <-obj.init.Done:
+		}
+	}()
+
+	for {
+		select {
+		case err := <-c:
```
purpleidea: err, ok :=

Look at other resources.

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+			if err == nil {
+				obj.init.Event()
+			} else {
```
purpleidea: inline this

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+		case c <- err:
+			if err == nil {
+				if opts.WaitIndex == 0 {
+					// WaitIndex = 0, which means that it is the first time we
+					// run the query. as we are about to change the WaitIndex to
+					// make a blocking query, we can consider the watch started.
+					obj.init.Running()
+				}
+				opts.WaitIndex = meta.LastIndex
+				goto POLL
```
purpleidea: no goto's please

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+		opts := &api.QueryOptions{RequireConsistent: true}
+		ctx, cancel := util.ContextWithCloser(context.Background(), obj.init.Done)
+		defer cancel()
+		opts = opts.WithContext(ctx)
+
+	POLL:
+		_, meta, err := kv.Get(obj.Key, opts)
+		select {
+		case c <- err:
+			if err == nil {
```
purpleidea: inline please

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+
+	c := make(chan error)
+	kv := obj.client.KV()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		opts := &api.QueryOptions{RequireConsistent: true}
+		ctx, cancel := util.ContextWithCloser(context.Background(), obj.init.Done)
```
purpleidea: This is a race, because you could call the context close at the same time as the below kv.Get runs.
So you need a separate signal that comes from the below for loop that runs when *it* closes (probably a defer) that then propagates into this goroutine.

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// CheckApply is run to check the state and, if apply is true, to apply the
+// necessary changes to reach the desired state. This is run before Watch and
+// again if Watch finds a change occurring to the state.
+func (obj *ConsulKVRes) CheckApply(apply bool) (bool, error) {
+	kv := obj.client.KV()
+	pair, _, err := kv.Get(obj.Key, nil)
+	if err != nil {
+		return false, err
```
purpleidea: errwrap.Wrapf

engine/resources/consul_kv.go
```diff
@@ -0,0 +1,186 @@
...
+
+// Cmp compares two resources and return if they are equivalent.
+func (obj *ConsulKVRes) Cmp(r engine.Res) error {
+	res, ok := r.(*ConsulKVRes)
+	if !ok {
+		return fmt.Errorf("not a %s", obj.Kind())
+	}
+
+	if obj.Key != res.Key {
+		return errors.New("the key param differs")
```
purpleidea: capitalize the fields `the Key param differs`

examples/lang/consul1.mcl
```diff
@@ -0,0 +1,4 @@
+consul:kv "love" {
+	key   => "mgmt/love",
```
purpleidea: Don't keys need to start with a `/` (slash?) If so shouldn't we validate them in Validate. Maybe we don't want to allow \0 or some other chars too?

roidelapluie: It does not, and https://www.consul.io/docs/agent/kv.html : `Keys, like objects are not restricted by type and can include any character.`

engine/resources/consul_kv.go
roidelapluie: so if Init fails we call close?

engine/resources/consul_kv.go
roidelapluie: yes

engine/resources/consul_kv.go
purpleidea: Probably not, but just to be safe.

engine/resources/consul_kv.go
roidelapluie: I will just take the name as the key by default. The actual path of the value is more complex in consul.

engine/resources/consul_kv.go
roidelapluie: why?

examples/lang/consul1.mcl
roidelapluie: everything is allowed

engine/resources/consul_kv.go
purpleidea: The problem is we want the NAME to be a unique ID of sorts.

So if you had two different consul resources at two different servers, they'd have a conflict if key was the same.

engine/resources/consul_kv.go
roidelapluie: ok, so I understand what you mean.

But, closing the context while kv.Get runs is what we want. kv.Get will only return when a value has changed (this is the actual watch). So when "done" is called, we want to cancel the running Get.

engine/resources/consul_kv.go
roidelapluie: fixed

purpleidea:
After doing some pair programming with @roidelapluie on this, it's now merged in: 3e16d1da4647743337725662bdd632a46501336d

PR #578, opened 2019-12-31
lang: funcs: Add epsilon checks to operator_polyfunc

dbakong:
## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/lang/funcs/operator_polyfunc.go
+++ b/lang/funcs/operator_polyfunc.go
@@ -39,6 +39,9 @@ const (
 )
 
 func init() {
+	// epsilon definition
+	var epsilon = math.Nextafter(1.0, 2.0) - 1
+
 	// concatenation
 	RegisterOperator("+", &types.FuncValue{
 		T: types.NewType("func(a str, b str) str"),
@@ -169,9 +172,15 @@ func init() {
 	RegisterOperator("==", &types.FuncValue{
 		T: types.NewType("func(a float, b float) bool"),
 		V: func(input []types.Value) (types.Value, error) {
-			// TODO: should we do an epsilon check?
+			// epsilon check
+			epsilonBool := math.Abs(input[0].Float()-input[1].Float()) <= epsilon
+			if epsilonBool {
+				return &types.BoolValue{
+					V: epsilonBool,
+				}, nil
+			}
 			return &types.BoolValue{
-				V: input[0].Float() == input[1].Float(),
+				V: false,
 			}, nil
 		},
 	})
@@ -207,9 +216,15 @@ func init() {
 	RegisterOperator("!=", &types.FuncValue{
 		T: types.NewType("func(a float, b float) bool"),
 		V: func(input []types.Value) (types.Value, error) {
-			// TODO: should we do an epsilon check?
+			// epsilon check
+			epsilonBool := math.Abs(input[0].Float()-input[1].Float()) > epsilon
+			if epsilonBool {
+				return &types.BoolValue{
+					V: epsilonBool,
+				}, nil
+			}
 			return &types.BoolValue{
-				V: input[0].Float() != input[1].Float(),
+				V: false,
 			}, nil
 		},
 	})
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by purpleidea** [comment]
@dbakong I think you're on the right track, the epsilon stuff should be done at the top and only once. As either a var or a const or an init() or a combination. See what works =D

**Comment by dbakong** [comment]
@purpleidea thanks, I moved it to the top.

File: lang/funcs/operator_polyfunc.go [hunk 1]
```diff
@@ -39,6 +39,11 @@ const (
 )
 
 func init() {
+	epsilon := func() float64 {
```
purpleidea: I bet you can do even better! this computation still repeats each time.

dbakong: updated!

lang/funcs/operator_polyfunc.go
```diff
@@ -38,6 +38,11 @@ const (
 	operatorArgName = "op" // something short and arbitrary
 )
 
+var epsilon = func() float64 {
```
purpleidea: 1) How can you simplify this?

2) It still runs for each use of the operator.

Can we do better?

lang/funcs/operator_polyfunc.go
```diff
@@ -169,9 +172,14 @@ func init() {
 	RegisterOperator("==", &types.FuncValue{
 		T: types.NewType("func(a float, b float) bool"),
 		V: func(input []types.Value) (types.Value, error) {
-			// TODO: should we do an epsilon check?
+			// epsilon check
+			if math.Abs(input[0].Float()-input[1].Float()) <= epsilon {
```
purpleidea: Should this be `<` instead of `<=` ?

purpleidea: I think the more elegant way to write this is to start off with something like:

`result := false`

And then in you condition you do: `result = true`

And at the end you put in the bool into the &BoolValue as result. (Similarly for other cases.)

lang/funcs/operator_polyfunc.go
```diff
@@ -207,9 +215,14 @@ func init() {
 	RegisterOperator("!=", &types.FuncValue{
 		T: types.NewType("func(a float, b float) bool"),
 		V: func(input []types.Value) (types.Value, error) {
-			// TODO: should we do an epsilon check?
+			// epsilon check
+			if math.Abs(input[0].Float()-input[1].Float()) > epsilon {
```
purpleidea: `>` or `>=` ?

lang/funcs/operator_polyfunc.go
```diff
@@ -39,6 +39,9 @@ const (
 )
 
 func init() {
+	// epsilon definition
```
purpleidea: This is an unhelpful comment. Explain *why* this is the case, since the code is not intuitive.

purpleidea (changes requested):
Since you have this pattern:

`math.Abs(input[0].Float()-input[1].Float()) > epsilon`

Repeated very often, maybe it makes sense to have it be a helper function? Write this as a lambda, right under the epsilon definition. Secondly, check the `>` above. Is it that or do we want `>=` ?

Being precise about `>` vs `>=` is *exactly* the point of this problem. Review *all* the instances in your patch and clean up the other issues mentioned and resubmit!

I hope this is fun!
Thanks!

purpleidea:
@dbakong The epsilon helper doesn't make sense... Where is `input` coming from?

Put the patch back to the way it was before and get all the < > <= >= stuff perfect. LMK here when ready.

dbakong:
@purpleidea I think I got the last missing operator update. Please review.

PR #582, opened 2020-01-07
Feat/pgraph testing

karthiksedoc:
* Fixed formatting issue of pgraph/pgraph_test.go
* Formatted commit message as guideline

### DIFF
--- a/pgraph/pgraph_test.go
+++ b/pgraph/pgraph_test.go
@@ -940,5 +940,5 @@ func TestGetName(t *testing.T) {
 	graph := G.GetName()
 	if graph == nil {
 		t.Errorf("Empty graph returned, expecting graph: %v", graph)
-		}
+	}
 }
--- a/pgraph/pgraph_test.go
+++ b/pgraph/pgraph_test.go
@@ -936,9 +936,10 @@ func TestSetValue(t *testing.T) {
 }
 
 func TestGetName(t *testing.T) {
-	G, _ := NewGraph("g9")
+	name := "g9"
+	G, _ := NewGraph(name)
 	graph := G.GetName()
-	if graph == nil {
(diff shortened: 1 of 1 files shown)


pgraph/pgraph_test.go
```diff
@@ -934,3 +934,11 @@ func TestSetValue(t *testing.T) {
 		t.Errorf("expecting value of %s at %s position, got %v", value, key, v)
 	}
 }
+
+func TestGetName(t *testing.T) {
+	G, _ := NewGraph("g9")
+	graph := G.GetName()
+	if graph == nil {
+		t.Errorf("Empty graph returned, expecting graph: %v", graph)
```
purpleidea: errors start with lower case messages

pgraph/pgraph_test.go
```diff
@@ -934,3 +934,11 @@ func TestSetValue(t *testing.T) {
 		t.Errorf("expecting value of %s at %s position, got %v", value, key, v)
 	}
 }
+
+func TestGetName(t *testing.T) {
+	G, _ := NewGraph("g9")
+	graph := G.GetName()
+	if graph == nil {
```
purpleidea: inline this: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

purpleidea: What type is `graph` here? =D

karthiksedoc: it's a short string, as per variable naming guideline change to s?

pgraph/pgraph_test.go
karthiksedoc: thanks for info :) I'll correct this.

pgraph/pgraph_test.go
karthiksedoc: this is sorted, thanks for pointing :)

PR #586, opened 2020-02-04
engine: resources: Add a test case for resource owner check

cheftony:
This adds FileOwnerExpect as a new Step which allows validating if the owner was set properly on a resource.

### DIFF
--- a/engine/resources/resources_test.go
+++ b/engine/resources/resources_test.go
@@ -174,7 +174,7 @@ func FileExpect(p, s string) Step { // path & string
 	}
 }
 
-// FileExpect takes a path and a string to write to that file, and builds a Step
+// FileWrite takes a path and a string to write to that file, and builds a Step
 // that does that to them.
 func FileWrite(p, s string) Step { // path & string
 	return &manualStep{
--- a/engine/resources/resources_test.go
+++ b/engine/resources/resources_test.go
@@ -23,9 +23,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/user"
 	"path"
+	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


engine/resources/resources_test.go
```diff
@@ -174,7 +177,24 @@ func FileExpect(p, s string) Step { // path & string
...
-// FileExpect takes a path and a string to write to that file, and builds a Step
+// FileOwnerExpect takes a path and a uid to expect from that file, and builds
+// a Step that checks that out of them.
+func FileOwnerExpect(p, o string) Step {
+	return &manualStep{
+		action: func() error { return nil },
+		expect: func() error {
+			fileInfo, _ := os.Stat(p)
+			owner := fileInfo.Sys().(*syscall.Stat_t).Uid
+			i, _ := strconv.ParseUint(o, 10, 32)
```
purpleidea: err?

engine/resources/resources_test.go
```diff
@@ -174,7 +177,24 @@ func FileExpect(p, s string) Step { // path & string
...
 
-// FileExpect takes a path and a string to write to that file, and builds a Step
+// FileOwnerExpect takes a path and a uid to expect from that file, and builds
+// a Step that checks that out of them.
+func FileOwnerExpect(p, o string) Step {
+	return &manualStep{
+		action: func() error { return nil },
+		expect: func() error {
+			fileInfo, _ := os.Stat(p)
+			owner := fileInfo.Sys().(*syscall.Stat_t).Uid
```
purpleidea: Some of this can panic, so it would probably be safer to do it the "right" way where we check as we go...
Probably not urgent in a test, but a good learning experience anyways and gives us a clean message instead of a panic.

engine/resources/resources_test.go
```diff
@@ -195,6 +215,15 @@ func ErrIsNotExistOK(e error) error {
 	return errwrap.Wrapf(e, "unexpected error")
 }
 
+// GetUID returns the UID of the running user.
```
purpleidea: Is this what you want?

https://github.com/purpleidea/mgmt/blob/be39fbeff6812e35454f0381f9ad551ac4f776c2/engine/util/util.go#L332

cheftony: This function gets the uid from the user defined in the `$USER` environment variable. I updated the comment, let me know if it makes more sense !

engine/resources/resources_test.go
cheftony: I changed the way we get the file attributes, should be better!

engine/resources/resources_test.go
cheftony: Fixed

cheftony:
> There's one test failure because of a new wrapping test we added (reflow docstring comments to 80 please) but otherwise, this looks great. A few small fixes, but nothing major.
> 
> The important thing: do you understand the test infra or have any more questions about it? (Both golang's and this specific test harness.)

Fixed!

purpleidea:
Okay, this is merged, thanks!

I still think you could have used the util GetUID, but what you did is fine as well. As long as you see why, then it's okay.

Cheers!

PR #589, opened 2020-02-05
Fix tests

ffrank:
Prometheus tests are busted.

Make their output more readable, and make sure they don't leave lingering mgmt processes upon failing.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
-[submodule "vendor/github.com/coreos/etcd"]
+[submodule "vendor/github.com/etcd-io/etcd"]
 	path = vendor/go.etcd.io/etcd
-	url = https://github.com/coreos/etcd/
+	url = https://github.com/etcd-io/etcd
 [submodule "vendor/google.golang.org/grpc"]
 	path = vendor/google.golang.org/grpc
 	url = https://github.com/grpc/grpc-go
@@ -15,7 +15,7 @@
 	url = https://github.com/purpleidea/go-systemd
 [submodule "vendor/honnef.co/go/augeas"]
 	path = vendor/honnef.co/go/augeas
-	url = https://github.com/dominikh/go-augeas/
+	url = https://github.com/dominikh/go-augeas
 [submodule "vendor/github.com/grpc-ecosystem/go-grpc-prometheus"]
 	path = vendor/github.com/grpc-ecosystem/go-grpc-prometheus
 	url = https://github.com/grpc-ecosystem/go-grpc-prometheus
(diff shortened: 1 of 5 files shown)


ffrank:
This is WiP.

Also please disregard the git thing, it got in here by accident and will be rebased away in time 🙈

ffrank:
Ready for merge.

It fixes the consistent test failures. The etcd race is not yet addressed, but will be easier to debug, and actual regressions will be more visible again.

.gitmodules
```diff
@@ -15,7 +15,7 @@
 	url = https://github.com/purpleidea/go-systemd
 [submodule "vendor/honnef.co/go/augeas"]
 	path = vendor/honnef.co/go/augeas
-	url = https://github.com/dominikh/go-augeas/
```
purpleidea: Why are we removing the trailing slashes?

test/shell/exec-usergroup.sh
```diff
@@ -76,3 +76,6 @@ cleanup
 run_usergroup_test "exec-usergroup-${GROUP}.yaml" "nobody" "${GROUP}"
 run_usergroup_test "exec-usergroup-user.yaml" "nobody" ""
 run_usergroup_test "exec-usergroup-group-${GROUP}.yaml" "" "${GROUP}"
+
+# avoid race against rm command from the shell test wrapper
```
purpleidea: What race is that?

test/shell/load0.sh
```diff
@@ -33,7 +33,7 @@ import "sys"
 
 file "${tmpdir}/loadavg" {
 	content => fmt.printf("load average: %f, %f, %f\n", \$x1, \$x5, \$x15),
-	state => $const.res.file.state.exists,
+	state => \$const.res.file.state.exists,
```
purpleidea: Nice catch, thanks!

test/shell/prometheus-2.sh
```diff
@@ -2,17 +2,39 @@
 
 . "$(dirname "$0")/../util.sh"
 
+TEMPFILE=`mktemp mgmt-tests-XXXXXXXX`
+
+cleanup()
```
purpleidea: I'm fine with having these here, but would putting them in the above listed util.sh help, and/or is there something there that's already useful?

.gitmodules
ffrank: Some git compat issue, but this is no longer part of this patchset anyway.

test/shell/exec-usergroup.sh
ffrank: Not 100% certain it's a race, but I saw this spectacular failure: https://travis-ci.org/purpleidea/mgmt/jobs/647560853?utm_medium=notification&utm_source=github_status

purpleidea:
Okay, thanks! I've merged 2/3 of the commits via cherry-pick to git master. The submodules one should be changed I guess, but not sure why the trailing slashes are being changed.

ffrank:
I have no recollection of what was going on here, and cannot imagine it might be of value today. Closing.

PR #592, opened 2020-02-24
lang: Add fuzz target to lang Makefile

HerrSpace:
As requested in https://github.com/blynn/nex/pull/55#issuecomment-589965901 here is a rough idea how to fuzz the mgmt mcl lexer/parser. I'm not sure where how you'd like to move it so just take it as inspiration :)


lang/Makefile
```diff
@@ -44,3 +44,9 @@ else
 	goyacc parser.y
 endif
 	@ROOT="$$( cd "$$( dirname "$${BASH_SOURCE[0]}" )" && cd .. && pwd )" && $$ROOT/misc/header.sh 'y.go'
+
+fuzz:
```
purpleidea: This would need a .PHONY entry since it's a fake target.

lang/Makefile
```diff
@@ -44,3 +44,9 @@ else
...
+	mkdir -p corpus
+	find .. -name \*.mcl -exec cp {} corpus \;
+	go-fuzz-build
```
purpleidea: Should we add this dependency to the misc/make-deps.sh script? If anything gated by an --optional flag?

lang/fuzz.go
```diff
@@ -0,0 +1,11 @@
+package lang
```
purpleidea: You'll need the standard headers or this will fail the test.

.gitignore
```diff
@@ -19,3 +19,7 @@ releases/
 .*.sw[op]
 # prevent `echo foo 2>1` typo errors by making this file read-only
 1
+lang/corpus/
```
purpleidea: Can we put everything inside of a fuzz/ sub directory (inside `lang/`) to keep it all self-contained? It could have its own Makefile in that dir, and .gitignore

Then we can chain the fuzz stuff in from the lang Makefile with `@$(MAKE) --quiet -C fuzz`

lang/fuzz.go
HerrSpace: like this?

purpleidea:
This is merged, thank you!

PR #593, opened 2020-02-29
lang: Fix wrong go-fuzz bin name

HerrSpace:
I missed renaming this after moving the fuzz.go from the lang package
to its own fuzz package.


purpleidea:
LGTM.

Issue #591, opened 2020-02-24
Parser panic on []struct types with duplicate member names


HerrSpace:
## Versions:

* mgmt version (eg: `mgmt --version`):
git commit be39fbeff6812e35454f0381f9ad551ac4f776c2

* operating system/distribution (eg: `uname -a`):
mac os 10.15.2

* golang version (eg: `go version`):
go version go1.13.8 darwin/amd64

## Description:
Parsing `$d []struct{x int;x int}` via the MCL parser causes the following panic:

```
panic: malformed struct length

goroutine 1 [running]:
github.com/purpleidea/mgmt/lang/types.(*Type).String(0xc0001f7450, 0x14, 0xc0001f7450)
	/Users/space/repos/go/src/github.com/purpleidea/mgmt/lang/types/type.go:520 +0xa1b
github.com/purpleidea/mgmt/lang.(*yyParserImpl).Parse(0xc000107000, 0x4c802c0, 0xc0001f72c0, 0x0)
	parser.y:1168 +0xdd3d
github.com/purpleidea/mgmt/lang.yyParse(...)
	yaccpar:153
github.com/purpleidea/mgmt/lang.LexParse(0x4c74ea0, 0xc0001ece10, 0xc000010018, 0x402eefa, 0xc0001bdef8, 0xc000010018)
	/Users/space/repos/go/src/github.com/purpleidea/mgmt/lang/lexparse.go:95 +0xbe
github.com/purpleidea/mgmt/fuzz.Fuzz(0xc000114000, 0x19, 0x219, 0x19)
	/Users/space/repos/go/src/github.com/purpleidea/mgmt/fuzz/fuzz.go:18 +0x7e
main.main()
	/Users/space/repos/go/src/github.com/purpleidea/mgmt/main.go:21 +0x164
exit status 2
```

Looking at `parser.y` it seems we crash in the string conversion here:
```
|	OPEN_BRACK CLOSE_BRACK type
	// list: []int or [][]str (with recursion)
	{
		posLast(yylex, yyDollar) // our pos
		$$.typ = types.NewType("[]" + $3.typ.String())
	}
```

Looking at `types.String` we see we crash in the second check here:
```go
	case KindStruct: // {a bool; b int}
		if obj.Map == nil {
			panic("malformed struct type")
		}
		if len(obj.Map) != len(obj.Ord) {
			panic("malformed struct length")
		}
```

And indeed, when trying to parse the type manually via `types.NewType` we get some object like:
```
[...]

PR #594, opened 2020-02-29
lang: Prevent struct types with duplicate field names

HerrSpace:
The previous fix for #591 in 70eecd5 didn't address all issues
concerning duplicate struct field names. It still crashed for inputs
like `$d []struct{x int; x float}`. Note the different types but
duplicate names.

### DIFF
--- a/lang/interpret_test/TestAstFunc2/struct-duplicate4/main.mcl
+++ b/lang/interpret_test/TestAstFunc2/struct-duplicate4/main.mcl
@@ -1,0 +1,1 @@
+$d []struct{x int;x float}
--- a/lang/parser.y
+++ b/lang/parser.y
@@ -1180,14 +1180,14 @@ type:
 		strs := []string{}
 		for _, arg := range $3.args {
 			s := fmt.Sprintf("%s %s", arg.Name, arg.Type.String())
-			if _, exists := names[s]; exists {
+			if _, exists := names[arg.Name]; exists {
 				// duplicate field name used
 				err := fmt.Errorf("duplicate struct field of `%s`", s)
 				// this will ultimately cause a parser error to occur...
 				yylex.Error(fmt.Sprintf("%s: %+v", ErrParseSetType, err))
 				break // we must skip, because code continues!
 			}
-			names[s] = struct{}{}
+			names[arg.Name] = struct{}{}
 			strs = append(strs, s)
 		}
 
--- a/lang/interpret_test/TestAstFunc2/struct-duplicate4.output
+++ b/lang/interpret_test/TestAstFunc2/struct-duplicate4.output
@@ -1,0 +1,1 @@
+# err: err1: parser: `syntax error: unexpected $end, expecting EQUALS` @1:6


lang/parser.y
```diff
@@ -1179,16 +1179,18 @@ type:
 		names := make(map[string]struct{})
 		strs := []string{}
 		for _, arg := range $3.args {
-			s := fmt.Sprintf("%s %s", arg.Name, arg.Type.String())
-			if _, exists := names[s]; exists {
+			field := fmt.Sprintf("%s %s", arg.Name, arg.Type.String())
+			if _, exists := names[arg.Name]; exists {
```
purpleidea: Woops, yeah this was a bug on my part, and this is the correct fix, thanks!

lang/parser.y
```diff
@@ -1179,16 +1179,18 @@ type:
 		names := make(map[string]struct{})
 		strs := []string{}
 		for _, arg := range $3.args {
-			s := fmt.Sprintf("%s %s", arg.Name, arg.Type.String())
```
purpleidea: Not sure this rename is necessary.

lang/parser.y
```diff
@@ -1179,16 +1179,18 @@ type:
...
+			if _, exists := names[arg.Name]; exists {
 				// duplicate field name used
-				err := fmt.Errorf("duplicate struct field of `%s`", s)
+				err := fmt.Errorf("duplicate struct field of `%s`", field)
 				// this will ultimately cause a parser error to occur...
 				yylex.Error(fmt.Sprintf("%s: %+v", ErrParseSetType, err))
-				break // we must skip, because code continues!
+				// we must immediately get out of here to avoid other parts
+				// of the parser from using this type here.
+				goto ret1
```
purpleidea: I'm not convinced this is appropriate-- it means we're depending on something in goyacc that could be unstable. Even without this, it should work fine.

lang/interpret_test/TestAstFunc2/struct-duplicate4/main.mcl
```diff
@@ -0,0 +1 @@
+$d []struct{x int;x float}
```
purpleidea: Awesome, thanks for adding a test!

lang/parser.y
HerrSpace: Hu, seems like I just forgot to `make lang` before retesting this, it actually _does_ work with the break in place, so I have removed my change for now.

That said I do get a few hundred execs per second less in go-fuzz using break vs return 1.  I find both goyacc as well as the plan9 yacc man page lacking, but I do think returning non zero ints anywhere in the parser or lexer should be fine to signal failed parsing/lexing. Is that correct? Why not just do that?

purpleidea:
LGTM, but your test doesn't work. I know why, but you can figure it out =D

purpleidea:
Great, thank you!

purpleidea:
(merged)

lang/parser.y
purpleidea: > That said, why not just return 1 after any call to yylex.Error? I find both goyacc as well as the plan9 yacc man page lacking, but I do think returning non zero ints anywhere in the parser or lexer should be fine to signal failed parsing/lexing. Is that correct?

It's a good question. I don't know the answer, but I found the yylex.Error somewhere I believe in some sample code somewhere, so I got the impression it was correct like that. If I would *guess*, here's what I would say:

Letting the parser "finish" gives it a chance to potentially find a more serious problem, and/or a second problem, and in theory, display multiple useful errors. I'm not sure it currently can do this, but it might be somewhere we could get to.

PR #597, opened 2020-04-14
Update minimum go version to 1.13

ahmedalhulaibi:
I was running through the Quick Start and ran into a problem when trying to build from source using `go1.12`. One of the dependencies `github.com/hashicorp/go-multierror` is using `go1.13` new `errors` package methods `errors.Is` and `errors.As`

This PR addresses this by updating minimum required go version in the Quick Start Guide to `go1.13`. 

I have also upated the `make-deps.sh` script to check if the installed version of go is below the minimum required version. This check only applies to the minor version and ignores the major and patch versions.

The regex associated with capturing the installed version of go has also been updated to capture only the minor version `go1.14.2 -> 14` and will also capture the minor version even when the patch version is not output `go1.8 -> 8`

### DIFF
--- a/docs/quick-start-guide.md
+++ b/docs/quick-start-guide.md
@@ -37,7 +37,7 @@ You'll need some dependencies, including `golang`, and some associated tools.
 
 #### Installing golang
 
-* You need golang version 1.11 or greater installed.
+* You need golang version 1.13 or greater installed.
 	* To install on rpm style systems: `sudo dnf install golang`
 	* To install on apt style systems: `sudo apt install golang`
 	* To install on macOS systems install [Homebrew](https://brew.sh)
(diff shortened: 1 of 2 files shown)


purpleidea:
Hi, thanks for the patch!

We're actually staying on 1.12 for a little bit longer. Basically I run the Fedora N-1 stock golang version. I should actually bump it up one bit, but I'm just waiting till I finish the branch I'm working on.

In the meantime, there is already a patch to fix the issue for older versions: https://github.com/purpleidea/mgmt/commit/fc839d29830829b760729f08f170b0772df41f35

If you have any other mgmt issues, please let me know!

purpleidea:
Merged in git master now. Thanks!!

purpleidea:
(Sorry for the delays!)

ahmedalhulaibi:
No problem! Thanks for following up on this

PR #599, opened 2020-04-28
Update pgraph_test.go

Pendragon10:
I'm still not hundred percent on this stuff so please go easy on me lol. Let me know what mistakes I made or how I can make it better.

## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/pgraph/pgraph_test.go
+++ b/pgraph/pgraph_test.go
@@ -934,3 +934,21 @@ func TestSetValue(t *testing.T) {
 		t.Errorf("expecting value of %s at %s position, got %v", value, key, v)
 	}
 }
+
+
+//Dov's new test
+
+//Testing the naming of graphs
+func TestSetName(t *testing.T) {
+	name := "Dov's Graph"
+	
+	var g string
+	g == name
+	
+	if !g == name {
+		t.Errorf("The graph's name is wrong because it should be %s and instead it is %s.", name, g)
+	}
+	
+	return
+}
+

### CODE REVIEW

File: pgraph/pgraph_test.go [hunk 1]
```diff
@@ -934,3 +934,21 @@ func TestSetValue(t *testing.T) {
...
 }
+
+
+//Dov's new test
+
+//Testing the naming of graphs
+func TestSetName(t *testing.T) {
+	name := "Dov's Graph"
+	
+	var g string
```
purpleidea, line 945: Should this test pass or fail, and what happens and why?

pgraph/pgraph_test.go
```diff
@@ -934,3 +934,21 @@ func TestSetValue(t *testing.T) {
 		t.Errorf("expecting value of %s at %s position, got %v", value, key, v)
 	}
 }
+
+
+//Dov's new test
+
+//Testing the naming of graphs
+func TestSetName(t *testing.T) {
+	name := "Dov's Graph"
```
purpleidea, line 943: Since we want to ultimately merge this patch, referring to "Dov" isn't helpful. Usually we only refer to well-known names like "einstein" or "monty python" or leave it generic "graph1" or "foo".

Pendragon10:
So, to go one by one. I will remove my name from the test. This test will
always pass which is why I don't think I am understanding how to test this
function. It will always pass because I created a variable and simply made
another variable equal to it. Then I checked to see if they are equal so of
course, it will always pass. Finally, the go "zero" values I know what they
are. It's the default value given to the different variable types when they
aren't given a value. So for a string it would be empty quotation marks.

Pendragon10:
Should I attempt a new test on a different function and forget this one, or continue on this one?

purpleidea:
@Pendragon10 You can reply directly to the individual comments... Try it.

Read: https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md to see if you should continue this PR or start a new one =D

PR #600, opened 2020-05-12
lang: funcs: Add macfmt function

cheftony:
* Add new `net` package that implements `mcl` function related to networking.
* Add `mcl` function that takes a mac address with hyphens as an argument and converts it to a format with colons.

### DIFF
--- a/lang/funcs/core/net/macfmt_func.go
+++ b/lang/funcs/core/net/macfmt_func.go
@@ -15,21 +15,39 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-package core
+package corenet
 
 import (
-	// import so the funcs register
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/convert"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/datetime"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/deploy"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/example"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/example/nested"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/fmt"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/math"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/net"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/os"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/regexp"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/strings"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/sys"
-	_ "github.com/purpleidea/mgmt/lang/funcs/core/world"
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/funcs/simple"
+	"github.com/purpleidea/mgmt/lang/types"
 )
+
+func init() {
+	simple.ModuleRegister(ModuleName, "macfmt", &types.FuncValue{
+		T: types.NewType("func(a str) str"),
+		V: MacFmt,
+	})
+}
+
+// MacFmt takes a MAC address with hyphens and converts it to a format with
+// colons.
+func MacFmt(input []types.Value) (types.Value, error) {
+	mac := input[0].Str()
+
+	// Check if the MAC address is valid.
+	if len(mac) != len("00:00:00:00:00:00") {
+		return nil, fmt.Errorf("invalid MAC address length: %s", mac)
+	}
+	_, err := net.ParseMAC(mac)
+	if err != nil {
+		return nil, err
+	}
+
+	return &types.StrValue{
+		V: strings.Replace(mac, "-", ":", -1),
+	}, nil
+}
--- a/lang/funcs/core/net/macfmt_func_test.go
+++ b/lang/funcs/core/net/macfmt_func_test.go
@@ -1,0 +1,54 @@
+// Mgmt
+// Copyright (C) 2013-2020+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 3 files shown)


lang/funcs/core/net/mac_fmt_func.go
```diff
@@ -0,0 +1,50 @@
...
+		V: MacFmt,
+	})
+}
+
+// MacFmt takes a mac address with hyphens and converts it to a format with
+// colons.
+func MacFmt(input []types.Value) (types.Value, error) {
+	mac := input[0].Str()
+
+	// Check if the MAC is valid
```
purpleidea, line 40: comments should be either full sentences (capitalized, with period at the end) or neither.

lang/funcs/core/net/mac_fmt_func.go
```diff
@@ -0,0 +1,50 @@
...
+	_, err := net.ParseMAC(mac)
+	if err != nil {
+		return nil, err
+	}
+
```
purpleidea, line 45: Should we check anything else for sanity? Ensure the length is correct for example? Specify why or why not, there is a specific reason, and add it to the code comments.

lang/funcs/core/net/mac_fmt_func.go
```diff
@@ -0,0 +1,50 @@
...
+	m := strings.Replace(mac, "-", ":", -1)
```
purpleidea, line 46: unnecessary var

lang/funcs/core/net/mac_fmt_func_test.go
```diff
@@ -0,0 +1,52 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package corenet
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
```
purpleidea, line 23: cute idea, but I'm not sure we need the extra dependency for a string compare.

lang/funcs/core/net/mac_fmt_func_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
```
purpleidea, line 24: Note for future: the import statements are grouped into three blocks. built-in, local, and other.

lang/funcs/core/net/mac_fmt_func_test.go
```diff
@@ -0,0 +1,52 @@
...
+		out     string
+		wantErr bool
+	}{
+		{"Valid mac with hyphens", "01-23-45-67-89-AB", "01:23:45:67:89:AB", false},
+		{"Valid mac with colons", "01:23:45:67:89:AB", "01:23:45:67:89:AB", false},
+		{"Invalid mac", "", "", true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
purpleidea, line 41: t.Run can contain a goroutine, so beware the bug you've introduced:

https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

lang/funcs/core/net/mac_fmt_func_test.go
```diff
@@ -0,0 +1,52 @@
...
+			m, err := MacFmt([]types.Value{&types.StrValue{V: tt.in}})
+			if (err != nil) != tt.wantErr {
+				t.Errorf("corenet.MacFmt() error = %v, wantErr %v", err, tt.wantErr)
```
purpleidea, line 44: you don't need to specify `corenet`, we know where the code is, and this makes the package name more difficult to rename because there's extra cruft we'd have to remember existed.

lang/funcs/core/net/mac_fmt_func.go
cheftony: There is `net.ParseMac()` that check if we have a valid MAC. If not, it returns an error. However I am not sure if it's relevant to add another validation on top of that unless we don't want anything longer than a EUI-48 MAC ?

lang/funcs/core/net/mac_fmt_func_test.go
cheftony: Added `tt := tt` to mitigate the bug as recommended in the link.

lang/funcs/core/net/mac_fmt_func.go
purpleidea: Bingo-- the broad definition of a mac address that the upstream golang lib uses is probably not what we want or need in our version.

lang/funcs/core/net/mac_fmt_func_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			m, err := MacFmt([]types.Value{&types.StrValue{V: tt.in}})
+			if (err != nil) != tt.wantErr {
+				t.Errorf("func MacFmt() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if m == nil && tt.wantErr {
```
purpleidea, line 47: Isn't this a bug?

lang/funcs/core/net/mac_fmt_func_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			m, err := MacFmt([]types.Value{&types.StrValue{V: tt.in}})
+			if (err != nil) != tt.wantErr {
+				t.Errorf("func MacFmt() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if m != nil {
```
purpleidea, line 47: Technically the function could return nil, and would escape our tests. You don't need to fix this, but it's technically a mistake I think.

lang/funcs/core/net/mac_fmt_func.go
```diff
@@ -0,0 +1,53 @@
...
+	})
+}
+
+// MacFmt takes a MAC address with hyphens and converts it to a format with
+// colons.
+func MacFmt(input []types.Value) (types.Value, error) {
+	mac := input[0].Str()
+
+	// Check if the MAC address is valid.
+	if len(mac) != 17 {
```
purpleidea, line 42: random constants aren't fun to understand, so I'd recommend instead doing `!= len("00:00:00:00:00:00")`

purpleidea:
Merged, thank you! \o/

PR #605, opened 2020-06-15
lang: structs: Fix duplicate resource field bug

adamsigal:
Add extra verification in Init() to prevent a resource with
multiple instances of the same field from being validated.

## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### CODE REVIEW

**Comment by purpleidea** [comment]
@adamsigal FYI, this was never reviewed because the tests failed. The convention is that a PR will be ignored if the tests failed *unless* you leave a comment explaining why the failure is not your fault or you need help understanding it.

**Comment by adamsigal** [comment]
> @adamsigal FYI, this was never reviewed because the tests failed. The convention is that a PR will be ignored if the tests failed _unless_ you leave a comment explaining why the failure is not your fault or you need help understanding it.

I actually realised that the tests didn't pass, that's why I didn't mention it on IRC ¯\_(ツ)_/¯

File: examples/lang/exec_duplicate.mcl [hunk 1]
```diff
@@ -0,0 +1,7 @@
+exec "exec_duplicate" {
+	cmd => "/usr/bin/env",
```
purpleidea, line 2: Why do we want to add an example of what's not allowed?

lang/interpret_test/TestAstFunc1/metaparams0.graph
```diff
@@ -25,7 +25,6 @@ Vertex: bool(true)
 Vertex: bool(true)
 Vertex: bool(true)
 Vertex: bool(true)
-Vertex: bool(true)
```
purpleidea, line 28: Looks right here! I trust you figured out how the test stuff works? FWIW I don't write these .graph files manually, I look at the output from running the test with -v and if it seems plausible I use that.

lang/structs.go
```diff
@@ -311,10 +311,19 @@ func (obj *StmtRes) Init(data *interfaces.Data) error {
...
 		return err
 	}
+	typeMap := make(map[string]struct{})
 	for _, x := range obj.Contents {
 		if err := x.Init(data); err != nil {
 			return err
 		}
+
+		// checks if the current ResField type has already been added to typeMap
+		_, isPresent := typeMap[x.String()]
```
purpleidea, line 321: I think you have the right idea... Except x.String() is probably not right. You want the field name, not the representation of that whole thing, right? This is trickier because obj.Contents is an interface IIRC... So explain what's going on please =D

lang/structs.go
```diff
@@ -311,10 +311,19 @@ func (obj *StmtRes) Init(data *interfaces.Data) error {
 	if err := obj.Name.Init(data); err != nil {
 		return err
 	}
+	typeMap := make(map[string]struct{})
 	for _, x := range obj.Contents {
 		if err := x.Init(data); err != nil {
 			return err
 		}
+
+		// checks if the current ResField type has already been added to typeMap
```
purpleidea, line 320: Do you want this before or after the x.Init and why?

examples/lang/exec_duplicate.mcl
adamsigal: i could have sworn you told me to, but if you don't want it I'll just remove it.

lang/interpret_test/TestAstFunc1/metaparams0.graph
adamsigal: Yeah exactly I took your advice and did just that.

lang/structs.go
adamsigal: StmtRes.string() returns the kind of resource as a string (structs.go line 275: `// kind of resource, eg: pkg, file, svc, etc...`), so it seems to me like a logical thing to compare. 

Essentially here, I use a map called `typeMap`. I use a map because of its simple method for checking if an object already exists within it. Here (line 320), putting the current string in typeMap returns a boolean, which I save in `isPresent`, that indicates if we have already seen this type of field. If we have seen it, I return an error, if not, the type is saved into the map to be compared against the next fields to be processed.

I see what you mean that this is kind of an indirect way of testing it, but I can't easily think of a more "correct" way to do it.

lang/structs.go
adamsigal: Since you're asking, I'm guessing that you think it should be put before the x.Init(). I imagine we would do that to not waste computation on an object that shouldn't be initialised anyway due to the disallowance of duplicate fields in mcl.

lang/structs.go
adamsigal: After looking at this, I still feel that this is valid since StmtRes objects are identified by their `kind` with a string, and that's what I test against to find duplicates.

purpleidea:
Heya, I've had to close this because I replaced it with a finished version.

I also added the `meta` dupes checking, so my version is a bit more complicated, but in case you're curious in the learning experience...

Here:

https://github.com/purpleidea/mgmt/commit/a45b7a1d8872d9fc3abbe8bccbb99df3aeea248f#diff-475d0aedee767145087b385c3ea097461792ffe83edac562c8263bb0e51e69b0R321

You can see how we look at the type of the thing, and switch into it before looking at it. We can't use `foo.String()` because:

1) That's a bigger representation that wouldn't guarantee us unique fields, and 2) We don't want to examine Edges for example.

If you have questions or want to jump back in, please LMK =D

PR #612, opened 2020-10-01
Create codeql-analysis.yml

purpleidea:
Testing https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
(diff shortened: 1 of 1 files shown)


purpleidea:
It seems to pass, but if you look at the logs, it doesn't do anything to build the project properly, because important dependencies like `go-bindata` for example are not installed. I think if we added a `make deps` step, then this might work better, but it's not clear if those failures hurt the code analysis much.

So I'm going to say this is low-priority, but if someone wants to pick this up and play with it, you're welcome to it.

roidelapluie:
You need to get it merged to have the actual results to compare to.

PR #615, opened 2020-11-24
Assert that 'metadata.yaml' is not parsed as raw mcl

frebib:
Contradictory to expectations, `mgmt run lang metadata.yaml` would
produce an error similar to the following, which is likely unexpected
from the user perspective:

    2020-11-24 12:24:08.330968 I | cli: lang: lexing/parsing...
    2020-11-24 12:24:08.331153 I | run: error: cli parse error: could not generate AST: parser: `syntax error: unexpected DOT` @1:1

Produce a user-friendly warning instead, hinting with the supported
usage:

    2020-11-24 13:15:01.686659 I | run: error: cli parse error: could not activate an input parser: unexpected raw code 'metadata.yaml'. Did you mean './metadata.yaml'?

### DIFF
--- a/lang/inputs.go
+++ b/lang/inputs.go
@@ -318,6 +318,12 @@ func inputCode(s string, fs engine.Fs) (*ParsedInput, error) {
 		}
 	}
 
+	// check if code is `metadata.yaml`, which is obviously not correct here
+	if s == interfaces.MetadataFilename {
+		return nil, fmt.Errorf("unexpected raw code '%s'. Did you mean './%s'?",
+			interfaces.MetadataFilename, interfaces.MetadataFilename)
+	}
+
 	wd, err := os.Getwd() // NOTE: not meaningful for stdin unless fs is an OsFs
 	if err != nil {
 		return nil, errwrap.Wrapf(err, "can't get working dir")


lang/inputs.go
```diff
@@ -318,6 +318,12 @@ func inputCode(s string, fs engine.Fs) (*ParsedInput, error) {
 		}
 	}
 
+	// Check if code is `metadata.yaml`, which is obviously not correct here
```
purpleidea, line 321: nit: Please either add a period at the end of the comment so it's a regular sentence, or uncapitalize the first letter.

Full explanation:
https://github.com/purpleidea/mgmt/blob/master/docs/style-guide.md#comments

lang/inputs.go
```diff
@@ -318,6 +318,12 @@ func inputCode(s string, fs engine.Fs) (*ParsedInput, error) {
...
+	if s == interfaces.MetadataFilename {
+		return nil, fmt.Errorf("unexpected raw code '%s'. Did you mean "+
```
purpleidea, line 323: why did you add the `+` instead of just having a single string?
(you can keep this all on oneline, no need to break at 80 for such a thing)

purpleidea (commented):
Patch looks great, and thank you! One small fix please, and so I piggybacked on a nit alongside it.

Cheers =D

lang/inputs.go
frebib: I wasn't sure. gofmt strangely doesn't break this line even though it's a lot longer than 80 chars. I've added what I think is a reasonable break. lmk what you think

purpleidea:
LGTM =D
I have no idea what's up with travis at the moment, so I'll wait to see if it does its thing, and then I'll merge or run the tests locally!

Thanks again!

purpleidea:
Travis passed!

https://travis-ci.org/github/purpleidea/mgmt/builds/745641420

All merged =D
Thank you and welcome to the project! \o/

PR #618, opened 2020-12-08
Drop support for Go 1.9 in `make build`

frebib:
#### Drop support for Go 1.9 in `make build`

docs/development.md says the minimum required Golang version is 1.13 at
the time of writing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,13 +170,7 @@ GOOS=$(firstword $(subst -, ,$*))
 GOARCH=$(lastword $(subst -, ,$*))
 build/mgmt-%: $(GO_FILES) $(MCL_FILES) | bindata lang funcgen
 	@echo "Building: $(PROGRAM), os/arch: $*, version: $(SVERSION)..."
-	@# reassigning GOOS and GOARCH to make build command copy/pastable
-	@# go 1.10+ requires specifying the package for ldflags
-	@if go version | grep -qE 'go1.9'; then \
-		time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags "-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS); \
-	else \
-		time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags=$(PKGNAME)="-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS); \
-	fi
+	@time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags=$(PKGNAME)="-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS)
 
 # create a list of binary file names to use as make targets
 crossbuild_targets = $(addprefix build/mgmt-,$(subst /,-,${GOOSARCHES}))


frebib:
At this point should 1.14 be the "supported" version? I'm not sure what kind of compatibility you want to support. Upstream supports 3 minor releases, which at present is 1.13-1.15

Makefile
```diff
@@ -172,6 +172,10 @@ build/mgmt-%: $(GO_FILES) $(MCL_FILES) | bindata lang funcgen
 	@echo "Building: $(PROGRAM), os/arch: $*, version: $(SVERSION)..."
 	@time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags=$(PKGNAME)="-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS)
 
+.PHONY: install
```
purpleidea, line 175: Is it common to add a new .PHONY here instead of the main one at the top? IOW does this work as expected?

purpleidea:
One question, otherwise LGTM. (Not entirely sure why you want `go install` but I don't mind having it there if it's useful for you. (Note we use go build -i which is similar.)

purpleidea:
PS: We'll move to 1.14 shortly.

frebib:
Yeah I can rename the target. The main justification for this is because there are several build steps outside of the go toolchain that are required before `go build`. It would be much easier if the Go build used `go generate` for all external build commands, but this works well enough as an alternative.

purpleidea:
np, just ping when patch is ready

Makefile
frebib: Oh I forgot there was a huge PHONY block at the top. I can move this up there.
Afaik you can have as many or few PHONY targets as you like. I've always tended to group the PHONY just above each respective rule but I will update to follow the existing style here

frebib:
ping @purpleidea
Addressed the comments

Makefile
```diff
@@ -170,13 +170,7 @@ GOOS=$(firstword $(subst -, ,$*))
...
 build/mgmt-%: $(GO_FILES) $(MCL_FILES) | bindata lang funcgen
 	@echo "Building: $(PROGRAM), os/arch: $*, version: $(SVERSION)..."
-	@# reassigning GOOS and GOARCH to make build command copy/pastable
-	@# go 1.10+ requires specifying the package for ldflags
-	@if go version | grep -qE 'go1.9'; then \
-		time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags "-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS); \
-	else \
-		time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags=$(PKGNAME)="-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS); \
-	fi
+	@time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags=$(PKGNAME)="-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS)
```
purpleidea, line 173: Why did you add the `@` ... Why would we want to silence this important command?

Makefile
```diff
@@ -172,6 +172,9 @@ build/mgmt-%: $(GO_FILES) $(MCL_FILES) | bindata lang funcgen
 	@echo "Building: $(PROGRAM), os/arch: $*, version: $(SVERSION)..."
 	@time env GOOS=${GOOS} GOARCH=${GOARCH} go build -i -ldflags=$(PKGNAME)="-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" -o $@ $(BUILD_FLAGS)
 
+goinstall: $(GO_FILES) $(MCL_FILES) | bindata lang funcgen
+	time go install -i -ldflags=$(PKGNAME)="-X main.program=$(PROGRAM) -X main.version=$(SVERSION) ${LDFLAGS}" $(BUILD_FLAGS)
```
purpleidea, line 176: Don't we want the `env GOOS=${GOOS} GOARCH=${GOARCH}` as with the `go build` ?

purpleidea:
Sorry, I just left two review comments. Not sure if I didn't see those points earlier or if they're new.
As for the .PHONY -- if it still works as a new header, I'm fine with that approach! Anything you can do better than I did previously, is of course reasonable to propose! Don't necessarily follow my bad habits when it makes sense to branch out!

Makefile
frebib: The `@` was already present before the `if` which is a single statement due to the continuation backslashes. I can drop it if you'd prefer

Makefile
frebib: Hm, I'll review where those come from. To the user it would make most sense for the `go install` command to build with the native host GOOS and GOARCH which are implied if you don't pass the variables. I can't really see any reason why one would want to `go install` a binary that they can't run on the local machine. wdyt?

Makefile
purpleidea: Quite right, thanks, you can leave as is =D

Makefile
purpleidea: Yeah, that's a fair point. LGTM, I'll merge shortly!

purpleidea:
LGTM

frebib:
Dropped the `go install` commit, I don't think it's necessary.

purpleidea:
Rebased to master and merged, thanks!

PR #620, opened 2020-12-09
misc: Add github CONTRIBUTING support

evrardjp:
Without this patch, there is no CONTRIBUTING.{md,txt,rst}
file.

This is a problem as it might not be obvious, for a first time
contributor, to check in .github/ for the PR template to know
how to contribute.

This fixes it by creating a CONTRIBUTING.txt file, which simply
points to the necessary existing documentation.

### DIFF
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -1,0 +1,1 @@
+Please see .github/PULL_REQUEST_TEMPLATE.md


purpleidea:
I'm not convinced about this patch, sorry. The first entry in the README under docs is our quick start guide. If that should be improved upon, please let me know. But we don't want to reference github-specific things where possible. IOW, imagine if we moved to gitlab overnight...

So if we wanted to add a `CONTRIBUTING` file, we'd want to tell it to point to our quick start guide for example. But I'm not sure we need that if our README is good enough. I'm going to close this PR for now, but happy to discuss more and reconsider.

evrardjp:
I think the quick start doesn't mention commit title expectations, which is only present in the github issue template. For me, that's even less browsable than a CONTRIBUTING file, which is what I consider the most old school. I don't have a strong take on this, because I now know the expectations.

purpleidea:
> CONTRIBUTING file, which is what I consider the most old school.

This is the way to convince me ;)

... Happy to accept a patch, although it should meet the above requirements, to avoid duplication or referencing github specific things.

This patch is not my priority at the moment, but I welcome everyone's opinions on what they think is best for the community.

PR #622, opened 2020-12-09
Ensure make test shell block can pass

evrardjp:
Without this patch, TEST_BLOCK="shell" make test will fail in
CI, with  "no Go files in
go/src/github.com/purpleidea/mgmt/bindata".

This is a problem, as it will prevent us to move to github actions.

This should fix it, by creating an empty golang file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ tmp/
 *WIP
 *_stringer.go
 bindata/*.go
+!bindata/doc.go
 mgmt
 mgmt.static
 # crossbuild artifacts
--- a/bindata/doc.go
+++ b/bindata/doc.go
@@ -1,0 +1,19 @@
+// Mgmt
+// Copyright (C) 2013-2019+ James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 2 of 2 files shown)


purpleidea:
This is _not_ a bug-- the issue is that you didn't build the generated files before you ran the tests. To do this run `make`. This is done automatically in our CI.

evrardjp:
I ran `make test`, like the documentation suggests.
if I check the Makefile, `make test` has a dep on build, which should be good enough. Because I ran this in CI, I expected that the `make build` ran and generated the proper content. Apparently I was wrong, and there is something no streamlined in the way the make runs in CI. Why the different code paths?

purpleidea:
> why the different code paths?

I don't know exactly what you're referring to, but if I was to guess, then the point is that you can run either:

1) `make test` (from CI)
2) `make` (build)
3) ./test/some_test_name.sh (run one individual test as user)

And any of these will ensure that the necessary deps are built.

If I didn't answer all your questions, please re-ask. I'm heading to bed shortly.

purpleidea:
> ./test/some_test_name.sh (run one individual test as user)

Note, you can also auto-complete these with make test-<tab>

evrardjp:
I will dig into why `make test`, in CI, didn't generate the proper artifacts.

evrardjp:
> If I didn't answer all your questions, please re-ask. I'm heading to bed shortly.

No software is worth cutting your bed time! Talk to you another day :)

frebib:
fwiw, this could be made a bit clearer using `go generate` to invoke the generator, or even `make` so `go build` can perform all of the compilation in one step

Issue #572, opened 2019-11-14
Move to golang modules instead of git submodules


roidelapluie:
I would really love to see git modules go away in favour of golang modules.

I am trying to build mgmt inside $customer infra and will have to maintain an internal fork of mgmt with modules to be able to build it.

Go modules enable more reproducible builds, and they are pushed by golang upstream, we should use them.

purpleidea:
@roidelapluie Is the only problem you're having the lack of reproducibility or is there anything else?

I was planning to solve the reproducibility issue, and I was likely going to use git modules for it, however the plan was to add those mod files into a subdir such as go.mod.<tag version> and a `make restore` and `make addtag` script. Because I wanted git master to stay free of those, but still make it possible to reproducibly build tagged releases.

If you'd like to write such a script, that's welcome.

roidelapluie:
Our pipelines internally use go mod, so that would be the easiest for us. Also normally we do go mod vendor and commit vendor in our internal repo.

jlucktay:
Hi there, speaking as someone that only just today cloned `mgmt` and started hacking through the quick start guide, FWIW my 2c is that going through that process without a `go.mod` file in the project root was my single biggest source of frustration in getting a binary built.

The maturity of Go modules has improved since this issue was raised; at the time Go 1.13 had only just come out, and then it wasn't until [1.14](https://golang.org/doc/go1.14) dropped in Feb 2020 that they actively encouraged users to migrate.

Keeping `vendor/` sub-directories as submodules is a pattern I hadn't come across before.
Switching to Go modules would definitely enable reproducible builds of git tags, if that is your main concern @purpleidea. 🙂

purpleidea:
@jlucktay Can you describe specifically, in technical terms "my single biggest source of frustration in getting a binary built." ?
We also have binary releases which you can download and use directly.
Thanks!

jlucktay:
It centred around the `make-deps.sh` script.
When it calls `go get -v -t -d ./...` it would use either 1) a version of an external package already downloaded in the local `$GOPATH`, or 2) pull whatever the latest version is from online, both of which are potentially incorrect/inconsistent, rather than their versions/commit hashes being explicitly defined/compatible.

I couldn't find any git submodule maintenance commands defined in the Makefile or the helper scripts either, so beyond the one initial `git clone --recursive` call listed in the quick start guide, there didn't seem to be any way to bring the `mgmt` workspace back to a known good state in terms of those vendor sub-directories.

[...]

PR #623, opened 2020-12-18
use go mod to build

guilhem:
- remove submodules
- Updates dependencies to reflect imports


purpleidea:
Hi! Thanks for the patch. Unfortunately we don't want to move to go modules at this time, because we want to develop on git master. We'd like to be able to do this in a different way. More discussion here: https://github.com/purpleidea/mgmt/issues/572

If you're blocked for some reason because of this please let me know why.

Thanks!

Issue #624, opened 2021-01-01
Type unification always fails with embedded structs in a resource due to field naming contradiction


frebib:
## Versions:

* mgmt version (eg: `mgmt --version`): 01c2131436dc77bb9ad922c62440efab2ff7ec60 (master HEAD)
* operating system/distribution (eg: `uname -a`): Arch
* golang version (eg: `go version`): 1.15.6

## Description:

This bug (oversight? I'm not sure) was actually already discovered 2 years ago https://github.com/purpleidea/mgmt/issues/432#issuecomment-450700956 in a similar issue that may actually be the same root problem as this. The error thrown is 
```
could not unify types: can't unify, invariant illogicality with equals: struct fields differ
```
Using the example from that comment we can see that the resource struct has a struct child, which in theory mcl should support just fine:
```go
type IPRange struct {
	From string `lang:"from" yaml:"from"`
	To   string `lang:"to" yaml:"to"`
}

// DHCPDRes is a dhcp daemon resource.
type DHCPDRes struct {
	traits.Base // add the base methods without re-implementation
	traits.Groupable
	traits.Refreshable

	init *engine.Init

	// Range defines the upper and lower limits for ip address assignments.
	Range IPRange `lang:"range" yaml:"range"`
	...
```
with some matching mcl ala
```ruby
dhcpd "foo" {
    range => struct{
        from => "192.168.42.100/24",
        to => "192.168.42.200/24",
    },
...
```
logic would dictate that this should work fine. The structs are tagged with the `lang` tag to indicate that the mcl fields should be matched to the struct field with the same tag: mcl `range.from` should match to the Go field `Range.From`. This is exactly how it works for resource struct fields. Even the `lang` struct tags are superfluous because the "default" behaviour is to use the struct field name in lowercase. That is the expected behavior and if it's not already explicitly documented, it probably should be.

The lang tags are the first clue as to what the issue is here. Looking at the code, we'd expect to find something doing roughly:
```
list fields in the go struct
enumerate their "mcl" struct field names
match the mcl fields to the go struct fields according to the name mappings
happy days
```
[...]

PR #625, opened 2021-01-05
lang: Add struct field matching test case for #624

frebib:
Struct field matching between lowercase/uppercase names even with
`lang:"name"` tags highlighting the issue in #624, with the hope that it
is fixed so the test case can be updated.

### DIFF
--- a/engine/resources/test.go
+++ b/engine/resources/test.go
@@ -82,6 +82,13 @@ type TestRes struct {
 	} `lang:"mixedstruct" yaml:"mixedstruct"`
 	Interface interface{} `lang:"interface" yaml:"interface"`
 
+	// See TestAstFunc2/struct-filename0
+	// https://github.com/purpleidea/mgmt/issues/624
+	NamedStruct struct {
+		// Go field name explicitly differs from lang/mcl tag
+		Named string `lang:"named" yaml:"named"`
+	} `lang:"namedstruct" yaml:"namedstruct"`
+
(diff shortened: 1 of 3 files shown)


purpleidea:
I'm assuming you *don't* want this merged for now, and this is for example purposes, and because if you did, it's missing a \n and mcl uses tabs not spaces for indentation.

frebib:
I can fix the whitespace errors - good eye. (Maybe I'll have to look at the formatter task at some point. No idea how, mind you)
Up to you if you want to merge this. I need to spend some more time on the fix.
If it ends up being merged, I will update the test case along with the fix

purpleidea:
On Thu, Jan 7, 2021 at 4:59 PM Joe Groocock <notifications@github.com> wrote:
>
> I can fix the whitespace errors - good eye. (Maybe I'll have to look at the formatter task at some point. No idea how, mind you)
> Up to you if you want to merge this. I need to spend some more time on the fix.
> If it ends up being merged, I will update the test case along with the fix

I'm slightly confused though-- why are we adding this test that
confirms a fail... Don't we want this to pass??

frebib:
You're right. This test demonstrates the failure, but hopefully we wont need it soon :)

purpleidea:
@frebib You shouldn't close it-- if anything you should reopen it and have it look like what's *supposed* to happen-- it will be failing, but someone soon (or in the future) can always rebase it to master once a fix is committed, and that will confirm the fix! =D

PR #626, opened 2021-01-30
lang: move engine/util funcs into lang package

frebib:
This avoids circular import references in the lang/types package when
calling these functions from lang/types. It could be argued that these
never really belonged in the engine/util package to begin with.

This is almost a direct code move, except for a few functions that have
been renamed for brevity and the function signature adjusted for
reusability particularly with regards to passing reflect.Type instead of
a concrete type, which makes the function much more useful in the
general case.


frebib:
For what it's worth, I do plan to refactor and simplify these functions some more in future PRs, but for now I'm trying to keep the changes as small as possible. There are improvements to be made to the consistency of lang struct handling on a project level, but one step at a time. This move also includes a small amount of refactoring too, but I can always change this up to do perform a direct move in one PR then resubmit the refactor later. I actually wrote the refactor before I moved the code, hence why it's bundled.

purpleidea:
> For what it's worth, I do plan to refactor and simplify these functions some more in future PRs, but for now I'm trying to keep the changes as small as possible.

I approve this part =D

...

Before I dig deeper into the code changes, is there a reason why we shouldn't or can't put the util stuff in a "lang/types/util" package instead? (eg: since I don't think they refer to each other do they?)

One other small note: there seem to be some changes which are different from "the straight move". Is there a reason for this, or can we split the two into separate commits in the odd chance we need to bisect or look into past history one day?

frebib:
>  is there a reason why we shouldn't or can't put the util stuff in a "lang/types/util" package instead?
I've never been sure whether to have `package/util.go` or `package/util/whatever.go`. I think you're right that these util functions don't reference anything, so moving them is probably preferable. Go's lack of cyclic imports still frustrates me, and I've learned that it's best to use either one monolithic package, or lots of small nested packages. I'm still fighting other similar cyclic import related issues so in the interest of avoiding more of those in the future I'll move it to a util subpackage.

> there seem to be some changes which are different from "the straight move".
Correct, there are. I started refactoring the function signatures a little before I realised they needed moving. The change isn't too significant so I'll back-peddle and split the move/refactor into two commits so it's clearer what the change is.

frebib:
Seemingly this change is not blocking, so I may revisit this later.
I'd still like to tidy this logic up at some point

PR #627, opened 2021-01-30
lang: Fix panic in lang/types/ValueOf() for Struct

frebib:
Replace use of reflect.Value.Len() with NumField() which is intended to
return the number of fields in reflected Struct value.

Len should only be used for Array, Chan, Map, Slice and String types.

Add some trivial sanity check tests for ValueOf() for the simple and
complex container types.

### DIFF
--- a/lang/types/value_test.go
+++ b/lang/types/value_test.go
@@ -21,6 +21,7 @@ package types
 
 import (
 	"fmt"
+	"reflect"
 	"sort"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


lang/types/value_test.go
```diff
@@ -590,5 +591,51 @@ func TestStruct2(t *testing.T) {
...
+		// TODO: implement ValueOf tests for TypeFunc
+	}
+
+	for value, gotyp := range testCases {
+		// get reflect.Value, then call ValueOf() for types.Value
+		val, err := ValueOf(reflect.ValueOf(gotyp))
+		if err != nil {
+			t.Errorf("ValueOf(%+v) returned err %s", gotyp, err)
+		}
+		// use string representation comparison as maps are non-deterministic in order
```
purpleidea, line 635: fun fact: I'm the reason that this is possible: https://github.com/golang/go/issues/21095

purpleidea:
This looks perfect, and a nice catch too! Thank you =D

purpleidea:
Woops. I didn't merge this, because it fails tests. One sec, I'll post output

purpleidea:
Merged in https://github.com/purpleidea/mgmt/commit/845d7ff188480737070c19a6178502e517746120 \o/

PR #629, opened 2021-01-30
lang: Implement Value.Into() to reflect.Value representations

frebib:
Into() mutates a given reflect.Value and sets the data represented by
the types.Value into the storage represented by the reflect.Value.

Into() is the opposite to ValueOf() which converts reflect.Value into
types.Value, and in theory they should (almost) bijective with some edge
case exceptions where the conversion is lossy.

Simply, it replaces reflect.Value.Set() in the broad case, giving finer
control of how the reflect.Value is modified and how the data is set.
types.Value.Value() is now also a redundant function that achieves the
same outcome as Into(), but with less type specificity.

---

lang: convert StmtRes to engine.Res with types.Into()

Replace existing field-mapping code with calls to types.Into() to
reflect the mcl data into the Go resource struct with finer granularity
and accuracy, and less reliance on the magic reflect.Set() function.o

One major advantage over reflect.Value.Set() is Into() allows tailoring
the data that is set, specifically when coercing mcl struct values into
Golang struct values, the fields can be appropriately mapped based on
the lang tag on the struct field. With reflect.Value.Set() this was not
at all possible as there was a contradiction of logic given the
following rules:

- mcl struct fields must have lowercase names
- Golang struct fields with lowercase names are unexported
- Golang reflection does not allow modifying unexported fields

Prior to this change, it was impossible to map an mcl inline struct in a
resource to the matched Golang counterpart, even if the lang tag was
present on the struct field. This can be demonstrated with the following
trivial example:

```perl
test "name" {
    key => struct{
        name => "hello",
    },
}
```

and the accompanying Golang resource definition:

```go
type TestRes struct {
    traits.Base
    traits.Edgeable

    Key struct {
        Name string `lang:"name"`
    } `lang:"key"`
}
```

Due to the mismatch in field names in the embedded struct, the type
unifier failed and refused to match mcl 'name' to Go 'Name' due to the
missing mapping logic.

---

lang: map Go struct fields using `lang` struct tag

Converting a reflect.Type of KindStruct did not respect the `lang` tag
on struct fields incidating how fields from mcl structs should be mapped
even though resource field names did. This patch should allow structs
with mapped fields to be respected.

### DIFF
--- a/examples/lang/virt2.mcl
+++ b/examples/lang/virt2.mcl
@@ -29,26 +29,24 @@ virt "mgmt4" {
 	state => "running",
 	transient => false,
 	boot => ["hd", ],
-	# can't add this part until we fix the unification bug
-	#disk => [
-	#	struct{
-	#		source => "~/.local/share/libvirt/images/fedora-23-scratch.qcow2",
-	#		type => "qcow2",
-	#	},
-	#],
-	# add the rest for unification bug
-	#osinit => "",
-	#cdrom => [
-	#],
-	#network => [
-	#],
-	#filesystem => [
-	#],
-	#auth => struct{
-	#	username => "",
-	#	password => "",
-	#},
-	#hotcpus => true,	# this is the default
-	#restartondiverge => "",
-	#restartonrefresh => false,
+	disk => [
+		struct{
+			source => "~/.local/share/libvirt/images/fedora-23-scratch.qcow2",
+			type => "qcow2",
+		},
+	],
+	osinit => "",
+	cdrom => [
+	],
+	network => [
+	],
+	filesystem => [
+	],
+	auth => struct{
+		username => "",
+		password => "",
+	},
+	hotcpus => true,	# this is the default
+	restartondiverge => "",
+	restartonrefresh => false,
 }
--- a/lang/structs.go
+++ b/lang/structs.go
@@ -604,8 +604,8 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
 		return nil, errwrap.Wrapf(err, "cannot create resource kind `%s` with named `%s`", obj.Kind, resName)
 	}
 
-	s := reflect.ValueOf(res).Elem() // pointer to struct, then struct
-	if k := s.Kind(); k != reflect.Struct {
+	sv := reflect.ValueOf(res).Elem() // pointer to struct, then struct
+	if k := sv.Kind(); k != reflect.Struct {
 		panic(fmt.Sprintf("expected struct, got: %s", k))
 	}
 
@@ -613,7 +613,7 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
 	if err != nil {
 		return nil, err
 	}
-	ts := reflect.TypeOf(res).Elem() // pointer to struct, then struct
+	st := reflect.TypeOf(res).Elem() // pointer to struct, then struct
 
 	// FIXME: we could probably simplify this code...
 	for _, line := range obj.Contents {
@@ -638,25 +638,19 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
 			return nil, errwrap.Wrapf(err, "resource field `%s` did not return a type", x.Field)
 		}
 
-		fieldValue, err := x.Value.Value() // Value method on Expr
-		if err != nil {
-			return nil, err
-		}
-		val := fieldValue.Value() // get interface value
-
 		name, exists := mapping[x.Field] // lookup recommended field name
-		if !exists {
+		if !exists {                     // this should be caught during unification.
 			return nil, fmt.Errorf("field `%s` does not exist", x.Field) // user made a typo?
 		}
 
-		f := s.FieldByName(name) // exported field
-		if !f.IsValid() || !f.CanSet() {
-			return nil, fmt.Errorf("field `%s` cannot be set", name) // field is broken?
+		tf, exists := st.FieldByName(name) // exported field type
+		if !exists {
+			return nil, fmt.Errorf("field `%s` type does not exist", x.Field)
 		}
 
-		tf, exists := ts.FieldByName(name) // exported field type
-		if !exists {                       // illogical because of above check?
-			return nil, fmt.Errorf("field `%s` type does not exist", x.Field)
+		f := sv.FieldByName(name) // exported field
+		if !f.IsValid() || !f.CanSet() {
+			return nil, fmt.Errorf("field `%s` cannot be set", name) // field is broken?
 		}
 
 		// is expr type compatible with expected field type?
@@ -668,65 +662,15 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
 			return nil, errwrap.Wrapf(err, "resource field `%s` of type `%+v`, cannot take type `%+v", x.Field, t, typ)
 		}
 
-		// user `pestle` on #go-nuts irc wrongly insisted that it wasn't
-		// right to use reflect to do all of this. what is a better way?
-
-		// first iterate through the raw pointers to the underlying type
-		ttt := tf.Type // ttt is field expected type
-		tkk := ttt.Kind()
-		for tkk == reflect.Ptr {
-			ttt = ttt.Elem() // un-nest one pointer
-			tkk = ttt.Kind()
+		fv, err := x.Value.Value() // Value method on Expr
+		if err != nil {
+			return nil, err
 		}
 
-		// all our int's are src kind == reflect.Int64 in our language!
-		if obj.data.Debug {
-			obj.data.Logf("field `%s`: type(%+v), expected(%+v)", x.Field, typ, tkk)
+		// mutate the struct field f with the mcl data in fv
+		if err := types.Into(fv, f); err != nil {
+			return nil, err
 		}
-
-		// overflow check
-		switch tkk { // match on destination field kind
-		case reflect.Int, reflect.Int64, reflect.Int32, reflect.Int16, reflect.Int8:
-			ff := reflect.Zero(ttt)          // test on a non-ptr equivalent
-			if ff.OverflowInt(val.(int64)) { // this is valid!
-				return nil, fmt.Errorf("field `%s` is an `%s`, and value `%d` will overflow it", x.Field, f.Kind(), val)
-			}
-
-		case reflect.Uint, reflect.Uint64, reflect.Uint32, reflect.Uint16, reflect.Uint8:
-			ff := reflect.Zero(ttt)
-			if ff.OverflowUint(uint64(val.(int64))) { // TODO: is this correct?
-				return nil, fmt.Errorf("field `%s` is an `%s`, and value `%d` will overflow it", x.Field, f.Kind(), val)
-			}
-
-		case reflect.Float64, reflect.Float32:
-			ff := reflect.Zero(ttt)
-			if ff.OverflowFloat(val.(float64)) {
-				return nil, fmt.Errorf("field `%s` is an `%s`, and value `%d` will overflow it", x.Field, f.Kind(), val)
-			}
-		}
-
-		value := reflect.ValueOf(val) // raw value
-		value = value.Convert(ttt)    // now convert our raw value properly
-
-		// finally build a new value to set
-		tt := tf.Type
-		kk := tt.Kind()
-		if obj.data.Debug {
-			obj.data.Logf("field `%s`: start(%v)->kind(%v)", x.Field, tt, kk)
-		}
-		//fmt.Printf("start: %v || %+v\n", tt, kk)
-		for kk == reflect.Ptr {
-			tt = tt.Elem() // un-nest one pointer
-			kk = tt.Kind()
-			if obj.data.Debug {
-				obj.data.Logf("field `%s`:\tloop(%v)->kind(%v)", x.Field, tt, kk)
-			}
-			// wrap in ptr by one level
-			valof := reflect.ValueOf(value.Interface())
-			value = reflect.New(valof.Type())
-			value.Elem().Set(valof)
-		}
-		f.Set(value) // set it !
 	}
 
 	return res, nil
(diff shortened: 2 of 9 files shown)


lang/types/value.go
```diff
@@ -593,14 +669,26 @@ func (obj *ListValue) Copy() Value {
...
+}
 
-	for _, x := range obj.V {
-		val = reflect.Append(val, reflect.ValueOf(x.Value())) // recurse
+// Into sets the value of the given reflect.Value to the value of this obj
+func (obj *ListValue) Into(val reflect.Value) {
+	val = FollowPtr(val)
+	count := len(obj.V)
+	if count > val.Cap() {
+		// TODO: should this just use len(obj.V) for cap
```
frebib, line 680: Want a second opinion here. Is this premature optimisation? Should I just drop the whole power-of-two thing and just use `len(obj.V)` instead? Is the additional memory allocation wasteful?

lang/types/value.go
```diff
@@ -967,10 +1080,11 @@ func (obj *FuncValue) Value() interface{} {
...
 			panic(fmt.Sprintf("function panic: %+v", err))
 		}
-		return []reflect.Value{reflect.ValueOf(result.Value())} // only one result
+		out := reflect.New(v.Type().Out(0))
+		result.Into(out)
+		return []reflect.Value{out} // only one result
 	}
-	val := reflect.MakeFunc(typ, fn)
-	return val.Interface()
+	v.Set(reflect.MakeFunc(v.Type(), fn))
```
frebib, line 1087: I'm not actually 100% sure how the FuncValue type works so this is mostly a guess. It does pass tests though, assuming there are tests for this

lang/types/value.go
frebib: Decided to keep this after a discussion with James. We're mostly indifferent, but this more closely resembles what the Go runtime does internally

lang/structs.go
```diff
@@ -638,27 +638,21 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
 			return nil, errwrap.Wrapf(err, "resource field `%s` did not return a type", x.Field)
 		}
 
-		fieldValue, err := x.Value.Value() // Value method on Expr
-		if err != nil {
-			return nil, err
+		name, exists := mapping[x.Field] // lookup recommended field name
+		if !exists {                     // this should be caught during unification. TODO: panic?
```
purpleidea, line 642: Please remove the `TODO: panic` ... Even if the issue is caused by "programmer error" it's still nicer to normal error and bubble it up with an obscure error message, instead of panicing, and potentially shutting down unsafely. There are a few exceptions though.

lang/structs.go
```diff
@@ -668,65 +662,16 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
...
-			kk = tt.Kind()
-			if obj.data.Debug {
-				obj.data.Logf("field `%s`:\tloop(%v)->kind(%v)", x.Field, tt, kk)
-			}
-			// wrap in ptr by one level
-			valof := reflect.ValueOf(value.Interface())
-			value = reflect.New(valof.Type())
-			value.Elem().Set(valof)
+		// mutate the struct field f with the mcl data in fv
+		err = types.Into(fv, f)
```
purpleidea, line 671: `if err := Into(...); err != nil {` ...

lang/structs.go
```diff
@@ -604,16 +604,16 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
 		return nil, errwrap.Wrapf(err, "cannot create resource kind `%s` with named `%s`", obj.Kind, resName)
 	}
```
purpleidea, line 606: commit message has: `function.o` ?

lang/types/type_test.go
```diff
@@ -1474,4 +1474,5 @@ func TestComplexCmp0(t *testing.T) {
 
 func TestTypeOf0(t *testing.T) {
 	// TODO: implement testing of the TypeOf function
```
purpleidea, line 1476: commit message: "should (almost) bijective" missing a word? `be`??

examples/lang/virt2.mcl
```diff
@@ -29,26 +29,24 @@ virt "mgmt4" {
...
-	#hotcpus => true,	# this is the default
-	#restartondiverge => "",
-	#restartonrefresh => false,
+	disk => [
+		struct{
+			source => "~/.local/share/libvirt/images/fedora-23-scratch.qcow2",
+			type => "qcow2",
+		},
+	],
+	osinit => "",
```
purpleidea, line 38: Out of curiosity, did you test all this out? ;) Don't worry about it though, we can merge as is =D

purpleidea (commented):
What an excellent patch-set. Thanks @frebib !

I think the choice of how you split the commits was excellent. I made a few review comments over 2-3 commits, so if you find it tricky git-skills wise to squeeze these changes in, lmk and I'll help you with git hacks!

Overall the patches look great. I had a few nitpicks (sorry) and the only one real substantial thing was the panic vs. error propagation stuff. I think we can do more error bubbling and less panic-ing if you agree. This might cause a few of your tests to need to be changed to look for an error instead of a panic, sorry! Let me know if it makes sense to move away from panic or if there was a necessary reason.

One disclaimer: I didn't smoke test this manually yet, or run the `gofmt`/`govet` tests yet, but barring a final check, this looks good! We'll get automatic CI working again shortly!

lang/types/value.go
```diff
@@ -203,6 +203,182 @@ func ValueOf(v reflect.Value) (Value, error) {
 	}
 }
 
+// Into mutates the given reflect.Value with the data represented by the Value
```
purpleidea, line 206: Last nit: add a trailing period please. If you would like to be even more wordy and explicit that this modifies the reflect.Value even if this errors, then that would be welcome as well.

lang/types/value.go
```diff
@@ -203,6 +203,182 @@ func ValueOf(v reflect.Value) (Value, error) {
...
+		inlist := func(i reflect.Kind, l []reflect.Kind) bool {
+			for _, n := range l {
+				if i == n {
+					return true
+				}
+			}
+			return false
+		}
+		if !inlist(rv.Kind(), kinds) {
+			panic(fmt.Sprintf("cannot Into() a %s into reflect %s", v.Type(), kinds))
```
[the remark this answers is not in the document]
frebib: The logic here is to show a clear divide between user errors, like the integer overflow errors, and type matching errors such as trying to `Into()` an int into a string container. I'm emulating what the reflect library would do if you tried to perform the equivalent action.
If the `Into()` function ever panics then it's certainly a programming error. Any type mismatches should be caught in the unification stage.

lang/types/value.go
```diff
@@ -203,6 +203,182 @@ func ValueOf(v reflect.Value) (Value, error) {
...
+			rv.Set(reflect.New(typ))
+		}
+		rv = rv.Elem() // un-nest rv from pointer
+	}
+
+	// capture rv and v in a closure that is static for the scope of this Into() call
+	// mustInto ensures rv is in a list of compatible types before attempting to reflect it
+	mustInto := func(kinds ...reflect.Kind) {
+		// sigh. Go can be so elegant, and then it makes you do this
+		inlist := func(i reflect.Kind, l []reflect.Kind) bool {
```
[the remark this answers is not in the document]
frebib: Good point. I'll do one better and drop the inner function; it's not required at all

lang/types/value_test.go
```diff
@@ -639,3 +640,276 @@ func TestValueOf0(t *testing.T) {
...
+					X: ptrstr("string pointer"),
+					Y: ptrto(555).(*int),
+				},
+			},
+		},
+	}
+
+	for index, tc := range testCases {
+		name := fmt.Sprintf("test Into() %s #%d", reflect.TypeOf(tc.container).Elem(), index)
+		t.Run(name, func(t *testing.T) {
```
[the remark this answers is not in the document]
frebib: It looks that by default t.Run() is sync, unless t.Parallel() is called, in which case it runs them async. I'll add the `tc := tc` statement anyway just to be safe. Performance isn't critical in tests

lang/types/value.go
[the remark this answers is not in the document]
purpleidea: > If the Into() function ever panics then it's certainly a programming error. Any type mismatches should be caught in the unification stage.

I completely agree! And it would be a bug. But for database-safety (and other reasons) when we hit this bug, an `error` will cause a safe shutdown, where as a panic would kill everything even if it's mid-write.

So for this reason we error.

It's better to panic in a situation before a bug that would otherwise *cause* a database corruption or some other bad thing. But then only if you can't bubble errors properly.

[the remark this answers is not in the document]
frebib: Okay I'll replace all `panic()` with `return fmt.Errorf(..`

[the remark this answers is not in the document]
frebib: Any calls to reflect will still panic though, if the errors make it that far, which for at least half of the types it will do. Should I try to catch the other errors too?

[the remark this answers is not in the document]
purpleidea: It's true that the reflect lib panics in a number of cases. It's always good to check the panic condition if you know what it is when you can. Eg: if something will panic on calling a function with a nil arg, check if the arg is nil first. =D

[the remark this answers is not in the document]
frebib: For the simple types: `SetInt`, `SetBool` etc I didn't add any type checks at all because I was relying on reflect causing the panics instead. Should I add some `mustInto()` checks for those too?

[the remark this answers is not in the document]
purpleidea: Can't hurt AFAICT, right?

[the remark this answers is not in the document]
frebib: A tiny amount of performance due to extra if statements, but should be fine

lang/types/value_test.go
```diff
@@ -639,3 +640,276 @@ func TestValueOf0(t *testing.T) {
...
+					}
+				}()
+			}
+
+			// Into() the value into the reflect.Value
+			err := Into(tc.value, rvo)
+
+			// check for non/expected errors
+			if !tc.shouldErr && err != nil {
+				t.Errorf("function Into() returned an error: %s", err)
```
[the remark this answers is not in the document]
frebib: Yep. I was under the impression that `t.Error()` was fatal, but it's not. Fixed

lang/structs.go
frebib: Sure, removed. I'd hope we can remove most of these checks at some point because a lot of them are redundant

lang/structs.go
frebib: Oops, that was a typo. Good catch

lang/types/type_test.go
frebib: I have no what you're talking about :grin:

examples/lang/virt2.mcl
frebib: I tried running the example and the  unification works properly now, as expected, but the resources failed due to missing files. I didn't debug further than that

lang/types/value.go
frebib: Done

frebib:
All comments addressed, I think.
Main changes are:
- all missing `.` and capital letters were fixed
- `Into()` will prefer to return an `error` instead of `panic()` in every case. If it `panic()`s then it'll be from the `reflect` library, and is likely indicative of a bug in the `Into()` implementation.

frebib:
🙈 Had to fix-up some `go vet` complaints about incorrectly reflowed comments, but I think we're good now. Running a full `make test` on this just to be sure

lang/types/value.go
```diff
@@ -203,6 +203,201 @@ func ValueOf(v reflect.Value) (Value, error) {
...
+				// when calling our function with the Call method, then
+				// we get the error output and have a chance to decide
+				// what to do with it, but when calling it from within
+				// a normal golang function call, the error represents
+				// that something went horribly wrong, aka a panic...
+				panic(fmt.Errorf("function panic: %+v", err))
+			}
+			out := reflect.New(rv.Type().Out(0))
+			// convert the lang result back to a Go value
+			if err = Into(result, out); err != nil {
```
purpleidea, line 385: `err :=`

lang/structs.go
```diff
@@ -668,65 +662,15 @@ func (obj *StmtRes) resource(resName string) (engine.Res, error) {
...
-			kk = tt.Kind()
-			if obj.data.Debug {
-				obj.data.Logf("field `%s`:\tloop(%v)->kind(%v)", x.Field, tt, kk)
-			}
-			// wrap in ptr by one level
-			valof := reflect.ValueOf(value.Interface())
-			value = reflect.New(valof.Type())
-			value.Elem().Set(valof)
+		// mutate the struct field f with the mcl data in fv
+		if err = types.Into(fv, f); err != nil {
```
purpleidea, line 671: `err :=`

lang/types/value_test.go
```diff
@@ -639,3 +640,276 @@ func TestValueOf0(t *testing.T) {
...
+			} else if tc.shouldErr && err == nil {
+				t.Errorf("function Into() didn't return an error but one was expected")
```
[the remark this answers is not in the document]
purpleidea: What happened to this return and the one above it?

purpleidea (commented):
I'm happy to merge this as-is, but I think you might have forgotten one or two returns, and there are two `err := ...` which should be probably used instead of `err =` ... Basically, you always want to limit the scope whenever you can, since you never know what someone might add above you.

frebib:
I've fixed up the `err :=` instances. It looks like I forgot to stage/commit the additional `return` statements; they were still in my working tree. Hopefully I didn't miss anything

purpleidea:
19:09 < purpleidea> frebib: Your PR failed because the first word in the commit 
                    message AFTER the prefixes needs to be capitalized...
19:09 < purpleidea> eg: lang: Move StructTag const etc
19:09 < purpleidea> OR
19:09 < purpleidea> example: lang: Reinstate mcl blah

Can you capitalize the first word after the prefixes and resubmit please? Sorry about that.

frebib:
Fixed up the capitalisation in the commit messages

purpleidea:
A very fantastic patchset merged. Thank you again!

PR #630, opened 2021-02-07
etcd: Bump etcd + friends to the latest upstream version

frebib:
This allows dropping the pinned grpc-prometheus and grpc-gateway
libraries as git master works fine for now.


frebib:
Hm I think this is not actually working as intended. Upstream renamed the package from `clientv3` to `client/v3` so I guess it's using the GOPATH code instead of this vendored code

frebib:
Nope never mind, those changes are in master, after this pinned version. That's something fun to look forward to next time

purpleidea:
Why did you pick grpc version 1.29, instead of `v1.26.0` ? Just curious...

I think it might be a bit dangerous to unpick some of these GRPC deps, but I'm going to go with it for now =D

Merged =D

frebib:
I could have sworn that I looked at the latest tag and not master, but I guess not. They did a lot of refactoring by the looks of things.
https://github.com/etcd-io/etcd/blob/master/go.mod#L31
If it causes any issues we can put the pin back

PR #631, opened 2021-02-07
docker: Bump docker dependencies, add containerd

frebib:
These dependencies are maintained because the upstream repos bundle
vendor directories into the repos, which cause namespacing issues during
build. Git submodules don't strip the vendor directory whereas most
vendoring tools would.

This PR requires 
https://github.com/frebib/containerd/commit/231d09c8988e17f10853232bdee2441a4b4f3541
https://github.com/frebib/distribution/commit/a778dab7d0f94fcd7950a8eb439fa1accb445839
https://github.com/frebib/docker/commit/6c6f9d8cf0bae08f7c91c6794906eba3ebac9205

I think this is probably the only major downside to using git submodules over the equivalent vendoring tools that take language-specific problems into account, such as vendored dependencies inside vendored dependencies.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,9 +28,9 @@
 [submodule "vendor/github.com/purpleidea/distribution"]
 	path = vendor/github.com/docker/distribution
 	url = https://github.com/purpleidea/distribution
-[submodule "vendor/github.com/purpleidea/go-connections"]
-	path = vendor/github.com/docker/go-connections
-	url = https://github.com/docker/go-connections
 [submodule "vendor/github.com/hashicorp/go-multierror"]
 	path = vendor/github.com/hashicorp/go-multierror
 	url = https://github.com/hashicorp/go-multierror
+[submodule "vendor/github.com/containerd/containerd"]
+	path = vendor/github.com/containerd/containerd
+	url = https://github.com/purpleidea/containerd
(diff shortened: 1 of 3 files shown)


purpleidea:
So this is a bit confusing... Are the docker people still copying in all their deps? :facepalm: 
And should I also change it to `moby/moby` ?
Lastly, did they break API? (hence the new `, nil` in there?)
Cheers!

frebib:
Yeah it seems they are still vendoring, unfortunately. Leave it as docker/docker as that's what the imports still use.
It looks like the API did change, but honestly I'm glad they did because the Docker API is terrible so changes hopefully mean improvements. When toying with the docker-compose resource I ran into some limitations of the old Docker library that was pinned so hopefully this is somewhat improved.
I noticed that I need to update the testes for this too, as it's failing because I missed adding a nil. I'll update the PR soon

frebib:
Fixed the tests. They should pass now too

frebib:
Oops

purpleidea:
I merged this. I think I pushed all the correct refs up to my branch, but even if I didn't, I'm sure the GH-sha1-duplication-bug will mean it's fine.

(See: the upstream took your change: https://github.com/containerd/containerd/commit/231d09c8988e17f10853232bdee2441a4b4f3541 -- note they added a warning message about it after enough people got confused)

Cheers!

PR #635, opened 2021-02-12
lang: Improve string interpolation

purpleidea:
The original string interpolation was based on hil which didn't allow
proper escaping, since they used a different escape pattern. Secondly,
the golang Unquote function didn't deal with backslash escaping either.

This patch replaces both of these so that string interpolation works
properly. This removes the ability to allow inline function calls in a
string, however this was an incidental feature, and it's not clear that
having it is a good idea.

There are still some small bugs or at least some uncertainties in what
should be expected, but this is definitely due to my lack of experience
with ragel, and should be easy to improve with a small patch and a test.
In any case, this is much better than the original hil implementation.


purpleidea:
@roidelapluie Sorry it took so long, but I have a fix for https://github.com/purpleidea/mgmt/issues/521
What's missing: I need _you_ to please have a look and make sure I didn't get any behaviour wrong. If you have any test cases to add (you can look and see how simple it is to add one liners) please do! Also please check I did them all the way we want!

Thanks!

examples/lang/os.mcl
```diff
@@ -9,19 +9,21 @@ file "${tmpdir}/execinfo" {
 	content => fmt.printf("mgmt is at %s\n", os.executable()),
 }
 
+$home = os.getenv("HOME")
 file "${tmpdir}/mgmtenv" {
 	state => $const.res.file.state.exists,
-	content => os.expand_env("$HOME sweet ${os.getenv(\"HOME\")}\n"),
```
frebib, line 14: Certainly worth a mention in the documentation that this specific behavior isn't allowed (any more)

lang/interpolate.go
```diff
@@ -35,12 +43,72 @@ type Pos struct {
...
+			}
+			exprs = append(exprs, expr)
+
+		case interpolate.Variable:
+			expr := &ExprVar{
+				Name: t.Name,
+			}
+			exprs = append(exprs, expr)
+		default:
+			return nil, fmt.Errorf("unknown term (%T): %+v", t, t)
```
frebib, line 82: TIL `%T`
https://golang.org/pkg/fmt/#hdr-Printing

lang/interpolate.go
```diff
@@ -254,3 +323,42 @@ func concatExprListIntoCall(exprs []interfaces.Expr) (interfaces.Expr, error) {
...
+		switch v := x.(type) {
+		case *ExprStr:
+			if !last {
+				last = true
+				result = append(result, x)
+				continue
+			}
+
+			// combine!
+			last := result[len(result)-1] // there has to be at least one
```
frebib, line 344: This `last` here confused me because it shadows the `var last bool` above, but this is an `Expr`. Maybe pick a different name so it's clearly a different type?

lang/interpolate/parse.rl
```diff
@@ -0,0 +1,180 @@
...
+				x = "\\"
+			case "\"":
+				x = "\""
+			case "$":
+				x = "$"
+			//case "0":
+			//	x = "\x00"
+			default:
+				//panic(fmt.Sprintf("unhandled escaped: %s", s))
+				x = s // XXX: should we error cleanly instead?
```
frebib, line 144: It makes sense to raise an `error` of `invalid escape sequence '$rune'` or similar. That's what most compilers would do IMO

lang/interpolate/parse.rl
```diff
@@ -0,0 +1,180 @@
...
+		#		  )
+		#	>start
+		#	@{ l = Literal{Value: "XXX" } }
+		#	;
+
+		# Anything followed by a '\' is matched and used appropriately.
+		#escaped_lit = '\\' any @{ l = Literal{Value: data[fpc:fpc+1]} };
+		#//l = Literal{Value: "Z("+data[fpc:fpc+1]+"):["+ data[fpc-1:fpc+1] +"]" }
+
+		# works!
```
frebib, line 79: Why is this duplicated? It's almost the same as below? Is it here for documentation purposes? Should it be removed?

lang/interpolate/parse.rl
frebib: Mostly because a user might be expecting a particular behaviour and silent failures are a nightmare to debug

lang/interpolate_test.go
```diff
@@ -127,6 +128,66 @@ func TestInterpolate0(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		}
+		testCases = append(testCases, test{
+			name: "variable escaping 1",
+			code: `
+			test "t1" {
+				stringptr => "\${hello}",
```
frebib, line 154: This relies on external behaviours of the implicit global variables (which I think should probably be removed). Can you add an explicit `$hello = "hello"` above?

lang/interpret_test.go
```diff
@@ -1130,6 +1138,7 @@ func TestAstFunc2(t *testing.T) {
 			errs: errs{
 				fail1: fail1,
 				fail9: fail9, // TODO: rename
+				fail8: fail8, // TODO: rename
```
frebib, line 1141: `failInterpolate`?

lang/interpret_test.go
```diff
@@ -1322,9 +1332,25 @@ func TestAstFunc2(t *testing.T) {
...
-			if err != nil {
+			if (!fail || !fail8) && err != nil {
 				t.Errorf("test #%d: FAIL", index)
-				t.Errorf("test #%d: interpolate failed with: %+v", index, err)
+				t.Errorf("test #%d: Interpolate failed with: %+v", index, err)
+				return
+			}
+			if fail8 && err != nil {
+				// TODO: %+v instead?
+				s := fmt.Sprintf("%s", err) // convert to string
```
frebib, line 1342: `err.Error()`? I think that's what fmt.Sprintf() does here anyway

lang/interpret_test/TestAstFunc2/escaping4.output
```diff
@@ -0,0 +1 @@
+# err: err8: parser failed: cannot parse string: X: \z
```
frebib, line 1: IMO `invalid escape sequence`

misc/make-deps.sh
```diff
@@ -104,6 +106,23 @@ if [ $travis -eq 0 ]; then
 	fi
 fi
 
+if [ $travis -eq 1 ]; then
```
frebib, line 109: Really? Can't it be installed from some package repo?

roidelapluie (commented):
I have posted a few comments, otherwise LGTM, I did not have the time to test it myself.

examples/lang/os.mcl
purpleidea: Mentioned in new commit message. (Coming shortly.)

lang/interpolate.go
purpleidea: Yeah! Good catch-- I don't know who wrote that nonsense!

lang/interpolate/parse.rl
purpleidea: This was my WIP branch that I pushed to get it out early. I've cleaned this up.

lang/interpret_test.go
purpleidea: I'll send a cleanup patch. This would have been smarter :P

lang/interpret_test.go
purpleidea: Sure, why not =D Not doing the %+v it seems anyways.

lang/interpret_test/TestAstFunc2/escaping4.output
purpleidea: true, sadly it's not guaranteed that we know why. TBH I'm trying to merge the ragel bits in a "good enough" state and hopefully a smart ragel person fixes the things I got wrong by adding fancy errors to it all.

misc/make-deps.sh
purpleidea: Apparently not :(

lang/interpolate_test.go
purpleidea: Not sure what you mean...

lang/interpolate/parse.rl
```diff
@@ -0,0 +1,180 @@
+package interpolate
```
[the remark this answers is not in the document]
purpleidea: Okay-- I moved the header into the .rl and the cleanup script strips the ragel header. No idea why I need _two_ sed's to get it working. Brain tired for now.

lang/interpolate_test.go
frebib: So these are only parse tests, not parse+eval tests. You can ignore me :)
I thought these were referring to some static variables in the code that I now can't find

frebib (approved):

purpleidea:
Merged a variant with comments address. Woo. There actually still is a subtle bug or two in the interpolation, but nothing too complex, and I think it can be easily fixed by someone who is experienced with ragel. Have a look at the mcl tests if interested. I wanted to get all the big plumbing in, so that future fixes are small. This is still a big improvement over the initial bad hil implementation!

PR #636, opened 2021-02-12
test: Add GitHub Actions test support

frebib:
Contains some other small cleanups, and some necessary improvements to make the CI formatting a bit tidier and cleaner.

### DIFF
--- a/test/util.sh
+++ b/test/util.sh
@@ -22,12 +22,14 @@ TIMEOUT="$timeout --kill-after=360s --signal=QUIT 300s"
 
 in_ci() {
 	if [ $# -eq 0 ]; then
-		test -n "$CI" -o -n "$TRAVIS" -o -n "$JENKINS_URL"
+		test -n "$CI" -o -n "$GITHUB_ACTION" -o -n "$TRAVIS" -o -n "$JENKINS_URL"
 		return $?
 	fi
 
 	for var in "$@"; do
 		case "$var" in
+		github)
+			test -n "$GITHUB_ACTION" && return 0;;
 		travis)
 			test "$TRAVIS" = "true" && return 0;;
 		jenkins)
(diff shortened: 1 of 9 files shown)


frebib:
I can't request reviewers @purpleidea 

cc @evrardjp based on your work

frebib:
Seems the test-shell is failing. I've not looked into why yet but I presume it's either a flaky test or the test is broken
https://github.com/frebib/mgmt/runs/1889240885?check_suite_focus=true#step:6:129

frebib:
Whilst I think about it (more of a note to myself more than anything) we need a step to send an IRC message the same way that Travis does

frebib:
> Seems the test-shell is failing. I've not looked into why yet but I presume it's either a flaky test or the test is broken
> https://github.com/frebib/mgmt/runs/1889240885?check_suite_focus=true#step:6:129

I'm starting to think that these shell tests are a bit flaky. I ran them locally and the one that failed in CI passed fine locally, but a couple of others later on failed for unrelated and silly reasons (couldn't bind, port already in use)

test/util.sh
```diff
@@ -20,26 +20,43 @@ fi
...
 	"$@" || failures=$( [ -n "$failures" ] && echo "$failures\\n$@" || echo "$@" )
 }
 
 # travis expander helpers from:
 # https://github.com/travis-ci/travis-rubies/blob/build/build.sh
 fold_start() {
-	if env | grep -q -e '^TRAVIS=true$'; then
-		echo -e "travis_fold:start:$1\033[33;1m$2\033[0m"
+	if in_ci travis; then
+		echo -e "travis_fold:start:$1\033[33;1m${@:2}\033[0m"
```
purpleidea, line 61: I'll merge this, but what's this for?

purpleidea:
So I merged all of this. yay \o/
I snuck in one additional commit after commit #4, and I rebased #5 ... Hope that's okay.
One question about: https://github.com/purpleidea/mgmt/pull/636#discussion_r575913084
And also why did you disable the `race` tests, should we add those back in?
Lastly, it's true that some of the shell tests are flaky. I can explain, but we should just be sad about them for now.

Thank you for the patches!

purpleidea:
PS: since github is dumb, it apparently is not smart enough to show even a partial "merged". Sorry everyone.

PR #638, opened 2021-02-28
learning and tests

S-ign:
## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/lang/funcs/core/net/ip_port.go
+++ b/lang/funcs/core/net/ip_port.go
@@ -20,6 +20,7 @@ package corenet
 import (
 	"errors"
 	"fmt"
+	"net"
 	"strconv"
 
 	"github.com/purpleidea/mgmt/lang/funcs/simple"
@@ -33,20 +34,25 @@ func init() {
 	})
 }
 
-// IPPort returns the combind IP:(input[0]) and Port:(input[1]) arguments as
-// long as port input is within range 1-65536
+// IPPort returns the combind IPv4/IPv6:(input[0]) and Port:(input[1]). Returns
+// error if IPv4/IPv6 string is incorrect format or Port not in range 0-65536.
 func IPPort(input []types.Value) (types.Value, error) {
-	ip := input[0].Str()
+	ip := net.ParseIP(input[0].Str()) // Is nil if incorrect format.
+	ipStr := input[0].Str()
 	port := input[1].Str()
 	portInt, err := strconv.Atoi(port)
 
+	if ip == nil {
+		return &types.StrValue{V: ""}, errors.New("incorrect ip format")
+	}
 	if err != nil {
 		return &types.StrValue{V: ""}, fmt.Errorf("err converting str to int %v", err)
 	}
-	if portInt < 1 || portInt > 65536 {
-		return &types.StrValue{V: ""}, errors.New("port number must be between 1-65536")
+	if portInt < 0 || portInt > 65536 {
+		return &types.StrValue{V: ""}, errors.New("port not in range 0-65536")
 	}
+
 	return &types.StrValue{
-		V: ip + ":" + port,
+		V: ipStr + ":" + port,
 	}, nil
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lang/funcs/core/net/ip_port.go [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+	simple.ModuleRegister(ModuleName, "ip_port", &types.FuncValue{
+		T: types.NewType("func(ip, port str) str"),
+		V: IPPort,
+	})
+}
+
+// IPPort returns the combind IP:(input[0]) and Port:(input[1]) arguments
+// as long as port input is within range 1-65536
+func IPPort(input []types.Value) (types.Value, error) {
+	ip := input[0].Str()
```
frebib, line 39: We should validate the IP address too: https://golang.org/pkg/net/#ParseIP

lang/funcs/core/net/ip_port.go
```diff
@@ -0,0 +1,52 @@
...
+	port := input[1].Str()
+	portInt, err := strconv.Atoi(port)
+
+	if err != nil {
+		return &types.StrValue{V: ""}, fmt.Errorf("err converting str to int %v", err)
+	}
+	if portInt < 1 || portInt > 65536 {
+		return &types.StrValue{V: ""}, errors.New("port number must be between 1-65536")
```
frebib, line 47: nit: the wording here is slightly confusing given the logic. I would expect either `between -0-65536` (exclusive) or preferably `between 1-65535` (inclusive). The current wording is exclusive-inclusive which is probably not what the user would expect.

Side note, in many cases a port number of `0` can be valid as it's used to denote "any port". Perhaps we should allow that

lang/funcs/core/net/ip_port_test.go
```diff
@@ -0,0 +1,67 @@
...
+		name     string
+		ip       string
+		port     string
+		expected string
+		err      error
+	}{
+		// Success
+		{"success", "192.168.1.1", "80", "192.168.1.1:80", nil},
+		// Fail
+		{"fail", "192.168.1.1", "9483670", "", errors.New("port number must be between 1-65536")},
```
frebib, line 38: Can we get some IPv6 tests too?
Also some tests for invalid IP address/port combos if you add IP address parsing

frebib (changes requested):

lang/funcs/core/net/ip_port.go
S-ign: Oh ok, cool, didn't realize there was already an IP Parser, I'll add that! :)

lang/funcs/core/net/ip_port.go
S-ign: Thanks, I updated this, hopefully, the wording is a little better.

lang/funcs/core/net/ip_port_test.go
S-ign: net/ParseIP made this really easy to implement, thanks!

purpleidea:
Side note, re discussion in IRC:

I noticed that one of the tests failed in an unusual place in this PR... And I said, gah, I hope it's not a race. So I dug into it, and it turns out it was caused by a bug in your code! (yay, no our fault!) So the good news is the tests work, the bad news is I think we could have caught this issue in a more obvious place. So I added a new commit that should help new developers out by providing a clearer error message.

https://github.com/purpleidea/mgmt/commit/887d374c530915565b0257e6ac6f8c0cda232067

You'll want to rebase your branch against this commit which is now in git master.

Cheers!

purpleidea:
@S-ign Hey-- I'm not going to do a full review yet-- I'd like you to first make a small examples/lang/whatever.mcl that uses this function, and confirm you got it doing what you expect. This should replace the copied example that you committed in this PR.

When you've succeeded, please update this PR and put a comment here with the command you used to run this example =D

Thanks!

purpleidea:
(For anyone else reading, @S-ign came to us for a learning experience, so this is not the traditional review process.)

PR #642, opened 2021-03-03
test: Ensure branches are also testable

evrardjp:
test-commit-message runs on PR, but also on push in other branches
which aren't PRs. We need to test those too.

This is fixed by ensuring the same kind of behaviour than travis CI:
When a patch is put on a branch, it's using the branchname for
testing [1].

[1]:
https://docs-staging.travis-ci.com/user/environment-variables/#default-environment-variables

### DIFF
--- a/test/test-commit-message.sh
+++ b/test/test-commit-message.sh
@@ -123,7 +123,17 @@ then
 	done
 elif [[ -n "$GITHUB_SHA" ]]
 then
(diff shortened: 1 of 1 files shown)


evrardjp:
Follow up would be to do actual testing on pushes, instead of just _seem_ to test, and not actually test.

purpleidea:
Merged, thanks =D

PR #652, opened 2021-05-01
lang: Fix ValueOf() with nil pointer values

frebib:
Some forms of reflect.Value can cause ValueOf() to panic when there is a
nil pointer somewhere within the reflect.Value, whether that be a
container type like a struct, list or map, or just a raw nil pointer.

In these cases, ValueOf() attempted to dereference the pointer without
ever checking if it was nil. mgmt lang doesn't have pointers of any
kind, so the equivalent is the zero type of the value. In such cases,
gracefully attempt to handle by constructing a types.Value based on a
new instance of the type of the pointer reflect.Value instead.

### DIFF
--- a/lang/types/value.go
+++ b/lang/types/value.go
@@ -53,16 +53,28 @@ func ValueOfGolang(i interface{}) (Value, error) {
 }
 
 // ValueOf takes a reflect.Value and returns an equivalent Value.
-func ValueOf(v reflect.Value) (Value, error) {
-	value := v
+func ValueOf(value reflect.Value) (Value, error) {
+	// gracefully handle invalid values instead of panic()
+	if !value.IsValid() {
+		return nil, &reflect.ValueError{Method: "types.ValueOf", Kind: reflect.Invalid}
+	}
+
 	typ := value.Type()
 	kind := typ.Kind()
+
 	for kind == reflect.Ptr {
-		typ = typ.Elem() // un-nest one pointer
-		kind = typ.Kind()
+		// Prevent panic() if value is a nil pointer and instead return the zero
+		// value based on the type of the nil pointer.
+		if value.IsNil() {
+			t, err := TypeOf(typ)
+			if err != nil {
+				return nil, err
+			}
+			return t.New(), nil
+		}
 
 		// un-nest value from pointer
-		value = value.Elem() // XXX: is this correct?
+		return ValueOf(value.Elem())
 	}
 
 	switch kind { // match on destination field kind
@@ -199,7 +211,7 @@ func ValueOf(v reflect.Value) (Value, error) {
 		}, nil
 
 	default:
-		return nil, fmt.Errorf("unable to represent value of %+v", v)
+		return nil, fmt.Errorf("unable to represent value of %+v", value)
 	}
 }
 
--- a/lang/types/value_test.go
+++ b/lang/types/value_test.go
@@ -27,6 +27,28 @@ import (
 	"testing"
 )
 
+// mustValue converts a Go variable to a types.Value, or panics if any error.
+func mustValue(v interface{}) Value {
+	val, err := ValueOfGolang(v)
+	if err != nil {
+		panic(err)
+	}
+	return val
+}
+
+// Reflect variant of & on a variable. Creates a pointer in memory, sets the
+// destination, then returns the pointer.
+func ptrto(v interface{}) interface{} {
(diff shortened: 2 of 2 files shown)


purpleidea:
Optional nit about two sentences with missing periods and one capitalization perhaps, but I can merge as is if tests are relatively happy =D

Thanks for the contribution!

purpleidea:
(I don't know why my brain even sees these things.)

frebib:
> Optional nit about two sentences with missing periods and one capitalization perhaps, but I can merge as is if tests are relatively happy =D
> 
> Thanks for the contribution!

Haha I knew you'd pick up on this. I even try super hard when writing comments to try and do the right thing but I think I struggle to see the distinction. Will update

purpleidea:
I don't mind merging as-is.

lang/types/value.go
```diff
@@ -153,6 +164,14 @@ func ValueOf(v reflect.Value) (Value, error) {
 		values := make(map[string]Value)
 		for i := 0; i < l; i++ {
 			x := value.Field(i)
+			// It's likely that setting the field is desirable, so clone the
+			// field to prevent the recursive ValueOf() call failing. This is
+			// obviously slower so hope we don't need this often.
+			if !x.CanSet() {
+				c := reflect.New(x.Type()).Elem()
```
frebib, line 171: I'm still not 100% on this, but it seems to work as-is. I feel like I'm inferring `x.Kind() == Ptr` from `!x.CanSet()` which may or may not hold true in the majority of cases. I think the thing I'm not sure on is _why_ `x.CanSet() == false`. Debugging this isn't easy when it's all just pointers

frebib:
CI seems happy now. I fixed up the comments

lang/types/value.go
```diff
@@ -58,9 +58,20 @@ func ValueOf(v reflect.Value) (Value, error) {
 	typ := value.Type()
 	kind := typ.Kind()
 	for kind == reflect.Ptr {
+		// Not much that can be done here. Caller must pass a Settable Ptr value.
+		if !value.CanSet() {
```
purpleidea, line 62: I'm not 100% sure this is what you want. In situations like these, what I'd recommend doing, is adding a single simple test with the short, standalone bit that you want to test:

Like:
TestFrebibNewTest(t....) ...  {
v, err := ValueOf(XXX)
if err != nil ...
if v == nil ...
if err := v.Cmp(XXX); err != nil ...
}

I think that would help us better understand where you're stuck, and then we can go through and understand the reflect lib and which method you want a bit better.

purpleidea (changes requested):
I was bit sleepy last night so not really reading code properly. I think it's worth looking into this. I can help you solve it if you want.

I'm not convinced about `CanSet` initially, because it's not `value` that we want to set... In fact, we're building a new Value here, there's no modifying or "Set()" happening (in reflect terms) ... Of course I could be overlooking the use of CanSet to determine something else, so let's simplify this by defining exactly what's wrong with a standalone test case first.

(And I'm guessing something in your new docker resource code triggered this, so feel free to post that for a quick look in case the bug is there.)

lang/types/value.go
frebib: You're right. I was seriously overthinking this. Reworked it to a really simple change. Most of the diff is just adjusting the tests now.

frebib:
This still doesn't fix the crashes :(
reflect is doing something weird with nil pointers here that I don't understand

purpleidea:
Did you want a review on this?

One test is failing:

>   lang/types/value_test.go:617:26:warning: should drop = nil from declaration of var nilStrPtr; it is the zero value (golint)

purpleidea:
I wasn't 100% sure what test was failing previously that this patch was meant to fix, so I merged a patch with two new standalone tests... You'll note they both pass. This is what I had in mind just to make it very obvious to see what's going on...

33fda8605a7e46597651edc2834c2dae618ed6ee

All the table tests and wrapper functions are great for writing lots of tests, but in my simplistic brain this makes it easier to find/work on specific issues...

In any case, it all is passing. LMK how I can help.

frebib:
Tests should be fixed now. Assuming CI passes, this should be ready for review 🙏🏻

frebib:
After discussion, we came to the conclusion that this is probably not the correct fix. Instead the behaviour is correct, but the panic is not prevented/handled appropriately, to be fixed in https://github.com/purpleidea/mgmt/pull/655

PR #653, opened 2021-05-02
misc: Replace go-bindata with maintained fork

frebib:
As per [1] go-bindata was removed from GitHub and later replaced by the
community. jteeuwen/go-bindata has since been archived to represent this
state and now most communities use kevinburke/go-bindata instead as it
is more actively maintained.

[1]: https://github.com/jteeuwen/go-bindata/issues/5


purpleidea:
I'm okay with this commit, and I'll merge shortly, but we should actually move to golang's `embed` in `1.16`... I've played with it and it works well. I have to check if we can do everything we need with it, but I think it's likely.

purpleidea:
Merged!

PR #655, opened 2021-05-04
lang: Fix ValueOf() panic with nil pointer values

frebib:
Some forms of reflect.Value can cause ValueOf() to panic when there is a
nil pointer somewhere within the reflect.Value, whether that be a
container type like a struct, list or map, or just a raw nil pointer.

In these cases, ValueOf() attempted to dereference the pointer without
ever checking if it was nil. mgmt lang doesn't have pointers of any
kind, so these Golang values cannot be represented in mcl types in the
current form so return a helpful error to the user.

### DIFF
--- a/lang/types/value.go
+++ b/lang/types/value.go
@@ -18,6 +18,7 @@
 package types
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -25,6 +26,16 @@ import (
 	"strings"
 
 	"github.com/purpleidea/mgmt/util/errwrap"
+)
+
+var (
+	// ErrNilValue is returned when ValueOf() attempts to represent a nil
+	// pointer as an mcl value. This is not supported in mcl.
+	ErrNilValue = errors.New("cannot represent a nil Golang value in mcl")
+
+	// ErrInvalidValue is returned when ValueOf() is called on an invalid or
+	// zero reflect.Value.
+	ErrInvalidValue = errors.New("cannot represent invalid reflect.Value")
 )
 
 // Value represents an interface to get values out of each type. It is similar
@@ -65,12 +76,22 @@ func ValueOfGolang(i interface{}) (Value, error) {
 // a resource field. This is done with our "elvis" operator. When using this
 // function, if you pass in something with a nil value, then expect a panic or
 // an error if you're lucky.
-// FIXME: we should make sure we error instead of ever panic-ing on a nil value
 func ValueOf(v reflect.Value) (Value, error) {
+	// Gracefully handle invalid values instead of panic(). Invalid
+	// reflect.Value values can come from nil values, or the zero value.
+	if !v.IsValid() {
+		return nil, ErrInvalidValue
+	}
+
 	value := v
 	typ := value.Type()
 	kind := typ.Kind()
 	for kind == reflect.Ptr {
+		// Prevent panic() if value is a nil pointer and return an error.
+		if value.IsNil() {
+			return nil, ErrNilValue
+		}
+
 		typ = typ.Elem() // un-nest one pointer
 		kind = typ.Kind()
 
--- a/lang/types/value_test.go
+++ b/lang/types/value_test.go
@@ -20,6 +20,7 @@
 package types
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"reflect"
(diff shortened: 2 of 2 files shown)


frebib:
I should add a test for this. Please hold

frebib:
Assuming the tests pass, this should be ready for review.

lang/types/value.go
```diff
@@ -27,6 +28,10 @@ import (
 	"github.com/purpleidea/mgmt/util/errwrap"
 )
 
+var (
+	ErrNilValue = errors.New("cannot represent a nil Golang value")
```
purpleidea, line 32: Add a docstring to keep the linter happy please. and

```cannot represent a nil Golang value in mcl``` ?

File: lang/types/value.go [hunk 2]
```diff
@@ -65,17 +70,23 @@ func ValueOfGolang(i interface{}) (Value, error) {
 // a resource field. This is done with our "elvis" operator. When using this
 // function, if you pass in something with a nil value, then expect a panic or
 // an error if you're lucky.
-// FIXME: we should make sure we error instead of ever panic-ing on a nil value
-func ValueOf(v reflect.Value) (Value, error) {
-	value := v
+func ValueOf(value reflect.Value) (Value, error) {
+	// Gracefully handle invalid values instead of panic().
+	if !value.IsValid() {
```
purpleidea, line 75: Are you sure we want IsValid ?

lang/types/value.go
```diff
@@ -65,17 +70,23 @@ func ValueOfGolang(i interface{}) (Value, error) {
...
+		return nil, &reflect.ValueError{Method: "types.ValueOf", Kind: reflect.Invalid}
```
purpleidea, line 76: I am slightly afraid of returning an error from the reflect lib... Convince me it's okay, or change for a normal error please.

lang/types/value.go
```diff
@@ -65,17 +70,23 @@ func ValueOfGolang(i interface{}) (Value, error) {
...
+	}
+
 	typ := value.Type()
 	kind := typ.Kind()
+
```
purpleidea, line 81: kill extra nl

lang/types/value.go
```diff
@@ -212,7 +223,7 @@ func ValueOf(v reflect.Value) (Value, error) {
 		}, nil
 
 	default:
-		return nil, fmt.Errorf("unable to represent value of %+v", v)
```
purpleidea, line 215: what? how did this ever compile with `v` before?

lang/types/value.go
```diff
@@ -65,17 +70,23 @@ func ValueOfGolang(i interface{}) (Value, error) {
...
 	for kind == reflect.Ptr {
-		typ = typ.Elem() // un-nest one pointer
-		kind = typ.Kind()
+		// Prevent panic() if value is a nil pointer and return an error.
+		if value.IsNil() {
+			return nil, ErrNilValue
+		}
 
 		// un-nest value from pointer
-		value = value.Elem() // XXX: is this correct?
```
purpleidea, line 78: Why is this changing please?

purpleidea (changes requested):
Sorry for the nitpicky review. I'm more cautious in these scary "affects everything" bits of code.

lang/types/value.go
frebib: I renamed the first function arg because we were pointlessly declaring it.

lang/types/value.go
frebib: I'll add a note in the commit message. The functionality is ~ the same but the recursive call allows making the IsValid() check above on every pointer dereference. I'm not 100% if we need to do it (if the stdlib can ever return invalid reflect.Value values, that is) but I figured better safe than sorry. It's a pretty cheap check

lang/types/value.go
frebib: Just checking the docs, yes. .Elem() can return the zero value and IsValid() returns false for the zero value. Will add a comment to the code

lang/types/value.go
frebib: Reverted this back as it's not required

lang/types/value.go
frebib: Fixed

lang/types/value.go
frebib: Fixed

lang/types/value.go
frebib: Kept the IsValid check, but dropped the recursive call. Clarified with a comment above the check.

frebib:
Merged in 29f1c6f50e3ff031c35ab53187324023a9de3192

purpleidea:
Yes! Sorry I forgot to update this =D
Thanks for the patch!

PR #656, opened 2021-05-04
project: Add frebib to AUTHORS

frebib:


purpleidea:
Merged, yay \o/

What is with all the `J` names up in here, eh?

Now that you're in the `AUTHORS` file, you can display it proudly to all your friends, but you're also responsible for 100 good commits per week =D

Thanks for all the patches!

PR #658, opened 2021-05-04
lang: types: Clear map/list types during Into()

frebib:
Map and list types are now unconditionally initialised during an Into()
call to ensure that the only data within them after the operation is
that added by the Into() function.

Prior to this change, map/list types would likely not be cleared prior
to the data being inserted into them with a few exceptions. Nil
pointers or maps/lists that were not sufficient in capacity would be
reinitialised and used to replace the existing backing data store. In
some cases this wouldn't occur meaning any residual data existing in the
container before the Into() call could persist after the data copy
completes. This behaviour is wildly inconsistent and not ideal in the
vast majority of cases. It should be assumed that the Into() call will
preserve nothing and always produce a consistent and deterministic
output.

One bonus of this change is that Into() will now correctly set empty
lists to an empty slice instead of leaving a nil pointer, and the same
is likely true for maps too. It allows the user to distinguish between
nil and empty container types, as in certain cases those semantics can
make a difference.

### DIFF
--- a/lang/types/value_test.go
+++ b/lang/types/value_test.go
@@ -781,7 +781,7 @@ func TestValueInto0(t *testing.T) {
 	var l []string
 	var ll [][]string
 	var lptrlptr []*[]*string
-	var arr [10]string
+	var arr [3]string
 
 	var m map[string]int
 
@@ -860,7 +860,7 @@ func TestValueInto0(t *testing.T) {
 		{ // arrays are pretty much the same as slices
 			container: &arr,
 			value:     mustValue([]string{"1", "2", "3"}),
-			compare:   [10]string{"1", "2", "3"},
+			compare:   [3]string{"1", "2", "3"},
 		},
 		{
 			container: &ll,
@@ -892,6 +892,11 @@ func TestValueInto0(t *testing.T) {
 		{ // map[int]int into map[string]int
 			container: &m,
 			value:     mustValue(map[int]int{1: 2, 3: 4}),
+			shouldErr: true,
+		},
+		{ // [4]string  into [3]string
+			container: &arr,
+			value:     mustValue([4]string{"1", "2", "3", "4"}),
 			shouldErr: true,
 		},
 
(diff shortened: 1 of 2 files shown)


lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
...
 }
+
+func TestValueInto1(t *testing.T) {
+	// Ensure Into() clears existing elements out of the slice
+	goList := []string{"", "", "three"}
+	valList := &ListValue{
+		T: NewType("[]str"),
+		V: []Value{&StrValue{V: "one"}, &StrValue{V: "two"}},
+	}
+	err := Into(valList, reflect.ValueOf(&goList))
```
purpleidea, line 1019: `if err := ...; err != nil {`

lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
...
+	// Ensure Into() clears all existing elements out of the map
+	goMap := map[string]string{"3": "three"}
+	valMap := &MapValue{
+		T: NewType("map{str: str}"),
+		V: map[Value]Value{
+			&StrValue{V: "1"}: &StrValue{V: "one"},
+			&StrValue{V: "2"}: &StrValue{V: "two"},
+		},
+	}
+	err = Into(valMap, reflect.ValueOf(&goMap))
```
purpleidea, line 1036: `if err := ...; err != nil {`

lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
...
+func TestValueInto1(t *testing.T) {
+	// Ensure Into() clears existing elements out of the slice
+	goList := []string{"", "", "three"}
+	valList := &ListValue{
+		T: NewType("[]str"),
+		V: []Value{&StrValue{V: "one"}, &StrValue{V: "two"}},
+	}
+	err := Into(valList, reflect.ValueOf(&goList))
+	if err != nil {
+		t.Errorf("Into(%+v, %+v) failed: %s", valList, goList, err)
```
purpleidea, line 1021: add return?

lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
...
+	}
+	if !reflect.DeepEqual(goList, []string{"one", "two"}) {
+		t.Errorf("Into(%+v, %+v) did not clear existing values from the list", valList, goList)
```
purpleidea, line 1024: add return?

lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
...
+	valMap := &MapValue{
+		T: NewType("map{str: str}"),
+		V: map[Value]Value{
+			&StrValue{V: "1"}: &StrValue{V: "one"},
+			&StrValue{V: "2"}: &StrValue{V: "two"},
+		},
+	}
+	err = Into(valMap, reflect.ValueOf(&goMap))
+	if err != nil {
+		t.Errorf("Into(%+v, %+v) failed: %s", valMap, goMap, err)
```
purpleidea, line 1038: add return?

lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
...
+	}
+	if !reflect.DeepEqual(goMap, map[string]string{"1": "one", "2": "two"}) {
+		t.Errorf("Into(%+v, %+v) did not clear existing values from the map", valMap, goMap)
```
purpleidea, line 1041: add return?

purpleidea, line 1041: I think the error messages aren't supposed to be capitalized... So `func Into...` instead maybe?

lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
 		})
 	}
 }
+
+func TestValueInto1(t *testing.T) {
+	// Ensure Into() clears existing elements out of the slice
```
purpleidea, line 1013: period or uncapitalize please

lang/types/value_test.go
```diff
@@ -1008,6 +1008,40 @@ func TestValueInto0(t *testing.T) {
...
+	}
+	err := Into(valList, reflect.ValueOf(&goList))
+	if err != nil {
+		t.Errorf("Into(%+v, %+v) failed: %s", valList, goList, err)
+	}
+	if !reflect.DeepEqual(goList, []string{"one", "two"}) {
+		t.Errorf("Into(%+v, %+v) did not clear existing values from the list", valList, goList)
+	}
+
+	// Ensure Into() clears all existing elements out of the map
```
purpleidea, line 1027: Period or uncapitalize please

purpleidea (commented):
Cool idea for a patch! Hopefully this prevents some bug somewhere some day!

frebib:
Had to rework the implementation slightly (one of the existing tests was failing, showing a flaw in my patch!). It now takes arrays into account too and doesn't panic now. Reworked the tests too. Good for review again

purpleidea:
Sorry for the delay! All merged now, thanks =D

Issue #221, opened 2017-09-17
[LOVE] specify the user/group in the exec resource


purpleidea:
#### Welcome:

This is a beginner level patch for someone who is new to golang, and is new to mgmt. It is meant to encourage new programmers to get involved with mgmt. It will probably take you a few hours to a day or so to complete.

#### Intro:

We have an "exec" resource. We'd like to pick the user that commands run as. You'll want to be at least casually familiar with the https://github.com/purpleidea/mgmt/blob/master/docs/resource-guide.md first. Find the exec resource in git master and look at the code.

To implement this, you'll need to either recommend an alternate design, or use the one provided below. This hasn't been exhaustively investigated, so comments are welcome!

#### Design:

It seems the pattern to add this sort of thing might look like:

```
cmd := exec.Command(command, args...)
cmd.SysProcAttr = &syscall.SysProcAttr{}
cmd.SysProcAttr.Credential = &syscall.Credential{Uid: uid, Gid: gid}
```

I haven't tested this, but it looks plausible. You'll need to lookup the uid/gid beforehand. This is currently done in file.go for that code but it might sense to refactor these functions into resources/util.go first and modify to make them generic. This probably relates to https://github.com/purpleidea/mgmt/issues/214 (possible/likely dependency)

This should be fun code and good to orient yourself in a new project and around some new libraries! If you get stuck, ping purpleidea on Freenode IRC.

Read the contributions tips:
https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Happy hacking, and ping me for questions!

#### Bonus/Future work:

* Add automatic edges ( https://ttboj.wordpress.com/2016/03/14/automatic-edges-in-mgmt/ ) between this resource and the `user` and `group` resources. Group is being tracked in https://github.com/purpleidea/mgmt/issues/220 so as of current writing only user auto edges can be added at this time. Leave any ideas for other automatic edges in the comments.

* Add some tests! It would be nice to have something in resources/exec_test.go and maybe some fancier tests in https://github.com/purpleidea/mgmt/tree/master/test/shell as well.

purpleidea:
Over 80% of this is merged! w00t!

What remains is the automatic edges portion. 1/2 of which is here: https://github.com/purpleidea/mgmt/pull/270/files

kwilczynski:
Hi @purpleidea!

[...]

PR #660, opened 2021-05-09
engine: resources: exec: AutoEdge to User/Group/File

frebib:
Fixes https://github.com/purpleidea/mgmt/issues/221

This supersedes #270 and #536 

```
=== RUN   TestExecAutoEdge1
    exec_test.go:308: test: adding autoedges...
    exec_test.go:308: test: no auto edges were found for: group[somegroup]
    exec_test.go:308: test: autoedge: UIDS:
    exec_test.go:308: test: autoedge: UID0: exec[true]
    exec_test.go:308: test: autoedge: Match: file[/somefile] with UID: exec[true]
    exec_test.go:308: test: autoedge: Match: user[someuser] with UID: exec[true]
    exec_test.go:308: test: autoedge: Match: group[somegroup] with UID: exec[true]
    exec_test.go:308: test: autoedge: UIDS:
    exec_test.go:308: test: autoedge: UID0: exec[true]
    exec_test.go:308: test: autoedge: Match: user[someuser] with UID: exec[true]
    exec_test.go:308: test: autoedge: Match: group[somegroup] with UID: exec[true]
    exec_test.go:308: test: autoedge: Match: file[/somefile] with UID: exec[true]
    exec_test.go:308: test: autoedge: adding: file[/somefile] -> exec[true] (autoedge)
    exec_test.go:308: test: autoedge: UIDS:
    exec_test.go:308: test: autoedge: UID0: exec[true]
    exec_test.go:308: test: autoedge: Match: user[someuser] with UID: exec[true]
    exec_test.go:308: test: autoedge: adding: user[someuser] -> exec[true] (autoedge)
    exec_test.go:308: test: autoedge: UIDS:
    exec_test.go:308: test: autoedge: UID0: exec[true]
    exec_test.go:308: test: autoedge: Match: group[somegroup] with UID: exec[true]
    exec_test.go:308: test: autoedge: adding: group[somegroup] -> exec[true] (autoedge)
    exec_test.go:308: test: autoedge: UIDS:
    exec_test.go:308: test: autoedge: UID0: file[/somefile]
    exec_test.go:308: test: autoedge: Match: exec[true] with UID: file[/somefile]
    exec_test.go:308: test: autoedge: Match: user[someuser] with UID: file[/somefile]
    exec_test.go:308: test: autoedge: Match: group[somegroup] with UID: file[/somefile]
    exec_test.go:308: test: the auto edge list is empty for: user[someuser]
--- PASS: TestExecAutoEdge1 (0.00s)
PASS
ok  	github.com/purpleidea/mgmt/engine/resources	1.582s
```


engine/resources/user.go
```diff
@@ -411,6 +431,7 @@ func (obj *UserRes) UIDs() []engine.ResUID {
 	x := &UserUID{
 		BaseUID: engine.BaseUID{Name: obj.Name(), Kind: obj.Kind()},
 		name:    obj.Name(),
+		uid:     obj.UID,
```
frebib, line 434: It's not used in this PR but I imagine this will be desirable in the future.
I think it would be reasonable for any resource that takes a `User string` to attempt to `strconv.ParseInt` and treat that username as a uid if it's a valid UID. I suppose alternatively resources could add a User/UID `union` and only allow one or the other

misc/make-deps.sh
```diff
@@ -56,6 +56,7 @@ if [ -n "$YUM" ]; then
 	$sudo_command $YUM install -y graphviz || true # for debugging
 fi
 if [ -n "$APT" ]; then
+	$sudo_command $APT update -y
```
purpleidea, line 59: Fantastic! Can you wrap this in a `in_ci github` block, because we already have two other `apt update` runs in this file?

engine/resources/user.go
```diff
@@ -333,6 +333,26 @@ func (obj *UserRes) Cmp(r engine.Res) error {
...
+	uid  *uint32
+}
+
+// IFF aka if and only if they are equivalent, return true. If not, false.
+func (obj *UserUID) IFF(uid engine.ResUID) bool {
+	res, ok := uid.(*UserUID)
+	if !ok {
+		return false
+	}
+	if obj.uid != nil && res.uid != nil {
```
purpleidea, line 345: nit, but for consistency, the equivalent function in *GroupUID does name compare, and then the gui compare. Can you swap these two statements so they look otherwise closely equivalent please?

engine/resources/exec_test.go
```diff
@@ -257,3 +260,62 @@ func TestExecTimeoutBehaviour(t *testing.T) {
...
+	logf := func(format string, v ...interface{}) {
+		t.Logf("test: "+format, v...)
+	}
+	// run artificially without the entire engine
+	if err := autoedge.AutoEdge(g, debug, logf); err != nil {
+		t.Errorf("error running autoedges: %v", err)
+		return
+	}
+
+	// two edges should have been added
```
purpleidea, line 316: two or three?

engine/resources/exec_test.go
```diff
@@ -257,3 +260,62 @@ func TestExecTimeoutBehaviour(t *testing.T) {
...
+	r1 := &ExecRes{
+		Base: traits.Base{
+			Kinded: traits.Kinded{Xkind: "exec"},
+			Named:  traits.Named{Xname: "true"},
+		},
+		Cmd:   "/somefile",
+		User:  "someuser",
+		Group: "somegroup",
+	}
+	r2 := &UserRes{
```
purpleidea, line 281: Can you build r2, r3, and r4 with `NewNamedResource` instead?

engine/resources/exec_test.go
```diff
@@ -257,3 +260,62 @@ func TestExecTimeoutBehaviour(t *testing.T) {
...
+
+func TestExecAutoEdge1(t *testing.T) {
+
+	g, err := pgraph.NewGraph("TestGraph")
+	if err != nil {
+		t.Errorf("error creating graph: %v", err)
+		return
+	}
+
+	r1 := &ExecRes{
```
purpleidea, line 272: Can you build r2, r3, and r4 with `NewNamedResource` instead? You'll also have to convert it to the *ExecRes afterwards to be able to set these additional properties, but I think that will be cleaner if it works.

engine/resources/exec_test.go
```diff
@@ -257,3 +260,62 @@ func TestExecTimeoutBehaviour(t *testing.T) {
...
+	if err := autoedge.AutoEdge(g, debug, logf); err != nil {
+		t.Errorf("error running autoedges: %v", err)
+		return
+	}
+
+	// two edges should have been added
+	if i := g.NumEdges(); i != 3 {
+		t.Errorf("should have 3 edges instead of: %d", i)
+		return
+	}
```
purpleidea, line 320: Do you think you might be convinced to add some additional easy logic to check (1) that these are the edges we actually expect, and (2) that they're pointing in the correct directions? You should be able to see the pointers match what we expect. If you don't know how, I can write this addendum to the patch. Cheers

purpleidea (changes requested):
Thanks for the useful patches! I had one or two small things that I was hoping you wouldn't mind fixing up. Cheers

misc/make-deps.sh
frebib: We should run an `apt update` here every time regardless of where it runs. This file already suffers far too much from branching. I'll clean up the duplicate `apt update` calls instead

purpleidea:
FYI: I merged the first patch because I kept getting annoyed at failing tests and I wanted to push some tests tonight... So anything else you want to change, it's a delta on top of this now. Cheers

engine/resources/user.go
frebib: Swapped

frebib: GroupUID has gid check first, then name: https://github.com/purpleidea/mgmt/blob/master/engine/resources/group.go#L261
Want me to leave this, or swap both of them?

engine/resources/exec_test.go
frebib: Done

engine/resources/exec_test.go
frebib: seven.

purpleidea:
Rebased and merged! \o/
Thanks for the lovely tests!

PR #661, opened 2021-05-24
Add separator before revisions

viq:
Error message suggests doing so, and with this it seems the tests pass


viq:
Nevermind, that's wrong.

PR #662, opened 2021-05-24
Use more fields if they are there

viq:
It would work in case of `refs/heads/master` but in case of `refs/heads/feat/unify-functions` would return just `feat`, which doesn't match what we need. This should fix this.


viq:
Alternative instead of `cut` would be something like `awk '{ FS = "/" } ; { for (i=3; i<=NF; i++) print $i }'`

purpleidea:
This failed correctly.

PR #663, opened 2021-05-24
test: Fix awk usage for printing refs

viq:
This should print more fields than just 3rd one if they're present.

### DIFF
--- a/test/test-commit-message.sh
+++ b/test/test-commit-message.sh
@@ -130,7 +130,7 @@ then
 		ref=${GITHUB_BASE_REF}
 		head=${GITHUB_SHA}
 	else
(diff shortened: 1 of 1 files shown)


purpleidea:
I merged this, thanks!
I think it's an improvement-- it fails for bad PR's (#662) and passes with good PR's (this one)
However I checked out each commit as a feature branch: bug/viq-patch-broken and bug/viq-patch-notbroken and they both pass when actions runs them.
I've now pushed this to master to see what happens. I think it will probably pass too.
So an improvement, but there are still ghosts if you really want to hunt them down =D

purpleidea:
btw, welcome to the project! Sorry your first commit is perhaps less than super fun! Hopefully we can get you to work on some more fun ones!

PR #665, opened 2021-05-27
docs: Update IRC links to Libera.Chat

frebib:
#mgmtconfig has moved to Libera.Chat as the primary channel for IRC
communications. Update the documentation to reflect this.
Libera.Chat doesn't provide a first-party web portal but does recommend
a few in the linked documentation on the website. As there is no
suitable replacement for webchat.freenode.net, link to the "Choosing an
IRC client" page instead.


roidelapluie:
Note: https://web.libera.chat/ exists now.

PR #668, opened 2021-07-12
docs: Provide Libera.Chat webchat links over of irc://

frebib:
Only a few days after updating the documentation[1] following the move
to libera.chat, the webchat client was added, mirroring the behaviour of
the documentation prior to the change. Replace the ircs:// links with
clickable URLs to a usable browser chat client, which is more ideal for
beginners. Advanced users will know what to do to connect using their
external client as normal

[1]: 7d7e225823fd2920c425c9c9518f739a5989b541


frebib:
Apparently the commit message linter doesn't take kindly to `irc://`

purpleidea:
Great, thank you, merged!

PR #669, opened 2021-07-16
Set up golangci-lint linter based on gometalinter

ginywiny:
Test HIGHLY based on test-gometalinter, as it is used to lint the same files, except where golangci-lint is used instead of gometalinter. 

Note: When golangci-lint uses the --skip-files flag, it still lints the file, but simply doesn't display the linting results. Lastly, older version of the test was linting individual .go files instead of *.go, but this ran a lot slower. For this reason, *.go is used instead.

Comments and TODOs from the previous gometalinter test file were kept in case new linters are being used. 

- Update make deps script to install golangci-lint
- Create golangci-lint script based on test-gometalinter, create logic to entirely skip files without using  golangci-lint skip flag
- Update test script to run test-golangci-lint

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,3 +34,6 @@
 [submodule "vendor/github.com/containerd/containerd"]
 	path = vendor/github.com/containerd/containerd
 	url = https://github.com/purpleidea/containerd
+[submodule "vendor/github.com/hashicorp/consul"]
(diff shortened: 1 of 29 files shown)


purpleidea:
Looks like lots of tests are failing, and this is nine commits... I think you probably want to squash this down a bit =D
LMK if you're not sure why the tests are failing.

Cheers!

PR #670, opened 2021-08-07
Set up golangci-lint linter based on gometalinter

ginywiny:
_Creating a new PR because the previous one had a weird amount of strange commit history which went over my head to squash despite the many attempts taken._

Test HIGHLY based on test-gometalinter, as it is used to lint the same files, except where golangci-lint is used instead of gometalinter. 

**Notes:** 
1. When golangci-lint uses the --skip-files flag, it still lints the file, but simply doesn't display the linting results. 
2. **The build fails** because the revive linter has found issues with certain files. This is new as revive is a new linter being used to replace golint, as it runs more quickly. This will be fixed down the line. 
3. Comments and TODOs from the previous gometalinter test file were kept in case new linters are being used.

**What was done:**
1. Updated make deps script to install golangci-lint
2. Created golangci-lint script based on test-gometalinter
3. Updated test script to run test-golangci-lint


purpleidea:
I see:

>  ./test.sh: line 34: ./test/test-golangci-lint.sh: Permission denied

Did you u+x the file? Where do you see the list of lint errors we expect?

ginywiny:
@purpleidea Whoops, must've forgot to change that. Fixed it and squashed.

purpleidea:
Great! Please rebase to master and force-push and let's see what happens. (I pushed some fixes.)

.golangci.yml
```diff
@@ -0,0 +1,17 @@
...
+    - lang/types/kind_stringer.go
+    - lang/interpolate/parse.generated.go
+    - lang/interpolate/interpolate/parse.generated.go
+
+linters:
+  disable-all: true
+  enable:
+    - goimports
+    - revive # using revive instead of golint as it runs more quickly
+    - misspell
```
purpleidea, line 17: missing newline at the end of the file

misc/make-deps.sh
```diff
@@ -157,6 +157,9 @@ go get golang.org/x/lint/golint				# for `golint`-ing
 go get golang.org/x/tools/cmd/goimports		# for fmt
 go get github.com/kevinburke/go-bindata/go-bindata	# for compiling in non golang files
 go get github.com/dvyukov/go-fuzz/go-fuzz		# for fuzzing the mcl lang bits
+# heavily recommended to use curl to install instead of go get under https://golangci-lint.run/usage/install/
+curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.41.1
```
purpleidea, line 161: why is this version hardcoded? can we use the latest version? is there a safer way to do this so we don't pipe to shell? if not, please change this so it only runs in ci.

test.sh
```diff
@@ -67,6 +67,7 @@ if label-block "basic"; then
 	run-testsuite ./test/test-examples.sh
 	run-testsuite ./test/test-gotest.sh
 	run-testsuite ./test/test-gometalinter.sh
+	run-testsuite ./test/test-golangci-lint.sh
```
purpleidea, line 70: see above comment and potentially move to only run in ci.

test/test-golangci-lint.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# print current running test
+echo running "$0"
+
+command -v golangci-lint >/dev/null 2>&1 || { echo >&2 "golangci-lint not found"; exit 1; }
+
+# configure settings for test scripts
+ROOT=$(dirname "${BASH_SOURCE}")/..
+cd "${ROOT}" # Enter mgmt root
```
purpleidea, line 10: nit: comments are either full sentences and are capitalized and end with a period or aren't capitalized and don't end with a period.

test/test-golangci-lint.sh
```diff
@@ -0,0 +1,48 @@
...
+failures=''
+function run-test()
+{
+	$@ || failures=$( [ -n "$failures" ] && echo "$failures\\n$@" || echo "$@" )
+}
+
+# using .golangci.yml config file settings in ROOT
+gcl='golangci-lint run'
+
+# commented out from gometalinter linter test
```
purpleidea, line 22: are these five lines of comments valid?

test/test-golangci-lint.sh
```diff
@@ -0,0 +1,48 @@
...
+# aligncheck, dupl, errcheck, gas, goconst, gocyclo, gotype, unconvert
+
+# TODO: only a few fixes needed before using following linters:
+# deadcode, gosimple, ineffassign, interfacer, lll --line-length=200
+# safesql, staticcheck, structcheck, unparam, unused, varcheck
+
+for dir in `find * -maxdepth 9 -type d -not -path 'old/*' -not -path 'old' -not -path 'tmp/*' -not -path 'tmp' -not -path 'vendor/*' -not -path 'examples/*' -not -path 'test/*' -not -path 'interpolate/*'`; do
+
```
purpleidea, line 30: kill nl please

test/test-golangci-lint.sh
```diff
@@ -0,0 +1,48 @@
...
+done
+
+if [[ -n "$failures" ]]; then
+	echo 'FAIL'
+	echo 'The following tests have failed:'
+	echo -e "$failures"
+	echo
+	exit 1
+fi
+echo 'PASS'
```
purpleidea, line 48: missing nl at EOF

purpleidea (changes requested):
Patch looks good, just needs a few small changes.
Also check your $EDITOR, it's letting you save files without EOF newlines.

.golangci.yml
ginywiny: Will add.

test/test-golangci-lint.sh
ginywiny: Added these lines in case somebody were to move over any changes from gometalinter to golangci-lint. There were some linters which were entirely commented out, whereas some were still left as TODO, since they still require some fixes. More for clarity purposes, but they can be removed.

misc/make-deps.sh
ginywiny: What is unsafe about piping into shell?

test/test-golangci-lint.sh
ginywiny: Are they still valid?

misc/make-deps.sh
ginywiny: I have changed this to something else. Any thoughts of it now?

.golangci.yml
```diff
@@ -0,0 +1,17 @@
+# test-golangci-lint configuration file for golangci-lint
+
+run:
+  skip-files:
+    - lang/lexer.nn.go
+    - lang/y.go
+    - bindata/bindata.go
```
frebib, line 7: This is/will be removed in `https://github.com/purpleidea/mgmt/pull/671` so if that's merged first, this will need to be rebased and updated

misc/make-deps.sh
```diff
@@ -157,10 +157,35 @@ go get golang.org/x/lint/golint				# for `golint`-ing
...
+	# echo "THIS IS IN GOLANGCI-LINT!"
+	# ls $(go env GOPATH)/bin/golangci-lint
+	# 	# puts binary into $(go env GOPATH)/bin/golangci-lint
+	# 	# heavily recommended to install specific version of golangci-lint as per https://golangci-lint.run/usage/install/
+	# 	# used to avoid breaking builds when new linter errors are introduced from a new release (minor or major)
+	# sh -s -- -b $(go env GOPATH)/bin/golangci-lint $glc_version
+	# sh -s -- -b $(go env GOPATH)/bin/golangci-lint v1.42.0
+
+	# curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.42.0
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh > $(go env GOPATH)/bin/golangci-lint-install
```
frebib, line 182: I don't know what @purpleidea thinks about this, but it seems wise to run this check in Github Actions, at least when running in Github. Perhaps we can skip installation and running in GH and only do it in other CI systems? Then to provide the same functionality, drop in the relevant configuration into https://github.com/purpleidea/mgmt/blob/master/.github/workflows/test.yaml
https://golangci-lint.run/usage/install/#github-actions

frebib: To clarify (as explained on the linked page), this provides useful comments in PRs in context of where the issues are, instead of being hidden away in the actions tab

purpleidea:
@ginywiny Want to rebase to newest master and fixup?

ginywiny:
@purpleidea I will rebase and make a fix.

PR #673, opened 2021-09-28
lang: interpolate: Fix string interpolation of dollar symbols

frebib:
Dollar symbols were failing to parse when not followed by a non-brace,
non-dollar, non-EOF token and causing expected tests to fail. This
simplifies the rules to allow the remaining tests to succeed.

Fix and reinstate the final few failing tests, and add another.

Allow any escape sequence to be matched so that invalid sequences
produce a meaningful error message instead of a generic "cannot parse":

    ast: interpolate: interpolating: V: \?
    unhandled escape sequence token: \?

Tidy the related Makefile rule for generating the ragel parser.

### DIFF
--- a/lang/interpolate/parse.rl
+++ b/lang/interpolate/parse.rl
@@ -76,7 +76,7 @@ func Parse(data string) (out Stream, _ error) {
 		var = '${' var_name '}' ;
 
 		# Any special escape characters are matched here.
-		escaped_lit = '\\' ( 'a' | 'b' | 'f' | 'n' | 'r' | 't' | 'v' | '\\' | '"' | '$' )
+		escaped_lit = '\\' (any)
 		@{
 			switch s := data[fpc:fpc+1]; s {
 			case "a":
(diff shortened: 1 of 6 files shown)


lang/interpolate/parse.rl
```diff
@@ -76,7 +76,7 @@ func Parse(data string) (out Stream, _ error) {
 		var = '${' var_name '}' ;
 
 		# Any special escape characters are matched here.
-		escaped_lit = '\\' ( 'a' | 'b' | 'f' | 'n' | 'r' | 't' | 'v' | '\\' | '"' | '$' )
+		escaped_lit = '\\' (any)
```
purpleidea, line 79: don't we want to _not_ allow the non-registered escape characters?

lang/interpret_test/TestAstFunc2/escaping1/main.mcl
```diff
@@ -46,23 +46,17 @@ test "S: \\$ EOF" {}
...
-#test "Y: $${y1} EOF" {} # check there isn't double parsing
-# got: Y: $${y1} EOF
-# exp: Y: ${unused} EOF
-
-#test "Z: $$$" {}
-# got: <error>
-# exp: Z: $$$ EOF
+test "U: tab\t\\\ttabEOF" {}
+
+# test "V: " {}
```
purpleidea, line 51: Should this be commented?

purpleidea (commented):
Nice work! Two small comments. LMK when ready to merge.

lang/interpolate/parse.rl
purpleidea: and as it turns out a previous test i wrote caught this change and so this PR failed.

lang/interpret_test/TestAstFunc2/escaping1/main.mcl
frebib: Yeah, I couldn't come up with another test 😛 Decided to leave the empty case instead of defacing the alphabet

lang/interpolate/parse.rl
frebib: My justification is that allowing any char in here gives a better error message instead of a generic "sorry no can do". It's also a bit redundant to check the character twice.

frebib: Updated both the tests and the error message to be a little more human friendly

frebib:
Hopefully CI passes this time now. As long as you're happy with my comments I think this is good to go

purpleidea:
LGTM, great work!

purpleidea:
Great patch is now merged. Thanks again. @roidelapluie AFAICT we now have 0 interpolation bugs. I dare you to find one. Thanks for finding the first one =D

PR #674, opened 2021-09-29
lib: Move to go modules and shed a few tears

purpleidea:
The old system with vendor/ and git submodules worked great,
unfortunately FUD around git submodules seemed to scare people away and
golang moved to a go.mod system that adds a new lock file format instead
of using the built-in git version. It's now almost impossible to use
modern golang without this, so we've switched.

So much for the golang compatibility promise-- turns out it doesn't
apply to the useful parts that I actually care about like this.

Thanks to frebib for his incredibly valuable contributions to this
patch. This snide commit message is mine alone.

This patch also mixes in some changes due to legacy golang as we've also
bumped the minimum version to 1.16 in the docs and tests.


purpleidea:
I can't seem to make this work. Locally if I do this:

```
$ go-mod-upgrade 
   • Discovering modules...   
   ⨯ upgrade failed            error=Error running go command to discover modules: exit status 1 stderr=go list -m: loading module retractions for github.com/coreos/go-systemd@v0.0.0-20190321100706-95778dfbb74e: no matching versions for query "latest"
go list -m: loading module retractions for gopkg.in/fsnotify.v1@v1.4.7: version "v1.5.1" invalid: go.mod has non-....v1 module path "github.com/fsnotify/fsnotify" at revision v1.5.1
```

I get these scary errors. I've tried different things... Not sure how to fix this. It does build though. And the build errors here are different, but possible related??

 @frebib I tried your branch as well and I get the same error locally...

lib/cli.go
```diff
@@ -23,22 +23,35 @@ import (
...
 	// test for sanity
-	if program == "" || version == "" {
+	if cliArgs == nil {
+		return fmt.Errorf("this CLI was not run correctly")
+	}
+	if cliArgs.Program == "" || cliArgs.Version == "" {
 		return fmt.Errorf("program was not compiled correctly, see Makefile")
 	}
+	if cliArgs.Copying == "" {
+		return fmt.Errorf("program copyrights we're removed, can't run")
```
frebib, line 53: `s/we're/were/`

lib/cli.go
```diff
@@ -23,22 +23,35 @@ import (
...
 	_ "github.com/purpleidea/mgmt/langpuppet"
 	_ "github.com/purpleidea/mgmt/puppet"
 	_ "github.com/purpleidea/mgmt/yamlgraph"
 
-	"github.com/urfave/cli"
+	"github.com/urfave/cli/v2"
 )
 
+// CLIArgs is a struct of values that we pass to the main CLI function.
+type CLIArgs struct {
```
frebib, line 36: I noticed that we use `Cli` in `gapi`: https://github.com/purpleidea/mgmt/blob/master/gapi/gapi.go#L60
Should we change one or the other to be consistent?

frebib (commented):
@purpleidea I wonder if you have a stale cache or something. Pulling `github.com/fsnotify/fsnotify` in a blank new project works fine: https://gist.github.com/frebib/9944516ccbac6e55f593e6ed9a7d968b/raw

PR #679, opened 2021-11-01
Add systemd service file to release packages

dantefromhell:
According to [1] "/usr/lib/systemd/system/" is where systemd files from installed packages are stored in ArchLinux

Open question:
A] Does this change work for all distributions mgmt packages get build for?
B] Is the systemd unit file path for packages consistent across linux distros?

[1] https://wiki.archlinux.org/title/Systemd#Writing_unit_files


purpleidea:
@dantefromhell Hey, I'm missing more context about what's going on here and why you removed the --prefix flag. If you could add all that into the commit message (and then force push over this) it would help a lot. Thanks!

PR #681, opened 2021-11-11
New resource for hcloud instances - HetznerRes

JefMasereel:
First draft for a new resource supporting hcloud instances (Hetzner). The golang wrapper hcloud-go is used to interact with the Hetzner API. The resource is currently named HetznerRes, but this can be changed if needed. 

reference docs
- https://docs.hetzner.cloud
- https://pkg.go.dev/github.com/hetznercloud/hcloud-go/hcloud?utm_source=godoc

to be discussed
- [ ] general stylistic considerations? error messages, variable names, .. 
- [ ] extended functionality (server creation not yet written)
- [ ] aim for a standardized approach to cloud resources? I would like to (help) build resources for other providers in the future (eg. OVH via OpenStack API, using Gophercloud SDK). Happy to help write some documentation for this if needed, curious how you/others see this.


purpleidea:
(for anyone watching, we did a live review)
@JefMasereel ping when you're ready for round two
cheers!

frebib:
This is an issue with type conversion, where `TypeOf()` tries to recursively convert types in a depth-first fashion, but hits a loop causing the same path to be followed infinitely until the stack is full. In this case, `hcloud.Server` includes an `hcloud.Image`, which includes a `hcloud.Server`
https://github.com/hetznercloud/hcloud-go/blob/master/hcloud/server.go#L35
https://github.com/hetznercloud/hcloud-go/blob/master/hcloud/image.go#L25

JefMasereel:
This fork currently builds and runs without problems, although that still needs to be verified with formal testing. Manual use of the Hetzner resource seems to be stable. However, there are some caveats that might need to be addressed for applications at scale. Cfr code comments. 

Open for review, in the meantime I will start working on hetzner_test.go.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+package resources
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/hetznercloud/hcloud-go/hcloud"
```
purpleidea, line 27: Move this line at the bottom and separate it by itself. See: https://github.com/golang/go/wiki/CodeReviewComments#imports

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+const (
+	// HetznerPollLimit sets a lower limit on polling interval in seconds.
+	// Note that the hetzner API supports bursts at up to 3600 requests per hour.
+	// A higher polling interval is recommended when expecting high rates
+	// of operational http requests to the hetzner api.
+	HetznerPollLimit = 1
+	// WaitIntervalLimit sets a lower limit on wait intervals in seconds.
+	// Nonzero values disable burst requests to manage rate limit errors.
+	// A wait interval >= the polling interval is recommended.
+	WaitIntervalLimit = 0
```
purpleidea, line 42: These constants are in the same namespace as the other resources so adding a Hetzner prefix might be a good idea...

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+}
+
+func (obj *HetznerRes) Default() engine.Res {
+	return &HetznerRes{
+		WaitInterval:     WaitIntervalLimit,
+		ServerRescueMode: "disabled",
+	}
+}
+
+func (obj *HetznerRes) Validate() error {
```
purpleidea, line 133: Add the stock docstrings... See the other resources. Feel free to expand on or improve if necessary though.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+
+	// verify name (TODO: how to ensure unique names across server instances?)
```
purpleidea, line 135: What do you mean about this TODO?

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+
+	// check wait interval
+	if obj.WaitInterval < WaitIntervalLimit {
+		return fmt.Errorf("invalid wait interval (min = %d)", WaitIntervalLimit)
+	}
+
+	// verify state specifiers (only allow steady states)
+	switch obj.State {
+	case StateRunning, StateOff, StateAbsent:
+		// continue, state is valid.
```
purpleidea, line 165: Comments are either full sentences and are capitalized and end with proper punctuation, or are short comments that don't end in a period. Please correct all of these throughout. Larger guide here: https://github.com/purpleidea/mgmt/blob/master/docs/style-guide.md

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+
+// TODO: clear the root password info if rescue mode was enabled?
+func (obj *HetznerRes) Close() error {
+	obj.client = nil
+	obj.server = nil
+	return nil
+}
+
+func (obj *HetznerRes) Watch() error {
+	return fmt.Errorf("invalid Watch() call: Hetzner has no event API and requires polling")
```
purpleidea, line 211: If you want to do it this way and require that the user use the polling metaparam, you can also check if that's setup correctly in Validate()!

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+		return false, errwrap.Wrapf(err, "")
+	} else if !c {
+		checkOk = false
+	}
+	// skip property checks if not in the intended state
+	if !apply && !checkOk {
+		return false, nil
+	}
+	// Check if the active server requires updates
+	if obj.State == StateRunning {
```
purpleidea, line 232: unindent?

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+}
+
+// checkApplyServerState powers the server on/off, creates a new instance, or
+// deletes the existing one as needed, to ensure the correct state.
+func (obj *HetznerRes) checkApplyServerState(ctx context.Context, apply bool) (bool, error) {
+	if obj.init.Debug {
+		obj.init.Logf("checkApplyServerState(apply: %t)", apply)
+	}
+	var checkOk = true
+	if obj.server == nil {
```
purpleidea, line 257: You might be able to unindent a bunch here and improve readability but I didn't check for sure:

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+		return errwrap.Wrapf(err, "failed to collect Datacenter struct")
+	}
+	// TODO: add more flexible key selection
+	keylist, err := obj.client.SSHKey.All(ctx)
+	if err != nil {
+		return errwrap.Wrapf(err, "failed to collect SSHKey array")
+	}
+	// build serverconfig from given specs & defaults
+	obj.serverconfig = hcloud.ServerCreateOpts{
+		Name:             obj.Name(),        //string
```
purpleidea, line 579: if you want to leave comments with all the types, leave a space after the // so that nobody thinks they're magic strings and so it's easier to read.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,716 @@
...
+
+// waitForRescueMode waits until rescue mode is in the desired state (on/off).
+// It returns an error message if something fails, nil on healthy exit.
+func (obj *HetznerRes) waitForRescueMode(ctx context.Context) error {
+	if obj.init.Debug {
+		obj.init.Logf("waitForRescueMode()")
+	}
+	for {
+		select {
+		case <-ctx.Done():
```
purpleidea, line 685: the Done() case is typically the last one...

examples/lang/hetzner0.mcl
```diff
@@ -0,0 +1,12 @@
+# simple demo of the Hetzner resource
+hetzner "resourcetest" {
+	apitokenpath => "../mgmt-dev/apitoken.txt", # local file
+	servertype => "cx11",
+	datacenter => "hel1-dc2",
+	image => "debian-11",
+	state => "absent",
+	waitinterval => 5,
+	Meta:poll => 10,
```
purpleidea, line 9: It's new, but stylistically sensible to leave a newline and put the Meta: stuff at the bottom.

examples/lang/hetzner0.mcl
```diff
@@ -0,0 +1,12 @@
+# simple demo of the Hetzner resource
+hetzner "resourcetest" {
+	apitokenpath => "../mgmt-dev/apitoken.txt", # local file
```
purpleidea, line 3: You may want to consider instead using one of the mcl file reading functions, and instead have this struct API take a string... This way you can pass in a raw string _or_ a file, and you aren't forced to read from a file on disk... You can instead read from the deploy. Try using:

```
import "deploy"
deploy.readfile(...)
```

or
```
import "os"
os.readfile(...)
```

purpleidea (commented):
Very excited to see the progression, nice work!
I've kept my initial review more cursory to help kill off all the easy, distracting stylistic issues first. I'll follow up with a deeper structural review when all this is fixed and you're ready.
Some of the comments apply to other areas as well.
I didn't review the split across multiple commits. You can squash everything into a single commit if you'd like.
Looking fantastic, and LMK if you have any questions.

purpleidea:
(I've also triggered the test run-- I forget sometimes, so ping if you ever want it running again... It should catch some of the formatting/style issues for you automatically.)

engine/resources/hetzner.go
JefMasereel: I was not sure if I need to add explicit protection against name collisions (eg. two servers of the same name specified in a .mcl file under the same API token). As I understand it now, mgmt seems to do a good job already in catching such name collisions when generating the resource graph. Probably okay as is then?

examples/lang/hetzner0.mcl
JefMasereel: Thanks, I agree that this is a cleaner solution.

JefMasereel:
I made the requested changes. If everything is alright I will squash it all into a single commit, and then we should be ready to merge. Let's do one more structural review as you suggested?

engine/resources/hetzner.go
purpleidea: Correct-- mgmt will not allow to identical kind+name pairs. The hetzner resource will never see another hetzner resource with the exact same name. But it could see a different kind of resource with the same Name().

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+		}
+		checkOk = false
+		obj.init.Logf("requesting server creation")
+		if err := obj.createServer(ctx); err != nil {
+			return false, errwrap.Wrapf(err, "createServer failed")
+		}
+	}
+	// Make sure the server is in a steady state before further changes.
+	// TODO: test if hetzner api accepts queries while in transient state
+	if err := obj.waitForSteadyState(ctx); err != nil {
```
purpleidea, line 261: how long can this block for? is it guaranteed to eventually unblock? do we have a constant for a maximum wait time? or a param?

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+	// TODO: add permissions (poweron, poweroff, delete)
+	checkOk = false
+	if !apply {
+		return false, nil
+	}
+	switch obj.State {
+	case StateRunning:
+		obj.init.Logf("requesting server poweron")
+		if _, _, err := obj.client.Server.Poweron(ctx, obj.server); err != nil {
+			return false, errwrap.Wrapf(err, "client.server.poweron failed")
```
purpleidea, line 277: do we want `client.server.poweron`? This is for the end user, not a programmer. If it's for programmer wrap it in obj.init.Debug{} and obj.Logf

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+	if err := obj.getServerUpdate(ctx); err != nil {
+		return false, errwrap.Wrapf(err, "getServerUpdate failed")
+	}
+	if err := obj.waitForConvergedState(ctx); err != nil {
+		return false, errwrap.Wrapf(err, "waitForConvergedState failed")
+	}
+	return checkOk, nil
+}
+
+// checkApplyServerRebuild rebuilds the server instance as needed,
```
purpleidea, line 300: Can you expand in the docstrings somewhere what rebuild means. Eg: is it destroying the user data in that machine? If so, perhaps the user DOES NOT want that and so this should be guarded? For example, sometimes you can't transform a machine into a different type without deleting all the data... The virt resource does this, but *if* it needs to make a destructive change, then you also have to set a flag in the params. In that case "destructive" count as restarting. So data-at-rest loss is even worse. Restarting without expecting or being okay with it is also very bad. See:

```
	// RestartOnDiverge is the restart policy, and can be: `ignore`,
	// `ifneeded` or `error`.
	RestartOnDiverge string `lang:"restartondiverge" yaml:"restartondiverge"`
```

So we give the user the option to ignore, do it if needed, or error the resource because we weren't expecting this. Make sense?

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+// to ensure that rescue mode is available over SSH when needed.
+// TODO: add support for password authentication, now only ssh.
+// TODO: add checkapply for rescue mode specs when already enabled?
+// Right now you must fully disable and re-enable to update specs.
+func (obj *HetznerRes) checkApplyRescueMode(ctx context.Context, apply bool) (bool, error) {
+	if obj.init.Debug {
+		obj.init.Logf("checkApplyRescueMode(apply: %t)", apply)
+	}
+	// Check if rescue mode is already in the intended configuration.
+	var checkOk = true
```
purpleidea, line 337: checkOK and := as above

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+	if obj.server.RescueEnabled == obj.rescueEnable {
+		return true, nil
+	}
+	checkOk = false
+	if !apply {
+		return false, nil
+	}
+	// Otherwise, apply the necessary changes.
+	switch obj.rescueEnable {
```
purpleidea, line 346: why a switch?

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+		if key != res.ServerRescueSSHKeys[i] {
+			return fmt.Errorf("serverrescuekeys differs")
+		}
+	}
+	return nil
+}
+
+// cmpServerSpecs compares the server specifications between the virtual
+// HetznerRes resource and its corresponding Hetzner instance.
+// Returns true if Servertype, Datacenter and Image match.
```
purpleidea, line 433: Might also want to document why we'd get an error...

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+}
+
+// stateConverged checks if the target server is in the desired state.
+// Returns true if both values agree that the server is powered on or off,
+// or "absent" as intended. Returns false otherwise.
+func (obj HetznerRes) stateConverged() bool {
+	if obj.init.Debug {
+		obj.init.Logf("stateConverged()")
+	}
+	var statesMatch bool
```
purpleidea, line 460: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow
(or return flow)
basically programmers always want to avoid elseif when we can return early unless absolutely necessary

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+// a steady state ("running" or "off").
+func (obj *HetznerRes) waitForSteadyState(ctx context.Context) error {
+	if obj.init.Debug {
+		obj.init.Logf("waitForSteadyState()")
+	}
+	for {
+		if obj.server == nil {
+			return fmt.Errorf("server is unavailable")
+		}
+		if obj.server.Status == hcloud.ServerStatusRunning {
```
purpleidea, line 656: is it not possible for obj.server.Status to be in a third or fourth state? If so, do we want to know this? Perhaps it's something like "BuildingOnFire" ... so we'd want to not wait forever in that state... But we don't know it doesn't exist as a state yet? So perhaps we want a switch here, and a default: of "unknown state" and an error??

engine/util/util.go
```diff
@@ -276,6 +276,13 @@ func StructKindToFieldNameTypeMap(kind string) (map[string]*types.Type, error) {
 		//	continue
 		//}
 
+		// Skip unexported fields. These should never be mapped golang<->mcl.
```
purpleidea, line 279: I'll merge this commit from the open PR. Sorry for the delay.

purpleidea (commented):
Lots of great fixes since last review, thanks!

Please when you resubmit check all the tests have either passed or that if something failed, then it's not your fault. Like the .sh tests you can probably ignore, but all the golang and lint tests should pass.

One general point of review in case it is immediately not obvious: suppose I make a comment about some issue on line 42, while I might make that comment specific to only that line, the goal is for it to be a general comment overall and if it makes sense, you might want to also make sure you don't need to change the same kind of pattern elsewhere in the code. I've left a few of such instances in this review. So just keep in mind, when reading a comment, think "does this comment also apply elsewhere in the code?"

I think this is very close, but a few things we still need to tackle.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+	// State is the desired steady state of the server instance.
+	// Choose between "running", "off" or "absent".
+	State string `lang:"state"`
+
+	// WaitInterval is the sleep interval in seconds that is used when waiting
+	// for transient states to converge between intermediate operations.
+	// Note that the hetzner API supports bursts at up to 3600 requests per hour.
+	// Nonzero values disable request bursts, a zero value will allow it.
+	// A wait interval >= 1s is recommended for extended use.
+	WaitInterval uint32 `lang:"hetznerwaitinterval"`
```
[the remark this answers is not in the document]
JefMasereel: Ah okay, thanks for clarifying my misunderstanding. Will fix it.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+
+	// ServerRescueMode specifies the image used when enabling rescue mode.
```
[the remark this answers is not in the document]
JefMasereel: Ah yes, agreed. Much cleaner to make proper use of the default zero values.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+// represents one server instance, even if its client has access to multiple
+// instances. A resource in the `absent` state only exists as a local
+// blueprint, and does not exist as server instance on Hetzner's side.
+// Note: for now, all manually registered SSH keys are used when creating a
+// server. More selective use to be added soon.
+type HetznerRes struct {
+	traits.Base
+	init *engine.Init
+
+	// APITokenPath is the path of a file containing the project's API token.
```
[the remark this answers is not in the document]
JefMasereel: woops, good catch :)

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+	// Provide the full path, incl. filetype, and keep the token private.
+	// Consider where the hcloud CLI stores tokens in default operation.
+	APIToken string `lang:"apitoken"`
+	// ServerType determines how much ram, disk and cpu the resource has.
+	Servertype string `lang:"servertype"`
```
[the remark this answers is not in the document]
JefMasereel: Good points. To reply to them all in order:
- ServerType instead of Servertype is fine for me, it makes more sense to write it that way. 
- Hetzner does use the "datacenter" spelling in their API, so I think it is most consistent to stick with that?
- list of server options is tricky, see below

The full list of options is only available by querying it from the API. I think this was done so that Hetzner doesn't have to maintain their static documentation when changes are made. I can put the current list of options in the comments somewhere, or as a set of constants, but they can become outdated over time so they will require maintenance. 

Alternatively, I could add an API query to Init() after creating the client, which validates the server options? I wasn't sure if API interactions within Init() are supported, but it's probably the cleanest way of adding a sanity check on those inputs. It doesn't provide the user with any starting info though, unless we add a list of valid options to the error message if an invalid option is specified. 

A more creative solution could be to build a little golang project in an independent repo that requests all server options and dumps them into a file for manual lookups. I'd be happy to host that on my github, and then people can use copies as needed. Then we can just add a link to that repo in the resource documentation for reference. It's not ideal, but at least it solves the problem of people having to write their own scripts to obtain the full options. 

Personally I tend towards the last option since it doesn't overcomplicate the Hetzner resource itself, and offers a little open source contribution for hcloud-specific projects independently of mgmt. Let me know what you prefer.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+}
+
+// Init runs some startup code for this resource.
+func (obj *HetznerRes) Init(init *engine.Init) error {
+
+	// save init struct
+	obj.init = init
+
+	// initialize hcloud client
+	token := obj.APIToken
```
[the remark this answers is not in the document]
JefMasereel: Because I was getting tired probably ;) 
You're right though, it's unnecessary. Fixed it.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+	obj.client = hcloud.NewClient(hcloud.WithToken(token))
+
+	// convert ServerRescueMode to rescueEnable flag
+	obj.rescueEnable = (obj.ServerRescueMode != ServerRescueDisabled)
+
+	return nil
+}
+
+// Close runs some cleanup code for this resource.
+// TODO: clear the root password info if rescue mode was enabled?
```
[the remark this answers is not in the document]
JefMasereel: I will clarify that in the code comments. Right now I am not even reading the root password when enabling rescue mode, so that you can only connect to the server over SSH. The ssh key implementation seems to work as intended, so there is no urgent security threat at the moment. If there is ever a desire to add password support though, I already added those TODO's in the relevant places to remind me (or other devs) what to pay attention to.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,710 @@
...
+func (obj *HetznerRes) Close() error {
+	obj.client = nil
+	obj.server = nil
+	return nil
+}
+
+// Watch is not implemented for this resource, since the Hetzner API does not
+// provide any event streams. Instead, always use polling.
```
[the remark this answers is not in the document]
JefMasereel: That makes sense, thanks for the explanation. I will log a warning like that in Validate after making sure the interval values are valid at that time, let me know if I misunderstood something.

engine/resources/hetzner.go
JefMasereel: Agreed, this needs something to exit if it keeps running for too long. It doesn't have a limit yet, but a maximum wait time sounds reasonable. I will add a param with a default value of 10 times the wait interval, that seems reasonable. Let me know if you have counterproposals.

engine/resources/hetzner.go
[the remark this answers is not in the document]
purpleidea: Let's keep it simple for now, but at minimum docstring should have a URL that tells the reader where the list of current values can be found (some link to the hetzner website probably.)

> Hetzner does use the "datacenter" spelling in their API, so I think it is most consistent to stick with that?
Yup, use what they use.

engine/resources/hetzner.go
JefMasereel: Agreed. I like the three options you describe here, but I'm not sure if/how to extend that to other important permissions like creating and deleting servers. Do you consider that to be part of the RestartOnDiverge behavior? In that case I would rename it to something like "AllowDestructiveChanges". Thoughts? 

Creating new servers isn't really destructive, but it can be expensive if abused... Best to add a second permission field? This one seems less urgent though, it's harder to make big mistakes with server creation than deletion.

I also have TODOs for permissions related to powering on/of and enabling/disabling rescue mode. These are all non-destructive changes though, so probably fine to leave as is? Wanted to check now that we're on the topic. 

The most flexible solution I can see is adding a specific private permission for each different operation, and then cluster them at Init() so that you can adjust the coverage of public permissions in one place. If you feel like one or two permission params are enough then it would probably be easier to just hardcode those more plainly.

engine/resources/hetzner.go
JefMasereel: The if-else seemed less readable, but it's no hard preference. Changed it again to separate if statements.

engine/resources/hetzner.go
JefMasereel: I fixed the corresponding TODO to throw an error in case of unknown keynames, rather than silently ignoring them. Seems like the safer way to handle those exceptions, and now the code looks much cleaner too.

engine/resources/hetzner.go
JefMasereel: Hmm, this made me realize that I'm not explicitly looking at any of the transient states, even though some of them could be early indicators of trouble... Converting the current functionality here to a switch seems mostly a matter of style preference, but I agree that it could provide a cleaner framework to catch exceptions if any bugs ever need to be addressed. 

Will make the simple transition to a switch case for now, but I need to think a bit longer about those transient states.

engine/resources/hetzner.go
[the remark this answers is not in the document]
JefMasereel: Agreed with keeping things simple, but afaik there is no website that lists the current options. Besides my proposed solutions, all I can do is refer users to the API docs and have them collect valid options on their own using the CLI (hcloud) or the API  wrappers for golang (hcloud-go) or python (hcloud-python). 

https://docs.hetzner.cloud

engine/resources/hetzner.go
JefMasereel: Added error cases for the `unknown` and `deleting` states, since those really shouldn't happen there. 
Beyond that there shouldn't be many issues, it just needs to wait until the transients converge into a valid steady state.

engine/util/util.go
JefMasereel: No problem. I'll try to remember not squashing it into the final commit along with the other stuff.

engine/resources/hetzner.go
JefMasereel: Just to be sure, your comment is about the relevance of the Logf message for the end user? Or about the error? 

For the former case, I feel like it would be relevant to know when my instances are powering on/off.  I gave those messages the same logging relevance/level as server creation and deletion, because they all affect the operational availability of your server instance. But I find it a bit difficult to draw the line sometimes between standard logging and debug logging, so please correct me if you don't agree.

For the latter case, I rewrote the error messages to "failed to power server on" etc to make them a bit more readable.

JefMasereel:
I pushed some more updates, but I haven't fully addressed some of the open comments yet. To be continued:
- [ ] add reference to server options documentation for hcloud
- [ ] rename Servertype
- [ ] add wait limit
- [ ] add RestartOnDiverge
- [ ] clarify Logf vs Debug (cfr. client.server.poweron remark)

engine/resources/hetzner.go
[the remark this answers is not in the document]
JefMasereel: Doublechecked the Hetzner docs, and ServerType seems like the correct option. Changed it in the next commit.

engine/resources/hetzner.go
purpleidea: I think you probably just want to have a const at the very top with some super long, but not overly long time... 10 min perhaps?? Basically something that should never happen unless hetzner is broken... This way, we'll eventually make progress even if the mgmt user waits 10 min, they'll at least get an error... Blocking forever is bad. Make sense?

purpleidea (commented):
> add RestartOnDiverge protection

You probably don't want to name it this exactly... As long as you understand the point of it and you do the correct thing here to prevent a change in machine type by the end user in mcl, causing their data to get nuked if they didn't ask for that. Get it?

> clarify Logf vs Debug (cfr. client.server.poweron remark)

Use your judgment... Basically regular logf the minimum because it will be noisy, but if it's something useful wrap it in Debug for the programmer.

HTH, lmk if I missed anything else.

engine/resources/hetzner.go
JefMasereel: The default is set to five minutes now, should be alright.

engine/resources/hetzner.go
```diff
@@ -0,0 +1,790 @@
...
+	conditionSatisfied := false
+	timeout := time.After(time.Duration(obj.WaitTimeout) * time.Second)
+	for {
+		// Get up-to-date serverinfo.
+		if err := obj.getServerUpdate(ctx); err != nil {
+			return errwrap.Wrapf(err, "failed serverupdate request")
+		}
+		// Check if the appropriate exit condition is satisfied.
+		switch exitCondition {
+		case "convergence":
```
frebib, line 697: Can you use constants instead of these magic strings?

examples/lang/hetzner0.mcl
```diff
@@ -0,0 +1,40 @@
...
+$f = "../mgmt-dev/apitoken.txt"
+$token = os.readfile($f)
+
+# alternatively, read from the deploy:
+# $token = deploy.readfile(...)
+
+# or give the string directly:
+# $token = "..."
+
+hetzner "resourcedemo" {
```
frebib, line 14: Would it make sense to use a namespaced name for this resource, like `hetzner:vm` (or whatever nomenclature Hetzner uses for their instances) in case you/someone wants to add other Hetzner resources in the future, to say add SSH keys or other things that Hetzner provide

purpleidea:
Yes please =D

engine/resources/hetzner.go
JefMasereel: Yes of course. I was a bit hesitant to do it this way because the constants seemed out of place as global values for the resource. Alternatively, I could pass the exitCondition as functions, maybe that's even cleaner? Then it's immediately clear which exit condition is being used by the waiter, without having that intermediate string.

examples/lang/hetzner0.mcl
JefMasereel: Agreed, having something like hetzner:vm or hetzner:server would be great. We've already been thinking about a dedidated hetzner resource for ssh keys, although I haven't gotten to that implementation yet. 

What would be the appropriate naming structure for mcl lang then? I'm not clear yet on how that is generated.

engine/resources/hetzner.go
JefMasereel: I tried the functional approach, and personally I really like it. Let me know what you think. A wait statement now looks something like `err := obj.WaitFor(ctx, serverInSteadyState)` where `serverInSteadyState` is a function that checks whether the server is in steady state. I added a `HetznerServerCondition` type to make sure any functions of that conditional structure can be used as needed.

examples/lang/hetzner0.mcl
JefMasereel: Done. I named it hetzner:server since the hcloud-go package uses server terminology rather than VMs, but we can change it if you prefer something else. The filename is now `hetzner_server.go` and the resource is `HetznerServerRes`.

engine/resources/hetzner_server.go
```diff
@@ -0,0 +1,804 @@
...
+	// ServerRescueTypeLinux32 specifies the image to be used in rescue mode.
+	ServerRescueTypeLinux32 = "linux32"
+	// ServerRescueTypeLinux64 specifies the image to be used in rescue mode.
+	ServerRescueTypeLinux64 = "linux64"
+	// ServerRescueTypeFreeBSD64 specifies the image to be used in rescue mode.
+	ServerRescueTypeFreeBSD64 = "freebsd64"
+)
+
+func init() {
+	engine.RegisterResource("hetzner:server", func() engine.Res { return &HetznerServerRes{} })
```
dantefromhell, line 65: I would propose using a more specific resource name, maybe
"hetzner:cloud:vm" (personal preference) or "hetzner:vm"

Hetzner operates 2 lines of product lines: physical servers and virtual machines plus some surrounding services for each. Both product lines have their own API (hcloud [1] for VMs and robot [2] for physical machines) which AFAIK are custom-made.

My concern is that a too generic name for this resource might conflict with the implementation of future features, therefore my suggestion to change the resource name to something more specific.

@purpleidea 
I assume that naming of MCL resource types and their attributes should be static over mgmts lifetime to prevent breaking changes in the language. Is that a fair assumption?

[1] https://docs.hetzner.cloud/
[2] https://robot.your-server.de/doc/webservice/en.html

purpleidea: >  "hetznercloudvm" (personal preference) or "hetzner:vm"
heh, the latter is fine.

>  I assume that naming of MCL resource types and their attributes should be static over mgmts lifetime to prevent breaking changes in the language. Is that a fair assumption?

This is fair! We'll try and keep everything static once merged, although I urge anyone to recommend any changes or renames of either resource names, param names, or any behaviours before 1.0 or perhaps 0.1 as that will be when we provide an API guarantee.

examples/lang/hetzner0.mcl
purpleidea: See other comment from @dantefromhell ... hetzner:vm sounds sensible, but so does hetzner:server ... As long as you've addressed the comment about what the physical server variant would be called (just suggest a name so we know our current choice is hopefully future proof) then I'm fine with that.

Remember the goals in order:

1) It should make the most sense to _our_ users.
2) It should make sense to the hetzner API authors and internal hetzner folks. We'd like to keep things consistent in their world and use their naming conventions, but if that would be in conflict with what makes sense here, do what's best for mgmt users.

engine/resources/hetzner_server.go
JefMasereel: If we really have to maintain consistency across the mgmt language then I think it's also important to consider the same naming problem for other cloud providers that might be added in the future. If we want to add eg. OVH support at some point, maybe it's cleaner to turn the names around and work with server:hcloud, server:robot, server:ovh...?

This way you define the function first, and then the name of (one of) the supported provider(s). Thoughts? I do realize that this implies that the API token for example should be called token:hcloud, and in that case the token specifier might not be super clear to the user...

purpleidea:
>
> If we really have to maintain consistency across the mgmt language then I
> think it's also important to consider the same naming problem for other
> cloud providers that might be added in the future. If we want to add eg.
> OVH support at some point, maybe it's cleaner to turn the names around and
> work with server:hcloud, server:robot, server:ovh...?
>
> This way you define the function first, and then the name of (one of) the
> supported provider(s). Thoughts? I do realize that this implies that the
> API token for example should be called token:hcloud, and in that case the
> token specifier might not be super clear to the user...
>

I disagree, but I've reflected upon the point. If it makes sense to try to
unify various different resources, then someone can always make a cloud:vm
resource that automatically chooses one or something.

For now it's best IMO that each resource model the thing it's trying to as
faithfully as possible to that upstream as long as this model makes sense
for our users.

HTH

For Message ID: ***@***.***>

>

JefMasereel:
Made some changes based on the received feedback and squashed everything into a single commit. I still have a shortlist of potential additions or changes, but nothing that can't be added as later features. What is in the commit now seems to be ready for use, but please let me know if I missed anything.

JefMasereel:
> Made some changes based on the received feedback and squashed everything into a single commit. I still have a shortlist of potential additions or changes, but nothing that can't be added as later features. What is in the commit now seems to be ready for use, but please let me know if I missed anything.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+// serverconfig in hcloud-go format from resource params, requests a server
+// creation with that configuration, and waits until the creation is confirmed.
+// Errors occur when the server exists already, the client fails, or the wait
+// step exits early (context cancellation, client failure or timeout).
+// TODO: add createServer permission field
+func (obj *HetznerVMRes) createServer(ctx context.Context) error {
+	if obj.init.Debug {
+		obj.init.Logf("createServer()")
+	}
+	if obj.server != nil {
```
purpleidea, line 540: are you sure obj.server is always up-to-date before createServer, deleteServer, and so on are called? Do we have a calling convention to ensure this? Should this happen in this function? Does it happen just once at resource start? What if we poll for ten minutes... Maybe things changed by then? LMK if this is handled correctly somewhere or not.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+	if err := obj.waitFor(ctx, serverAbsent); err != nil {
+		return errwrap.Wrapf(err, "waitFor(serverAbsent) exited early")
+	}
+	return nil
+}
+
+// rebuildServer deletes the current server instance and creates a new one, in
+// accordance with the provided resource specifications.
+// NOTE: this is a destructive operation, don't call without AllowDeletion.
```
purpleidea, line 579: Then why aren't we checking AllowDeletion in _here_ ??

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+// getServerConfig converts resource fields to hcloud.ServerCreateOpts struct,
+// such that a server instance created with those Opts has the desired specs.
+func (obj *HetznerVMRes) getServerConfig(ctx context.Context) error {
+	if obj.init.Debug {
+		obj.init.Logf("getServerConfig()")
+	}
+	// default, volumes not supported (yet)
+	automount := false
+	// poweron at creation to reach `running` faster
+	startAfterCreate := (obj.State == StateRunning)
```
purpleidea, line 605: set a default and then:
xxx := default
`if obj.State == whatever {
  xxx = blah
}

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+	// build serverconfig from given specs & defaults
+	obj.serverconfig = hcloud.ServerCreateOpts{
+		Name:             obj.Name(),        // string
+		ServerType:       serverType,        // *ServerType
+		Image:            image,             // *Image
+		SSHKeys:          keylist,           // []*SSHKey
+		Location:         nil,               // *Location
+		Datacenter:       datacenter,        // *Datacenter
+		UserData:         "",                // string
```
purpleidea, line 632: This might be the correct way to kick-off mgmt in the new machine... Look at the other userdata in https://github.com/purpleidea/mgmt/blob/master/engine/resources/aws_ec2.go#L166
I think we might want to have this in the main struct or is it something else?

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	return nil
+}
+
+// getServerConfig converts resource fields to hcloud.ServerCreateOpts struct,
+// such that a server instance created with those Opts has the desired specs.
+func (obj *HetznerVMRes) getServerConfig(ctx context.Context) error {
+	if obj.init.Debug {
+		obj.init.Logf("getServerConfig()")
+	}
+	// default, volumes not supported (yet)
```
purpleidea, line 602: np, add `TODO: ` prefix to the comment then

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+			return nil
+		}
+		// Retry every WaitInterval until the exit condition is satisfied.
+		// Can exit early by timeout, context cancellation or an error.
+		select {
+		case <-time.After(time.Duration(obj.WaitInterval) * time.Second):
+			// retry confirmation
+		case <-timeout:
+			return fmt.Errorf("timeout: exit condition not confirmed after %d seconds", obj.WaitTimeout)
+		case <-ctx.Done():
```
purpleidea, line 709: incorrect: See

https://pkg.go.dev/context#Context

You want to either use or wrap the Err() method!

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	} else if obj.server.RescueEnabled == obj.rescueEnable {
+		rmOK = true
+	}
+	if obj.init.Debug {
+		obj.init.Logf("rescueModeOK() = %t", rmOK)
+	}
+	return rmOK, err
+}
+
+func serverExists(obj *HetznerVMRes) (exists bool, err error) {
```
purpleidea, line 786: These should be private methods on the struct, not take the struct as input.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	var err error = nil
+	conditionSatisfied := false
+	timeout := time.After(time.Duration(obj.WaitTimeout) * time.Second)
+	for {
+		// Get up-to-date serverinfo.
+		if err := obj.getServerUpdate(ctx); err != nil {
+			return errwrap.Wrapf(err, "failed serverupdate request")
+		}
+		// Check if the provided exit condition is satisfied.
+		conditionSatisfied, err = exitFunc(obj)
```
purpleidea, line 695: see my comment on serverExists... You don't need to pass in obj here.

Also, why not := ? maybe isOK or something shorter instead?

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	if obj.server != nil {
+		return fmt.Errorf("server already exists")
+	}
+	if err := obj.getServerConfig(ctx); err != nil {
+		return errwrap.Wrapf(err, "getServerConfig failed")
+	}
+	if _, _, err := obj.client.Server.Create(ctx, obj.serverconfig); err != nil {
+		return errwrap.Wrapf(err, "client.server.create failed")
+	}
+	if err := obj.waitFor(ctx, serverExists); err != nil {
```
purpleidea, line 549: this should be obj.serverExists instead of course... And similarly for the other function args to waitFor.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+	var err error = nil
+	conditionSatisfied := false
+	timeout := time.After(time.Duration(obj.WaitTimeout) * time.Second)
+	for {
+		// Get up-to-date serverinfo.
+		if err := obj.getServerUpdate(ctx); err != nil {
+			return errwrap.Wrapf(err, "failed serverupdate request")
+		}
+		// Check if the provided exit condition is satisfied.
```
purpleidea, line 694: You should add a comment here saying:

"The exitFunc does not take a content because it is guaranteed to never block or run for very long."

If this is not true, then of course it should take a context! And functions that can't use the context because they're very short and don't block should just ignore the using of that context of course!

Get it?

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	obj.rescueKeys = keylist
+	return nil
+}
+
+// waitFor provides a general function that returns nil when the provided exit
+// function is satisfied. It retries every WaitInterval until the condition is
+// satisfied. It can exit early in case the WaitTimeout is reached, the context
+// is cancelled or an error occurs. The exit function must be of type
+// HetznerVMCondition, and check a well-defined condition for the resource.
+func (obj *HetznerVMRes) waitFor(ctx context.Context, exitFunc HetznerVMCondition) error {
```
purpleidea, line 682: I'm not sure `exitFunc` is the correct name. Can you change or convince me please?

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	if obj.init.Debug {
+		obj.init.Logf("waitFor()")
+	}
+	var err error = nil
```
purpleidea, line 686: why are `err` and `conditionSatisfied` even defined here at all???

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+			return fmt.Errorf("wait interrupted by context")
+		}
+	}
+}
+
+// HetznerVMCondition provides a HetznerVM-specific type to define functions
+// that check a logical condition on a given HetznerVMRes. Return true if the
+// condition is satisfied, false otherwise. An error must be returned if the
+// condition's input requirements (if any) are not met.
+type HetznerVMCondition func(*HetznerVMRes) (bool, error)
```
purpleidea, line 719: Not sure this needs a type. You can nuke this type entirely unless there's a reason not to. It only exists in one place...

purpleidea (commented):
So very excellent to see this coming together. A few general comments:

You should add a paragraph or so in the commit message itself. After the online, you leave one newline, and then type the paragraph. This is useful for general commentary about the commit and so on. I'd strongly recommend mentioning the fact that this uses polling and not events for watch. I'd also mention this in the main struct for the resource.

Also commit message should say `hetzner:vm` instead of `hetzner_vm`.

One architectural question I'd like you to answer:
Support I took the exact same mcl code for a hetzner:vm and the same token, and ran it on two different computer more or less simultaneously... Would there be an unexpected or undesirable effects? Are there any API races which would be dangerous? If no, fantastic. State this in the struct resource description. If yes and they're fixable, please fix. If yes and they're not easily fixable, please state your expectations clearly in the resource struct.

Think of it this way: if you sell a power drill to someone, you tell them how to hold it. "put the pointy, spinny bit towards the wood, not your face". The big description at the top of the hetzner struct is that instruction manual. Be clear about what your expectations are for the user.

Lastly, a general comment that will hopefully help you personally as a programmer-- after reading a bunch of your code, here is some advice:

I get the feeling that you're not very thorough when you make changes and write code. For example, every time you write or make a small change, you have to constantly ask yourself: "if I do this, what else will the effect, and what will I have to also change if I change this" (and so on recursively... This is the annoying hard work of being a programmer. But that is one of the algorithms you should run in your brain. It takes time and patience, but if you practice this more, you will go from "good hacker" to "elegant code sculptor". I hope that is useful. It is meant as constructive feedback!

Thank you again for all your hard work. I've left a bunch of review comments. Please take the time to address them all-- remember, don't make any changes blindly. Push back if you think I'm wrong about something, and explain why.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	// project. Keep this token private! It provides full access to this
+	// project, so a leaked token will be vulnerable to abuse. Read it from
+	// a local file or the mgmt deploy, or provide it directly as a string.
+	APIToken string `lang:"apitoken"`
+
+	// ServerType determines the machine type as defined by Hetzner.
+	ServerType string `lang:"servertype"`
+	// Datacenter determines where the resource is hosted.
+	Datacenter string `lang:"datacenter"`
+	// Image determines the operating system to be installed.
```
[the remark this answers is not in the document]
JefMasereel: I proposed some solutions for this in an earlier discussion. There is no always-up-to-date list of options avalable online. What you can do is query a list of options from a live Hetzner project over the API. I don't have a complete list myself at the moment, but I will build a small golang project on my github to query all the opts and dump them into a local file in readable format. 

Would you then prefer a written list of options per variable in the code comments, or a link to the repo where I publish those options? In the latter case, developers might have an easier time pinging me for an update, or using that code to query an updated list themselves.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	WaitInterval uint32 `lang:"waitinterval"`
+
+	// WaitTimeout will cancel wait loops if they do not exit cleanly before
+	// the expected time in seconds, in order to detect defective loops and
+	// avoid unnecessary consumption of resources.
+	WaitTimeout uint32 `lang:"waittimeout"`
+
+	// ServerRescueMode specifies the image type used when enabling rescue mode.
+	// The supported image types are `linux32`, `linux64` and `freebsd64`.
+	// Alternatively, use `disabled` to disable rescue mode (default).
```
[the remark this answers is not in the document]
JefMasereel: It is less explicit, but it would work as well. Will do it your way, that is indeed more in line with the resource guide.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+	// warning logs to inform user of interval limits
+	obj.init.Logf("WARNING: Meta:poll must always be greater or equal than %d seconds", HetznerPollLimit)
+	obj.init.Logf("WARNING: waitinterval must always be greater or equal than %d seconds", HetznerWaitIntervalLimit)
+
+	// validate state param
+	switch obj.State {
+	case StateRunning, StateOff, StateAbsent:
+		// valid options
+	case "":
```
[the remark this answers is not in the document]
JefMasereel: This is new information to me, was this specified earlier or somewhere in the developer guides? Seems important to explicitly mention this somewhere. A little surprised that we didn't catch this earlier, but I understand why you would want to implement that behavior. Is it safe though, given the declarative style? 

I'm not sure how you would respond to a change in params like the Datacenter for example. Do you rebuild a vm in a different physical location and set it to the same unknown state as it was in before? Or do you set it to some default state? I feel like your proposed "let it be" behavior breaks some of that transitional logic, but please let me know if I'm missing something.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+
+// Init runs some startup code for this resource: initialize hcloud-go client,
+// and then build some internal flags from the given public fields.
+func (obj *HetznerVMRes) Init(init *engine.Init) error {
+
+	// save init struct
+	obj.init = init
+
+	// initialize hcloud-go client
+	obj.client = hcloud.NewClient(hcloud.WithToken(obj.APIToken))
```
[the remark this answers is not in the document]
JefMasereel: I'm adding those options, but I don't know how to access that `*lib.Main` object you refer to. From what I can tell it is related to the CLI operations, while I'm only interacting with the `*HetznerVMRes`. So far I can't really tell how to read those values, any tips would be appreciated. Otherwise I'll get back to it later.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	if err != nil {
+		return checkOK, errwrap.Wrapf(err, "confirmStateConvergence failed")
+	}
+	if stateConverged {
+		return checkOK, nil
+	}
+
+	// Request changes if the server is not in the intended state yet.
+	// TODO: add permissions? (poweron, poweroff, delete)
+	checkOK = false
```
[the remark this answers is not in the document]
JefMasereel: Yes, but looking back at the code I agree that it might need a clearer rewrite. Maybe a more specific flag like `serverExistsAsIntended` would improve readability? Will take a moment to consider the options. 

As for logic: the checkOk returned on line 316 is 
* `true` if the server did not require creation in the first check, and is already in the desired state (running/off). 
* `false` if the server was absent and required creation, or if it still needs to be pushed to the desired state (running/off).

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	// Make sure the server exists as intended before further checks.
+	// TODO: add create permission
+	if obj.server == nil {
+		if obj.State == StateAbsent {
+			return true, nil
+		}
+		if !apply {
+			return false, nil
+		}
+		checkOK = false
```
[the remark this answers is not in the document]
JefMasereel: See my reply to your comment on line 321. I'm adding some simple code comments to clarify the logic as well.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+			obj.init.Logf("requesting server poweroff")
+		}
+		if _, _, err := obj.client.Server.Poweroff(ctx, obj.server); err != nil {
+			return false, errwrap.Wrapf(err, "failed to power server off")
+		}
+	case StateAbsent:
+		switch obj.AllowDeletion {
+		case "error":
+			return checkOK, fmt.Errorf("server deletion not allowed")
+		case "ignore":
```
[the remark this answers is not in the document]
JefMasereel: Agreed, not the most user-friendly option. See my reply to your comment on line 124.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+		if err := obj.rebuildServer(ctx); err != nil {
+			return false, errwrap.Wrapf(err, "rebuildServer failed")
+		}
+	}
+	return checkOK, nil
+}
+
+// checkApplyRescueMode enables/disables rescue mode for the given keyset, to
+// ensure that rescue mode is available over SSH when needed.
+// NOTE: changing image type currently requires a manual re-enable.
```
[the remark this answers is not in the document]
JefMasereel: By "manual re-enable" I mean that you have to set the rescuemode param to `""` (disabled), run that mcl file, and then set a different image type in that param and run it again. That is currently required when you have enabled rescue mode with some image type, and want to switch to a different image type. 

I skipped that TODO earlier because in most use cases it wouldn't be an issue (in favor of other priorities), but I agree that it shouldn't become normal to ask the user to toggle any params manually to reset this property. 

It is probably best to split `checkApplyRescueMode` into two steps, something like `checkApplyRescueModeEnable` and `checkApplyRescueModeImage`. Adds another bit of complexity, but it's very similar to other checkApply steps so I don't expect too many issues.

examples/lang/hetzner_vm0.mcl
```diff
@@ -0,0 +1,42 @@
...
+$f = "../mgmt-dev/apitoken.txt"
+$token = os.readfile($f)
+
+# alternatively, read from the deploy:
+# $token = deploy.readfile(...)
+
+# or give the string directly:
+# $token = "..."
+
+hetzner:server "resourcedemo" {
```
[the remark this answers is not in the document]
JefMasereel: I eventually decided on hetzner:vm, although I'm still not fully convinced that one is better than the other. In terms of future-proofing the naming convention, I think that it might be better to use something like `hetzner:hcloud:server`. It names the provider, the service, and the resource type. But I also understand that this becomes a bit bulky. 

What do you think now that you've had some time to think about it? Maybe it's easier to settle this in a quick call?

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	}
+	if (obj == nil) != (r == nil) {
+		return fmt.Errorf("one resource is empty")
+	}
+	// compare types
+	res, ok := r.(*HetznerVMRes)
+	if !ok {
+		return fmt.Errorf("not a %s", obj.Kind())
+	}
+	// compare resource fields
```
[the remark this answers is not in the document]
JefMasereel: I didn't include `waitinterval`, `waittimeout` and `allowdeletion`, all other public params are there. I did it this way because these three only affect the interaction with the API and not the actual server instance, and the definition of Cmp in the resource guide seems to allow omissions in that case. 

Now I'm thinking it's safer to minimize exceptions and just include everything to avoid surprises for any future developments. If I understand correctly this affects whether or not two resources can be merged in a graph? I was unsure about that earlier, but if that's the case then it's probably important to include `AllowDeletion`, and the two time params would prevent some unexpected timing issues.  

I'm also not comparing any of the private params, since those are generated (and some can be updated) at or after initialization. Will keep those out of Cmp, but please let me know if you disagree.

engine/resources/hetzner_vm.go
JefMasereel: I looked into this earlier, and I believe it is implemented correctly, but I agree that it might be good to have a more consistent design pattern/rule for it, or at least add some documentation on my current design somewhere. As it is now, CheckApply opens with a serverupdate request, and then requests a new update any time a change is made. Those updates happen inside the functions that push those changes though (after the operation, before returning, to validate the outcome). It's not super clear when reading through CheckApply, but it's there.

What would you prefer to make this more clear? Not sure where to write this kind of information. For now I will just add a note to the CheckApply description.

engine/resources/hetzner_vm.go
JefMasereel: Agreed, will move those protections into these functions directly.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	// default, volumes not supported (yet)
+	automount := false
+	// poweron at creation to reach `running` faster
+	startAfterCreate := (obj.State == StateRunning)
+	// build serverconfig elements
+	serverType, _, err := obj.client.ServerType.GetByName(ctx, obj.ServerType)
+	if err != nil {
+		return errwrap.Wrapf(err, "failed to collect ServerType struct")
+	}
+	image, _, err := obj.client.Image.GetByName(ctx, obj.Image)
```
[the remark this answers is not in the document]
JefMasereel: Thanks for catching this! Somehow I missed that, will fix it of course. Wasn't addressed for these parameters, even though I use that behavior to check if the server has been deleted correctly...

engine/resources/hetzner_vm.go
JefMasereel: If we just need to accept a string param to make this work it's no problem to add that feature. I'm not familiar with how to use it, but from the aws docs it does look more useful than I thought. Thanks for pointing this out.

engine/resources/hetzner_vm.go
JefMasereel: Agreed, it's a bit rough. I'm considering to rename the whole thing to `waitUntil(ctx, checkCondition)`, with `checkCondition` replacing `exitFunc`. That way it matches the internal `conditionSatisfied` flag better, and it is more in line with general golang style. Counterproposals are welcome :)

engine/resources/hetzner_vm.go
JefMasereel: Agreed, will remove that.

engine/resources/hetzner_vm.go
JefMasereel: Oh wow, didn't know that. Will change it in all cases. Thanks!

engine/resources/hetzner_vm.go
JefMasereel: I've been using this type to define an input type for waitFor, to avoid misuse. Can this be done as safely without defining a type? I was having some trouble with these kinds of golang-specific constructs, please let me know if I'm missing something. Will keep the type for now, but will look for a better solution when I make the requested changes to waitFor.

engine/resources/hetzner_vm.go
JefMasereel: Agreed and understood. Still need to adjust those little checkCondition functions, but will keep it in mind.

engine/resources/hetzner_vm.go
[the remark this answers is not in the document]
JefMasereel: Tadaa! Seems like a decent solution for now. https://github.com/JefMasereel/hcloud-go-getopts.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+	ServerRescueMode string `lang:"serverrescuemode"`
+
+	// ServerRescueSSHKeys can be used to select a subset of keys that should be
+	// enabled for rescue mode operations over SSH. From all SSH keys known to
+	// the project client, choose a subset of keys by name, as an array of
+	// strings. New keys must first be added manually via the cloud console.
+	// An error is thrown if a given keyname is not recognized by the client.
+	ServerRescueSSHKeys []string `lang:"serverrescuekeys"`
+
+	// AllowDeletion is a server protection policy to prevent unwanted deletion
```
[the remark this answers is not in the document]
purpleidea: discussed in jitsi. tri state.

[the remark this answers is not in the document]
JefMasereel: Conclusion:
* Deletion is always allowed when the user explicitly sets `state => "absent"`.
* Deletion for implicit reasons (e.g. rebuilding to switch ServerType) must be protected.
  * `""` throws an error when a deletion is required but the state is not explicitly set to absent.
  * `"ifneeded"` allows implicit deletions if that is required by checkapply to reach all the specified resource fields.
  * `"ignore"` blocks any implicit deletions that might be requested by checkapply, but without error. Continue running.

The same protection should be implemented for poweroffs, to prevent unexpected reboots etc.

engine/resources/hetzner_vm.go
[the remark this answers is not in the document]
JefMasereel: Fixed in commit `0ab2406db9b4d5dac2348dd033d5fd37abd92a94`. Use `obj.init.Program` and `obj.init.Version`.

engine/resources/hetzner_vm.go
```diff
@@ -0,0 +1,804 @@
...
+		if key != res.ServerRescueSSHKeys[i] {
+			return fmt.Errorf("serverrescuekeys differs")
+		}
+	}
+	return nil
+}
+
+// cmpServerSpecs compares the server specifications between the virtual
+// HetznerVMRes resource and its corresponding Hetzner instance. Returns true if
+// ServerType, Datacenter and Image match. Returns an error is the server is
```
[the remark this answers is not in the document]
JefMasereel: if the server is...

engine/resources/hetzner_vm.go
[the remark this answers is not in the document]
JefMasereel: I added some logic to fix this issue, but I'll review it again later.

engine/resources/hetzner_vm.go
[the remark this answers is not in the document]
JefMasereel: withEndpoint() uses `endpoint = "https://api.hetzner.cloud/v1"` by default, and I haven't found any information about valid alternatives yet. Not sure when it would be relevant to specify this as a resource param. Ideas?

withDebugWriter is currently added as TODO, will try to implement that in a later PR.

JefMasereel:
Pushed the new updates, I think I covered all your feedback. Please let me know if I missed anything. Playing around with the mcl demo it seems to function as intended, but some automated testing would be helpful to make sure all cases are covered. Hopefully this provides a good enough basis to merge now and then start adding features over time?

purpleidea:
Please excuse my horrible delay in doing a final review and merge here! It's been great reading through all of this, and it's obviously much better to get this in and people hacking on it more than to keep it out. Thank you again and looking forward to seeing more hetzner patches in the future!

purpleidea:
(I rebased and merged in b26f842de19a31c36fb5149f292b9117dbf4bed7)

purpleidea:
\o/

Issue #684, opened 2021-11-25
type unification of hcloud type(s) in HetznerRes


JefMasereel:
## Versions:

* mgmt version: 0.0.22-19-gf7271de-dirty (built from PR#681)

* operating system/distribution: 5.15.4-arch1-1 x86_64 GNU/Linux

* golang version: go1.17.3 linux/amd64

## Description:

The mgmt binary was built from PR #681 on 23/11/2021, without any alterations from the official PR at that time. I wrote `hetznertest.mcl` for a first test, which is provided in the latest commit. 

While testing the new hetzner resource with a small .mcl file, mgmt runs into a stack overflow issue. I discussed this with @purpleidea, and the problem appears to be an infinite recursion during type unification, starting from the `HetznerRes.serverconfig` field. The error message is shown below. 

The `serverconfig` field is of type `*hcloud.ServerCreateOpts`, which is a rather convoluted struct containing all the information the hcloud client requires to create a new server via the cloud API. For now, it can be removed from the private fields in order to fix this issue, but it would be good to look into a stable solution eventually. 

Hetzner-related documentation can be found at
- https://docs.hetzner.cloud
- https://github.com/hetznercloud/hcloud-go
- https://pkg.go.dev/github.com/hetznercloud/hcloud-go/hcloud#Server

```
$ ./mgmt run lang examples/lang/hetznertest.mcl 
2021/11/25 10:07:38 cli: lang: lexing/parsing...
2021/11/25 10:07:38 cli: lang: init...
2021/11/25 10:07:38 cli: lang: interpolating...
2021/11/25 10:07:38 cli: lang: building scope...
2021/11/25 10:07:38 cli: lang: running type unification...
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc081000388 stack=[0xc081000000, 0xc0a1000000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x1d7892a, 0x2fd5b80})
	/usr/lib/go/src/runtime/panic.go:1198 +0x71
runtime.newstack()
	/usr/lib/go/src/runtime/stack.go:1088 +0x5ac
runtime.morestack()
	/usr/lib/go/src/runtime/asm_amd64.s:461 +0x8b

goroutine 1 [running]:
runtime.heapBitsSetType(0xc101d8d270, 0x50, 0x48, 0x1c853a0)
[...]

PR #685, opened 2021-12-01
engine: util: Skip unexported fields in struct field mapping

frebib:
Unexported fields should be ignored when mapping structs from Golang to
mcl, as this avoids issues where certain structs cannot be used in mcl
representation due to their structure that may be incompatible with
conversion. Disallowing private fields allows those to still be included
in the struct but ignored when mapping the types, allowing the mcl
representation of the type to stil be valid

Resolves: https://github.com/purpleidea/mgmt/issues/684

### DIFF
--- a/engine/util/util.go
+++ b/engine/util/util.go
@@ -276,6 +276,13 @@ func StructKindToFieldNameTypeMap(kind string) (map[string]*types.Type, error) {
 		//	continue
 		//}
 
+		// Skip unexported fields. These should never be mapped golang<->mcl.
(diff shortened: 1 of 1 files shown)


frebib:
So this is apparently a 1.17 feature 🤔

purpleidea:
Looks beautiful, thanks. How much do I have to pay to add a simple test? =D

purpleidea:
Sorry for the long delay in merging. If someone would like to write a test for this situation, it would certainly be appreciated.

PR #688, opened 2022-07-28
lang: funcs: add ExecFunc

gelisam:
Runs a string as a shell command, then produces each line from stdout.


gelisam:
This is a draft PR because the implementation is still a work-in-progress: I currently hardcode the command to be executed instead of taking it from the input stream.

gelisam:
@purpleidea, how does this look like so far? Here are the things I currently plan to improve, anything else?

[ ] TODO: Use the input as the command.
[ ] Don't store the Context in the struct, as per https://github.com/golang/go/wiki/CodeReviewComments#contexts
[ ] Do not use the name MCL in the documentation, as per https://github.com/gelisam/mgmt/blob/77346527f3c13c646f11d85780f9b38a3ce9bc98/docs/style-guide.md#product-identifiers
[ ] Make it clear when/weather goroutines exit if it's not obvious, as per https://github.com/golang/go/wiki/CodeReviewComments#goroutine-lifetimes

gelisam:
As discussed offline, when the input changes, it should emit all the lines from the old input's command until EOF, and only then start emitting the lines for the new input's command.

gelisam:
It was difficult to understand why `seq 10` was working but `ping -c 13 google.ca` was not, and it turns out the stderr message was useful in figuring out why, so I'll also log stderr in order to help users debug their commands.

gelisam:
@purpleidea this PR is now ready to review!

gelisam:
CI says that the following tests are failing:

1.  `./test/test-headerfmt.sh`  
    The problem there is that the copyright notice I copied from `lang/funcs/core/world/kvlookup_func.go` was only going up to 2021, but it should be going up to 2022. I have now fixed this in `lang/funcs/core/os/exec_func.go`, but the problem remains in `lang/funcs/core/world/kvlookup_func.go` and many other files and I don't think this PR is the right place to fix all of those other files.
2.  `./test/test-commit-message.sh`  
    This one was tricky to investigate. Running
    
        ./test/test-commit-message.sh
    
    prints `PASS` on my machine, yet fails on CI. Why? It turns out CI sets some environment variables, `GITHUB_SHA`, `GITHUB_REF`, and `GITHUB_BASE_REF`, and that if those variables aren't set the test silently passes instead of complaining that they are not set. Thankfully, there is a comment in the script listing those three variables and giving an example of what they should be set to. Unfortunately, that example is wrong; it claims it should be `GITHUB_REF=refs/heads/exec-func` but that doesn't work, it must be `GITHUB_REF=exec-func`. Here is the correct incantation for this PR:
    
    GITHUB_SHA=ef63f74f306cd90c2c40718bdb5bff327ac9f4b2 GITHUB_REF=exec-func GITHUB_BASE_REF=master ./test/test-commit-message.sh
    
    After some more fussing trying to figure out which part of that complex regex was failing, I eventually figured out that it's not sufficient for my commit message to begin with `lang: funcs: ...`, the `...` also has to start with a capital letter.
3.  `./test/test-govet.sh`
    This one complains that my comment "is not reflowed properly", but doesn't tell me how to fix it. My comment is already wrapped to 80 characters, which is what the comments in `./test/comment_parser.go` around that error message talk about, so I have no idea how I'm supposed to "reflow" my comment in order to make it match the test.

gelisam:
Turns out I was wrapping at 79 characters instead of 80.

lang/funcs/core/os/exec_func.go
```diff
@@ -0,0 +1,164 @@
...
+// Init runs some startup code for this function.
+func (obj *ExecFunc) Init(init *interfaces.Init) error {
+	obj.init = init
+	obj.ctx, obj.cancel = context.WithCancel(context.Background())
+	return nil
+}
+
+// Stream returns the changing values that this func has over time.
+func (obj *ExecFunc) Stream() error {
+	var wg sync.WaitGroup
```
purpleidea, line 86: Sneaky, but you should actually do:

`wg := &sync.WaitGroup{}`

If you look at the docs, I think it suggests this way is safer so that we can copy this thing around. But perhaps I've misunderstood things. All I know is using the aforementioned form seems to have always worked better.

lang/funcs/core/os/exec_func.go
```diff
@@ -0,0 +1,164 @@
...
+			// own or or due to obj.cancel().
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				cmd.Wait()
+			}()
+
+			// Log the lines from stderr, to help the user debug.
+			// Terminates when the process exits, on its own or or
```
purpleidea, line 128: or or

lang/funcs/core/os/exec_func.go
```diff
@@ -0,0 +1,164 @@
...
+			// due to obj.cancel().
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				stderrScanner := bufio.NewScanner(stderrReader)
+				for stderrScanner.Scan() {
+					obj.init.Logf(
```
purpleidea, line 136: It's okay to put all of this on one line unless you prefer it like this. Docstrings and comments usually wrap at 80 chars, but it's okay to let code lines that are sensible to run a bit longer.

main.go
```diff
@@ -1,5 +1,5 @@
 // Mgmt
-// Copyright (C) 2013-2021+ James Shubin and the project contributors
+// Copyright (C) 2013-2022+ James Shubin and the project contributors
```
purpleidea, line 2: fixed everywhere in git master, rebasing will drop this

lang/funcs/core/os/exec_func.go
```diff
@@ -0,0 +1,164 @@
...
+	// Kill the current process if it's still running.
+	defer obj.cancel()
+
+	for {
+		select {
+		case input, more := <-obj.init.Input:
+			if !more {
+				return nil
+			}
+			shellCommand := input.Struct()["shell_command"].Str()
```
purpleidea, line 101: If the input command changes (remember, input strings are a stream) we would probably want to stop exec-ing the previously running command and switch to the new one... Don't we want to kill that old one and then reset everything before running a new one here?

lang/funcs/core/os/exec_func.go
```diff
@@ -0,0 +1,164 @@
...
+						"ExecFunc: \"%v\": stderr: %v\n",
+						shellCommand,
+						stderrScanner.Text(),
+					)
+				}
+			}()
+
+			// Emit one value downstream for each line from stdout.
+			stdoutScanner := bufio.NewScanner(stdoutReader)
+			for stdoutScanner.Scan() {
```
purpleidea, line 146: I'm assuming `stdoutScanner` and `stderrScanner` both close when the command exits right? Thus breaking us out of this infinite loop. If you've just once added a printf in there saying "this loop exited" or something, then that's perfect for me, and I would recommend leaving a short comment mentioning that the loop closes when the command exits, so future readers know it gets cleaned up early.

purpleidea (commented):
A truly gorgeously written patch. Thank you for your contribution! \o/
I commented on a few things that I think we need to fix. Feel free (and I encourage) you to push back on any comments you disagree with!
Cheers!

lang/funcs/core/os/exec_func.go
```diff
@@ -0,0 +1,164 @@
...
+		case input, more := <-obj.init.Input:
+			if !more {
+				return nil
+			}
+			shellCommand := input.Struct()["shell_command"].Str()
+
+			// Run the command, connecting it to ctx so we can kill
+			// it if needed, and to two Readers so we can read its
+			// stdout and stderr.
+			cmd := exec.CommandContext(obj.ctx, "sh", "-c", shellCommand)
```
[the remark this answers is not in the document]
gelisam: What's wrong with using the shell? Are you concerned about security or performance?

I used the `sh` in order to make sure I am splitting the algorithm the same way the shell does. In particular, the linked code splits on whitespace, which is much simpler than what the shell does; on the input `foo "bar baz" 'hello world'`, the linked code produces `foo`, `"bar`, `baz"`, `'hello`, and `world'`, while the shell would produce `foo`, `bar baz`, and `hello world`.

Using the shell also provides more functionality to the user. For example, if you have a command like `bc` which expects its input from stdin rather than the command line, the with the shell the user can still use `os.exec(fmt.printf("echo %d %% 10 | bc", datetime.now()))` to control `bc`'s stdin.

In most programming languages, there are typically at least two functions for executing outside processes. This derives from C, where there is both `system` and the `execv` family of functions. High level languages often prefer to spawn a separate process than to replace the current process, as a result the second function is often called something other than "exec". But, like in C, languages typically provide one function with a simple API consisting of a single string input, often called "system", but sometimes also "shell" or "command", and a more comprehensive function which takes a list of strings, doesn't delegate to the shell, and allows the user to control other things such as a map of environment variables, a string to use for stdin, and an option to control what to do with stderr and with the exit code.

Since it would make sense to implement the simple function before the complex function, how about keeping the current shell-based implementation and renaming `exec` to "system" or "shell", thus keeping the name "exec" available for a future, more complex version?

lang/funcs/core/os/exec_func.go
gelisam: I do know that the loop exists, not because of a printf but because this loop needs to terminate in order for the loop which encloses it to run its next iteration, which is needed in order to handle the next command in the input stream, and `os.exec(fmt.printf("echo %d %% 10 | bc", datetime.now()))` works as expected.

That being said, I tested this loop and the stderr one immediately above just to be sure, and they both terminate. I then added a comment to this loop.

lang/funcs/core/os/exec_func.go
gelisam: Darn, what a misunderstanding!

So, this comment made me quite upset, enough that I took a break from contributing for a bit. Now that I am coming back to it, I think I understand what really happened.

I was originally planning to use the implementation you suggest in this comment, namely to kill the process when the input changes. I had told you about this plan early on, when it was still easy to change course, and I thought you had told me to pick a different course, namely to wait until the executable terminates on its own before starting the next one. I implemented what I thought you had asked, and was thus upset by the request to change it back, since I wouldn't have had to do this extra work if you let me to implement it that way to begin with.

I now understand that you were not asking me to change _my_ code so that I would send all the lines of stdout downstream, you were commenting on the fact that MCL doesn't currently send all the lines to the resources. That is, I wrote this:

```golang
// ExecFunc runs a string as a shell command, then produces each line from
// stdout. If the input string changes, the process is killed and the lines
// from the new process are produced.
//
// Note that in the likely case in which the process emits several
// lines one after the other, it is normal for some of the downstream nodes not
// to receive some of those lines. This is because MCL is not designed to
// process data streams, but rather to compute the desired state from the most
// recent values. Thus, the last line produced by the process will definitely
// be seen by the downstream nodes, and a line just before the process pauses
// is very likely to be seen as well.
```

In my mind the two paragraphs were tied: since it is only the most recent line which is guaranteed to be seen by the downstream nodes, if we know we will spawn a new process then there is no point continuing to stream the stdout of the previous process, as those lines are now known not to be the most recent value. You commented about the behaviour of MCL in the second paragraph, and I incorrectly assumed that you didn't want me to kill the old process. But I now see that you never said that and that I thus have no reason to be upset.

All this to say: I will update the code to kill the process.

lang/funcs/core/os/exec_func.go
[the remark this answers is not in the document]
purpleidea: > Since it would make sense to implement the simple function before the complex function, how about keeping the current shell-based implementation and renaming exec to "system" or "shell", thus keeping the name "exec" available for a future, more complex version?

It's a really good idea. As it happens the exec resource `exec` can easily have flags, and it has the `Shell` flags to let you specify a shell. https://github.com/purpleidea/mgmt/blob/3cea4223657cdbb32fcb0afc54bb736388dfbd5e/engine/resources/exec.go#L64
But assuming you are talking about this function, then yes, we can rename it as you suggest to `system` I think for the shell version.

lang/funcs/core/os/exec_func.go
purpleidea: As is usual in the world, communication is always the most challenging part!

For any future things: if you disagree with whatever I'm saying (maybe I'm just phrasing it poorly and we actually agree, or perhaps I am thinking about things the wrong way) I urge you to definitely push back, as the goal is _the right way_, not necessarily _my way_. HTH and thank you for picking it back up =D

examples/lang/exec_func.mcl
```diff
@@ -0,0 +1,70 @@
...
+#	[10 seconds pause]
+#	9
+#	[10 seconds pause]
+#	0
+#	[10 seconds pause]
+#	1
+#	[10 seconds pause]
+#	2
+#	[...]
+$n = os.exec(fmt.printf("echo \"(%d / 10) %% 10\" | bc", datetime.now()))
```
purpleidea, line 21: Well this is brilliant, and also fun.

lang/funcs/core/os/exec_func.go
```diff
@@ -0,0 +1,198 @@
...
+			// Log the lines from stderr, to help the user debug.
+			// Terminates when the process exits, on its own or
+			// due to cancel().
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				stderrScanner := bufio.NewScanner(stderrReader)
+				for stderrScanner.Scan() {
+					obj.init.Logf("ExecFunc: \"%v\": stderr: %v\n", shellCommand, stderrScanner.Text())
```
purpleidea, line 175: nit: consider changing the prefix to `exec` or (if we do the rename) to `system` to match what the user would expect it be called?

purpleidea (commented):
Gorgeous patch, ready to merge, modulo the rename to `system` if you want to do that first.
I also added a new commit in `master` to bump to golang 1.17 which means if you rebase our tests should hopefully be less broken.

Thank you again!

purpleidea:
Your glorious patch has been renamed to use `system` instead of `exec` as per your permission. I also added a second patch to fix a `go vet` issue which was breaking the build. Sorry the builds were rotted a bit previously. I've now fixed all of these and the basic tests should be stable now. Cheers!

Issue #610, opened 2020-09-22
Package for Nixos


nixinator:
Nixos and Mgmt Config might be combination for the following reasons (there are many more).

* Keeping secrets out of the /nix/store

* Coordinating and running changes to /etc/nixos/configuraton.nix (and import files). (on a git commit)

* Setting up tests infrastructure

* Integrating with Nixops

purpleidea:
Happy to have a package, but no resources available to build one atm.

nixinator:
I'm working on it :-).  It may not be possible currently because the way dependencies are built into the project. Nixos likes to build from golang projects from source (it is a build system after all). It can handle go modules, with buildGoModule. 

There a few other 'nixos' tools to gather dependencies , mgmt seems to use a combination of 'go get' , 'go submodules', which may make it a little more difficult to get it compile in a nix enviroment. I may need to pick your brains on this build process (but not much). 

The main goal is to get all the source dependencies into the source tree before nix builds them, as once the build has start no network access is allowed (to help with a deterministic build :-) ).

dep2nix converts a Gopkgs.lock file into a deps.nix file which is understood by nixpkgs's go abstraction thanks to go2nix effort.

go2nix is best suited for Go apps written before dep or modules were introduced.

vgo2nix Convert go.mod files to nixpkgs buildGoPackage compatible deps.nix files.

It's a matter of unpicking you build system, and see if any of these can help. I'll have a hack to see where i can get.

nixos declarative configuration +  mgmt config, might be an interesting combination.

purpleidea:
If this would make your life easier (and ours too perhaps)

We like building everything from git master to keep things fresh for now, but for packaging it's an issue. So I wanted to extend (via shell script I suppose) a "tag" script that saves a lock file with the currently used dep versions. So that it can be restored if you want to build a specific tag.

You can use the standard golang module lock files if you'd like, they just need to get committed and stored in a subdir and not the root.

If you can contribute that, it would be appreciated! I think it would then make building your nix package much easier. Ping if you'd like more details.

nixinator:
Ping!!! :-). Well if there is a coherent way to make the build more 'nix' friendly, then it's worth exploring. If it builds in nix/nixos, then would you be willing to add a default.nix and shell.nix (and maybe even a flake.nix) to the root of the project?? 

Nix is a bit alien when you first see it but....as they say in The Hitch Hikers Guide to the Galaxy... DON'T PANIC!.

[...]

PR #694, opened 2022-11-07
add flake for easier build support

urandom2:
This change adds support for nix flake based builds. This includes a dev shell required for running make targets like code generation or testing, and a package target for the output mgmt binary. We have also .gitignored the default output location, result.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,58 @@
+{
+  description = "Next generation distributed, event-driven, parallel config management!";
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:urandom2/nixpkgs/mgmt";
+  };
+  outputs = {
+    flake-utils,
+    nixpkgs,
+    self,
+  }:
+    flake-utils.lib.eachDefaultSystem (system: let
+      buildInputs = with pkgs; [
+        augeas
+        libvirt
+        libxml2
+      ];
+      nativeBuildInputs = with pkgs; [
+        gotools
+        nex
+        pkg-config
+        ragel
+      ];
+      pkgs = nixpkgs.legacyPackages.${system};
+    in {
+      devShells.default = pkgs.mkShell {
+        inherit buildInputs nativeBuildInputs;
+        packages = [pkgs.go];
+      };
+      packages.default = let
+        pname = "mgmt";
+        version = "0.0.22";
+      in
+        pkgs.buildGoModule {
+          inherit pname version buildInputs nativeBuildInputs;
+          ldflags = [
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,58 @@
...
+      ];
+      pkgs = nixpkgs.legacyPackages.${system};
+    in {
+      devShells.default = pkgs.mkShell {
+        inherit buildInputs nativeBuildInputs;
+        packages = [pkgs.go];
+      };
+      packages.default = let
+        pname = "mgmt";
+        version = "0.0.22";
```
urandom2, line 32: unfortunately, flakes cannot consume version strings from vcs, so you will be required to manually update the string before releases, or just leave the flake as something like `unstable-20221107155054` like so:
```suggestion
        version = "unstable-${self.lastModifiedDate}";
```

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+        gotools
+        nex
+        pkg-config
+        ragel
+      ];
+      pkgs = nixpkgs.legacyPackages.${system};
+    in {
+      devShells.default = pkgs.mkShell {
+        inherit buildInputs nativeBuildInputs;
+        packages = [pkgs.go];
```
urandom2, line 28: if there is anything I missed that is useful for testing/debugging/development, we should add it here

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+    nixpkgs,
+    self,
+  }:
+    flake-utils.lib.eachDefaultSystem (system: let
+      buildInputs = with pkgs; [
+        augeas
+        libvirt
+        libxml2
+      ];
+      nativeBuildInputs = with pkgs; [
```
urandom2, line 18: these are dependencies that are called during the build process explicitly

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:urandom2/nixpkgs/mgmt";
+  };
+  outputs = {
+    flake-utils,
+    nixpkgs,
+    self,
+  }:
+    flake-utils.lib.eachDefaultSystem (system: let
+      buildInputs = with pkgs; [
```
urandom2, line 13: these are dependencies that are linked against, think shared libraries

urandom2 (commented):
no pressure to merge or review this; just got it put together based upon nixos/nixpkgs#198963

per #610, suggestions for repo level changes are inline, but can be added to this change or broken out as a separate pr; I would suggest integrating `nix build` into your build process at least to verify that this logic does not break and ossify, but it is not required, and just let me know if you want a hand

PR #695, opened 2022-11-21
lang: Remove Expr::SetValue()

gelisam:
The method is used by the function engine to store the latest value which was emitted by `Func::Stream()`, but this is redundant because the engine also stores that value in the engine's own `table[vertex]`.

Moreover, most of the `SetValue()` implementations have a comment indicating that the function is a no-op, or is never called and is only implemented for consistency.

`Expr::Value()` is still useful because it is used during unification. For example, `printf("%d", x)` can tell the type checker that `x` must be an `int`, but only if the format argument's value is statically known. So it is still useful to have a method for asking for that value if it is known.

### DIFF
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -4770,18 +4770,6 @@ func (obj *ExprBool) Func() (interfaces.Func, error) {
 	}, nil
 }
 
-// SetValue for a bool expression is always populated statically, and does not
-// ever receive any incoming values (no incoming edges) so this should never be
-// called. It has been implemented for uniformity.
-func (obj *ExprBool) SetValue(value types.Value) error {
-	if err := types.TypeBool.Cmp(value.Type()); err != nil {
-		return err
-	}
-	// XXX: should we compare the incoming value with the stored value?
-	obj.V = value.Bool()
-	return nil
-}
-
 // Value returns the value of this expression in our type system. This will
 // usually only be valid once the engine has run and values have been produced.
 // This might get called speculatively (early) during unification to learn more.
@@ -4946,17 +4934,6 @@ func (obj *ExprStr) Func() (interfaces.Func, error) {
 	}, nil
 }
 
-// SetValue for an str expression is always populated statically, and does not
-// ever receive any incoming values (no incoming edges) so this should never be
-// called. It has been implemented for uniformity.
-func (obj *ExprStr) SetValue(value types.Value) error {
-	if err := types.TypeStr.Cmp(value.Type()); err != nil {
-		return err
-	}
-	obj.V = value.Str()
-	return nil
-}
-
 // Value returns the value of this expression in our type system. This will
 // usually only be valid once the engine has run and values have been produced.
 // This might get called speculatively (early) during unification to learn more.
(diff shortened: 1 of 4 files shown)


lang/ast/structs.go
```diff
@@ -7319,7 +7223,7 @@ type ExprCall struct {
 	expr interfaces.Expr // copy of what we're calling
 	orig *ExprCall       // original pointer to this
 
-	V types.Value // stored result (set with SetValue)
+	V types.Value // stored result
```
gelisam, line 7226: Why is `V` public if it is intended to be accessed with a getter and a setter?

lang/ast/structs.go
```diff
@@ -8318,23 +8222,10 @@ func (obj *ExprCall) Func() (interfaces.Func, error) {
...
-
 // Value returns the value of this expression in our type system. This will
 // usually only be valid once the engine has run and values have been produced.
 // This might get called speculatively (early) during unification to learn more.
 // It is often unlikely that this kind of speculative execution finds something.
-// This particular implementation of the function returns the previously stored
-// and cached value as received by SetValue.
 func (obj *ExprCall) Value() (types.Value, error) {
 	if obj.V == nil {
 		return nil, fmt.Errorf("func value does not yet exist")
```
gelisam, line 8231: Why fail when the value is `nil`? As the comment says, `Value()` is sometimes called speculatively during unification, so it should be fine to return `nil` in that case.

The test `TestAstFunc2/test_#0_` currently fails in CI with this error message.

lang/funcs/engine.go
```diff
@@ -450,10 +450,6 @@ func (obj *Engine) Run() error {
 				// XXX: maybe we can get rid of the table...
 				obj.table[vertex] = value // save the latest
 				node.mutex.Lock()
-				if err := node.Expr.SetValue(value); err != nil {
```
gelisam, line 453: This is the call to `SetValue()` which seems redundant to me, because the engine is already storing the latest value in `obj.table[vertex]` two lines above.

lang/funcs/engine.go
```diff
@@ -45,7 +45,7 @@ type State struct {
 	input  chan types.Value // the top level type must be a struct
 	output chan types.Value
 
-	mutex *sync.RWMutex // concurrency guard for modifying Expr with String/SetValue
+	mutex *sync.RWMutex // concurrency guard for accessing Expr.String()
 }
 
 // Init creates the function state if it can be found in the registered list.
```
gelisam, line 51: github won't let me attach a comment to a line which hasn't changed, but a few dozen lines below, at line 82, there's another occurrence of `SetValue` in a comment, and I'm not sure how to handle it:

```go
// String satisfies fmt.Stringer so that these print nicely.
func (obj *State) String() string {
	// TODO: use global mutex since it's harder to add state specific mutex
	//obj.mutex.RLock() // prevent race detector issues against SetValue
	//defer obj.mutex.RUnlock()
	// FIXME: also add read locks on any of the children Expr in obj.Expr
	return obj.Expr.String()
}
```

It's a commented-out line anyway, so I should probably just delete it, but the TODO next to it seems related so I'm not sure whether I should delete it too or if I would accidentally be removing an important TODO note.

purpleidea:
As a quick aside, I've been writing the new version of the engine to not use SetValue, and instead have obj.table be returned via a Table() method... Having said that, I'm still hitting some bugs that is blocking this from happening, and I'm not yet sure why. I guess this PR was prescient! ... If I can end up making it work that is!

PR #697, opened 2023-01-29
lang: CallExpr must generate a subgraph

gelisam:
FuncValues are now manipulating the graph instead of manipulating values, so the logic for calling a FuncValue must now follow suit.


gelisam:
This PR builds on top of https://github.com/purpleidea/mgmt/pull/696, and so the diff will be much easier to review after that one is merged.

This is the majority of the work @purpleidea and I did this weekend during our pairing session. It is relatively self-contained because it all pertains to porting `ExprCall` to the new representation introduced in #696.

gelisam:
force-pushed to remove a temporary commit which I did not mean to include in this PR.

gelisam:
The tests fail, but as in #696, this is expected, because a lot of `panic` calls remain.

gelisam:

PR #703, opened 2023-03-06
last few MergedGraph() implementations

gelisam:
Building on top of #701.

This implements MergedGraph() for ExprCall, ExprFunc, and ExprVar.


purpleidea:
Sweet! I'm making good progress on the new engine. The race detector helped me find some races today and I think I got those fixed. One last API addition to the new engine library, and then I can port it in and maybe we can put all of this together!

gelisam:

PR #708, opened 2023-04-20
solve issue of Add new golangci-lint test

royadaneshi:
I have create a new test file by golangci-lint similar to the test file that was by gometalinter.


royadaneshi:
I have create a new test file by golangci-lint similar to the test file that was by gometalinter so the gometalinter file can be removeed.

purpleidea:
Hi,

Thank you for the patch. Before I review, I recommend you squash both commits and change the commit message. More info can be found here: https://github.com/purpleidea/mgmt/blob/master/.github/PULL_REQUEST_TEMPLATE.md

As an aside, I don't think you need to change go.mod or go.sum for this patch.

Cheers

purpleidea:
This has been superseded by https://github.com/purpleidea/mgmt/pull/711
Let me know if you'd like to work on more commits.

Issue #686, opened 2022-01-27
Add a test for PR # 685


purpleidea:
https://github.com/purpleidea/mgmt/pull/685#issuecomment-1023139028

Would be great to add a simple test for this commit and also to double check that the test fails if that commit is reverted.

Commit: https://github.com/purpleidea/mgmt/commit/bf7e45439b044d18d0fdf3151fbdd7994875edce

Thanks to @frebib for the nice patch!

neverbeenthisweeb:
@purpleidea 
Hi.
I can help with this, I am aware of what should be tested in regards to commit [bf7e454](https://github.com/purpleidea/mgmt/commit/bf7e45439b044d18d0fdf3151fbdd7994875edce).
However, I still don't know what does the term `kind` means inside the `StructKindToFieldNameTypeMap(kind string) (map[string]*types.Type, error)`.
Can you help to pinpoint me some examples of how this struct `kind` looks? Maybe something like testdata? Thank you.

neverbeenthisweeb:
@purpleidea 
Hi again.
After playing around with the `StructKindToFieldNameTypeMap` method using tests under the `./engine` folder, it looks like to test the `StructKindToFieldNameTypeMap`, we should have a registered resource in the [registeredResources](https://github.com/purpleidea/mgmt/blob/master/engine/resources.go#L31) map.
Do we already have the convention to conveniently register a resource when doing tests?
However, I found this [TestRes](https://github.com/purpleidea/mgmt/blob/master/engine/resources/test.go#L33) struct. But this struct can't be imported in `./engine/util` as it will result to an import cycle.
Wdyt? Thanks.

purpleidea:
@JefMasereel can probably point to the issue he found that this patch fixed... After our memory is refreshed, we can use TestRes (and import it elsewhere) or do something different. Does this help?

purpleidea:
Original issue is here actually: https://github.com/purpleidea/mgmt/issues/684

neverbeenthisweeb:
@purpleidea 
Ok, I will take a look at #684 first.

Anyway, by "_...our memory is refreshed_", do you mean clearing the [registeredResources](https://github.com/purpleidea/mgmt/blob/master/engine/resources.go#L31)?

I think we can create a new file of `./engine/util/test.go`, then copy the implementation of `TestRes` (and its related dependencies) in `./engine/resource/test.go` to that new file. Although it looks like a duplication, I think it is tolearble, considering that:

[...]

PR #709, opened 2023-05-08
engine: util: Add test for StructKindToFieldNameTypeMap function

ofekatr:
### What this does

- Adds test for the `StructKindToFieldNameTypeMap` utility function, to ensure no private properties are being included in the result.

### More Information

- Relevant Issue: #686

### DIFF
--- a/engine/util/util_test.go
+++ b/engine/util/util_test.go
@@ -24,6 +24,9 @@ import (
 	"reflect"
 	"strconv"
 	"testing"
+
+	"github.com/purpleidea/mgmt/engine"
+	"github.com/purpleidea/mgmt/lang/types"
 )
 
 func TestUnknownGroup(t *testing.T) {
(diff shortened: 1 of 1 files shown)


frebib (approved):

engine/util/util_test.go
```diff
@@ -164,3 +168,77 @@ func TestStructTagToFieldName2(t *testing.T) {
...
+}
+
+func (t *testEngineRes) Name() string {
+	return "test-name"
+}
+
+func (t *testEngineRes) SetKind(_ string) {
+}
+
+func (t *testEngineRes) SetMetaParams(_ *engine.MetaParams) {
```
purpleidea, line 228: You don't need these underscores since the inputs aren't being used, you can omit them.

engine/util/util_test.go
```diff
@@ -164,3 +168,77 @@ func TestStructTagToFieldName2(t *testing.T) {
 	t.Logf("got output: %+v", m)
 	t.Logf("got error: %+v", err)
 }
+
+func TestStructKindToFieldNameTypeMap(t *testing.T) {
```
purpleidea, line 172: Can you move this test to the bottom of the file after the declaration of the resource please? Or even better, put it _inside_ the test block like how the other tests do?

engine/util/util_test.go
```diff
@@ -164,3 +168,77 @@ func TestStructTagToFieldName2(t *testing.T) {
...
+	}
+	require.NoError(t, err)
+	require.Equal(t, expected, res)
+}
+
+type testEngineRes struct {
+	PublicProp1  string
+	PublicProp2  string
+	privateProp1 string
+	privateProp2 string
```
purpleidea, line 190: Since we're adding this test, want to add a non-string prop too?

purpleidea (changes requested):
Thanks for the patch! I suggested some small changes. Please make them and send em my way! As for the commit message, could you instead have it start with: "engine: util: Add test for StructKindToFieldNameTypeMap function"

Cheers

engine/util/util_test.go
ofekatr: > put it inside the test block like how the other tests do? 

Unless I understand your meaning, struct functions can't be declared inside functions. If you have tests you can link for reference to explain the requested change, that would help

engine/util/util_test.go
ofekatr: Didn't know it was possible, thanks!
Fixed in 8ef2386

engine/util/util_test.go
ofekatr: The test was relocated beneath the resource in 8ef2386

purpleidea: > Unless I understand your meaning, struct functions can't be declared inside functions. If you have tests you can link for reference to explain the requested change, that would help

You are entirely correct, my mistake sorry. Standalone structs can, but I didn't realize the methods on them can't be! Woops!

engine/util/util_test.go
```diff
@@ -164,3 +168,77 @@ func TestStructTagToFieldName2(t *testing.T) {
...
+	k := "test-kind"
+
+	engine.RegisterResource(k, func() engine.Res { return &testEngineRes{} })
+	res, err := StructKindToFieldNameTypeMap(k)
+
+	expected := map[string]*types.Type{
+		"PublicProp1": types.TypeStr,
+		"PublicProp2": types.TypeStr,
+	}
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
ofekatr: Done in c4c168e

engine/util/util_test.go
```diff
@@ -164,3 +167,83 @@ func TestStructTagToFieldName2(t *testing.T) {
...
+
+func TestStructKindToFieldNameTypeMap(t *testing.T) {
+	k := "test-kind"
+
+	engine.RegisterResource(k, func() engine.Res { return &testEngineRes{} })
+	res, err := StructKindToFieldNameTypeMap(k)
+
+	expected := map[string]*types.Type{
+		"PublicProp1": types.TypeStr,
+		"PublicProp2": types.TypeStr,
```
purpleidea, line 239: Is this still the correct expected type?

engine/util/util_test.go
```diff
@@ -164,3 +167,83 @@ func TestStructTagToFieldName2(t *testing.T) {
...
+
+func (t *testEngineRes) MetaParams() *engine.MetaParams {
+	return nil
+}
+
+func (t *testEngineRes) Name() string {
+	return "test-name"
+}
+
+func (t *testEngineRes) SetKind(string) {
```
purpleidea, line 210: nit: (optional)
single line functions like this can be written like:

```func (t *testEngineRes) SetKind(string) {}```

purpleidea (commented):
Patch looks great, thank you very much. I think there is one small bug (I commented.) we'll see what CI tests tell us.

Note: there is a likelihood of some unrelated breakages in the tests atm, because I've let some stuff rot that will be replaced shortly.

In any case, could you fixup and rebase all of these commits into a single clean one?

Thanks!

engine/util/util_test.go
ofekatr: Fixed in eddd162

engine/util/util_test.go
ofekatr: Fixed in eddd162

purpleidea:
LGTM, thank you. Just waiting on CI.

purpleidea:
Clean patch, merged, thank you! \o/
If you'd like some more things to work on, please let me know!

Issue #487, opened 2019-03-16
Add new golangci-lint test


purpleidea:
We use gometalinter in our tests. We'd like to switch to https://github.com/golangci/golangci-lint since the former is deprecated. This should in theory improve testing speed as well.

Patches welcome!

(Note initially, the goal would be to add a new test similar to https://github.com/purpleidea/mgmt/blob/master/test/test-gometalinter.sh and in the future, I can remove the gometalinter test.)

(If the new gometalinter finds any new issues, then we can add those as commits before the test addition commit.)

amitech:
@purpleidea let me know if it is still relevant? I can take this one.

royadaneshi:
Hello   @purpleidea  
I added new test similar to the test-gometalinter.sh in golangci-lint and I created a pull request.

purpleidea:
This has been merged.

PR #676, opened 2021-10-12
Create golangci_lint test script

ginywiny:
Refer to previous PR for feat/create_golangci_lint_test_487.

PR #711, opened 2023-05-12
test: Add golangci-lint test

KaushalIngle:
Added the test for golangci-lint for #487


purpleidea:
I rebased to master and merged, thank you!

There is still more work to plumb this in so it runs automatically, and adding some excludes or fixups so it passes. Some similar patches exist in https://github.com/purpleidea/mgmt/pull/670 and https://github.com/purpleidea/mgmt/pull/676 if you're curious.

Let me know if you'd like to make more commits either here or elsewhere!

PR #704, opened 2023-03-09
add functions to the language

gelisam:
This PR is not yet complete. It exists to make it easy to look at and comment on the code for the `mcl-functions` feature branch, which completely reworks the way in which functions work in MCL.

## The problem

Today in `master`, builtins and lambdas use very different compilation strategies. A builtin is compiled to a single Func which receives values from upstream nodes and emits values downstream. A lambda, by contrast, is compiled away, as the lambda's body is inlined at every call site. Once this inlining is performed, only builtins remain, so the resulting function graph is the graph of their Funcs.

This does not work at all for MapFunc, a builtin which expects to receive a FuncValue as one of its inputs, because the above compilation strategy does not produce FuncValues, neither for builtins nor for lambdas. It would in fact be incorrect to construct a FuncValue from a builtin, because builtins can have an internal state and emit multiple values over time (e.g. `System("seq 3")` emits `"1"`, then `"2"`, then `"3"`), but a FuncValue can only transform input values into a single output value, so it cannot represent that complexity. And of course, lambdas can contain builtins (e.g. `func(x) { System(x) }`), so they cannot be represented by a FuncValue either.

## The solution

The function graph must be able to change over time, because if `func(x) { Map(System, xs) }` first receives a list of length 1 and then a list of length 2, then at first there must be one SystemFunc in the function graph, then later there must be 2 SystemFuncs.

In order to do this, a FuncValue must be able to do more than transform input values into a single output value. Instead a FuncValue must describe in which way the function graph needs to change in order to accommodate a new instance of the function which the FuncValue represents. This branch's FuncValue does this by receiving input Funcs, creating and adding new nodes to the function graph, and then returning the one node among those new creations which shall be treated as the output Func, that is, the one whose output values are the result of applying the function represented by the FuncValue to the inputs received from the input nodes.

Another big change is that the function engine now works on a graph of Funcs instead of a graph of Exprs. This allows FuncValues to add Funcs to the function graph, without having to wrap them in Expr wrappers which don't correspond to any MCL code the user wrote.

## Overview of the changes so far

1. FuncValue now behaves as described above.
1. A new type called SimpleFn behaves in the same way the old FuncValue did, because it is still useful in many places to represent a transformation from input values to a single output value.
1. Txn, an abstract API for modifying a function graph transactionally.
1. graphTxn, a thread-safe implementation of that API for modifying the function graph while the function engine is still running that graph.
[...]

PR #716, opened 2023-08-30
WIP discussion of function values to resources

purpleidea:


purpleidea:
Hey @gelisam I opened this PR (it's based of #704) to discuss passing function values to resources. In playing around first I hit an expected bug in the Into() function. This function pushes a bunch of mcl values INTO a golang struct (the resource). Of course the function part wasn't implemented.

So I started moving things around. And I have the import cycle of Txn <-> Types since I thought FancyFunc should maybe be part of the types package. Although it's a bit different because it has the unusual Txn aspect in there.

Before proceeding with a more complex refactor, I figured I'd check with you if you think this all makes sense or if you had a thought about how instead we'd pass in a simpler FuncValue to resources instead? (I don't expect they'll need the Txn, that's not the kind of functions we'd be passing to the resource I don't think.)

Cheers!

gelisam:
Notes from today's meeting:

* a "timeless" function is a function which takes a single list of input Values to a single output Value.
* a "timeful" function is the opposite: it plays shenanigans with time, e.g. by returning multiple values over time, or no value at all.
* in `test "name" { func1 => expr }`, it's fine if `expr` itself is not timeless, it can e.g. emit a different FuncValue every second. But every FuncValue it emits must be timeless.
* if a lambda body contains a timeful expression, or a variable whose value is obtained from a timeful expression, then it is considered timeful, even though it would be possible to compile the lambda so that it emits a different FuncValue every time the value of a captured variable changes. The reason for that is that in CallFunc, the sub-graph is only regenerated when the lambda expression is replaced with something else. When the captured variables inside the lambda change, the corresponding Funcs emit values downstream, into the existing sub-graph. New values flow through the sub-graph and the output of the sub-graph changes, but the sub-graph itself does not have to be replaced.

gelisam:
I propose a new static analysis phase after type checking and before `Expr.Graph()`. The input is a mapping from lambda-bound name to their type, along with an annotation indicating whether each function type inside that type is timeless or timeful. The output is a similarly-annotated type for the Expr. We then check if in `test "name" {func1 => expr}`, the annotated type is a timeless function, and if it returns a function, that function is itself timeless, etc.

gelisam:
Since you opened this PR, I guess this time I'll be the one pushing to a auxiliary branch and you'll be the one updating the official MR branch to incorporate my changes! I have pushed to `gelisam/functions-to-resources`. So far it's the same as `master`, because now that #704 has been merged, we should obviously start from the merged version.

gelisam:
Could you post the examples and counter-examples of what counts as timeless, both as a reminder of what the goal is and so we can use them as tests?

purpleidea:
I'll fix and throw away this branch. But if you'd prefer to have the PR branch, lmk.

> Could you post the examples and counter-examples of what counts as timeless, both as a reminder of what the goal is and so we can use them as tests?

In short:

* timeless:
any pure function
iter.map
strconv.itoa


* timeful:

world.schedule
datetime.now
os.system

purpleidea:
Here are the notes I took (may contain errors)

```
import "strconv"

$fn1 = func($x) { # pure! also good
	strconv.itoa($x)
}

$fn2 = if datetime.now() % 2 == 0 { # should be good! in fact, this has to work.

	func($x) { # middle point
		strconv.itoa($x)
	}

} else {

	func($x) {
		"hello"
	}

}

$fn3 = func($x) { # bad
	if datetime.now() % 2 == 0 { # this stream function makes things bad (it's also not pure)
		strconv.itoa($x)
	} else {
		"hello"
	}
}

$fn4 = func($x) {
	iter.map(...) # map is pure and okay here, as long as the input function is also.
}

$fn5 = func($x) { # bad!
	os.system("seq $x")
}


$fn6_prep = func($sam) { # this is ok or is it? -- the simple thing to do is that this is not allowed.
	func($x int) {
		$sam
	}
}

$fn6 = $fn6_prep(os.system("seq 10"))


$fn7_prep = func($sam) { # this is good

	if datetime.now() % 2 == 0 { # should be good! in fact, this has to work.

		func($x) { # middle point
			strconv.itoa($x)
		}

	} else {

		func($x) {
			"hello" + $sam
		}

	}

}

$fn7 = $fn7_prep("hello")


$fn8_prep = func($sam str) { # this is bad because it's simpler to do it that way and technically $fn8 is static but the insides are not timeless

	if datetime.now() % 2 == 0 {

		func($x) { # middle point
			strconv.itoa($x)
		}

	} else {

		func($x) {
			"hello" + $sam
		}

	}

}

$fn8 = $fn8_prep(os.system("seq 10"))


$fn9_prep = func($sam int) { # we actually agree this one is okay, but the reason it's okay is because $fn9 changes for each datetime.now() timeful change.

	if $sam % 2 == 0 {

		func($x) { # middle point
			strconv.itoa($x)
		}

	} else {

		func($x) {
			"hello"
		}

	}

}

$fn9 = $fn9_prep(datetime.now())


test "test1" {
	anotherstr => "hello",
	func1 => $fn3, // txn!
}



#
#	WHY DO WE WANT THIS?
#

import "prometheus"

# prometheus.connect is not pure because it has some sneaky side-effect somewhere...

$conn = func($x string) int {
	prometheus.connect($x) + 42 # not pure but timeless
}

test "test2" {
	func1 => $conn,
}


import "flagthing" # a built-in library

$flag1 = "flag1" # some arbitrary name

if flagthing.happened($flag1) {
	print "wow" {}
}

test "test2" {
	flagname => $flag1,
	func1 => $flagthing.start,
}

```

purpleidea:
@gelisam Just wanted to double-check if this patch with interface and one implementation is basically what you were expecting? If so, LMK and I can complete it for other functions too.

gelisam:
Looks good, yes!

gelisam:
Next steps:

* [ ] Fill in the `panic("TODO")` in `StmtFoo.TimeCheck()`
* [ ] Add a Timeless info field to `Func`, or just a `Func.TimeCheck()` method
* [ ] Use that `Func.TimeCheck` to fill in the `obj. Function` case in `ExprFunc.TimeCheck()`
* [ ] Fill in the `obj.Values` case in `ExprFunc.TimeCheck()`
* [ ] convert a FuncValue to a golang function, either via `Func.Call()` or by sending a single value to the sub-graph and waiting for the result.

gelisam:
* [ ] write tests exercising the new time check

purpleidea:
Not the priority, but I was playing with this code and implemented a few extra for fun. I also pushed in your commits to have them here.

An interesting realization I had is that we should eventually once all have been implemented, be able to remove each mostly identical custom Stream() implementation with a single one that is common to all functions. Or most functions!

PR #721, opened 2023-10-07
lang: ast: Clear env in VarExpr.Graph

gelisam:
Variables are defined at the top-level, where only obj.Scope variables are in scope. The lambda parameters which are in scope where the variable is used must not be in scope where the variable is defined. This test detects future mistakes in which we make a recursive call with the same environment as the use site instead of clearing the env.

The test passes with or without the fix, because an earlier phase (SetScope) already clears the env, thus triggering a compile-time error.


purpleidea:
Awesome! LGTM I will merge shortly!

Issue #722, opened 2023-12-02
Duplicate functions being created


purpleidea:
I noticed the following with our latest code. For example:

```
import "example"
$x = example.answer()	# gets called twice
test "test1" {
	int64 => $x,
}
test "test2" {
	int64 => $x,
}
```

Save in whatever.mcl and run with:

```
./mgmt run --tmp-prefix lang whatever.mcl
```

There are actually TWO functions started by the function engine. One for each $x. I know some of the time we need to copy the function and this will cause a second copy, but in this simple case, I think it's important that we don't copy it. Is there a reason we have to?

I think this is an important optimization, since I was seeing way too much memory usage for a situation when we had a small number of functions with their values everywhere, and well it explodes out of control.

Thanks!

/cc @gelisam

gelisam:
The reason `$x` is copied is in order to support polymorphism. Here both call sites have type `int`, but it is possible to use `$x` at two different types, in which case we need two copies.

In theory, it should be possible to use a single copy in a case like this where the two use site have the same type. However, in practice, this is quite difficult to achieve because this would introduce a circular dependency between the type-checker and the part of the code which performs this copy.

Consider a similar example in which the two use sites have a different type. If we were to use a single instance for both use sites, the type checker would propagate the type information from one use site to the other, and then type unification would fail because it would try to unify two different types, the types of the two use sites. To avoid this, the current code makes a copy first, and then performs type-checking on each copy separately.

In order to avoid the duplication, the code which makes the copy would need to check if the two use sites need to instantiate `$x` at the same type; if so, use one copy for both use sites, otherwise make two copies. But now we have a circular dependency: we need to know the type before deciding whether to make a copy, and we need to make a copy before performing type-checking!

gelisam:
Would it make sense to make two copies, perform type-checking on the two call sites separately, delete all the copies, and then create copies again using the new type information? That would make even more copies overall, but only at compile-time.

purpleidea:
I wrote a test for this scenario (which obviously fails at the moment). It's currently here: https://github.com/purpleidea/mgmt/commit/bbb2b1569284a28780a3fb967f54f785d1956e1a and you can cherry-pick it if it helps you to validate things faster.

purpleidea:
In case it's helpful, this did the right thing for all values, before we started this functions work. I was able to make it work with my "light copy" approach, in case that rings a bell.

Maybe we only want to do these extra instantiations when we're returning a function?

[...]

Issue #142, opened 2017-02-14
prometheus-3.sh is failing


purpleidea:
@roidelapluie 

https://travis-ci.org/purpleidea/mgmt/jobs/201582409#L734

Should I disable it or do you know why it's failing?

roidelapluie:
Because there is a bug in the code :)

```
  - name: file4
    path: "/tmp/mgmt/f4"
    content: |
      i am f4
    state: exists
    meta:
      noop: true
```

That one produces an error:

```
20:28:12 file.go:604: File[file4]: contentCheckApply(false)
20:28:12 file.go:666: File[file4]: chmodCheckApply(false)
20:28:12 file.go:705: File[file4]: chownCheckApply(false)
20:28:12 actions.go:417: File[file4]: CheckApply errored: could not Process() successfully: stat /tmp/mgmt/f4: no such file or directory
20:28:12 actions.go:635: File[file4]: Exited with failure: could not Process() successfully: stat /tmp/mgmt/f4: no such file or directory
```

roidelapluie:
@mildred :)

purpleidea:
So if we had used errwrap, that which stat error that was would be obvious

purpleidea:
but should be easy to track down... who's on it? me? mildred? roi?

purpleidea:
So this is because chmod/chown both call stat in noop mode...

purpleidea:
What should the correct fix be...

purpleidea:
Merged, thanks!

Issue #31, opened 2016-05-13
filter managed attributes


tuxmea:
There are use cases where one does not want to manage all attributes.
e.g. create a file in case that it is missing, manage user and mode but don't change the content when the file is there.
This could be achieved by filtering attributes on resource level making "mask" a new meta parameter.
In discussion with @purpleidea

purpleidea:
Technical note: the idea would be to have either a resource specific parameter, or a metaparameter, which takes a mask of resource specifications to skip when doing CheckApply. For example: we add a file if it's missing, but not if the content is wrong. So IOW, Mask=>FileContent...

daenney:
To me this kinda sounds like not properly doing configuration management though. Only partially managing a resource means you can end up with systems in which we believe things are the same (the file exists for example) but actually behaves completely differently (b/c one got a new config from a package upgrade and the other didn't b/c lets say a signing key expired and it can't update on that other node or someone hand-hacked a config in one place etc. etc.).

We've also seen really nasty stuff happen in Puppet when certain attributes are omitted (like owner and group) which can then make the file ownership flip-flop on `apply` runs based on if it's actually run by root (through a cron for example) or done through sudo by a user.

I can understand not wanting to manage the content of the file for example but I would be very careful implementing this as a more general purpose thing that can apply to any resource. It risks breaking a fundamental promise a configuration management engine gives, that everything is in a consistent state.

purpleidea:
> We've also seen really nasty stuff happen in Puppet when certain attributes are omitted (like owner and group) which can then make the file ownership flip-flop on apply runs based on if it's actually run by root (through a cron for example) or done interactively through sudo by a user.

This is a very insightful comment, and a good point, thanks!

As an aside we've got a WIP patch for user support: https://github.com/purpleidea/mgmt/pull/110

mildred:
I believe that in my implementation #110, if there is no `content` defined, it doesn't check for it at all. You might have some content or not, it shouldn't care: https://github.com/purpleidea/mgmt/blob/master/resources/file.go#L585-L588

mildred:
It will check for owner and mode separately than for content

purpleidea:
On Fri, Feb 10, 2017 at 6:03 PM, mildred <notifications@github.com> wrote:
>
[...]

PR #723, opened 2023-12-05
lang: ast: ExprBind is now monomorphic

gelisam:
An attempt at fixing #722


gelisam:
@purpleidea all right, let's see how many tests this break!

gelisam:
Test TestAstFunc2 0 fails with `could not unify types: call doesn't contain an expr pointer yet`. I think that means I'm not filling in `ExprCall.expr` because I'm not traversing the original definitions, only the copies. But now there aren't any copies anymore, so I should traverse and typecheck the original definitions.

gelisam:
Fixed, only 3 failures left! Including, hilariously, the `clear-env-on-var.txtar` test I just added in the previous MR. That one makes sense: the `$wat = $x` line is now scope-checked before `$f`'s body, so the test fails with a different error message.

The other two tests, `map-iterator1.txtar` and `polymorphic-lambda.txtar`, are specifically making sure that `$add = func($x) {$x + $x}` can be instantiated at two different types, so hopefully you agree that with the new assumption that StmtBinds are no longer polymorphic, it is now the tests which are wrong.

purpleidea:
> Fixed, only 3 failures left!

Amazing!

> Including, hilariously, the clear-env-on-var.txtar test I just added in the previous MR.

haha! The code $DEITY's love us!

Can you cherry-pick in my new test too please? It's here: https://github.com/purpleidea/mgmt/commit/bbb2b1569284a28780a3fb967f54f785d1956e1a

I'll look over those tests now, and I'll also check we have one for StmtFunc polymorphism =D

Are there any negative side-effects or important features we're leaving on the floor with this?

purpleidea:
I added all of this ontop and it's in bug/722 branch testing! (I fixed those tests, let's see if it works!)

purpleidea:
Here: https://github.com/purpleidea/mgmt/actions/runs/7096929277

purpleidea:
The AST tests pass, but unfortunately, we still have a double. My new test fails:

>     interpret_test.go:715: test #142: funcs: Erroring func `call`: panic in Stream of func `call`: should not get called twice

Basically we still have a copy.

gelisam:
Hmm, this is completely wrong.

`StmtBind.SetScope()`, `StmtBind.Unify()`, and `StmtBind.Graph()` were all no-ops, because the work was done in the copies. Now that I have removed the `ExprPoly` wrapper, there aren't any copies anymore, so the work isn't done anywhere! I have changed `StmtBind.SetScope()` to no longer be a no-op, but I forgot to change `StmtBind.Unify()`, and `StmtBind.Graph()` as well. It is thus quite surprising that so few tests are failing! In fact, when I change those two, a lot more tests fail!

purpleidea:
> It is thus quite surprising that so few tests are failing!

If any of these situations makes it obvious what new tests to add, of course, please go crazy! I must admit that the test suite was not written methodically, so there is likely some permutation of features which is not tested.

gelisam:
Ah, of course! `StmtBind.Unify()` might not call `Unify()` on the expression it binds, but `ExprVar.Unify()` does. So the test could only potentially fail if it contains `$x = ...` but no uses of `$x`. And that case doesn't fail either because the `$x = ...` doesn't contribute any constraints, so it can't cause type checking to fail.

Similarly, `StmtBind.Graph()` doesn't call `Graph()` on the expression it binds, but `ExprVar.Graph()` does. And since each call to `Graph()` generates a new `Func`, this explains why there are two `Func`s and thus why test 142 fails.

gelisam:
All right, here's the plan. We want to scope-check, type-check, and graph each the expression bound by `StmtBind` exactly once, regardless of how many use sites the variable has. So `StmtBind.SetScope()`, `StmtBind.Unify()`, and `StmtBind.Graph()` must all make a recursive call. I've done that already. What is missing is that `ExprVar.SetScope()` etc. should _only_ make a recursive call if their target _was_ a `PolyExpr`.

Currently there is no way to tell, because if `ExprVar.SetScope()` points to a `PolyExpr`, it replaces its target with a copy, and then `ExprVar.Unify()` and `ExprVar.Graph()` always make a recursive call. So we need `ExprVar` to track whether it has made a copy, perhaps by making it point to a copy of the `PolyExpr` instead of a copy of the `Expr` wrapped by the `PolyExpr`.

Or maybe... maybe `ExprVar` is _never_ making a copy now? Hmm... what about classes? We said that we _do_ want classes to be polymorphic, but that could mean two different things. Do we want to support this:

```
class c($id) {
  $int = $id(42)
  $str = $id("foo")
}
```

or that:
```
class c($x) {
  test printf("%v", $x + $x)
}
include c(42)
include c("foo")
```

or both? or neither?

gelisam:
Another question: in the following code,

```
$y = $x + $x
$x = 42
```

are the `StmtBind.Unify()` and `StmtBind.Graph()` calls done in the order `$y`, then `$x`, or in the order `$x`, then `$y`?

purpleidea:
This is a good question:
```
class c($id) {
  $int = $id(42)
  $str = $id("foo")
}
```
Since it violates our "lambdas in variables can't be polymorphic" rule from before, my answer would be: "it doesn't bother me if it's not allowed, and if it's easily possible, great, but if we need for it to be a compile error, that's okay.

(Side note, of course the class in this situation doesn't do anything since $int and $str aren't used, so this alone might not be a good test case without a res in there somewhere, as the compiler might nuke them since they're not used.)


This second example is definitely allowed, and most likely important to keep in.
```
class c($x) {
  test printf("%v", $x + $x)
}
include c(42)
include c("foo")
```

purpleidea:
```
$y = $x + $x
$x = 42
```

The above _is_ valid code, and you might recall the `Ordering` method of our AST. This basically figures out the DAG of what comes first, so that it always works even if you get the "logical ordering" wrong.

Additionally, there is a constant: `RequireTopologicalOrdering` which is currently false. If true, in the future, it would check if it was a valid toposort ordering or not, but I've left it unimplemented for now since I think it's more interesting and useful to not require code be in a logical order.

> are the StmtBind.Unify() and StmtBind.Graph() calls done in the order $y, then $x, or in the order $x, then $y?

I might not entirely be understanding this question. If my above explanation doesn't suffice, are you just asking for the AST recurse/descent order?

gelisam:
> lambdas in variables can't be polymorphic

That's not what we said at all!

There are four places where variables are bound, and two places where non-variable names are bound:
1. `$x = ...`
2. `func f($x) {...}`
3. `func($x) {...}`
4. `class c($x) {...}`
5. `func x(...) {...}`
6. `class x(...) {...}`

On master, 1, 4, 5 and 6 are polymorphic (and thus 2 and 3 are monomorphic). We've agreed that 1 should be monomorphic, and that 5 and 6 should be polymorphic, but we forgot to discuss 2, 3, and 4.

You are now talking about lambdas in variables, but remember, we can't have a different rule depending on what type of value ends up in the variable. Maybe you are saying that all variables should be monomorphic? That is, 1, 2, 3, and 4 should be monomorphic, while 4 and 5 should be polymorphic?

gelisam:
Let's write 5 tests covering those 5 cases

gelisam:
> If my above explanation doesn't suffice, are you just asking for the AST recurse/descent order?

I do remember that out-of-order definitions are allowed and that there is a function which calculates the topological order, but I don't remember what happens to that topological order. Does `StmtProg` use it to perform there recursive calls in that order rather than the order in which the definitions appear in the source code? Does it do so for `StmtProg.SetScope()`, `StmtProg.Unify()` and `StmtProg.Graph()`, or just for some of them?

purpleidea:
> Does StmtProg use it to perform there recursive calls in that order rather than the order in which the definitions appear in the source code?

Yes. Ordering() is only used once outside of the recursive ordering calls themselves-- it is used in StmtProg:SetScope. The other methods AFAICT don't need ordering. But if they do, we can change this.

gelisam:
I have just pushed a bunch of tests, looking good?

purpleidea:
> I have just pushed a bunch of tests, looking good?

All these tests look amazing!

I'd recommend for any of the ones that check a field, to use TestAstFunc3 instead of TestAstFunc2. All this means is you just move it one folder over. In addition, the test output will also check the resource fields (anotherstr => "what") stuff, *and* it will run the resource engine, so we get a slight bit more testing. IOW, any new tests that are still static, should probably go in TestAstFunc3.

purpleidea:
> I don't understand what you mean by tests which are "still static". Do you mean tests which only exercise the type checker?

Sorry, what I meant was the TestAstFunc* tests can only test a single resource graph. So basically anything which would produce more than one resource graph would be dynamic. So if datetime.now() was a value in a resource, then we'd get a continuous stream.

> I don't know why those tests use anotherstr => $x instead of test $x {}, I was just copying what your test-one-instance.txtar did.
> Your test-one-instance.txtar is in TestAstFunc2, not TestAstFunc3

Yeah it's fine for now, we'll fix it up later if we need to.

gelisam:
I have now implemented every step of [my plan](https://github.com/purpleidea/mgmt/pull/723#issuecomment-1848246681), stripping away the assumption that ExprVars always point to a PolyExpr and replacing it with a new assumption that ExprVars _never_ point to a PolyExpr. Alas, it is not enough.

Before, ExprVars could look up their target in `obj.scope.Variables[obj.Name]`. That target was a `PolyExpr` copy, and it was then the responsibility of ExprVar to call Graph() on this copy in order to obtain a Func which is unique to this use site.

Now, it is the responsibility of StmtBind to call Graph() on the one and only copy. The problem is: how does the ExprVar find the existing Func? It can only look up the Expr, not the Func!

Before the mcl-functions branch was merged, the answer to that question was easy, because there was only ever one Func for each Expr. Now that we have functions, this is no longer true: the Exprs in the body of a function correspond to multiple Funcs. But it is still true for StmtBind.

All right, I think I have a plan. StmtBind will wrap it Expr in a new SingletonExpr. The first time SingletonExpr.Graph() is called, it will store the resulting Func. Subsequent calls will return this Func.

purpleidea:
> All right, I think I have a plan.

\o/

PS: feel free to add private struct fields if needed like .expr as I previously did if that would help =D

gelisam:
a bunch of examples now fail with `could not unify types: 1 unconsumed generators`, what does this mean?

purpleidea:
> a bunch of examples now fail with could not unify types: 1 unconsumed generators, what does this mean?

Type unification is failing... Why? Well a generator is one of the special invariants which can generate new invariants based on runtime information. For example, once an earlier stage of unification determines something is a string, that generate can learn that, and then generate a simple set of invariants...

So a generator is not being used up. Are we adding something to the set of things to unify which shouldn't get unified?
Which examples should I look at? I can see if anything obvious sticks out.

gelisam:
The simplest failing test is probably `example.txtar`

purpleidea:
> The simplest failing test is probably example.txtar

That code looks like:

```test fmt.printf("answer: %d", 42) {}```

So yeah, indeed that's simple =D

Something is amiss. Reminder, your debug function is still in the code base.

> func DebugSolverState(solved map[interfaces.Expr]*types.Type, equalities []interfaces.Invariant) string

It should run, and help you know what's wrong with the AST perhaps? We can look at this together too if you want.

**Comment by purpleidea** [comment]
>         ?3 = ?3

This looks suspicious...

>         0xc00064e750 topLevel(func() { <built-in:printf> }) :: ?2

This too...

**Comment by purpleidea** [comment]
> in both cases, we have call:fmt.printf(str("answer: %d"), int(42)) :: ?. why isn't printf telling the type checker that it returns a str?

There could be an addition to printf Unify() that we're missing... I will look into it tonight, but I suspect something else.

**Comment by gelisam** [comment]
`0xc00064e750 topLevel(func() { <built-in:printf> }) :: ?2` is what I expect to see. I have added a new `ExprTopLevel` wrapper which captures the scope at the top-level. it is now wrapping all the built-ins, functions, and `StmtBind`s right-hand sides (RHS).

One benefit is that we no longer need to remember to call `target.Graph(emptyEnv)` instead of `target.Graph(env)`, this is easier because all the recursive calls now look the same. Another reason is that RHSs used to be wrapped in `ExprPoly`, which was capturing the top-level scope _and_ indicating to ExprVar that it should copy its target. We no longer want to copy the target, but we still need to capture the top-level scope, so I had to move the responsibility of capturing that scope out of `ExprPoly` and into this new `ExprTopLevel`.

**Comment by gelisam** [comment]
I have added an equality constraint between the `ExprTopLevel` and the `printf` it contains, that should be sufficient to propagate printf's type to the use site. Unless the type checker is matching on the `Expr` and doing something special if it's a built-in, and it can't do that anymore because the Expr is now an `ExprTopLevel` wrapping a built-in?

**Comment by purpleidea** [comment]
> Unless the type checker is matching on the Expr and doing something special if it's a built-in, and it can't do that anymore because the Expr is now an ExprTopLevel wrapping a built-in?

There's no special casing for any different type of Expr... If the wrapping is an indirection, you can add an equality saying that Expr1 == Expr2, and that would fix it.

I'll dig down into the branch more, but in the meantime, I'd recommend just trying to get it work without unification. Once we know it works, then we are more confident we got the right AST shapes and so on, and then the type unification can fall into place =D

**Comment by gelisam** [comment]
> in the meantime, I'd recommend just trying to get it work without unification

there are no variables in `example.txtar`. where can I put the type annotation? something like this?
```
import "fmt"
test (fmt.printf("answer: %d", 42) :: str) {}
```

Even if that was syntactically valid I feel like that's probably not going to solve the problem, because `test` already knows that it needs an str and `printf` already knows that it returns an str, so the type annotation is not adding any new information.

**Comment by gelisam** [comment]
> If the wrapping is an indirection, you can add an equality saying that Expr1 == Expr2, and that would fix it.

I did that already: https://github.com/purpleidea/mgmt/pull/723/commits/cd334bbfe5922f4f527ecc3259a04798b673c057

**Comment by gelisam** [comment]
I think I know what is going on. `printf` expects to see a `CallFuncArgsValue` invariant where the `Expr` in the function position is _exactly_ the `printf` node, and the `TopLevel` wrapper breaks this because the `Expr` in the function is now a `TopLevel` wrapping the `printf` node.

**Comment by purpleidea** [comment]
> I think I know what is going on. printf expects to see a CallFuncArgsValue invariant where the Expr in the function position is exactly the printf node, and the TopLevel wrapper breaks this because the Expr in the function is now a TopLevel wrapping the printf node.

Oh!

I highly recommend we leave the printf unification code alone, HOWEVER, if it's appropriate, we can change the expr that CallFuncArgsValue lists. Maybe that's what you want instead? Since this top-level expr, is actually "the thing" now? -- That in and of itself raises an eyebrow... Does it have to be that way?

**Comment by gelisam** [comment]
> I highly recommend we leave the printf unification code alone

Good news, I can't!

I tried modifying the printf unification code to pattern-match on the Expr in the function position to see if it is an ExprTopLevel, and if so to check if the wrapped Expr is the printf expr. But that would create a circular dependency between the module which defines `printf` and the module which defines `ExprTopLevel`.

**Comment by gelisam** [comment]
> Since this top-level expr, is actually "the thing" now? -- That in and of itself raises an eyebrow... Does it have to be that way?

I don't understand what you are asking. Which thing?

**Comment by gelisam** [comment]
> it's appropriate, we can change the expr that CallFuncArgsValue lists

it worked, thanks for the suggestion!!

**Comment by purpleidea** [comment]
> it worked, thanks for the suggestion!!

\o/

You did the hard part figuring it out! I just know how some of the bits are connected.

It reminds me about something we've discussed previously, which seems true over and over again:

Our programming brains work in such different ways, it's a real pleasure working with you because either I can't solve something, or you can't, but together we can do anything! o/ \o

**Comment by purpleidea** [comment]
FYI: I added a few new commits to git master (not added in this branch) which among other things, add the start of an mcl and txtar formatting check. TL;DR: indents with spaces instead of tabs will fail the test.

The good news is, that your EDITOR might also see the improved .editorconfig file and automatically use tabs for you now.

**Comment by gelisam** [comment]
all right, rebasing on master then

**Comment by gelisam** [comment]
only 9 failing tests left!

**Comment by purpleidea** [comment]
\o/

**Comment by gelisam** [comment]
3 of them fail because they expect the type error to be `2 unconsumed generators` but the actual error is `1 unconsumed generators`.
1. `lookup3.txtar`
2. `printfempty.txtar`
3. `printfunificationerr0.txtar`

Looking at the code, I don't understand why these tests expect two errors. Does it seem acceptable to change the expected error to be `1 unconsumed generators`?

**Comment by gelisam** [comment]
3 more fail with `field arg0 does not exist`, but the MCL code doesn't mention `arg0` anywhere. what's `arg0`?
1. `deploy-readfile0.txtar`
2. `deploy-readfile1.txtar`
3. `slow-unification2.txtar`

**Comment by gelisam** [comment]
The last 3 fail because I've somehow broken SetScope's recursion detector. I am guessing that `$x = $x + 1` no longer triggers the recursion detector because one of those two `$x`s is wrapped in an ExprTopLevel and the other isn't, something like that. What can you tell me about SetScope's recursion detector?

**Comment by purpleidea** [comment]
> Looking at the code, I don't understand why these tests expect two errors. Does it seem acceptable to change the expected error to be 1 unconsumed generators?

Yes absolutely. Something has clearly changed how the type unification solver sees the world, but this is not necessarily an issue or bad. The important thing is the test still fails with a type unification error as we expect. You can safely update these!

**Comment by purpleidea** [comment]
> what's arg0?

I suspect it's coming from here...

https://github.com/purpleidea/mgmt/blob/a6d22a5a4b4e8a4fa25bf5f68f113dd686d72e00/lang/ast/structs.go#L7825

In this instance, it looks like we don't know what to name the args, which are used to build the types for unification, so we invent an arbitrary name... Maybe this should use the standard a, b, c, convention instead...

This shouldn't affect unification though because `func(a str) bool` should match `func (arg0 str) bool` just fine.

So maybe this is a unification bug elsewhere?

**Comment by purpleidea** [comment]
> what's arg0?
(continued)

    deploy-readfile0.txtar
    deploy-readfile1.txtar
    slow-unification2.txtar

Of note, all of these use `import` to pull in a subsequent file... So maybe there's an issue with that and how the stuff pulled in gets scoped... This happens in SetScope from StmtProg...

**Comment by purpleidea** [comment]
> What can you tell me about SetScope's recursion detector?

Well, there are two ways to find this. The first: look for TopologicalSort! (in ast/structs.go of course) you might want the first one for import related stuff:

https://github.com/purpleidea/mgmt/blob/a6d22a5a4b4e8a4fa25bf5f68f113dd686d72e00/lang/ast/structs.go#L3089

Or this one for the ordering stage stuff (more likely $x = $x I think):

https://github.com/purpleidea/mgmt/blob/a6d22a5a4b4e8a4fa25bf5f68f113dd686d72e00/lang/ast/structs.go#L3711

Lastly for class recursion it's here:

https://github.com/purpleidea/mgmt/blob/a6d22a5a4b4e8a4fa25bf5f68f113dd686d72e00/lang/ast/structs.go#L4490

**Comment by gelisam** [comment]
> So maybe this is a unification bug elsewhere?

I have just fixed two unification bugs introduced by ExprTopLevel, so that seems likely!

**Comment by gelisam** [comment]
My fix for the loop detection makes me think that polymorphic code might still evade detection. Do we have tests for loops like `func f() {f()}` and `class c() {include c()}`?

**Comment by purpleidea** [comment]
> Do we have tests for loops like func f() {f()} and class c() {include c()}?
```
$ ack 'include' -l
class-capture1.txtar
class-capture2.txtar
class-capture3.txtar
class-capture4.txtar
class-capture5.txtar
class-capture6.txtar
class-capture7.txtar
class-capture8.txtar
class-recursion1.txtar
class-recursion2.txtar
class-shadowing1.txtar
class-shadowing2.txtar
class-shadowing3.txtar
func-gen1.txtar
func-gen2.txtar
slow-unification1.txtar
slow-unification2.txtar
```

```
$ ack '#err' class-* func-gen* slow-unification*
```

(no results)

Apparently not for classes!

And I can't recall doing any recursive funcs!

**Comment by purpleidea** [comment]
Woops, typo:

```
ack '# err' class-* func-gen* slow-unification*
class-capture7.txtar
12:# err: errSetScope: class `bar` does not exist in this scope

class-recursion1.txtar
9:# err: errSetScope: recursive reference while setting scope: not a dag

class-recursion2.txtar
12:# err: errSetScope: recursive reference while setting scope: not a dag

class-shadowing3.txtar
15:# error prone, and also require a higher-order FRP, which would add complexity

func-gen1.txtar
14:# err: errSetScope: func `fun1` does not exist in this scope

func-gen2.txtar
17:# err: errSetScope: func `fun2` does not exist in this scope
```

**Comment by purpleidea** [comment]
So yes a few basic ones for classes, and none for functions that I know of.

**Comment by purpleidea** [comment]
Wait, I was wrong again:

```
ls *recur*
class-recursion1.txtar             stmtfunc-recursive-double.txtar
class-recursion2.txtar             stmtfunc-recursive-no-operators.txtar
lambdafunc-recursive-double.txtar  stmtfunc-recursive.txtar
lambdafunc-recursive.txtar

```

**Comment by gelisam** [comment]
Excellent, the two cases I was worried about are covered then.

It's pretty strange that `stmtfunc-recursive-double.txtar` and `stmtfunc-recursive-no-operators.txtar` have the exact same contents, probably a mistake?

**Comment by gelisam** [comment]
Only 4 failing tests left!

```
test #27 FAIL TestAstFunc2 27   (const1.txtar) could not unify types: only recursive solutions left
test #30 FAIL TestAstFunc2 30   (deploy-readfile0.txtar) field arg0 does not exist
test #31 FAIL TestAstFunc2 31   (deploy-readfile1.txtar) field arg0 does not exist
test #124 FAIL TestAstFunc2 124 (slow-unification2.txtar) field arg0 does not exist
```

Let's focus on the unification error and hope that it solves the other 3. The `const1` test is very short:
```
# this magic string variable should exist and be "exists"
test $const.res.file.state.exists {}
```
the constraints on master are correct:
```
        0xc000685100 var(const.res.file.state.exists) :: ?1
        0xc000684940 str("exists") :: ?2
        ?2 = str
        ?1 = ?2
```
whereas on this branch the `?2 = str` is missing:
```
        0xc00030fcc0 var(const.res.file.state.exists) :: ?1
        0xc00030e7e0 str("exists") :: ?2
        ?1 = ?2
```
I don't really understand what `const.res.file.state.exists` is supposed to be. Builtins like `printf` require an `import "fmt"`, but this test has no imports, so I guess `$const` is a special builtin which doesn't need to be imported? The comment certainly implies so.

I am guessing that magic variables use a separate codepath which bypasses the code I have added for the non-magic codepath. Maintaining both codepaths is of course more effort than supporting a single codepath; is this extra effort worth it? What is this magic variable providing over an equivalent imported variable, and where are magic variables defined?

**Comment by gelisam** [comment]
The `slow-unification2.txtar` failure is very unusual! I tried to simplify the test file, and this version:
```
-- main.mcl --
import "datetime"

$now = 0 
print "test1" { 
        msg => if datetime.weekday($now) == "friday" { "TGIF YAY!" } else { "meh..." },
} 
print "test2" { 
        msg => "FLIP",
} 
-- OUTPUT --
Vertex: print[test1]
Vertex: print[test2]
```
Passes 55% of the time. The original version only passes 10% of the time.

Since none of my changes in this MR involve threads, I am guessing that this is an existing race condition which used to work fine because the timings of the threads was just right, and this MR changes the timings just enough to expose the race condition. The timings probably vary from machine to machine, too, so I bet you and CI will get different results than me.

I am thus inclined to disregard the `field arg0 does not exist` failures as existing bugs which already exist on master, even though the timings are such that master passes ~100% of the time.

purpleidea:
> It's pretty strange that stmtfunc-recursive-double.txtar and stmtfunc-recursive-no-operators.txtar have the exact same contents, probably a mistake?

Yeah definitely. I probably meant to change the operators in the no-operators version. We could either delete `stmtfunc-recursive-no-operators.txtar` or change it if you find something valuable to do so.

purpleidea:
> I don't really understand what const.res.file.state.exists is supposed to be. Builtins like printf require an import "fmt", but this test has no imports, so I guess $const is a special builtin which doesn't need to be imported? The comment certainly implies so.

So $const is a magic hierarchy of variables... The one in question is simply a string "exists".
If that test fails with a raw string "exists", then I don't know what's wrong. Something you changed I guess. _However_ if it works with a raw string, then we know where to narrow things down... This would be in the importing variables code path.

The libraries for that is here: https://github.com/purpleidea/mgmt/blob/master/lang/funcs/vars/vars.go

Each of them is basically any constant types.Value... And then this function: VarPrefixToVariablesScope is run in lang/lang.go (or lang/interpret_test.go for tests) and that gets used in the top-level scope.

So if it's not working, SetScope is broken.

purpleidea:
>  I guess $const is a special builtin which doesn't need to be imported? The comment certainly implies so.

It's part of the top-level scope. It's actually an ExprStr or and ExprBool or whatever the appropriate one is.

>  Maintaining both codepaths is of course more effort than supporting a single codepath

Are you saying it would be smarter to do "import const" and then const.whatever... ? I actually really like this idea.

purpleidea:
> Passes 55% of the time. The original version only passes 10% of the time.

There is indeed a race with the function engine unfortunately. I have to spend some time to debug it :(

> the field arg0 does not exist failures

But I wouldn't expect to see this error as the intermittent failure. Is that what you see?

gelisam:
> Are you saying it would be smarter to do "import const" and then const.whatever... ? I really like this idea.

In this MR, I added `ExprTopLevel` to `FuncPrefixToFunctionsScope`, I think that's the `printf` codepath? Thus, when I saw that `fmt.printf` worked but `$const` didn't, my first thought was that `$const` probably doesn't use `FuncPrefixToFunctionsScope`, so that's why it's broken.

I know you added a secret `$purpleidea` variable, using a completely separate codepath from every other variable, so I bet that variable is broken as well. I assumed that `const` and `$purpleidea` were using the same codepath, but now that you mention `VarPrefixToVariablesScope`, I see that the more important difference between `fmt.printf` and `$const` isn't the `fmt.`, it's the `$`. So the fix must be in `VarPrefixToVariablesScope`!

gelisam:
Oh, I know what's wrong! `ExprVar` used to copy its target and then call `Unify()` on it. Now, it no longer copies its target, and also it no longer calls `Unify()` on it, because that would cause the constraints to be added once per use site rather than once per definition. Conversely, `StmtBind` used not to call `Unify()` on its definition, and now it does. So for variables bound by `StmtBind`, the `Unify()` call is now made in a different spot, but importantly, the call is still made.

The variables introduced by `VarPrefixToVariablesScope` are not introduced by `StmtBind`, so there is currently nothing which calls `Unify()` on them. How was `Unify()` called on them before the mcl-functions branch? Does it make sense to call `Unify()` on _all_ the magic variables? If so, from where? Does the caller of the top-level `StmtProg.Unify()` have access to the list of all the magic `Expr`s? If it does not make sense to call `Unify()` an the unused magic variables, is there a function which lists the magic variables used by the program?

purpleidea:
> How was Unify() called on them before the mcl-functions branch?

They were in the scope, so it got called when one used one.

> Does it make sense to call Unify() on all the magic variables?

Just the ones that are consumed somewhere I think.

> If so, from where?

Whatever pulls them in from the scope calls Unify, no?

> Does the caller of the top-level StmtProg.Unify() have access to the list of all the magic Exprs?

Unsure.

> If it does not make sense to call Unify() an the unused magic variables, 

It does not, I believe...

> is there a function which lists the magic variables used by the program?

I mean, the premise of a top-level scope, that we propagate through... Is this not right? If it is, then why can't the top-level scope contain something. If not, then well... I don't know how imports and everything else work...

So I still think it makes sense to have some top-level variables. $hostname being an important one, and for most of them, it probably make sense to move them to a "const" package.

gelisam:
> They were in the scope, so it got called when one used one. [...] Whatever pulls them in from the scope calls Unify

Ok, I'll change it back so that `ExprVar` calls unify instead of `StmtBind`. Don't complain that there is one copy of the `Unify()` constraints for each use site!

purpleidea:
This is merged in #725 \o/

Issue #717, opened 2023-08-30
Discussion on graph shape


purpleidea:
We left off here...

> if we want to write an optimization which notices that a sub-graph is no longer changing

What was the advantage of doing that rather than just not generating a subgraph via Txn to begin with for the nodes that won't ever need it? IOW, what is the downside of just generating those function graphs to be the same shape as they currently are in git master? (For the non-lambda functions of course.)

> It might also be necessary for expressions to be able to express that they will never create a sub-graph

We can add a field to function Info to say whether or not they will need Txn or not. I don't mind manually labelling each function. It's actually easy, the default can be false, and then we have a single function that needs labelling as true so far (iter.map).

gelisam:
> What was the advantage of doing that rather than just not generating a subgraph via Txn to begin with

My suggestion is to write a more general optimization, one which will work both for static expressions and for dynamic expressions which stop changing mid-way through the execution. Let's detect when a sub-graph stops changing, and then let's rewrite the nodes around it so that it looks as if that sub-graph was always part of the original graph. I think this optimization fits better with the existing optimization when Stream() indicates that it will no longer send values downstream.

gelisam:
> It's actually easy

Because the functions themselves are the easy part, they are the base case! The more interesting part of the analysis is to look at an _expression_ which returns a function, and to determine whether it is constant or not.

gelisam:
You were right, I have changed my mind about this: let's generate the optimized sub-graph right away rather than rewrite the graph at runtime!

The reason I changed my mind is because I now see that if we do, then we can use the same static analysis for this feature and for the "send a function to a resource" feature. This static analysis determines if a given function or expression is "timeless", meaning that it doesn't do shenanigans like sending more than one value downstream or waiting until it receives multiple inputs before sending that single output.

There are several variants to this question:
1. For a Func, we want to know that _if_ the Func receives one record of input values, then it will emit one output value.
2. For a FuncValue, we want to know that _if_ the input Funcs emit a single value, then the output Func will emit a single value.
3. For an Expr, we want to know that the Func returned by `Expr.Graph()` will only ever output a single value. This implies that the Func itself is timeless, and so are its inputs, and so on transitively up to the Funcs with in-degree zero.

[...]

PR #724, opened 2023-12-16
optimize the graph shape to not create sub-graphs when possible

gelisam:
Working on #717

### DIFF
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -7787,14 +7787,19 @@ func (obj *ExprFunc) SetValue(value types.Value) error {
 // This particular value is always known since it is a constant.
 func (obj *ExprFunc) Value() (types.Value, error) {
 	if obj.Body != nil {
-		env := make(map[string]interfaces.Func) // XXX ??? SAM will decide what to do here
+		// We can only return a Value if we know the value of all the ExprParams.
+		// We don't have an environment, so this is only possible if there are no
+		// ExprParams at all.
+		if err := obj.CheckParamScope(make(map[interfaces.Expr]struct{})); err != nil {
+			// return the sentinel value
+			return nil, ErrCantSpeculate
+		}
+
 		return &full.FuncValue{
 			V: func(innerTxn interfaces.Txn, args []interfaces.Func) (interfaces.Func, error) {
-				// Extend the environment with the arguments.
+				// There are no ExprParams, so we start with the empty environment.
+				// Extend that environment with the arguments.
 				extendedEnv := make(map[string]interfaces.Func)
-				for k, v := range env {
-					extendedEnv[k] = v
-				}
 				for i, arg := range obj.Args {
 					extendedEnv[arg.Name] = args[i]
 				}
@@ -7813,7 +7818,6 @@ func (obj *ExprFunc) Value() (types.Value, error) {
 			},
 			T: obj.typ,
 		}, nil
-
 	} else if obj.Function != nil {
 		return structs.FuncToFullFuncValue(obj.function, obj.typ), nil
 
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -8612,10 +8612,10 @@ func (obj *ExprCall) Graph(env map[string]interfaces.Func) (*pgraph.Graph, inter
 					},
 				}.Init())
 				args := []interfaces.Func{}
-				for _, arg := range obj.Args { // []interfaces.Expr
+				for i, arg := range obj.Args { // []interfaces.Expr
 					g, f, err := arg.Graph(env)
 					if err != nil {
-						return nil, nil, errwrap.Wrapf(err, "could not even XXX")
+						return nil, nil, errwrap.Wrapf(err, "could not get graph for arg %d", i)
 					}
 					args = append(args, f)
 					txn.AddGraph(g)
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -7824,7 +7824,21 @@ func (obj *ExprFunc) Value() (types.Value, error) {
 	} else if obj.Function != nil {
 		// TODO: if the builtin is timeless, use SimpleFnToFuncValue instead of
 		// FuncToFullFuncValue
-		return structs.FuncToFullFuncValue(obj.Title, obj.function, obj.typ), nil
+
+		// Copy obj.function so that the underlying ExprFunc.function gets
+		// refreshed with a new ExprFunc.Function() call. Otherwise, multiple
+		// calls to this function will share the same Func.
+		exprCopy, err := obj.Copy()
+		if err != nil {
+			return nil, errwrap.Wrapf(err, "could not copy expression")
+		}
+		funcExprCopy, ok := exprCopy.(*ExprFunc)
+		if !ok {
+			// programming error
+			return nil, errwrap.Wrapf(err, "ExprFunc.Copy() does not produce an ExprFunc")
+		}
+		valueTransformingFunc := funcExprCopy.function
+		return structs.FuncToFullFuncValue(obj.Title, valueTransformingFunc, obj.typ), nil
 	}
 	// else if /* len(obj.Values) > 0 */
 	panic("what to do here")
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -7220,7 +7220,6 @@ func (obj *ExprFunc) Copy() (interfaces.Expr, error) {
 
 // Ordering returns a graph of the scope ordering that represents the data flow.
 // This can be used in SetScope so that it knows the correct order to run it in.
-// XXX: do we need to add ordering around named args, eg: obj.Args Name strings?
 func (obj *ExprFunc) Ordering(produces map[string]interfaces.Node) (*pgraph.Graph, map[interfaces.Node]string, error) {
 	graph, err := pgraph.NewGraph("ordering")
 	if err != nil {
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -7687,7 +7687,7 @@ func (obj *ExprFunc) Graph(env map[string]interfaces.Func) (*pgraph.Graph, inter
 	var funcValueFunc interfaces.Func
 	if obj.Body != nil {
 		funcValueFunc = structs.FuncValueToConstFunc(&full.FuncValue{
-			V: func(innerTxn interfaces.Txn, args []interfaces.Func) (interfaces.Func, error) {
+			Timeful: func(innerTxn interfaces.Txn, args []interfaces.Func) (interfaces.Func, error) {
 				// Extend the environment with the arguments.
 				extendedEnv := make(map[string]interfaces.Func)
 				for k, v := range env {
@@ -7716,8 +7716,10 @@ func (obj *ExprFunc) Graph(env map[string]interfaces.Func) (*pgraph.Graph, inter
 		// an output value, but we need to construct a node which takes no
 		// inputs and produces a FuncValue, so we need to wrap it.
 
+		// TODO: if the builtin function is known to be timeless, use Timeless
+		// instead of Timeful
 		funcValueFunc = structs.FuncValueToConstFunc(&full.FuncValue{
-			V: func(txn interfaces.Txn, args []interfaces.Func) (interfaces.Func, error) {
+			Timeful: func(txn interfaces.Txn, args []interfaces.Func) (interfaces.Func, error) {
 				// Copy obj.function so that the underlying ExprFunc.function gets
 				// refreshed with a new ExprFunc.Function() call. Otherwise, multiple
 				// calls to this function will share the same Func.
@@ -7797,7 +7799,7 @@ func (obj *ExprFunc) Value() (types.Value, error) {
 		}
 
 		return &full.FuncValue{
-			V: func(innerTxn interfaces.Txn, args []interfaces.Func) (interfaces.Func, error) {
+			Timeful: func(innerTxn interfaces.Txn, args []interfaces.Func) (interfaces.Func, error) {
 				// There are no ExprParams, so we start with the empty environment.
 				// Extend that environment with the arguments.
 				extendedEnv := make(map[string]interfaces.Func)
@@ -7820,8 +7822,9 @@ func (obj *ExprFunc) Value() (types.Value, error) {
 			T: obj.typ,
 		}, nil
 	} else if obj.Function != nil {
-		return structs.FuncToFullFuncValue(obj.function, obj.typ), nil
-
+		// TODO: if the builtin is timeless, use SimpleFnToFuncValue instead of
+		// FuncToFullFuncValue
+		return structs.FuncToFullFuncValue(obj.Title, obj.function, obj.typ), nil
 	}
 	// else if /* len(obj.Values) > 0 */
 	panic("what to do here")
@@ -8709,7 +8712,7 @@ func (obj *ExprCall) Graph(env map[string]interfaces.Func) (*pgraph.Graph, inter
 					args = append(args, f)
 					txn.AddGraph(g)
 				}
-				outputFunc, err := exprFuncValue.Call(txn, args)
+				outputFunc, err := structs.CallFuncValue(exprFuncValue, txn, args)
 				if err != nil {
 					return nil, nil, errwrap.Wrapf(err, "could not construct the static graph for a function call")
 				}
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -8553,9 +8553,32 @@ func (obj *ExprCall) Graph(env map[string]interfaces.Func) (*pgraph.Graph, inter
 		graph.AddVertex(paramFunc)
 		funcValueFunc = paramFunc
 	} else {
-		// The function being called is a top-level definition. The parameters which
-		// are visible at this use site must not be visible at the definition site,
-		// so we pass an empty environment.
+		// The function being called is a top-level definition.
+                
+		// Optimization: in the common case in which the function is
+		// statically-known, generate a single static sub-graph.
+                exprValue, err := obj.expr.Value()
+                if err == nil {
+			exprFuncValue, ok := exprValue.(*full.FuncValue)
+			if ok {
+				struct GraphTxn {
+					Graph pgraph.Graph
+				}
+				txn := GraphTxn()
+				outputFunc, err := exprFuncValue.Call(txn, []Func)
+				if err != nil {
+					return nil, nil, errwap.Wrapf(err, "could not construct the static graph for a function call")
+				}
+				return txn.Graph, outputFunc, nil
+			}
+		}
+
+		// Otherwise, generate a CallFunc node, which will dynamically
+		// re-create the sub-graph as needed.
+                
+                // The parameters which are visible at this use site must not
+                // be visible at the definition site, so we pass an empty
+                // environment.
 		emptyEnv := map[string]interfaces.Func{}
 		exprGraph, topLevelFunc, err := obj.expr.Graph(emptyEnv)
 		if err != nil {
--- a/lang/ast/structs.go
+++ b/lang/ast/structs.go
@@ -8809,13 +8809,7 @@ func (obj *ExprCall) Value() (types.Value, error) {
 		args = append(args, a)
 	}
 
-	_ = funcValue
-	panic("IMPLEMENT THIS")
-	// we now have a full.FuncValue and a []Value. We can't call the existing
-	//   Call([]Func) Func
-	// method on the FuncValue, we need a speculative
-	//   Call([]Value) Value
-	// method. so the next step is to implement that method.
+	return structs.CallTimelessFuncValue(funcValue, args) // speculative
 }
 
 // ExprVar is a representation of a variable lookup. It returns the expression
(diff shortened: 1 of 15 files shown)


gelisam:
Here is a corner case.
```
$apply1 = func($f, $x) {
  $f($x)
}
$apply2 = func($f, $x) {
  $f($f($x))
}
$apply = if ($now % 2 == 0) then $apply1 else apply2
$id = func($y) { $y }
test $apply($id, "foo")
```
Every second, `$apply` will switch from one constant expression to another. Because `$apply` changes, the `$apply($id, "foo")` call needs a dynamic sub-graph. With the current implementation, each of those sub-graphs will in turn use `CallFunc`s for the `$f($x)` calls, thus resulting in sub-sub-graphs. Since `$id` is statically-known, it should theoretically be possible to for each instantiation of the sub-graph to use the optimization, that is, to skip the `CallFunc`s and avoid the need for sub-sub-graphs. Unfortunately, since `Value()` uses the scope and does not have `Graph()`'s `sctx` argument, it is not possible to determine what `$f` refers to, and thus `Value()` must return `nil`.

Is it worth to add the `sctx` argument in order to address this case, or is this too small of a corner case?

purpleidea:
> Is it worth to add the sctx argument in order to address this case, or is this too small of a corner case?

I do not know, but let's please definitely add this as a test =D

If for weird examples like this we have a dynamic graph, it is okay. What's the `sctx` argument??

gelisam:
Another corner case:
```
$f1 = func($x) { $x }
$f2 = func($x) { $x + $x }
$id = func($y) { $y }
$f = if $id(true) then $f1 else $f2
test $f("foo")
```
No higher-order functions this time. In order for `$id(true)` to be statically-known, we need to:
1. speculatively obtain a `full.FuncValue` for `$id`
2. speculatively obtain a `Value` for `true`
3. call the `full.FuncValue` on the input `Value` in order to obtain an output `Value`

The hard step is (3), because the type
```
func (obj *full.FuncValue) Call(txn Txn, args []Func) (Func, error)
```
is not what we need. We need
```
func (obj *full.FuncValue) SpeculativeCall(args []Value) (Value, error)
```

is it worth implementing `SpeculativeCall` in order to support this use case?

purpleidea:
> is it worth implementing SpeculativeCall in order to support this use case?

This case (or something that looks like it) seems more plausible, but still in the 20%.

I think our goal for `step one` should be to initially make 80% of common mcl code use the simple graph shape, as long as we can extend our optimizations out as a `step two` without re-writing all our plumbing for `step one`.

Let's add this as a test btw =D

gelisam:
> What's the `sctx` argument??

`SetScope`'s second argument:
```
func (obj *Expr) SetScope(scope *Scope, sctx map[string]Expr) error
```
in the `mcl-functions` branch I named it `context`, I don't know why you renamed it to `sctx`. In
```
$f = func($x) {
  func($y) {
    ...
  }
}
```
it maps `$x` and `$y` to different `ExprParam`s.

gelisam:
Yet another corner case. Lambdas are always constant, but in
```
$add = func($x) {
  func($y) {$x + $y}
}
test $add("foo")("bar")
```
the nested `FuncExpr.Value()` will not be able to produce a `full.FuncValue` representing `func($y) {$x + $y}`, because once again `.Value()` does not receive an `sctx`.

Actually, this might not be a corner case at all... The `FuncExpr` doesn't know whether it is a nested lambda, so it doesn't know whether the variables in its body are supposed to refer to top-level definitions or to parameters from surrounding lambdas. Thus, it is never safe for `FuncExpr.Value()` to return a `FuncValue`! If `FuncExpr.Value()` tried to return a `FuncValue` in the case in which all the variables of its body are in `obj.scope`, then
```
$x = "not this x"
$add = func($x) {
  func($y) {$x + $y}
}
test $add("foo")("bar")
```
would `FuncExpr.Value()` would incorrectly generate a `FuncValue` corresponding to `func($y) {$x + $y}` where `$x` refers to `"not this x"` instead of referring to the outer lambda's parameter.

gelisam:
It thus looks like `Expr.Value()` should take an `sctx` parameter. Does it still make sense to use the same `Expr.Value()` for obtaining the `Value` to give to a resource, or would it make more sense to split `Expr.Value()` into `Expr.ValueForResource()` and `Expr.SpeculativeValue(sctx)`?

purpleidea:
> Does it still make sense to use the same Expr.Value() for obtaining the Value to give to a resource, or would it make more sense to split Expr.Value() into Expr.ValueForResource() and Expr.SpeculativeValue(sctx)?

Hmmm...

Is it still possible to do a `step one` where we don't do any `sctx` stuff, just to merge it and have _some_ easy optimizations done, and grow our confidence in the optimization for those easy things?

The advantage of doing this in two steps are:

1) There's a bigger chance I more fully understand what's going on.
2) There's a chance we have a new realization that precludes needing the sctx thing (maybe unlikely)

The down side is we have extra commits in master, but that just means more points! \o/

It will also give us a chance to briefly verify if the signature of SetScope:

```SetScope(scope *interfaces.Scope, sctx map[string]interfaces.Expr)```

Still makes sense or if this should be a single combined arg or something different.

Actually a good question is: Are you currently happy/okay with the design of SetScope or is it broken in your mind as a stage?

**Comment by gelisam** [comment]
right, right, you want me to make these kinds of decisions on my own... I say we keep a single `Expr.Value(sctx)`.

Except we make this speculative call from `Graph(env)`, not `SetScope`, so it needs to be an `env: []Func`, not an `sctx: []Expr`.

**Comment by purpleidea** [comment]
> right, right, you want me to make these kinds of decisions on my own

You're in charge, I'm meant to be here to complain if my spider sense goes off, or if I feel you're writing pure/functional code instead of elegant, imperative, faster golang code ;)

**Comment by gelisam** [comment]
> Is it still possible to do a step one where we don't do any sctx stuff, just to merge it and have some easy optimizations done, and grow our confidence in the optimization for those easy things?

Yes, if we give up on optimizing inside lambdas. `ExprCall.Graph()` will have to check if `sctx` is non-empty, and if so, assume that calling `Value()` on the function would give the wrong result, and thus use dynamic sub-graphs.

**Comment by purpleidea** [comment]
> if so, assume that calling Value() on the function would give the wrong result

So currently we use Value() here for printf... Does this mean that this is currently incorrect? Would we have to plumb through the sctx parameter? (This call happens during unification to inspect the format string.)

https://github.com/purpleidea/mgmt/blob/6a6546db8d3d481cfaf5adcb71c2329d1284cbd3/lang/funcs/core/fmt/printf_func.go#L280

**Comment by gelisam** [comment]
I'd try this:
```
$s = "%d"
$f = func($s) {
  print($s, 42)
}
test $f("%s") {}
```
I suspect that since Value doesn't have an sctx parameter, when printf calls `Value()` on the `$s`, it might see the top-level `$s = "%d"` and incorrectly _accept_ this program instead of failing with a type error which says that we don't know the value of the format string and thus can't check if 42 has the right type.

**Comment by purpleidea** [comment]
> I'd try this:

Indeed there's an issue! Actually I found a sort-of related bug. Hmmm...

Well here's the interesting point: plumbing through sctx would be very complicated... Which leads me to think that the more appropriate solution is one in which sctx is "captured" within the Expr AST node, so that we don't need to plumb it through the API...

**Comment by gelisam** [comment]
That would not work because the same Expr is used for several lambda calls. You can only store one sctx in the Expr itself, but you can pass a different sctx each time Graph() or Value() is called.

**Comment by purpleidea** [comment]
Yes, but if we have to have an sctx inside of Unify, that means we need to plumb this into the unification engine...

I feel like the right approach might involve an Initial Expr* whatever ast node, with it's own sctx, but then when it's used we copy it and have a new sctx contained within for each different Expr?

Or something else?

**Comment by gelisam** [comment]
Remember, we used to do that, but we had to change course in order to support `map`! Before mcl-functions, each `ExprCall` copied the body of the lambda it called, and then `SetScope` was able to make each `$x` use site inside a lambda `func($x) {...}` point to the argument `$f(arg)` of the corresponding call site. We did all that before type checking.

We can't do that with `map`, because the number of call sites is not known until runtime, after type checking. So we had to switch to a different implementation strategy in which a `full.FuncValue` generates a sub-graphs. Thanks to those sub-graphs, it is possible to change the shape of the function graph at runtime, so that we end up with one `CallFunc` for each call site even though the number of call sites is not known until runtime.

But we only create more Funcs, we don't create more Exprs! And we don't need to: lambdas are monomorphic, so the `$x` is `func($x) {...}` has only one type, so it only needs one copy.

**Comment by purpleidea** [comment]
From our phone conversation:

https://github.com/purpleidea/mgmt/blob/6a6546db8d3d481cfaf5adcb71c2329d1284cbd3/lang/ast/structs.go#L8802
https://github.com/purpleidea/mgmt/blob/6a6546db8d3d481cfaf5adcb71c2329d1284cbd3/lang/ast/structs.go#L8685

Both of these ExprParam and ExprPoly return nil, nil. Which we probably don't want.

This might come from here: https://github.com/purpleidea/mgmt/blob/6a6546db8d3d481cfaf5adcb71c2329d1284cbd3/lang/ast/structs.go#L8539

**Comment by purpleidea** [comment]
(I will add a patch to add that test in and add errors on those two nils.)

**Comment by gelisam** [comment]
> From our phone conversation

More importantly for this MR, since `ExprVar.SetScope()` does
```
obj.scope.Variables[obj.Name] = sctx[obj.Name]
```
There is no need to add `sctx` to `Value()` after all.

**Comment by purpleidea** [comment]
Here's the patch (just testing)

https://github.com/purpleidea/mgmt/commit/a23312a913cb5263a2cbb90b2c4c3403276b2886

Sort of related:

```
import "fmt"
$format = "%d"	# should get ignored
$fn = func($format) {
	fmt.printf($format, 42)
}
test $fn("%s") {}
# should error at unification if possible, otherwise at runtime
```

So btw for this test, with the dynamic printf enabled ( `PrintfAllowNonStaticFormat = true`) this fails at runtime, and with it off, it fails during type unification. I guess that's fine for now. Of course if all the format strings are "%d", then this still can't unify with `PrintfAllowNonStaticFormat = false`. Which is too bad, but seems fine for now.

Why do I mention any of this? Because speculatively, I would expect this func to be run at compile time...

**Comment by purpleidea** [comment]
(Merged in master.)

**Comment by gelisam** [comment]
> I would expect this func to be run at compile time

Wait, do you want to evaluate statically-known expressions at compile-time or to generate a graph with no sub-graphs when possible? It's not the same thing!

**Comment by gelisam** [comment]
Huh, the code on this branch doesn't even compile? 7b5466e fails with
```
lang/funcs/facts/func.go:90:18: obj.Fact.Call undefined (type Fact has no field or method Call)
```

Instead of starting from a broken #716, I'll start from master. I'll cherry-pick from #716 later if that turns out to be useful for this MR.

**Comment by gelisam** [comment]
I'm confused, why does this say that I have closed the PR??

Oh! I have reverted `gelisam/graph-shape` to point to `master`, so now it looks like `master` contains all the commits from `gelisam/graph-shape` and thus that the branch has been merged! Well played GitHub, this is a clever way to avoid having to store one boolean per PR indicating whether it has been merged or not.

**Comment by purpleidea** [comment]
> lang/funcs/facts/func.go:90:18: obj.Fact.Call undefined (type Fact has no field or method Call)

Oh, hm weird. Maybe I forgot to include something in my commit? if you need it LMK and I'll make sure it's working.

**Comment by gelisam** [comment]
I have created tests based on the counter-examples in this thread, let's see if GitHub will allow me to reopen the PR...

**Comment by gelisam** [comment]
I have now made sure the new tests pass on CI, not just on my machine, before I start modifying the code.

Did we decide whether we would implement FakeTxn or if we would change Graph() to receive a Txn instead of returning a Graph?

**Comment by purpleidea** [comment]
> Did we decide whether we would implement FakeTxn or if we would change Graph() to receive a Txn instead of returning a Graph?

We didn't decide, and I'm also undecided. If you feel strongly about one or the other, let's do that. I can't visualize all the implications yet.

**Comment by purpleidea** [comment]
> I can't visualize all the implications yet.

Which really means, we should go with whatever you think is best, and hopefully we don't discover some reason that makes us need to change our minds after of course =D

**Comment by gelisam** [comment]
Last time we met, I was confused about `Expr.Graph()`'s `env` parameter. It represents the subset of variables which are currently in scope which come from a function parameter. I was worried that the a call site's `env` might be passed to the implementation, which would be bad if there is an `$x` which is in scope at both places, but it's not the same `$x`. Like this:
```
$make_exclaimer = func($x) {
  func() {$x + "!"}  # at the definition site, $x is "foo"
}
$use_exclaimer = func($f, $x) {
  $f() + $x  # at the call site, $x is "bar"
}
test $use_exclaimer($make_exclaimer("foo"), "bar") {}
```

Luckily, I wrote some comments in `ExprFunc.Graph()` and `ExprCall.Graph()` which clarify how I am handling the situation.

There are actually two separate cases. In the easy case, the definition site is a top-level definition:
```
$x = "foo"
$exclaimer = func() {$x + "!"}  # at the definition site, $x is "foo"
$use_exclaimer = func($f, $x) {
  $f() + $x  # at the call site, $x is "bar"
}
test $use_exclaimer($exclaimer, "bar") {}
```

In that case the definition site's `env` should be empty, and the call site clears out the `env` when making the recursive `Graph()` call, so everything is fine.

In the more complicated case, the definition site is inside another function, in this case `$make_exclaimer`. This time there is no recursive call to `Graph()`, so there is no opportunity for the call site's `env` to be accidentally given to `ExprFunc.Graph()`. Instead, the `Func`s for the parameters have already been created earlier and the `env` is holding on to them. It is when that already-created `Func` was created that the definition site's `env` was captured. The call site doesn't call this `Func` directly, instead it adds an `CallFunc` of its own, and it is only a runtime when that `Func` sends a `full.FuncValue` to that `CallFunc` that the `env` hidden inside this `full.FuncValue` comes into play.

**Comment by gelisam** [comment]
For `Value`, this is a bit more complicated because we don't know whether if we're in the easy case or the complicated case. So the first step is to figure that out.

It should be possible to figure it out by traversing the body and check that all of the `ExprVar`s refers to an `ExprParam` which is either introduced by the lambda itself or by a nested lambda. If so, we are in the easy case and `env` should be empty. Otherwise, we are in the more complicated case, which we can keep for later.

**Comment by gelisam** [comment]
Actually, `ExprCall.Graph()` only calls,`Value()` in that easy case, so in order to validate the approach, we can just unconditionally use the empty `env`. But that would cause runtime crashes in other circumstances, so we definitely can't skip the check in the name of only supporting the optimization for the rest case.

**Comment by gelisam** [comment]
@purpleidea I have now dealt with the `env` in `ExprFunc.Value()`. The optimization should now work in the most common cases, but I can't check because the code doesn't compile. It wasn't compiling for me before I made my changes either.

**Comment by purpleidea** [comment]
Sweet. New branch available with your changes and fixes in feat/graph-shape. More comments here: https://github.com/purpleidea/mgmt/issues/717#issuecomment-1909226779

**Comment by purpleidea** [comment]
Sorry I've been a bit AFK, I'm mid conferencing and gave my first presentation today. One more on Monday, and then a workshop on Wednesday =D Thanks for all your great work. Ping me for anything you're stuck on or busy work that you need me to refactor/golangify/etc.

**Comment by gelisam** [comment]
> Sorry I've been a bit AFK

Me too, I was in NYC for work! I have also been elected as one of the administrators of my building, and it's taking a lot of time. I was actually working on mgmt to relax a bit from all that 😅

> I'm mid conferencing and gave my first presentation today. One more on Monday, and then a workshop on Wednesday =D

Woohoo! I hope everybody is as excited about mgmt as you were hoping!

> Ping me for anything

Not a chance, focus on your conference!

**Comment by gelisam** [comment]
165 / 206 tests pass, not too bad but there is still a lot of work to do!

**Comment by gelisam** [comment]
up to 183 / 206!

**Comment by purpleidea** [comment]
Sweet! Remember TestAstFunc1 doesn't matter since that's just shapes which
will change.

**Comment by gelisam** [comment]
Oh, I didn't even look at TestAstFunc1, those are all TestAstFunc2 failures.

Here's the next failure:
```
$f = func($x) {$x + "!"}
test $f("a") {}
test $f("b") {}
```

Since the problem only occurs when the function is called twice, I suspect that we're trying to Revert () the nodes added by the first call, so they're no longer available for the second call.

**Comment by gelisam** [comment]
up to 195 / 206!

**Comment by gelisam** [comment]
The next failure is `chained-vars.txtar`, slightly tweaked to make the labels unique:
```
import "fmt"

$zero = 0
$one = $zero + 1
$two = $one * 2 # needs a chain to panic

test fmt.printf("%d%d%d", $zero, $one, $two) {}
```

The test always generates the wrong Func graph, but it is non-deterministic, different runs generate different incorrect Func graphs:

failure1
failure2
failure3
failure4
failure5
failure6

I am thus guessing that when we generate the static graph and capture it via the Txn which isn't connected to the engine, the code which generates the static graph spawns some threads and not all the nodes have been added by those threads when the graph is extracted from the Txn.

purpleidea:
I'm trying to think what would race there, that feels a bit unusual unless
there was something concurrent added in this branch. I'll look at the diff
tonight. Could there be a bug or an API mismatch somewhere in the TXN code?

I'll be back in a few days and on a regular hacking schedule again.

Thanks for your work here!

gelisam:
I think the problem is as follows.

For dynamic graphs, it is normal for a Func to hold onto the Txn and to use it to add more nodes asynchronously over time. For the optimization, I need the opposite: I need the `full.FuncValue` to add all of its nodes synchronously, before we extract the static graph from the Txn. I assumed that in those cases in which we know that the graph is static, the `full.FuncValue` must be adding all of its nodes synchronously. But I was probably mistaken: the graph will quickly reach an equilibrium, but it might do so asynchronously.

gelisam:
I tested my hypothesis by sleeping for three seconds before extracting the graph for the Txn, but that did not fix the problem. Thus, it is _not_ the case that the graph quickly reaches an equilibrium.

Maybe it's not threads, but the random order in which the elements of a map are traversed?

gelisam:
Let's see, so `ExprInt.Graph()` is called on `0`, and a ConstFunc is added to the Func graph.

Then, `ExprCall.Graph()` is called on `$zero + 1`. `ExprFunc.Value()` is called speculatively on `+`, and it successfully returns a `full.FuncValue. In this branch, there are now two kinds of `full.FuncValue`s: timeless and timefull. You and I know that `+` is timeless, but we haven't taught mgmt about this fact yet, so the returned `full.FuncValue` is actually timefull.

Timeless or timefull, the optimization fires regardless, so the `full.FuncValue` is only called once. This particular `full.FuncValue` comes from `FuncToFullFuncValue`, which adds `_operator`'s three input edges, synchronously. Looks like there is a step missing to actually add the `_operator` itself, that's fine here since adding an edge also adds the input and output nodes, but this is going to cause problems for functions which take zero arguments.

Everything so far seems to work as intended, since that `_operator` node and its input edges are always present in the GraphViz representation.

Next, `ExprCall.Graph()` is called on `$one * 2`. `ExprFunc.Value()` is called speculatively on `*`, and presumably it again successfully returns a timefull `full.FuncValue`. Thus, the optimization fires again.

The `FuncToFullFuncValue` adds another `_operator` and its three input edges, synchronously. But this time, something goes wrong, because only a random subset of those edges appears in the GraphViz representation.

What are the differences between the `+` and `*` cases? I can't think of any. They both use the function `_operator`, they both have one argument which is a variable and whose target is already in the Func graph when `ExprCall.Graph()` is called, and one argument which is a number and is not yet in the Func graph. The one obvious difference is that `$one`'s target is an `_operator` which already has input edges coming into it, instead of a `ConstFunc` with no inputs, but what does that change?

Another difference is that `$two` comes later in the Ordering dag than `$one`. So if there are two threads running and there is a race, maybe the first thread always has time to process `$one` but is still in the middle of processing `$two` when the second thread wins the race. And the `printf`, which is just completely missing from the Func graph, never even gets a chance to get processed. The fact that adding a three second sleep just before extracting the graph from the Txn didn't change anything argues against this idea, but since I'm running out of ideas, er, maybe there's a different place where sleeping would make a difference?

purpleidea:
> The FuncToFullFuncValue adds another _operator and its three input edges

This jumped out at me... Is the pointer of _operator unique or is there a chance you forgot to copy it or something and it's actually the _same_ operator pointer as previously used?

If that's not it, then don't worry about this, we'll debug together after I look at the code more. I'm back in two days and I'll be focusing more on coding and not demoing/etc. Cheers

gelisam:
> is there a chance you forgot to copy it or something and it's actually the same operator pointer as previously used?

I did forget initially, and one of the tests caught it, and that mistake has since been fixed. But who knows, maybe I fixed it incorrectly, just enough to fix that other test but not well enough to fix this one as well?

Nah, there are clearly two `_operator` nodes on that GraphViz representation, I don't think that's it. But it's worth trying a variant with a built-in other than `_operator`!

gelisam:
Nope, same behaviour if I use `contains` instead of `*`

gelisam:
Printing the graphs extracted from each Txn, I see some unexpected results. I expected the first Txn to contain 0, 1, `+`, and `_operator1`, the second Txn to contain `_operator1`, 2, `*`, and `_operator2`, and the third Txn to contain `"%d%d%d"`, 0, `_operator1`, `_operator2`, and `printf`. Instead, each Txn contains what I expect _plus_ part of three previous Txn. The third Txn's graph is the one shown in GraphViz.

There are thus two bugs and one source of non-determinism. The first bug is that each Txn contains the dependencies of the inputs, when they should only contain one node per input. The source of non-determinism is that only a subset of the spurious dependencies are added. Finally, the second bug is that the last Txn's graph seems to be the final graph, when instead the final graph should be the union of the graphs from all the Txn's.

PR #725, opened 2023-12-27
Proposed cleanup merge for https://github.com/purpleidea/mgmt/pull/723

purpleidea:


purpleidea:
@gelisam Here's my proposed merge for your excellent #723. I did some small cleanups and squashing, and everything passes now. NOTE: that due to an existing race (not yet fixed) it does sometimes deadlock in CI, but it seems to work great locally at least for now.

purpleidea:
I was wrong about the reason for the failure. It's actually because we now use `test.one_instance_a()` in more than one test... When called individually, things are fine, but because that fact uses global state to check if it's called twice, when we run the tests all together it panics! (correctly so!)

haha!

So let me fix that up so we can still test this behaviour but not fail things because of it. Neat!

purpleidea:
Fixed!

@gelisam If you don't object, I will merge this shortly. I've kept the main commit in your name even though it is squashed and has some slight formatting and other changes.

purpleidea:
Merged \o/

PR #735, opened 2024-01-30
lang/types: Fail build if stringer run from GOROOT

ollytom:
If stringer is run from GOROOT, it can't find any packages; same error as in https://go.dev/issue/31843. And since GOCACHE is always ignored (see issue above) we can have a bare go generate command.

### DIFF
--- a/lang/types/Makefile
+++ b/lang/types/Makefile
@@ -27,6 +27,10 @@ clean:
 
 kind_stringer.go: type.go
 	@echo "Generating: type kind strings..."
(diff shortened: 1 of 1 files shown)


purpleidea:
Awesome, thank you!
Assuming this didn't break CI, I will merge shortly with a two char commit message change so it passes CI. (as seen in bug/stringer)

purpleidea:
Merged in d47869ac2f1addf23a55bc0a6f4c5f2ea25ecfb4

Thanks and welcome officially to the project!

PR #738, opened 2024-02-09
Use Go 1.20 `slices` package

frebib:
This package is a drop-in replacement for a lot of the custom (duplicated) functions that perform basic slice operations via iteration.

### DIFF
--- a/pgraph/pgraph.go
+++ b/pgraph/pgraph.go
@@ -21,6 +21,7 @@ package pgraph
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


pgraph/pgraph.go
```diff
@@ -387,11 +388,11 @@ func (g *Graph) Sprint() string {
 		str += fmt.Sprintf("Vertex: %s\n", v)
 	}
 	for _, v1 := range g.VerticesSorted() {
-		vs := []Vertex{}
+		var vs VertexSlice
```
purpleidea, line 391: Please leave these alone for now. Explicitly seeing the list is more logical at least for me.

pgraph/pgraph_test.go
```diff
@@ -323,34 +323,6 @@ func TestDeleteVertex2(t *testing.T) {
 	}
 }
 
-func TestVertexContains1(t *testing.T) {
-	v1 := NV("v1")
-	v2 := NV("v2")
-	v3 := NV("v3")
-	if VertexContains(v1, []Vertex{v1, v2, v3}) != true {
```
purpleidea, line 330: You can leave all the tests in, let's just test the new methods you're using!

purpleidea (commented):
Brilliant idea, I love it!

pgraph/pgraph_test.go
frebib: Is there much point in testing stdlib functionality? The point of this change was to reduce the amount of code we're carrying. Less code -> easier to understand/reason about

purpleidea:
Let's it at least prove it does the same thing at least once.

purpleidea:
This would need a rebase, please do so if you're still interested. I'm out of model tokens and energy to do so for now, but it might also be simpler to just ask the models for such a patch. Closing this for now. Cheers!

PR #742, opened 2024-03-01
lang: Use `mcl` struct tag instead of `lang`

frebib:
mgmt's language is named `mcl`, and to avoid conflicts with the very generic term `lang` and to give mgmt more of an identity, it should use the `mcl` nomenclature to hopefully encourage other software to integrate more seamlessly.

### DIFF
--- a/docs/resource-guide.md
+++ b/docs/resource-guide.md
@@ -75,8 +75,8 @@ type FooRes struct {
 
 	init *engine.Init
 
-	Whatever string `lang:"whatever" yaml:"whatever"` // you pick!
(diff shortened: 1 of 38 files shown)


purpleidea:
I did very strongly consider doing this when I first added the struct tags, but I kept it generic for two reasons:

1. I figured there was a chance I change the name of the language. This might actually happen.
2. More than one different language could be built with the same resources. This might actually happen.

If there is an actual conflict with someone else using the same struct tag such that it's a conflict that's blocking us, please give an example, and I'll consider changing it, but otherwise I'm going to unfortunately close this for now, sorry.

Issue #525, opened 2019-06-05
Mount absent behaves erratically


ffrank:
## Versions:

* mgmt version (eg: `mgmt --version`): `mgmt version 0.0.19-30-ge15f110`

* operating system/distribution (eg: `uname -a`): `Linux dev-app01-ffrank 4.4.0-140-generic #166-Ubuntu SMP Wed Nov 14 20:09:47 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`

* golang version (eg: `go version`): `go version go1.11.5 linux/amd64`

## Description:

I'm mounting a simple filesystem using the following script
```
mount "/mnt/acceptance_mount" {
	state => "exists",
	device => "/tmp/loopdevice",
	type => "ext2",
	options => {
		"rw"=>"",
		"loop"=>"",
	},
}
```

I'm trying to unmount it. The following script does unmount it, but leaves it in fstab:
```
mount "/mnt/acceptance_mount" {
        state => "absent",
        device => "/tmp/loopdevice",
}
```

Same when also specifying the type
```
mount "/mnt/acceptance_mount" {
        state => "absent",
        device => "/tmp/loopdevice",
        type => "ext2",
```

However, when specifying the `options`, the entry *will* be removed from fstab. But the filesystem will not be unmounted.
```
mount "/mnt/acceptance_mount" {
        state => "absent",
        device => "/tmp/loopdevice",
        type => "ext2",
        options => {
                "rw"=>"",
                "loop"=>"",
        },
}
```
Omitting the type from the final example does not change the outcome.

What *will* change the outcome is changing the options to a set that does not match the existing mount.
```
mount "/mnt/acceptance_mount" {
        state => "absent",
        device => "/tmp/loopdevice",
        type => "ext2",
        options => {
                "ro"=>"",
                "loop"=>"",
        },
}
```
When specifying `ro` rather than `rw`, the fstab is not touched, nor is the filesystem unmounted.

In summary: Mount resource with state=>absent will
* unmount but not update fstab when resource specifies no options
* update fstab but not unmount when resource does specify matching options
* do nothing when the resource specifies non-matching options

I have not been able to come up with a script that cleanly removes the mount.

purpleidea:
@ffrank TBH I have not properly tested the resource, so there could be bugs. @jonathangold wrote it, but maybe he'll know if this sounds like something he overlooked.

I'll have to circle back to this after I finish function values in core.

[...]

PR #744, opened 2024-03-06
Rewrite the mount resource to fix at least one existing issue

ffrank:
Let's go. It's still not perfect, but simpler and more correct.


ffrank:
Fixes #525

ffrank:
The wrapping of resources is based on how NspawnRes does it.

But I wonder: It would feel much better to just run Watch of both the wrapped file and svc, and in the mount Watch, only wait for the svc to finish, and then also finish.

purpleidea:
Hey great to see this! I had a quick look first. Lots of nice work there. Here are the open issues to discuss:

1. This removes fstab support-- I actually think that is still valuable to have, at least for now-- so how do we handle that? Do we have a separate resource? Do we have it be an optional for this resource? Do we have mount:systemd and mount:fstab ? Also see 2...

2. Is the systemd variant of this resource (as shown here) similar enough to the svc resource that they should just be combined there? I think architecturally they're quite similar. I would need to investigate more.

3. I'm not convinced I want shell tests to stay around. So we should consider a better approach if possible. This won't block your patch, but I want to mention it in case you have ideas.

Thanks Felix!

ffrank:
1. It's a big hassle in terms of required code. I will say, what was there seemed to do a good enough job. The unit tests were nicely elaborate, BUT most of them never ran (they were written as methods of MountRes...). Then again, this feature opens us up to many additional edge cases I believe. 
2. The whole point here is to generate the systemd unit first, which svc does not (and should not?) do. I was considering to move the systemd stuff to some common utility code, so that mount and svc could both rely on that. But with a very small patch to svc, it was embeddable as you had proposed originally. This seems like the right approach to me atm.
3. That shell test is basically a placeholder. It does not run. I might go ahead and propose and prototype a better approach some time, and when I do, this test should not be forgotten. So having the placeholder is valuable from my point of view.

engine/resources/mount.go
```diff
@@ -243,36 +323,36 @@ func (obj *MountRes) Watch(ctx context.Context) error {
...
-	recWatcher, err := recwatch.NewRecWatcher(fstabPath, false)
-	if err != nil {
-		return err
-	}
-	// close the recwatcher when we're done
-	defer recWatcher.Close()
-
 	obj.init.Running() // when started, notify engine that we're running
 
+	// // watch the unit file
```
ffrank, line 328: This should probably be uncommented, but I want better end-to-end testing facilities first. These wrapped resources make life a little bit complicated.

engine/resources/mount.go
```diff
@@ -300,86 +380,49 @@ func (obj *MountRes) Watch(ctx context.Context) error {
...
-	}
-	obj.init.Logf("mountCheckApply(%t)", apply)
+	obj.init.Logf("finished CheckApply on unit file (return %v)", checkOK)
 
-	if obj.State == "exists" {
-		// Reload mounts from /etc/fstab by performing a `daemon-reload` and
-		// restarting `local-fs.target` and `remote-fs.target` units.
-		if err := mountReload(ctx); err != nil {
-			return false, errwrap.Wrapf(err, "error reloading /etc/fstab")
+	if !checkOK {
```
ffrank, line 409: ah shoot this needs an `&& apply`

purpleidea:
FYI as per Felix good idea we're in https://github.com/purpleidea/mgmt/issues/745 for now until we agree on how to proceed.

PR #747, opened 2024-03-10
Design prototype: Replacement for shell tests

ffrank:
We need a robust and powerful mechanism to implement end-to-end acceptance tests, i.e. make mgmt run actual code, and examine the state of the system afterwards (possibly even DURING the run).

This prototype was written quick-and-dirty. It was **specifically not engineered** and I am not looking for feedback on the code.

Instead, I'd like thoughts on the approach and design, specifically:
* is a YAML specification the right approach?
* obvious flaws in the way tests are specified?

But any discussion based on this starting point are welcome.

As said, quick and dirty, if this results in a suggestion for a fundamentally different approach, that is fine also. Happy to scrap this prototype in favor of a whole new one.

### DIFF
--- a/test/mgmtest/file-mode.yml
+++ b/test/mgmtest/file-mode.yml
@@ -1,0 +1,46 @@
+test:
+  title: File mode
+  inputs:
+    yaml0: yamlfile
+  vars:
+    stat: stat # TODO: in OSX use a different value
+               # or just supply these in mgmtest itself
+timeline:
+  - input:
+      target: yaml0
+      content: |
+        ---
+        graph: mygraph
+        resources:
+          file:
+          - name: dir
+            path: /tmp/mgmt/
+            state: exists
+          - name: file1
+            path: "/tmp/mgmt/f1"
+            content: |
+              i am f1
+            state: exists
+          - name: file2
+            path: "/tmp/mgmt/f2"
+            content: |
+              i am f2
+            state: exists
+            mode: 0741
+          - name: file3
+            path: "/tmp/mgmt/f3"
+            content: |
+              i am f3
+            state: exists
+            mode: '0614'
+  - launch:
+      command: "./mgmt run --no-watch --tmp-prefix yaml {{.yaml0}}"
+  - converge:
+      time: 5
+      checks:
+        - shell: ls -l /tmp/mgmt
+        - shell: test -e /tmp/mgmt/f1
+        - shell: test -e /tmp/mgmt/f2
+        - shell: test -e /tmp/mgmt/f3
+        - shell: test $({{.stat}} -c%a /tmp/mgmt/f2) = 741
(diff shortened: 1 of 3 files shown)


ffrank:
Can be tested using

```
go run test/mgmtest/mgmtest.go test/mgmtest/file-mode.yml
```

The mount.yml does not work with current master because MountRes does not support tmpfs. (It does work when rebased to the remount branch as seen in #744 ). It does show how this approach improves upon shell scripting (I hope).

I went for YAML because why not. We could also write something that is more native to golang, but frankly, the information density in YAML is pretty nice, and I expect that we can do anything we need here.

purpleidea:
There is already an integration package which could be greatly expanded if that's the right approach.

There is already a `step` test system for resources ( https://github.com/purpleidea/mgmt/blob/master/engine/resources/resources_test.go#L66 ) which could be greatly expanded if that's the right approach.

Lastly there is TestAstFunc3 which could be greatly expanded if that's the right approach.

I think the _next_big_step_ in testing real-time systems is actually using `mgmt` itself to do so, and to do that we need more pieces first.

In the meantime we probably need a better mechanism for testing resources in pure golang.

I'm closing this PR for now, please keep design discussions in matrix until our mailing list is back up.

Thanks!

ffrank:
> There is already an integration package which could be greatly expanded if that's the right approach.

Integration testing is not acceptance testing; both are useful.

> There is already a `step` test system for resources ( https://github.com/purpleidea/mgmt/blob/master/engine/resources/resources_test.go#L66 ) which could be greatly expanded if that's the right approach.

Yes, that one is also a really good integration/unit test approach that we should keep.

> Lastly there is TestAstFunc3 which could be greatly expanded if that's the right approach.

Ah cheers, TIL. This is also a different level of granularity though.

> I think the _next_big_step_ in testing real-time systems is actually using `mgmt` itself to do so, and to do that we need more pieces first.

Interesting idea, fair enough. If that's on the horizon, doesn't really make sense to invest in the approach sketched in this PR.

> In the meantime we probably need a better mechanism for testing resources in pure golang.

Yes, definitely agree.

> I'm closing this PR for now, please keep design discussions in matrix until our mailing list is back up.

Chat is a terrible place for that, or is there some kind of indexed archive where we could later look up what was said wrt. a certain topic?

purpleidea:
> Integration testing is not acceptance testing; both are useful.

:+1: 

> Yes, that one is also a really good integration/unit test approach that we should keep.

There's obviously some value in the ideas you PR'ed here, but please read through some of the existing code first, and then we can circle back here if needed.

> Interesting idea, fair enough. If that's on the horizon, doesn't really make sense to invest in the approach sketched in this PR.

Not sure how far away it is. Not short though.

> Chat is a terrible place for that, or is there some kind of indexed archive where we could later look up what was said wrt. a certain topic?

I agree. Matrix is the best we have for now. I'm waiting on a mailing list host.

PR #748, opened 2024-03-12
Solve issue with variable names like $in or $struct

ffrank:
Make variables like `$foo` and `$in` and `$this.smells.funny` into single lexical tokens. This avoids lexing errors when the variable name without the dollar sign looks like a reserved word.

(Frankly, I believe this is the more correct way to do this, and if not the only way, it is probably the easiest.)

### DIFF
--- a/lang/interpret_test/TestAstFunc2/lexer-parser0.txtar
+++ b/lang/interpret_test/TestAstFunc2/lexer-parser0.txtar
@@ -1,16 +1,16 @@
 -- main.mcl --
 import "fmt"
-# unfortunately, for now `in` is a reserved keyword, see:
-# https://github.com/purpleidea/mgmt/issues/728
 $map = 55
-$fn = func($in) { # in is a special keyword
+# in is a special keyword, but accepted in a variable name
+$fn = func($in) {
 	13
 }
 test fmt.printf("%d", $fn(0)) {}
-func fn($in) { # in is a special keyword
+func fn($in) {
 	42 + $map
 }
 test fmt.printf("%d", $fn(0)) {}
 test fmt.printf("%d", fn(0)) {}
 -- OUTPUT --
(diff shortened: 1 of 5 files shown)


ffrank:
As an aside, I may have found a race in https://github.com/ffrank/mgmt/actions/runs/8247933356/job/22557152150

purpleidea:
Neat! I wasn't sure this was possible while also not blocking some other parsing changes I have. I appreciate the patch, but I will have to delay merging it to make sure it doesn't conflict with some other parsing stuff I have pending. Of note, if you look at the git history, you'll see I had something similar in until I realized it was causing a conflict.

Lastly, there's a good chance I get rid of the $foo style naming entirely (not yet) and moving to just `foo`. TBD.

The race: yes there is one that I know about that needs fixing.

purpleidea:
> The proposed patch avoids some (all?) of this by allowing the lexer to consume every variable name as a single token, and provide the information that the parser needs to make proper decisions (i.e., is it a dotted or undotted name token).

Yes this is very clever, thank you. I'm slightly worried that doing this might make it more difficult in the future to add parser features without being ambiguous. I will look through my WIP feature branches and have a look.

ffrank:
FWIW I think it is really **really** awkward to make the dots and particles of qualified names into fully qualified citizens on the parser level.

This is currently a valid resource

```
print "it" {
        msg => fmt.printf("%d", $c . a),
}
```

as is this

```
print "it" {
        msg => fmt.printf("%d", $c

.

                                                a),
}
```

I argue that `$c.a` should always be a token, as should `fmt.printf`.

Yes even this is accepted right now (this patch only addresses the variable):

```
import "fmt"

class container {
        $a = 1
}
include container as c

print "it" {
        msg => fmt . printf("%d", $c . a),
}
```

PR #763, opened 2024-06-17
engine: resources: file: Remove Validate owner/group Checks

### LINKED ISSUE #762: File Resource Checks Host for User/Group in Validation
## Versions:

* mgmt version (eg: `mgmt --version`):
`mgmt` built with this [commit](https://github.com/purpleidea/mgmt/commit/296fc484ba76da3232da25e18b285fa15707e0c1)

* operating system/distribution (eg: `uname -a`):
Bookworm (via Docker Desktop on an ARM64 Mac)

* golang version (eg: `go version`):
`go version go1.22.4 linux/arm64`

## Description:
When attempting to create a file that is owned by a user/group that are being managed within the resource graph, `mgmt` fails early with an error during the `Validate` stage (before any resources have been managed) because the user/group do not exist on the host (checks are [here](https://github.com/purpleidea/mgmt/blob/3c661ab6745002c71aff97edfdaa22a876a0cd2c/engine/resources/file.go#L355-L361)).

A sample `mcl` that demonstrates this is below:
```
group "mygroup" {
  state => "exists",
  gid => 1234,
}

user "myuser" {
  state => "exists",
  uid => 1234,
  group => "mygroup",
}

file "/tmp/myfile" {
  state => "exists",
  content => "mycontents",
  owner => "myuser",
  group => "mygroup",
}
```
When run, `mgmt` logs an error and then is not able to manage the resources:
```
user@host: mgmt run lang example.mcl
...start up
main: graph validate failed: file[/tmp/myfile] did not Validate: user lookup error (myuser): user: unknown user myuser
main: waiting...
...and then nothing happens
```

cianyleow:
The owner/group of a file should not be validated on the host until runtime. This removes the checks in Validate() that were happening before the execution of the resource graph (and therefore bound to fail if the system was being bootstrapped).

Fixes #762

### DIFF
--- a/engine/resources/file.go
+++ b/engine/resources/file.go
@@ -352,13 +352,6 @@ func (obj *FileRes) Validate() error {
 			return fmt.Errorf("can't set Owner or Group on this platform")
 		}
 	}
-	if _, err := engineUtil.GetUID(obj.Owner); obj.Owner != "" && err != nil {
-		return err
-	}
-
-	if _, err := engineUtil.GetGID(obj.Group); obj.Group != "" && err != nil {
-		return err
-	}
 
 	// TODO: should we silently ignore this error or include it?
 	//if obj.State == FileStateAbsent && obj.Mode != "" {


purpleidea:
Patch looks perfect to me. Thank you!

(Will merge later today/tomorrow.)

If you're looking for a next patch please LMK.

purpleidea:
This has been merged, thank you! As a bonus you can forever gloat that your first mgmt patch only removed code =D

PR #765, opened 2024-07-07
examples: lang: Fix env0 example

Hiexy:
Change the function call from `getenv` to `defaultenv`.

### DIFF
--- a/examples/lang/env0.mcl
+++ b/examples/lang/env0.mcl
@@ -4,20 +4,20 @@
 import "fmt"
 import "sys"
 
-$x = sys.getenv("TEST", "321")
+$x = sys.getenv("TEST")
 
 print "print1" {
 	msg => fmt.printf("the value of the environment variable TEST is: %s", $x),
 }
 
-$y = sys.getenv("DOESNOTEXIT", "321")
+$y = sys.defaultenv("DOESNOTEXIT", "321")
 
 print "print2" {
 	msg => fmt.printf("environment variable DOESNOTEXIT does not exist, defaulting to: %s", $y),
 }
 
-$z = sys.getenv("EMPTY", "456")
+$z = sys.defaultenv("EMPTY", "456")
 
 print "print3" {
-	msg => fmt.printf("same goes for epmty variables like EMPTY: %s", $z),
+	msg => fmt.printf("same goes for empty variables like EMPTY: %s", $z),
 }


examples/lang/env0.mcl
```diff
@@ -4,19 +4,19 @@
 import "fmt"
 import "sys"
 
-$x = sys.getenv("TEST", "321")
+$x = sys.defaultenv("TEST", "321")
 
 print "print1" {
 	msg => fmt.printf("the value of the environment variable TEST is: %s", $x),
```
purpleidea, line 10: I suspect this message alludes to this being a regular "getenv" call... So it has one too many args.

examples/lang/env0.mcl
```diff
@@ -4,19 +4,19 @@
...
 
 print "print2" {
 	msg => fmt.printf("environment variable DOESNOTEXIT does not exist, defaulting to: %s", $y),
 }
 
-$z = sys.getenv("EMPTY", "456")
+$z = sys.defaultenv("EMPTY", "456")
 
 print "print3" {
 	msg => fmt.printf("same goes for epmty variables like EMPTY: %s", $z),
```
purpleidea, line 22: noticed a typo here too!

examples/lang/env0.mcl
Hiexy: Updated.

Hiexy:
I updated the commit message again, and I am using this option `--force-with-lease` to push the amended commit messag, but I am not sure if it is the proper way to do so.

purpleidea:
> I updated the commit message again, and I am using this option --force-with-lease to push the amended commit messag, but I am not sure if it is the proper way to do so.

Indeed it is the proper way. For most patches (especially small ones) we don't need to preserve the history of our work on this change-- IOW, suppose we made some mistakes along the way of developing this patch, we don't need to record those in git master forever.

So --force pushing over a feature branch is actually the ideal scenario.

If this was a very large patch, particularly with multiple commits, it might be preferable to add new commits until the entire PR was fixed. This _can_ in some rare instances help the submitter and reviewer keep track of what changes they've made and so on. Only at the end does the patchset typically get squashed into one or more commits.

HTH

Patch looks perfect now, will merge shortly.

PS: Bonus points to look cool (just a lesson for future you)

Sending small typo fix style patches is definitely useful, but you get more hacker style points when the same patchset also adds a test to permanently check for that kind of bug as well. It's not always possible, but when it is, it's cooler.

purpleidea:
Merged, thank you!

PR #779, opened 2024-11-28
Draft: fix: allow runtime to be ANY

karpfediem:
Found a bug within IsMyArch in packagekit.go
it gets passed 'any' which is mapped by archUtil to 'ANY'
This is then directly compared with the runtime.GOARCH

Digged deep enough with logging statements to show this is where it returns false (usually without error)
10:58:55 main.go:88: engine: pkg\[unattended-upgrades\]: packagekit: PackagesToPackageIDs() \[IsMyArch\] b: false, err: goarch 'ANY', GOARCH 'amd64'

The code after this skips evaluating this package further, so it has them still marked as not found.

### DIFF
--- a/engine/resources/packagekit/packagekit.go
+++ b/engine/resources/packagekit/packagekit.go
@@ -972,5 +972,5 @@ func IsMyArch(arch string) (bool, error) {
 	if goarch == archUtil.Any { // special value that corresponds to noarch
 		return true, nil
 	}
-	return goarch == runtime.GOARCH, nil
+	return goarch == runtime.GOARCH || goarch == "ANY", nil
 }
--- a/engine/resources/packagekit/packagekit.go
+++ b/engine/resources/packagekit/packagekit.go
@@ -969,7 +969,7 @@ func IsMyArch(arch string) (bool, error) {
 		// if you get this error, please update the PkArchMap const
 		return false, fmt.Errorf("arch '%s', not found", arch)
 	}
-	if goarch == archUtil.Any { // special value that corresponds to noarch
+	return goarch == runtime.GOARCH || goarch == "ANY" || goarch == archUtil.Any, nil
 		return true, nil
 	}
 	return goarch == runtime.GOARCH || goarch == "ANY", nil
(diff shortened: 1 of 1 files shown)


karpfediem:
might want to consolidate this into a single return statement, and think about what the Value of `archUtil.Any` should be
(currently `Any = "*"`)

engine/resources/packagekit/packagekit.go
```diff
@@ -972,5 +972,5 @@ func IsMyArch(arch string) (bool, error) {
 	if goarch == archUtil.Any { // special value that corresponds to noarch
```
frebib, line 972: I wonder if this should be catching this case 🤔 
This code could be optimised a little bit I guess:
```suggestion
	return goarch == runtime.GOARCH || goarch == "ANY" || goarch == archUtil.Any, nil
```

karpfediem:
Let's still discuss the value of archUtil.Any

purpleidea:
Also please log whatever the goarch value you're seeing here is:

https://github.com/purpleidea/mgmt/blob/master/engine/resources/packagekit/packagekit.go#L740C1-L740C27

karpfediem:
> Also please log whatever the goarch value you're seeing here is:
> 
> https://github.com/purpleidea/mgmt/blob/master/engine/resources/packagekit/packagekit.go#L740C1-L740C27


```
19:51:44 main.go:88: engine: pkg[ssh]: packagekit: PackagesToPackageIDs() pkg: git, ver: 1:2.39.5-0+deb12u1, arch: amd64, data: manual:debian-stable-main
```

Full Log with more log statements: [out.log](https://github.com/user-attachments/files/17951512/out.log)

purpleidea:
If "all" is passed, why does this patch change anything?

purpleidea:
Here's the fix I recommend: https://github.com/purpleidea/mgmt/tree/bug/packagekit

If you confirm this fixes the issue, I will merge.

karpfediem:
LGTM

purpleidea:
merged that commit, thanks for looking into this!

PR #780, opened 2025-01-21
Feature/auto edge httpfile

### LINKED ISSUE #734: automatic edge between http:file and file
Automatic edges add dependencies between different resources.

For example:

```mcl
$f2 = "/tmp/some_file"
file $f2 {
        content => "i'm a cool file in /tmp\n",
}

http:server ":8080" {
}

http:file "/file2" {
        path => $f2,

        Depend => File[$f2],    # TODO: add autoedges
}

```

We have an API to add these automatically. Add this feature for these two resources! Keep in mind that path can be a dir as well.
The examples/lang/http*.mcl examples should make this all easier to test.

knobby-tires:
Fixes #734 

## What this adds: 
Automatic edge handling for the HTTPFile resource. The goal is to make sure that when an HTTP file is served from a path, the file resource is properly tracked as a dependency.

## How it does that: 
`HTTPFileResAutoEdges` struct: Handles dependecy tracking by keeping a list of related resources and checks if they exist.

`AutoEdges()` method:  If the resource is serving from a file path, it creates an automatic edge to the corresponding file resource.
If  the `data` field  is used instead, no edges are created.

## How I tested: 
After setting things up I ran the following curl commands 

```
curl http://localhost:8080/test1  # Check if /test1 works
curl http://localhost:8080/test2  # Check if /test2 works
curl http://localhost:8080/dir/index  # Check if specific file works
```


purpleidea:
Thanks for the patch! A few quick nits to help you get through CI and the easy stuff:

1) Missing an EOL in your example file.
2) Make everything a single commit.
3) Get rid of the go.mod changes.
4) Comments should be full sentences (capitalized/ends with punctuation) or short one-liners usually that aren't capitalized.

knobby-tires:
Awesome, thanks for the fast reply. I will fix all of that tomorrow afternoon.

purpleidea:
(The race tests will likely fail, this is OK, but check out the other tests which should pass if there are no issues.)

knobby-tires:
It looks like I forgot to run gofmt on http.go, I fixed that and amended it to the commit.

purpleidea:
Your mcl file comments need the same fixups and it will also fail without tab indentation. (There's also a trailing newline at the end if you want a nit.)

purpleidea:
CI needs a slightly different commit message, eg:

> resources: Add auto edge support between http_file and file resources

And you have one `vet` issue. I also noticed you're missing a newline at the end of a file.

purpleidea:
Thanks for the patch, but did you test that it actually creates the autoedges? You may wish to add a test case once you get it working in TestAstFunc2/ ... at the moment I don't see any edges except one.

LMK about your use-case if you want some help.

knobby-tires:
Sorry, I'm still working on this. Just getting my bearings again after picking this back up. I will definitely take your advice about the test case and will let you know if I need help, thank you!

PR #781, opened 2025-01-22
util: Test

ahmad-abuziad:
The goal of this PR to test cover util/util.go and increase it from 64.8% to 100% since almost all the util functions are testable
Screenshot From 2025-01-22 13-43-38

### DIFF
--- a/util/util_test.go
+++ b/util/util_test.go
@@ -2193,3 +2193,72 @@ func TestFlattenListWithSplit(t *testing.T) {
 		})
 	}
 }
+
+func TestRebase(t *testing.T) {
+	type input struct {
+		path string
+		base string
+		root string
+	}
+	type want struct {
+		rebasedPath string
+		err         error
+	}
+
+	tests := []struct {
+		name  string
+		input input
+		want  want
+	}{
+		{
+			name: "rebased to absolute directory",
+			input: input{
+				path: "/usr/bin/foo",
+				base: "/usr/",
+				root: "/usr/local/",
+			},
+			want: want{
+				rebasedPath: "/usr/local/bin/foo",
+				err:         nil,
+			},
+		},
+		{
+			name: "rebased to relative directory",
+			input: input{
+				path: "/var/lib/dir/file.conf",
+				base: "/var/lib/",
+				root: "",
+			},
+			want: want{
+				rebasedPath: "dir/file.conf",
+				err:         nil,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		gotStr, gotErr := Rebase(tt.input.path, tt.input.base, tt.input.root)
+
+		if gotStr != tt.want.rebasedPath {
+			t.Errorf("got rebased path: %v, want rebased path: %v", gotStr, tt.want.rebasedPath)
+		}
+		if gotErr != tt.want.err {
+			t.Errorf("got error: %v, want error to be: %v", gotErr, tt.want.err)
+		}
+	}
+
+	t.Run("root doesn't end with /", func(t *testing.T) {
+		// TODO(ahmadabuziad) in Rebase return predefined error instead e.g.
+		// e.g. var ErrRootNotDirectory = errors.New("root is not a directory")
+		// so it would be clearer and easier to test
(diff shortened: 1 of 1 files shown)


purpleidea:
These will keep failing because of the commit message. Look at the tests. Lmk if you need more help.

ahmad-abuziad:
> These will keep failing because of the commit message. Look at the tests. Lmk if you need more help.

I will look into it.
I will stick with the format since i started with it, except the last commit will contain a message that doesn't fail.

purpleidea:
You can easily rename a bunch with: EDITOR=vim git rebase -i <commit id before the first one>

ahmad-abuziad:
coverage increased from 64.8% to 94.4
image

util/util_test.go
```diff
@@ -1636,3 +1639,960 @@ func TestSortMapStringValuesByUInt64Keys(t *testing.T) {
...
+			input: []string{},
+			want:  "",
+		},
+		{
+			name:  "single path",
+			input: []string{"/path/to"},
+			want:  "/path/to",
+		},
+		{
+			name:  "single path doesn't start with /", // TODO(ahmadabuziad): need to double check the desired output with @purple idea
```
purpleidea, line 2139: What should we do here? But if it's very important put `XXX: message` and remove your username and mine please

util/util_test.go
```diff
@@ -1636,3 +1639,960 @@ func TestSortMapStringValuesByUInt64Keys(t *testing.T) {
...
+		if gotStr != tt.want.rebasedPath {
+			t.Errorf("got rebased path: %v, want rebased path: %v", gotStr, tt.want.rebasedPath)
+		}
+		if gotErr != tt.want.err {
+			t.Errorf("got error: %v, want error to be: %v", gotErr, tt.want.err)
+		}
+	}
+
+	t.Run("root doesn't end with /", func(t *testing.T) {
+		// TODO(ahmadabuziad) in Rebase return predefined error instead e.g.
```
purpleidea, line 2253: Remove username please.

util/util_test.go
```diff
@@ -1636,3 +1639,960 @@ func TestSortMapStringValuesByUInt64Keys(t *testing.T) {
...
+		if gotStr != "" {
+			t.Errorf("got: %v, want empty string", gotStr)
+		}
+
+		if gotErr.Error() != "must be absolute" {
+			t.Errorf(`got error: %v, want error "must be absolute"`, gotErr.Error())
+		}
+	})
+
+	//TODO(ahmadabuziad): edge cases currently panic, handle edge cases. "/", ""
```
purpleidea, line 2293: Same as above...

purpleidea (changes requested):
Thanks for all the tests.

For a first start, can you please rename the commit messages so they have the right format, typically `util: Test blah blah...` in this case.

Next see the other misc things, then we'll do a final review. Thank you!

Sorry for the delay, I've been at the FOSDEM+CfgMgmtCamp conference stuff.

util/util_test.go
ahmad-abuziad: - We should return empty string "" (undefined behavior) same as the next test case.
When sending multiple paths and one of them is not absolute the current behavior returns an empty string, so for the case when single path is not absolute it should return an empty string. (the change may break existing usage - it showed me there is no usage for this function)

- What does `XXX: message` format mean?
- removed usernames

purpleidea:
I expect to still see failures, test is here: https://github.com/purpleidea/mgmt/blob/master/test/test-commit-message.sh#L17

So for example a valid commit is:

foo: Hello something

an invalid commit is:

foo: hello something

Note the capitalization.

purpleidea:
Sorry for the delays... So this was very hard to review because I couldn't cherry-pick out anything without massive git conflicts because it was all in the same file...

So I had to squash everything down, leaving out a few problematic tests and with one small fixup.

Merged, thank you!

Note: I had to drop the time-based tests since they aren't guaranteed to not be flaky in slow CI. If I mis-read this please let me know

purpleidea:
Btw nice catch on some of the corner cases. The interesting one is actually mentioned in the docs so I commented that out.

Issue #606, opened 2020-06-16
Feature: Url Parser for mcl


dbakong:
''' 
It turns out that it is useful to "parse" strings that represent URL's. Golang makes this a bit easier because it has: https://golang.org/pkg/net/url/#Parse -- we'd like to do something similar. Here's the catch:

0) Look up "pure functions", "side effects", and "functional programming" and spend 30 min understanding the basics.

1) Design the "API" for whatever our URL parsing should look like. Should it be one function or many, and what is/are the signature(s) ? Explain this clearly in a new GH issue.

2) Have a really, properly comprehensive set of test cases for this. Remember to test absurd things, or null/empty cases. In particular, the Parse function in golang has some surprises, so make sure you catch these!
'''

####Intro:

Mgmt has a language - mcl - and it currently doesn't have a function to parse urls

####Design: 
The functionality would rely on Golang's default url parsing package - specifically [url.Parse](https://golang.org/pkg/net/url/#Parse) to allow mcl to parse a given url and return various elements of the url, including the scheme, host and path. The function would take the FQDN version of a url.

The tests for the patch would cover:
- The suggested return type
- Output from the parsed url
- Valid input
- Expected errors

The signature of the function would be similar to the below:
ParseUrl(input []types.Value) (types.Value, error) - the function would both accept and return string values. Some errors handled would include:
- Non-FQDN urls
- Empty url
- Invalid port

purpleidea:
@dbakong I was expecting to see the recommended `mcl` signature, and some mcl examples...

purpleidea:
@dbakong The example doesn't match the signature, and the scheme is only one small part...

dbakong:
@purpleidea -  The example above was only meant to show a small example of the url parser in action. I'll add more examples as I implement the patch. I'll work on matching the signature to the current example.

frebib:
https://golang.org/pkg/net/url/#Parse
Could this be used along with the simple function api and the funcgen code?
I'm not sure how that handles functions that return errors, though

purpleidea:
@frebib Correct, this could use the simple function API since it's completely pure. As for errors, you'd need to decide how to represent that in mgmt. Either a `isValidURI() bool` to know if it would error, and for the `Parse` function, on error you empty string (likely my preference) or if you want to hard error and shutdown the language. (Probably not recommended.)

[...]

PR #783, opened 2025-01-31
lang: core: util: add a new func for URL parsing

lourencovales:
This is a first attempt to adding a new function for URL parsing, using go's net/url package and the simple API. This is still a barebones implementation, there's possibility to expose more information that the stdlib exposes.

Related: https://github.com/purpleidea/mgmt/issues/606

### DIFF
--- a/lang/interpret_test/TestAstFunc2/url_parser.txtar
+++ b/lang/interpret_test/TestAstFunc2/url_parser.txtar
@@ -1,0 +1,22 @@
+-- main.mcl --
+
+import "net"
+import "fmt"
+
+$url1 = "https://www.example.com/search?q=mgmt"
+$url2 = "ftp://example.com/apath"
+$url3 = "https://www3.weird.url/?qu=ery"
+
+$url_parse1 = net.url_parser($url1)
+$url_parse2 = net.url_parser($url2)
+$url_parse3 = net.url_parser($url3)
+
+test [fmt.printf("%s %s %s %s", $url_parse1->scheme, $url_parse1->host, $url_parse1->path, $url_parse1->raw_query),] {}
+test [fmt.printf("%s %s %s %s", $url_parse2->scheme, $url_parse2->host, $url_parse2->path, $url_parse2->raw_query),] {}
+test [fmt.printf("%s %s %s %s", $url_parse3->scheme, $url_parse3->host, $url_parse3->path, $url_parse3->raw_query),] {}
+
+-- OUTPUT --
+
+Vertex: test[https www.example.com /search q=mgmt]
+Vertex: test[ftp example.com /apath ]
+Vertex: test[https www3.weird.url / qu=ery]
--- a/lang/core/net/url_parser.go
+++ b/lang/core/net/url_parser.go
@@ -1,0 +1,93 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+// Additional permission under GNU GPL version 3 section 7
+//
+// If you modify this program, or any covered work, by linking or combining it
+// with embedded mcl code and modules (and that the embedded mcl code and
+// modules which link with this program, contain a copy of their source code in
+// the authoritative form) containing parts covered by the terms of any other
+// license, the licensors of this program grant you additional permission to
+// convey the resulting work. Furthermore, the licensors of this program grant
+// the original author, James Shubin, additional permission to update this
+// additional permission if he deems it necessary to achieve the goals of this
+// additional permission.
+
+package corenet
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/purpleidea/mgmt/lang/funcs/simple"
+	"github.com/purpleidea/mgmt/lang/types"
+	"github.com/purpleidea/mgmt/util/errwrap"
+)
+
+const (
+	urlParserFieldScheme   = "scheme"
+	urlParserFieldHost     = "host"
+	urlParserFieldPath     = "path"
+	urlParserFieldRawQuery = "raw_query"
+)
+
+var urlParserReturnType = fmt.Sprintf(
+	"struct{%s str; %s str; %s str; %s str}",
+	urlParserFieldScheme,
+	urlParserFieldHost,
+	urlParserFieldPath,
+	urlParserFieldRawQuery,
+)
+
+func init() {
+	simple.ModuleRegister(ModuleName, "url_parser", &simple.Scaffold{
+		T: types.NewType(fmt.Sprintf("func(str) %s", urlParserReturnType)),
+		F: URLParser,
+	})
+}
+
+// URLParser takes a URL as a string, and extracts the different components of
+// the URL: scheme, host, path, and query. The function will error out if the
+// given URL doesn't contain a valid URL, as this is a requirement of the
+// underlying net/url library. Note that additional fields may be added to the
+// output in the future.
+func URLParser(ctx context.Context, input []types.Value) (types.Value, error) {
+	u, err := url.Parse(input[0].Str())
+	if err != nil {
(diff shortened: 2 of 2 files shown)


purpleidea:
I expect the basic tests will fail. See if you can figure out why by reading the logs, and ping when a fix is in. Looks like you're on the right track, but no hints =D This is as much about learning how CI works and so on.

purpleidea:
This will still fail for the same reason.

purpleidea:
So your commit message will fail the tests, and secondly, since this is pre-merge, we can also hide any mistakes. So you will need to rebase both of these commits into a single one.

Remember to look at a `git diff` right before you push to see if every single byte you changed is what you wanted. HTH

purpleidea:
Lastly, for this patch consider adding a short test in lang/interpret_test/TestAstFunc2/ folder.

lourencovales:
Oops, I missed the part where you were asking for tests all in one commit. In any case, I fixed the basic tests, and the previous commit I had also fixed the function signature, which was also causing trouble with the tests. I'll work on tests now. Do you also want an all-in-one commit in this case?

purpleidea:
> Do you also want an all-in-one commit in this case?

Yeah it's probably good to have everything here in one commit. A single .mcl test will probably be good enough!

lang/core/util/url_parser.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/purpleidea/mgmt/lang/funcs/simple"
+	"github.com/purpleidea/mgmt/lang/types"
+	"github.com/purpleidea/mgmt/util/errwrap"
+)
+
+const structComp = "struct{Scheme str; Host str; Path str; Query str}"
```
purpleidea, line 42: I think the field names should be lowercase. Let's name this var urlParserReturnType too.

lang/core/util/url_parser.go
```diff
@@ -0,0 +1,81 @@
...
+		F: URLParser,
+	})
+}
+
+// URLParser takes an URL as a string, and finds the different components of
+// said URL - scheme, host, path, and query. The function will error out if the
+// given URL doesn't contain a scheme, as this is a limitation of the underlying
+// net/url library.
+func URLParser(ctx context.Context, input []types.Value) (types.Value, error) {
+	i := input[0].Str()
```
purpleidea, line 56: Nit: Short letter vars have an unwritten convention: use `s` instead here. `i` is commonly for integers or counters.
OR since we don't need the var elsewhere, substitute this into the url.Parse call directly.

lang/core/util/url_parser.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := v.Set("Scheme", &types.StrValue{V: u.Scheme}); err != nil {
+		return nil, errwrap.Wrapf(err, "invalid scheme value")
+	}
+	if err := v.Set("Host", &types.StrValue{V: u.Host}); err != nil {
+		return nil, errwrap.Wrapf(err, "invalid host value")
+	}
+	if err := v.Set("Path", &types.StrValue{V: u.Path}); err != nil {
+		return nil, errwrap.Wrapf(err, "invalid path value")
+	}
+	if err := v.Set("Query", &types.StrValue{V: u.RawQuery}); err != nil {
```
purpleidea, line 76: should this be `query` or `raw_query` ?

lang/core/util/url_parser.go
```diff
@@ -0,0 +1,81 @@
...
+	u, err := url.Parse(i)
+	if err != nil {
+		return nil, errwrap.Wrapf(err, "error parsing the URL")
+	}
+	if u.Scheme == "" {
+		return nil, errwrap.Wrapf(err, "empty schemes are invalid")
+	}
+
+	v := types.NewStruct(types.NewType(structComp))
+	if err := v.Set("Scheme", &types.StrValue{V: u.Scheme}); err != nil {
```
purpleidea, line 67: optional: You _could_ store the field names as a constant, like urlParserFieldScheme and so on... The value of this is that they can be used in the type above and it enforces a logical connection that the two are the same "string".

Know what I mean?

lang/core/util/url_parser.go
```diff
@@ -0,0 +1,81 @@
...
+	simple.ModuleRegister(ModuleName, "url_parser", &simple.Scaffold{
+		T: types.NewType(fmt.Sprintf("func(str) %s", structComp)),
+		F: URLParser,
+	})
+}
+
+// URLParser takes an URL as a string, and finds the different components of
+// said URL - scheme, host, path, and query. The function will error out if the
+// given URL doesn't contain a scheme, as this is a limitation of the underlying
+// net/url library.
```
purpleidea, line 54:
```suggestion
// URLParser takes a URL as a string, and extracts the different components of
// the URL: scheme, host, path, and query. The function will error out if the
// given URL doesn't contain a valid URL, as this is a requirement of the underlying
// net/url library. Note that additional fields may be added to the output in the future.
```

(How about this? Make sure to wrap it correctly.)

lang/interpret_test/TestAstFunc2/url_parser.txtar
```diff
@@ -0,0 +1,19 @@
...
+
+$url1 = "https://www.example.com/search?q=mgmt"
+$url2 = "ftp://example.com/apath"
+$url3 = "https://www3.weird.url/?qu=ery"
+
+
+test [fmt.printf("%v", util.url_parser($url1)),] {}
+test [fmt.printf("%v", util.url_parser($url2)),] {}
+test [fmt.printf("%v", util.url_parser($url3)),] {}
+
```
purpleidea, line 14: nit: You could also do:

```
$whatever = util.url_parser($url1)
test [$whatever->scheme,] {}
```

for example...

purpleidea (commented):
Patch is excellent, thank you!

I've nitpicked this patch like crazy since we're on a journey to learn excellence. I could happily merge this as-is, but here are some suggestions to fix. Keep in mind, anytime me, or someone else, suggests something you don't agree with, speak up! Often maintainers make mistakes. Of course do so carefully since it's often their project, but if they miss something that you didn't, let them know why.

lang/core/util/url_parser.go
lourencovales: I had the field names with lowercase, but then the test I built complained that the struct had an exported field:

`interpret_test.go:802: test #226: funcs: Erroring func printf@0xc000980000: panic in Stream of func printf@0xc000980000: struct has unexported field: scheme`

Ack'd on the variable name.

lang/core/util/url_parser.go
lourencovales: fair on both points; in this situation it does make more sense to avoid declaring a new variable.

lang/core/util/url_parser.go
lourencovales: The URL type on the net/url package does not export a "query" field: https://pkg.go.dev/net/url#URL
They do have a Query function, but this would make a logic here a tad more complex - still doable, but up to you.

lang/core/util/url_parser.go
lourencovales: not sure I understand here. Are you suggestion doing something like this?

```
const urlParserFieldScheme = "Scheme"
const urlParserFieldHost = "Host"
...
const urlParserReturnType = "struct{" + urlParserFieldScheme + " str; " + urlParserFieldHost + " str; ...
...
if err := v.Set(urlParserFieldScheme, &types.StrValue{V: u.Scheme}); err != nil {
...
```

lang/core/util/url_parser.go
lourencovales: This does read better, appreciate the corrections!

lang/interpret_test/TestAstFunc2/url_parser.txtar
lourencovales: What would be the advantage here? Not sure I grasp the implications.

lang/core/util/url_parser.go
purpleidea: My comment is about how to name this in the struct... What do we want the user to see?

lang/core/util/url_parser.go
purpleidea: Sort of yes. Hint: If you followed this practice, you would likely not have hit a bug!
(Although I'd recommend fmt.Sprintf instead of + to make this readable.)

lang/interpret_test/TestAstFunc2/url_parser.txtar
purpleidea: Functions in mcl can only return one value (which is the case for most programming languages). Golang happens to allow multiple btw, neat! In any case, we return a struct. But the user _probably_ wants to access just one or two values from that struct. So prove that they can get them out =D

purpleidea:
Review done!

lourencovales:
@purpleidea should be a-ok now

purpleidea:
Looks perfect, except can we put it into the net package instead of util please? Rebase to master too please and then I think it's merge time!

lourencovales:
Please review @purpleidea :-)

purpleidea:
merged \o/

woo! Thank you =D

PR #785, opened 2025-02-03
engine: resources: Adding logic for svc unit state

lourencovales:
This is a small patch that adds logic for checking what's the state of the unit file and making the CheckApply function more robust

### DIFF
--- a/engine/resources/svc.go
+++ b/engine/resources/svc.go
@@ -333,7 +333,15 @@ func (obj *SvcRes) CheckApply(ctx context.Context, apply bool) (bool, error) {
 
 	var running = (activestate.Value == dbus.MakeVariant("active"))
 	var stateOK = ((obj.State == "") || (obj.State == "running" && running) || (obj.State == "stopped" && !running))
-	var startupOK = true // XXX: DETECT AND SET
+
+	startupstate, err := conn.GetUnitPropertyContext(ctx, svc, "UnitFileState")
+	if err != nil {
+		return false, errwrap.Wrapf(err, "failed to get unit file state")
+	}
+
+	enabled := (startupstate.Value == dbus.MakeVariant("enabled"))
+	disabled := (startupstate.Value == dbus.MakeVariant("disabled"))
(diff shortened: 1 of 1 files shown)


purpleidea:
Okay, give me an example of how you'd smoke/manual test this, and does it do what you expect?

lourencovales:
As far as I understand, this check I added deals with comparing the unit state (enabled/disabled) vs. what is set in the service resource struct (SvcRes.Startup); if they match, the startupOK variable is set to true, and the check on line 357 is set to true as well and all is good (at least as far as this patch is concerned). I now notice that I didn't really add any logic for what happens when startupOK is false, which I guess is the next step.

In any case, a quick smoke test here would be having a .mcl and a .service file, and the .mcl would mostly revolve around 1) setting the values expected through the svc util and 2) using exec to issue systemctl commands to change the unit state externally and seeing how mgmt handle those cases.

purpleidea:
Try and answer the question with only TWO bash shell commands that you would type.

lourencovales:
systemctl enable unit / systemctl disable unit ?

purpleidea:
bingo!

lourencovales:
So I guess we'd roughly want something like this for a quick smoke test?

```
svc "unit" {
    startup => "enabled"
}

exec "cmd" {
    cmd => "/usr/bin/systemctl"
    args => ["disable", "unit",]
}
```

purpleidea (commented):
Apart from one nit fix, as long as this works, it looks good. Fix the commit message, and then we can merge this.

lourencovales:
@purpleidea made the small fix you recommended, and this seems to work: the startup value of the unit file is made coherent with the value declared on the mcl file. What doesn't seem to work is achieving coherence during runtime - i.e., if I change the value after mgmt is started up, nothing happens. But at least nothing seems to have broken (tests all passed locally), and the very basic is done.

purpleidea:
Okay, I rebased this to master and pushed as:

b5384d127834acd3676f40cb72ffdcfebb021c77

The second part of this patch is to make sure you can see the events of the enabled/disabled state. We can work on that in a separate PR. \o/

PS: You may wish to update your git config user name so it shows better in `git log`.

PR #789, opened 2025-02-05
resources: Add shell to user resource

jfindlay:


engine/resources/user.go
```diff
@@ -261,6 +267,9 @@ func (obj *UserRes) CheckApply(ctx context.Context, apply bool) (bool, error) {
 		if obj.HomeDir != nil {
 			args = append(args, "-d", *obj.HomeDir)
 		}
+		if obj.Shell != nil {
+			args = append(args, "-d", *obj.Shell)
```
purpleidea, line 271: I think you want `--shell` and also you want to handle the "remove shell" case.

purpleidea:
@jfindlay Thanks for the update. I think we probably want to use `--shell` to be more precise. Also maybe @frebib could have a quick review?

purpleidea:
This is done in 82489c3fe0b894ac1511430dda0f297c0d986d1a

PR #790, opened 2025-02-10
utl: add DownloadFile function

ahmad-abuziad:
**Enhancement**
- automatically create directories if don't exist, now it fails if any directory doesn't exist in the _filePath_
- if _filePath_ is empty string, create a file with random name under tmp directory and return a string(path to the file)
- parameter named to _filePath_ with capital P to avoid confusion with the filepath package


ahmad-abuziad:
@purpleidea review please

ahmad-abuziad:
@purpleidea please review

purpleidea:
Where is this utility function being used?

PR #795, opened 2025-02-23
exec_test: Fix wrong err variable being checked

hades:
## Tips:

```topic: Capitalized message with no trailing period```

or:

```topic, topic2: Capitalized message with no trailing period```

```
```

```
gofmt -w yourcode.go
```

```
git checkout master
git pull origin master
git checkout your-feature
git rebase master
```

```
git add -p		# add new changes
```

### DIFF
--- a/engine/resources/exec_test.go
+++ b/engine/resources/exec_test.go
@@ -412,7 +412,7 @@ func TestExecTimeoutBehaviour(t *testing.T) {
 	}
 
 	exitErr, ok := err.(*exec.ExitError) // embeds an os.ProcessState
-	if err != nil && ok {
+	if ok {
 		pStateSys := exitErr.Sys() // (*os.ProcessState) Sys
 		wStatus, ok := pStateSys.(syscall.WaitStatus)
 		if !ok {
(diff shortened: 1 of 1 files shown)


purpleidea:
Unless you know something different, these are equivalent.

purpleidea:
Some might argue actually technically less correct too.

hades:
> Unless you know something different, these are equivalent.

err can be any type, not necessarily ExitError, no?

hades:
> Some might argue actually technically less correct too.

Elaborate.

purpleidea:
(Closing this, but if you think I made a mistake, feel free to re-open.)

hades:
Well, for one, err cannot be nil (see line 411). So at the very least the check should be removed from both clauses.

W.r.t. the correctness, I don't think comparing a value to nil (which is the actual value in case when type assertion fails according to the manual) counts as "touching" it. However, I don't know enough of golang to claim I have a better understanding here.

Would you like me to remove the check instead?

purpleidea:
Indeed it looks like it can't be nil. This was likely just copy+pasta from early ugly exec code, and is likely similar in shape to the exec.go stuff. Up to you if you want to clean it up.

hades:
Updated the change.

purpleidea:
Added commit prefix and merged in: 63d7b8e51e136c11a4728c2c95f5044aa0747f52

PR #796, opened 2025-02-24
Fix typos, YAML problems, and clarify one comment

hades:


purpleidea:
wow, amazing work on all the typo fixing-- honestly, I've always been good at spelling, so I don't know how all of these crept in! I bet if you `git blame` they're mostly other people, right? =D

I have a WIP branch with a "spelling" test, but it always had so many false positives. If you have something better, lmk!

purpleidea:
(first commit merged, thanks, remaining two pending)

hades:
> If you have something better, lmk!

The spelling errors I fixed here were found by codespell[1]. Unlike spellcheckers for normal text, it uses a blacklist approach, so it looks for common typos and misspellings, which generates less false positives.

I plan to include it in the pre-commit checks.

[1] https://github.com/codespell-project/codespell/tree/main

purpleidea:
> pre-commit checks

Can we have it so pre-commit checks are actually shell files to run from test/test-* so that we can have the _same_ scripts be usable in our test suite if we want them there too?

(I'd like to avoid duplicating the logic in two places.)

I also have a test-spelling.sh stub which I never pushed which we can add codespell to if you like?

hades:
> Can we have it so pre-commit checks are actually shell files to run from test/test-* so that we can have the _same_ scripts be usable in our test suite if we want them there too?
> 
> (I'd like to avoid duplicating the logic in two places.)

Yes, I would like to avoid duplication too. However, I propose to go the other way, and migrate some of the existing test-*.sh (e.g. gofmt, yamlfmt, mclfmt, basically anything based on find_files/bad_files pattern) to pre-commit and then have GitHub CI run pre-commit on pull requests too. This way the logic will be in one place, the pre-commit config.

> I also have a test-spelling.sh stub which I never pushed which we can add codespell to if you like?

Is there anything there that you want to keep other than spelling? If yes, I'll see if I can reimplement it with pre-commit.

purpleidea:
> migrate some of the existing test-*.sh to pre-commit

Actually I do _not_ want this, sorry. I do not want to add a python dependency to this project. We'd like to be as close as possible to bash+golang as we can, and furthermore, keeping everything as a ./test.sh runner lets us easily move this to additional CI's. Similarly, pre-commit could be seen as a "CI" and so any logic you want in there, please make sure it's first in test/*

purpleidea:
(To be clear, this doesn't mean you can't use pre-commit or add the config file for it of course, but we aren't going to depend on it for our test suite or add specific things that only work there.)

purpleidea:
I think all is merged, thank you!

PR #797, opened 2025-02-26
lang: core: os: Add is_virtual function

lourencovales:
This is a basic implementation of a detection method for whether mgmt is running in a virtualized or containerized environment. We achieve this by doing two types of checks: on one hand, we check if the DMI/SMBIOS info can give us information about being on a virtualized env; on the other, we check for the presence of known files (and its contents) can confirm if we're inside a container. Either of these situations will cause the function to return true, with the default case being false. All of these checks are relatively naive and can be improved by looking at the main inspiration for this implementation, which was systemd's own virtualization check.

### DIFF
--- a/lang/core/os/is_virtual.go
+++ b/lang/core/os/is_virtual.go
@@ -1,0 +1,159 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+// Additional permission under GNU GPL version 3 section 7
+//
+// If you modify this program, or any covered work, by linking or combining it
+// with embedded mcl code and modules (and that the embedded mcl code and
+// modules which link with this program, contain a copy of their source code in
+// the authoritative form) containing parts covered by the terms of any other
+// license, the licensors of this program grant you additional permission to
+// convey the resulting work. Furthermore, the licensors of this program grant
+// the original author, James Shubin, additional permission to update this
+// additional permission if he deems it necessary to achieve the goals of this
+// additional permission.
+
+package coreos
+
+import (
+	"context"
+	"errors"
+	"os"
+	"runtime"
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/funcs/simple"
+	"github.com/purpleidea/mgmt/lang/types"
+	"github.com/purpleidea/mgmt/util/errwrap"
+)
+
+// The Desktop Management Interface (DMI) is a framwork for exposing system data
+// to software. We will make use of it to determine whether we're using a
+// virtualization platform. dmiFilesSlice is a slice of file paths which might
+// contain information related to the presence of a virtualization platform.
+// This is where we might find the values present in virtualizationVendorSlice.
+var dmiFilesSlice = []string{
+	"/sys/class/dmi/id/product_name",
+	"/sys/class/dmi/id/sys_vendor",
+	"/sys/class/dmi/id/board_vendor",
+	"/sys/class/dmi/id/bios_vendor",
+	"/sys/class/dmi/id/product_version",
+}
+
+// virtualizationVendorSlice is a slice of strings that might be found in DMI
+// related files (dmiFilesSlice) during the checks performed for the presence of
+// virtualization.
+var virtualizationVendorSlice = []string{
+	"Amazon EC2",
+	"Apple Virtualization",
+	"BHYVE",
+	"Bochs",
+	"Google Computer Engine",
+	"Hyper-V",
+	"innotek GmbH",
+	"KubeVirt",
+	"KVM",
+	"OpenStack",
+	"Oracle Corporation",
+	"Parallels",
+	"QEMU",
+	"VMware",
+	"Xen",
+}
+
+func init() {
+	simple.ModuleRegister(ModuleName, "is_virtual", &simple.Scaffold{
+		T: types.NewType("func() bool"),
+		F: IsVirtual,
+	})
+}
+
+// IsVirtual is a shim for the isVirtual function
+func IsVirtual(ctx context.Context, input []types.Value) (types.Value, error) {
+	b, err := isVirtual(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &types.BoolValue{V: b}, nil
+}
+
+// isVirtual is a function that executes two types of checks: first, it checks
+// whether we're running on Linux. If that's the case, we run checks related
+// with the presence of virtualization platforms. If any of those checks returns
+// true, then so does this function. Otherwise, it's assumed that it's not a
+// virtualized environment.
+func isVirtual(ctx context.Context) (bool, error) {
+	// if we implement detection for OS other than Linux, this logic will have
+	// to change
+	if runtime.GOOS != "linux" {
+		return false, errors.New("operating system is not Linux")
+	}
+
+	cpuInfoCheck, err1 := cpuInfoCheck(ctx)
+	if err1 == nil && cpuInfoCheck {
+		return true, nil
+	}
+
+	dmiFileCheck, err2 := dmiFileCheck(ctx)
+	if err2 == nil && dmiFileCheck {
+		return true, nil
+	}
+
+	if err1 != nil || err2 != nil {
+		return false, errwrap.Append(err1, err2)
+	}
+
+	return false, nil
+
+}
+
+// In an x86 system, there's a check to detect virt envs. The Linux kernel adds
+// the "hypervisor" flag to the CPU flags.
+func cpuInfoCheck(ctx context.Context) (bool, error) {
+	cpuInfo, err := os.ReadFile("/proc/cpuinfo")
+	if err != nil {
+		return false, err
+	}
+	if strings.Contains(string(cpuInfo), "hypervisor") {
+		return true, nil
+	}
+	return false, nil
+}
+
+// Check if any of the slices of virtualizationVendorSlice are present in any of
+// the DMI files contained on dmiFilesSlice. If that's the case, then we return
+// this function as true. This approach was inspired on systemd's work for a
+// similar purpose.
+// https://github.com/systemd/systemd/blob/main/src/basic/virt.c#L158
+func dmiFileCheck(ctx context.Context) (bool, error) {
+	for _, dmiFile := range dmiFilesSlice {
+		dmiFileContent, err := os.ReadFile(dmiFile)
+		if err != nil && !os.IsNotExist(err) {
+			return false, err
+		} else if err != nil {
+			continue
+		}
+
+		for _, vendor := range virtualizationVendorSlice {
+			if strings.Contains(string(dmiFileContent), vendor) {
+				return true, nil
+			}
+		}
+	}
+
+	return false, nil
+}
--- a/examples/lang/is_virtual.mcl
+++ b/examples/lang/is_virtual.mcl
@@ -1,0 +1,8 @@
+import "fmt"
+import "os"
+
+$s = fmt.printf("%t", os.is_virtual())
+
+print "is_virtual" {
+	msg => $s,
+}


lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,138 @@
...
+	default:
+		return &types.BoolValue{V: false}, nil
+	}
+}
+
+// We make use of systemd's work for detecting virtualization platforms, and
+// check if any of the keys of virtualizationVendorMap is present on any of the
+// DMI files present on dmiFilesSlice. If that's the case, then we return this
+// func as true. https://github.com/systemd/systemd/blob/main/src/basic/virt.c
+func virtCheck() bool {
```
purpleidea, line 105: Don't be afraid to have this signature include an error as well, and a context as input, if that makes sense.

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,138 @@
...
+func containerCheck() bool {
+	_, err := os.Stat("/run/.containerenv") // Podman
+	if err == nil {
+		return true
+	}
+	_, err = os.Stat("/.dockerenv") // Docker
+	if err == nil {
+		return true
+	}
+	wsl, _ := os.ReadFile("/proc/sys/kernel/osrelease") // WSL
```
purpleidea, line 129: Don't we care about errors? Or what they are? For example, IO error vs ErrNotFound?

purpleidea:
I think being "in a container" in and of itself, should not necessarily warrant is_virtual being true. But maybe this is a hot topic?

lang/core/os/is_virtual.go
lourencovales: For the first two checks, we only care if the file exists, and this is the simpler and cheaper way to check for it that I could think of. If there are errors, I don't think we should care what they are: if they're ErrNotFound, this just means that we're not in that type of container; if they're something else, I don't think this is the place to handle whatever else is causing this simple check to fail!

Whenever I tried to leverage something that makes use of errors to check if the file exist (making use of errors.Is(err, fs.ErrExist)), I couldn't find a way that would attempt to create the file and fail if it already existed.

lourencovales:
> I think being "in a container" in and of itself, should not necessarily warrant is_virtual being true. But maybe this is a hot topic?

I had read https://github.com/purpleidea/mgmt/issues/737 to mean that a container would be considered the same way. I guess the answer here depends on the scope of this function, what would this be used for - is there a situation where we want to treat virt and containers differently? If so, we could easily divide this into two different checks - one for virt, one for containers.

lang/core/os/is_virtual.go
purpleidea: > 	wsl, _ := os.ReadFile("/proc/sys/kernel/osrelease") // WSL

This doesn't tell you if the file exists or not. Since an empty file would still be zero value for the output var. Right?

purpleidea:
> two different checks - one for virt, one for containers

I completely agree. But for this patch let's just focus on `is_virtual` for now.

purpleidea:
Remember our lesson about `const` from last patch too. Maybe that will change how you write some of this code.

lourencovales:
@purpleidea I've remade the PR to reflect your comments, lmk if this is up to snuff

lourencovales:
I realized that I need to redo the commit message as well - leaving a comment here as a way to remind myself before the next commit

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,121 @@
...
+			return returnTrue, nil
+		}
+	}
+
+	// We make use of systemd's work for detecting virtualization platforms, and
+	// check if any of the keys of virtualizationVendorMap is present on any of
+	// the DMI files present on dmiFilesSlice. If that's the case, then we
+	// return this func as true.
+	// https://github.com/systemd/systemd/blob/main/src/basic/virt.c
+	for _, dmiFile := range dmiFilesSlice {
```
purpleidea, line 109: Is this really the right logic for detecting this? We loop through those files?

If so at least inline the code please. See: https://go.dev/wiki/CodeReviewComments#indent-error-flow

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,121 @@
...
+	// If we implement detection for OS other than Linux, this logic will have
+	// to change
+	opersys := runtime.GOOS
+	if opersys != "linux" {
+		return nil, errors.New("we're not running on Linux, exiting")
+	}
+
+	// If we're on x86, we can make use of a simple check to detect virt envs
+	cpuInfo, err := os.ReadFile("/proc/cpuinfo")
+	if err == nil {
```
purpleidea, line 98: Can you see how to remove the extra nesting here?

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,121 @@
...
+
+var dmiFilesSlice = []string{
+	"/sys/class/dmi/id/product_name",
+	"/sys/class/dmi/id/sys_vendor",
+	"/sys/class/dmi/id/board_vendor",
+	"/sys/class/dmi/id/bios_vendor",
+	"/sys/class/dmi/id/product_version",
+}
+
+var (
```
purpleidea, line 71: We don't want this, but what you _might_ want to achieve the _idea_ of this is you implement the logic as a helper function, eg:

func isVirtual(ctx context.Context) (bool, error) {

}

which is the only thing called by the real IsVirtual function.

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,121 @@
...
+	"errors"
+	"os"
+	"runtime"
+	"strings"
+
+	"github.com/purpleidea/mgmt/lang/funcs/simple"
+	"github.com/purpleidea/mgmt/lang/types"
+)
+
+var virtualizationVendorMap = map[string]string{
```
purpleidea, line 43: This at least needs a comment in the standard form explaining what this is. Same about the other var.

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,121 @@
...
+// check whether we're running on Linux. If that's the case, we run checks
+// related with the presence of virtualization platforms. If any of those checks
+// returns true, then so does this function. Otherwise, we assume that we're not
+// in a virtualized environment.
+func IsVirtual(ctx context.Context, input []types.Value) (types.Value, error) {
+	// If we implement detection for OS other than Linux, this logic will have
+	// to change
+	opersys := runtime.GOOS
+	if opersys != "linux" {
+		return nil, errors.New("we're not running on Linux, exiting")
```
purpleidea, line 93: error messages are not "talking" do us, we just state the issue in a bland way, eg: "$osname not supported". As a follow-up, there's no reason to believe we're "exiting".

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// If we're on x86, we can make use of a simple check to detect virt envs
```
purpleidea, line 96: What's the rule about comments?

purpleidea (changes requested):
Okay here's some feedback! hth

lang/core/os/is_virtual.go
lourencovales: My C is terrible, but it seems to be logic being applied here: https://github.com/systemd/systemd/blob/main/src/basic/virt.c#L158

lang/core/os/is_virtual.go
lourencovales: OK, I understand what was meant now.

lang/core/os/is_virtual.go
purpleidea: It's okay to steal their idea and copy it and reference "this is the way systemd does it" (which you should in your comment) but you should also describe the logic here, since in general our code base should be able to standalone without having to refer to some other git repo which might change/move/etc.

And of course the primary goal is not to get this function, the goal is to understand what we're doing =D

lourencovales: I'm confused now: when you say "you should also describe the logic here", is this not what I did in the comment for this block? I'm a bit at a loss on how to proceed here

purpleidea: Explain _why_ looking in these files does what we want.

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,143 @@
...
+	return false, nil
+}
+
+// Check if any of the keys of virtualizationVendorMap are present in any of the
+// DMI files contained on dmiFilesSlice. If that's the case, then we return this
+// function as true. This approach was inspired on systemd's work for a similar
+// purpose. https://github.com/systemd/systemd/blob/main/src/basic/virt.c#L158
+func dmiFileCheck(ctx context.Context) (bool, error) {
+	for _, dmiFile := range dmiFilesSlice {
+		dmiFileContent, err := os.ReadFile(dmiFile)
```
purpleidea, line 132: Name two distinct ways that this can error.

Q1: How do we differentiate between the two?
Q2: What do we do differently for each case?

Hint: You will need to do this to write this code correctly.

purpleidea (commented):
See if you can solve this Q1/Q2 and then ping me let's setup a time to chat for 30 min on next topic and remind me to discuss comments!

lang/core/os/is_virtual.go
lourencovales: For Q1: the file might not exist; or if they do, we might not have permissions to read it.
For Q2: first situation, we might want to just try another file (which was the approach I had in mind with the L133-135); for the second, we would need to escalate privileges, but I've no way how to deal with it other than quitting with an error saying as much.

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func init() {
+	simple.ModuleRegister(ModuleName, "is_virtual", &simple.Scaffold{
+		T: types.NewType("func() bool"),
+		F: IsVirtual,
+	})
+}
+
+// IsVirtual is a shim for the isVirtual function,
```
purpleidea, line 88: This is the doc string which will automatically get used in the documentation. So this one should be good =D

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,160 @@
...
+	cpuInfoCheck, err1 := cpuInfoCheck(ctx)
+	dmiFileCheck, err2 := dmiFileCheck(ctx)
+
+	switch {
+	case cpuInfoCheck:
+		return true, nil
+	case dmiFileCheck:
+		return true, nil
+	case err1 != nil || err2 != nil:
+		return false, errwrap.Append(err1, err2)
```
purpleidea, line 119: Did you check what `errwrap.Append(nil, err2)` or `errwrap.Append(err1, nil)` returns?

I think the logic of this function can be simplified too.

What do you think of:

cpuInfoCheck, err1 := cpuInfoCheck(ctx)
if err1 == nil && cpuInfoCheck {
    return true, nil
}

dmiFileCheck, err2 := dmiFileCheck(ctx)
if err2 == nil && dmiFileCheck {
    return true, nil
}

if err1 == nil && err2 == nil {
    return false, nil
}

return ...

What do you think?

My unwritten rule is you don't want to use switch unless you really have a lot of options.

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,160 @@
...
+	"/sys/class/dmi/id/board_vendor",
+	"/sys/class/dmi/id/bios_vendor",
+	"/sys/class/dmi/id/product_version",
+}
+
+// virtualizationVendorSlice is a slice of strings that might be found on DMI
+// related files (dmiFilesSlice) during the checks performed for the presence of
+// virtualization.
+var virtualizationVendorSlice = []string{
+	"KVM",
```
purpleidea, line 62: Does the order matter? If so, OK, tell us about this. If not, then sort it. Should this be a map?

examples/lang/virtualization.mcl
```diff
@@ -0,0 +1,8 @@
+import "os"
```
purpleidea, line 1: is_virtual.mcl please, thanks for adding it!

examples/lang/virtualization.mcl
```diff
@@ -0,0 +1,8 @@
...
+import "fmt"
+
+$s = fmt.printf("%v", os.is_virtual())
```
purpleidea, line 4: What's better than %v here?

purpleidea (commented):
Here you go! Hope you are learning. Patch is great, but let's be perfect!

examples/lang/virtualization.mcl
lourencovales: %b, since this will only be a boolean

lang/core/os/is_virtual.go
lourencovales: Upon reading a bit, seems like maps are better than slices for this purpose, so I am changing this to a map[string]struct{}

lourencovales:
@purpleidea finally pushed the latest version, please review!

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if err1 == nil && err2 == nil {
+		return false, nil
+	}
+
+	if err1 != nil || err2 != nil {
+		return false, errwrap.Append(err1, err2)
+	}
+
+	return false, nil
```
purpleidea, line 124: How do you get to this line?

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,163 @@
...
+func dmiFileCheck(ctx context.Context) (bool, error) {
+	for _, dmiFile := range dmiFilesSlice {
+		dmiFileContent, err := os.ReadFile(dmiFile)
+		if err != nil && !os.IsNotExist(err) {
+			return false, err
+		} else if err != nil {
+			continue
+		}
+
+		for vendor := range virtualizationVendorMap {
```
purpleidea, line 155: What's the advantage to using a map here and what's the downside?

purpleidea (commented):
So everytime you write a patch. Think slowly through it.
And after you commit and BEFORE you push your commit, read through your commit. Every single line... Get good at reading code, and try and "run" the code in your head.
Does the code do what you expect? Is the code what you wanted to push? Do you still have any tunnel vision from writing it?

HTH

lang/core/os/is_virtual.go
```diff
@@ -0,0 +1,163 @@
...
+// This is where we might find the values present in virtualizationVendorMap.
+var dmiFilesSlice = []string{
+	"/sys/class/dmi/id/product_name",
+	"/sys/class/dmi/id/sys_vendor",
+	"/sys/class/dmi/id/board_vendor",
+	"/sys/class/dmi/id/bios_vendor",
+	"/sys/class/dmi/id/product_version",
+}
+
+// virtualizationVendorMap is a slice of strings that might be found on DMI
```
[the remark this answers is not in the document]
lourencovales: Leftovers from a previous incarnation. Will correct.

lang/core/os/is_virtual.go
lourencovales: You never do. I noticed that this was redundant - but this made the error handling more obvious/complete. I can remove L116-118, if that's better.

lang/core/os/is_virtual.go
lourencovales: Advantage might be that lookup is cheaper than if I was using a slice; downside is that this is worse if I need to add keys to the map (which is expensive, relatively speaking)

purpleidea: What lookup is happening here?

purpleidea: And what's the downside here.

lourencovales: > What lookup is happening here?

The range is looking over the keys of the virtualizationVendorMap map; which makes me think that another advantage over slices is that a range over slices has to lookup both index and values, while with a map we can do it with only the value - or at least this is my intuition :-D

> And what's the downside here.

If it's not what I mentioned before (viz., a situation where I'd need to add items to this map), then I don't know what you're referring to here.

purpleidea:
\o/

woo hoo merged in f2d9219218eea41589da3a0f875e85cdfcdf4b3d

PR #799, opened 2025-03-11
Add some Module examples

karpfediem:
I will be adding some more example modules here that I am using, which I can share.

This is a Work In Progress for now and I'm happy to improve upon these.

---
This MR includes some commits to allow me to work on NixOS, which I will remove before changing from `Draft` state.
I just don't want to rebase all the time while still working on this.
The changes of this MR should remain limited to the `modules/` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,15 @@ sites/
 .*.sw[op]
 # prevent `echo foo 2>1` typo errors by making this file read-only
 1
+# Devenv
+.devenv*
+devenv.local.nix
+
+# direnv
+.direnv
+
+# pre-commit
+.pre-commit-config.yaml
+
+result
+graphs
--- a/devenv.lock
+++ b/devenv.lock
@@ -1,0 +1,116 @@
+{
+  "nodes": {
+    "devenv": {
+      "locked": {
+        "dir": "src/modules",
+        "lastModified": 1740460834,
+        "owner": "cachix",
+        "repo": "devenv",
+        "rev": "9e4003b2702483bd962dac3d4ff43e8dafb93cda",
+        "type": "github"
+      },
+      "original": {
+        "dir": "src/modules",
+        "owner": "cachix",
+        "repo": "devenv",
+        "type": "github"
+      }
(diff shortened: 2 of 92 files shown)


purpleidea:
Nice! If you want to rename the bash commit properly, I could cherry-pick/merge that one.

For the mcl stuff, needs tabs instead of spaces.

karpfediem:
Making tests work properly under NixOS took me a bit longer than anticipated.
I will add module descriptions soon, but at another time.

purpleidea:
Overall the whole thing fails, but if the shebang commits and other bits are ready to be cherry-picked, maybe put those in a separate PR so I can see it passes and I can pull those in early.

purpleidea:
(Doesn't need to pass the race tests, just the basic and shell.)

karpfediem:
> Overall the whole thing fails, but if the shebang commits and other bits are ready to be cherry-picked, maybe put those in a separate PR so I can see it passes and I can pull those in early.

Sorry, but i think it's not ready yet afterall. There are many more cases of [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) filepaths.
For now i have fixed my own issues using a shell with emulated FHS paths.
I would like to rather address it all at the same time in a separate issue after giving it some more thought.

purpleidea:
FYI: I cherry-picked/merged/etc most of this. You may wish to rebase at some point or open a new PR.

modules/debian/apt_repository_sources/main.mcl
```diff
@@ -0,0 +1,47 @@
...
+		  sources => [$source_default, $source_default_updates, $source_security,],
+		}
+		file "/etc/apt/sources.list" {
+		  state => $const.res.file.state.exists,
+		  content => golang.template(deploy.readfile("/files/sources.list.tmpl"), $data),
+		}
+	}
+}
+
+include lists()
```
purpleidea, line 47: FYI: side effects in modules won't be allowed. The compiler will complain if you try and import this module.

PR #801, opened 2025-03-20
Streamline finding files in scripts

karpfediem:
Replace most usages of `find` by `git ls-files`

Some reasons for using git over find:
- Respects .gitignore
  - no need to redefine/copy-paste each ignored directory with every `find` call (configured only once)
- Scans only relevant files (only lists files Git knows about)
- git index is already optimized for this, while `find` needs to traverse directories

Here are the options in use explained in the man page of `git ls-files`:
```
       -c, --cached
           Show all files cached in Git’s index, i.e. all tracked files. (This is the default if
           no -c/-s/-d/-o/-u/-k/-m/--resolve-undo options are specified.)
       -o, --others
           Show other (i.e. untracked) files in the output
       --exclude-standard
           Add the standard Git exclusions: .git/info/exclude, .gitignore in each directory, and
           the user’s global exclusion file.
```
---
I have also added the vendor directory to the .gitignore file, since it appeared in several find commands, but i could not figure out if this directory is actually (still) in use or not.

---

I was not sure about it but I have generally ignored the `-maxdepth` arguments which was previously 9.
The awk command under `yamlfmt:` (https://github.com/karpfediem/mgmt/blob/0d41ad9f786d045eefc32e843268503c80c7ec2a/Makefile#L272) serves to limit the `-maxdepth` to the same one as in the find command (find . -maxdepth 3) 
```sh
			| awk -F'/' 'NF<=4' \
```
If this depth isn't important either that line can be removed as well.

---

I also have a question regarding this line https://github.com/karpfediem/mgmt/blob/0d41ad9f786d045eefc32e843268503c80c7ec2a/Makefile#L38

Does this comment still apply? E.g. is the cause the slow find command or the longer tests due to more files?

---
Please let me know I should explain anything else in more detail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .envrc
 old/
 tmp/
+vendor/
 *WIP
 *_stringer.go
 mgmt
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,9 @@ SHELL = /usr/bin/env bash
 .PHONY: funcgen
 .SILENT: clean
 
-# a large amount of output from this `find`, can cause `make` to be much slower!
-GO_FILES := $(shell find * -name '*.go' -not -path 'old/*' -not -path 'tmp/*')
-MCL_FILES := $(shell find lang/ -name '*.mcl' -not -path 'old/*' -not -path 'tmp/*')
+# a large amount of output from this `find`, can cause `make` to be much slower! TODO: does this comment still apply? E.g. is the cause the slow find command or the longer tests due to more files?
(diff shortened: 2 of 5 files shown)


karpfediem:
Disadvantages: nix build is not git aware :upside_down_face: 
We could pivot to using `fd` (https://github.com/sharkdp/fd) as a `find` replacement
But using git ls-files makes things more complicated for me... lol

purpleidea:
Thanks anyways! Yeah I did consider using git ls-files and maybe even did previously, but when you're developing it isn't in git yet, so you can't do stuff.

PR #802, opened 2025-03-20
Use /usr/bin/env for a more generic shebang

karpfediem:
Replaces all static file paths such as `#!/bin/bash` in shebangs by prefixing the binary with `#!/usr/bin/env`

This should be preferable due to more widespread compatibility, especially with [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) non-compliant systems such as NixOS.

In my experience it's universal enough that it is generally the recommended way. It's more universal than /bin/bash. It works even on busybox and termux.

However this requires that those binaries are in the user's PATH.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@
 # additional permission if he deems it necessary to achieve the goals of this
 # additional permission.
 
-SHELL = /usr/bin/env bash
+SHELL = bash
 .PHONY: all art cleanart version program lang path deps run race generate build build-debug crossbuild clean test gofmt yamlfmt format docs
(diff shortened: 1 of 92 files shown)


purpleidea:
LGTM! Will merge shortly, thanks! Hope this helps you hacking, and we'll be waiting for when you upgrade away from Nix ;)

karpfediem:
Added another small fix, which requires less patching when packaging for nix

purpleidea:
Okay I squashed these two and pushed in: 7e8ced534f34fc177c3e5507f6bdb3921785713c

Thanks!

PR #805, opened 2025-04-08
lang: core: iter: add a range function

lourencovales:
This commit implements a range function that mimicks python's range built-in by having a start, stop, and range argument.

At the moment, the Stream function is not function correctly - and hence this draft to get some help. I'm also still missing the integration with the for function, as well as a comprehensive test suit.

### DIFF
--- a/examples/lang/range.mcl
+++ b/examples/lang/range.mcl
@@ -1,0 +1,19 @@
+import "iter"
+import "fmt"
+
+print "range" {
+	msg => fmt.printf("%v", iter.range(5)),
+}
+
+for $i, $v in iter.range(34, 50) {
+	print [fmt.printf("index is: %d, value is: %d", $i, $v),] {
+		Meta:autogroup => false,
+	}
+}
+
+for $i, $v in iter.range(10) {
+	print [fmt.printf("index is: %d", $i),] {
+		msg => fmt.printf("value is: %d", $v),
+		Meta:autogroup => false,
+	}
+}
--- a/lang/core/iter/range.go
+++ b/lang/core/iter/range.go
@@ -1,0 +1,260 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


lang/core/iter/range.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+	}
+	if len(args) == 3 { // we have all the args
+		result, err = obj.run(ctx, args[0].Int(), args[1].Int(), args[2].Int())
+		if err != nil {
+			return nil, err
+		}
+	}
+	// TODO: should call the for function here and return the result
+	return &types.ListValue{V: result}, nil
```
purpleidea, line 222: Why not build this in run() and return directly above?
Also look at some examples, it wasn't built correctly, it needs a type.

lang/core/iter/range.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+// Run is the private helper function that calculates the range according to the
+// inputs provided.
+func (obj *RangeFunc) run(ctx context.Context, start, stop, step int64) ([]types.Value, error) {
```
purpleidea, line 227: maybe name this `loop`

lang/core/iter/range.go
```diff
@@ -0,0 +1,262 @@
...
+			return []types.Value{}, nil
+		}
+		// calculating the lenght needed for initializing the list
+		length = (stop - start + step - 1) / step
+	} else {
+		if start <= stop {
+			// empty since step is negative and start < stop
+			return []types.Value{}, nil
+		}
+		// calculating the lenght needed for initializing the list
```
purpleidea, line 245: length*

examples/lang/range.mcl
```diff
@@ -0,0 +1,23 @@
+import "iter"
+import "fmt"
+
+print "range" {
+	msg => fmt.printf("%v", iter.range(1, 200, 3)),
+}
+
+#range (1 2 3)
```
purpleidea, line 8: for $i, $x in iter.range(3) {
        print "..." {
                msg => "...",
        }
}

purpleidea (changes requested):
Looking really close! Nice work.

While I think at least initially we should keep it as is, for future reference (tricky to understand) we can use the wrapped https://github.com/purpleidea/mgmt/blob/master/lang/core/list/list_concat.go#L59 pattern here. (Let's not do that today though.)

lourencovales:
@purpleidea I've submitted changes and the test suite. Moved this to an open PR as well, lmk what you find.

purpleidea (commented):
Looks great! If you are in the mood to push slightly, I think we can inline the code in loop() a bit better. We can pair on that together if you like.

purpleidea:
I reordered a method or two and made a nitpick changes and merged this! woo ! thanks =D

PR #810, opened 2025-04-18
engine: resources: file: Support symbolic links

xtexx:
Support managing symbolic links.

```
file "/tmp/symlink2" {
	source => "dir1/",
	symlink => true,
}
```

- Now file resources also generates auto-edges to `Source`.
- FileUID now sees `PATH/` and `PATH` as equivalent.

### DIFF
--- a/engine/resources/file.go
+++ b/engine/resources/file.go
@@ -1439,7 +1439,7 @@ func (obj *FileUID) IFF(uid engine.ResUID) bool {
 	if !ok {
 		return false
 	}
-	return obj.path == res.path
+	return obj.path == res.path || obj.path == res.path+"/" || obj.path+"/" == res.path
 }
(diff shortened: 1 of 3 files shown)


purpleidea:
Hi, thank you very much for your patch, and please accept my apology for the delay in my review.

After some consideration, the patch is unfortunately not precisely correct as there are some tricky edge cases that were not covered and it was difficult to explain. As a result, I took some time to rewrite this from scratch for you and others.

I may not have handled all the scenarios you had in mind, so please test and let me know if my versions suits your needs or if there is anything additional you require. As well there are lots of weird edge cases with symlinks, so if you are an expert here, please let me know of any mistakes I made!

If you'd like to work on something else, we'd love that too! Thanks

PR #812, opened 2025-04-24
engine: resources: Add a WatchFiles func to exec

lourencovales:
This adds a function that keeps track of changes to a given file inside the exec resource.

Keep this as a draft PR for now because it seems like we're generating more events than we should, looking for some feedback into this.

### DIFF
--- a/engine/resources/exec.go
+++ b/engine/resources/exec.go
@@ -119,6 +119,9 @@ type ExecRes struct {
 	// WatchCwd is the Cwd for the WatchCmd. See the docs for Cwd.
 	WatchCwd string `lang:"watchcwd" yaml:"watchcwd"`
 
+	// WatchFiles is a list of files that will be kept track of.
+	WatchFiles []string `lang:"watchfiles" yaml:"watchfiles"`
+
 	// WatchShell is the Shell for the WatchCmd. See the docs for Shell.
 	WatchShell string `lang:"watchshell" yaml:"watchshell"`
 
@@ -197,6 +200,12 @@ func (obj *ExecRes) Validate() error {
 		return fmt.Errorf("the Args param can't be used when Cmd has args")
 	}
 
+	for _, file := range obj.WatchFiles {
+		if file != "" && !strings.HasPrefix(file, "/") {
+			return fmt.Errorf("the path (`%s`) in WatchFiles must be absolute", file)
+		}
+	}
+
 	if obj.Creates != "" && !strings.HasPrefix(obj.Creates, "/") {
 		return fmt.Errorf("the Creates param must be an absolute path")
 	}
@@ -238,10 +247,13 @@ func (obj *ExecRes) Cleanup() error {
 
 // Watch is the primary listener for this resource and it outputs events.
 func (obj *ExecRes) Watch(ctx context.Context) error {
-	defer obj.wg.Wait()
+	wg := &sync.WaitGroup{}
+	defer wg.Wait()
 
 	ioChan := make(chan *cmdOutput)
 	rwChan := make(chan recwatch.Event)
+	filesChan := make(chan recwatch.Event)
+
 	var watchCmd *exec.Cmd
 	if obj.WatchCmd != "" {
 		var cmdName string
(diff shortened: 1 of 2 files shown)


engine/resources/exec.go
```diff
@@ -119,6 +119,9 @@ type ExecRes struct {
 	// WatchCwd is the Cwd for the WatchCmd. See the docs for Cwd.
 	WatchCwd string `lang:"watchcwd" yaml:"watchcwd"`
 
+	// WatchFile is the file that the WatchFile routine will track.
+	WatchFile string `lang:"watchfile" yaml:"watchfile"`
```
purpleidea, line 123: Don't we want this to be a list of files?

I have a hint to implementing this if you feel stuck.

engine/resources/exec.go
```diff
@@ -281,6 +285,45 @@ func (obj *ExecRes) Watch(ctx context.Context) error {
...
+				}
+
+				if !ok {
+					err := fmt.Errorf("channel shutdown")
+					event = recwatch.Event{Error: err}
+					shutdown = true
+				}
+
+				select {
+				case rwChan <- event:
```
purpleidea, line 316: You may not want to piggyback this onto the existing rwChan...

purpleidea (commented):
Looks great, let's change it up a bit though =D

lourencovales:
@purpleidea this should in better shape now

engine/resources/exec.go
```diff
@@ -281,6 +286,46 @@ func (obj *ExecRes) Watch(ctx context.Context) error {
 		}
 	}
 
+	for _, file := range obj.WatchFiles {
+		// since we're only dealing with files, recurse can be set to false
```
purpleidea, line 290: Why can't we be dealing with directories? From a API/UI point of view, how would we let the user differentiate between the two?

engine/resources/exec.go
```diff
@@ -281,6 +286,46 @@ func (obj *ExecRes) Watch(ctx context.Context) error {
...
+		recWatcher, err := recwatch.NewRecWatcher(file, false)
+		if err != nil {
+			return err
+		}
+		defer recWatcher.Close()
+
+		obj.wg.Add(1)
```
purpleidea, line 297: There's a bit of an issue that needs investigating...

_We_ definitely want to use a local wg. Not an obj.wg and it should have the same lifetime as the Watch function.

But there's some legacy code in obj.wg which we should see if we (probably I) ported correctly or is a bug/unoptimal scenario from years ago.

At least for us, we don't want obj.wg 

Do you understand why?

purpleidea (commented):
As you may have already figured, we probably want this to look similar to the: https://github.com/purpleidea/mgmt/blob/master/engine/resources/file.go#L477 code

engine/resources/exec.go
lourencovales: I thought this was the point of the function - to only deal with files. I can implement the ability for taking a directory as argument if that's a desired feature.
As for the other question, a directory would have to end with a forward slash (/), while a file wouldn't.

engine/resources/exec.go
lourencovales: If we're using obj.wg we're using a shared the same wait group instance across goroutines. Would that be a source of funkyness? If so, we definitely want a local wg.

lourencovales:
> Looks really nice, some small nits to look into. The most important question, does this feel natural to you? Does it make sense?

The whole "dance" around go routines became much more understandable after working on this issue, that's for sure. That's my biggest takeaway from this PR.

engine/resources/exec.go
purpleidea: Correct =D

engine/resources/exec.go
purpleidea: We want the local wg, since it's lifetime is only limited to the Watch function. If we had a helper function watch called, then we'd want something shared as well, but we don't here.

lourencovales:
@purpleidea we should be good to go now!

lourencovales:
@purpleidea this should in better shape now, let me know if there's anything missing

purpleidea:
I see a bug. What if Watchfiles of one res is len 4 and the other is 3?

lourencovales:
> I see a bug. What if Watchfiles of one res is len 4 and the other is 3?

@purpleidea I fixed this, and I also added another check in the Validate function that I noticed was also missing. Let me know if you spot anything else

purpleidea:
Thank you!

I merged a slightly retouched version. Have a look at the diff =D

lourencovales:
Thank you for merging the patch. I'll keep in mind the engineUtil.StrListCmp util :-)

purpleidea:
> Thank you for merging the patch. I'll keep in mind the engineUtil.StrListCmp util :-)

Thank _you_ for the patch! (I forget about that util too.) But note the change I made in Validate.

PR #814, opened 2025-05-07
modules: misc: Add cryptpad module

lourencovales:
This is a module to deploy cryptpad locally - as it is, this is only intended for dev purposes, not for production environments

### DIFF
--- a/modules/contrib/cryptpad/main.mcl
+++ b/modules/contrib/cryptpad/main.mcl
@@ -1,0 +1,51 @@
+import "fmt"
+
+class deploy() {
+
+	$pkgname = "cryptpad"
+	$installdir = "/opt/"
+	$gitbranch = "2025.3.1-rc"
+	$gitclonecommand = fmt.printf("git clone -b %s --depth 1 https://github.com/cryptpad/cryptpad.git %s", $gitbranch, $pkgname)
+	$shell = "/bin/bash"
+
+	exec "cryptpad-fetch" {
+		cwd => $installdir,
+
+		ifshell => $shell,
+		ifcmd => fmt.printf("test ! -d %s%s", $installdir, $pkgname),
+
+		shell => $shell,
+		cmd => $gitclonecommand,
+	}
+
+	exec "cryptpad-install" {
+		cwd => fmt.printf("%s%s", $installdir, $pkgname),
+
+		ifshell => $shell,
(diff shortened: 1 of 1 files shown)


purpleidea:
Does it work if you import it as a module and run it? (Hint: it probably won't.)

If you want to fix it up, let's move it into modules/contrib/cryptpad/ instead. You probably want it named main.mcl and you'll also need a metadata.yaml file in that same dir.

lourencovales:
@purpleidea I've made the changes you requested

purpleidea:
Can you show me a small stub of mcl that you tested that imports it to use it? You should likely get an error trying to import it.

lourencovales:
I think I figured out what I was missing here. So, in essence: 1) the .mcl needs to be called "main.mcl", 2) the different resources need to be inside a class statement. With this, I can do `import "modules/contrib/cryptpad/"` and all is (seemingly) well. 
But since I'm already messing with classes anyway, does it make sense to split this into different classes? Because for now this is a pretty "naive" module, it only serves the purpose of deploying the software and nothing else; but say I want to update instead of installing, I could achieve this with classes and reusing some of the workflow here right?

purpleidea:
The biggest "bug" here is that this won't import as a module since you can't have any side-effects by running an import.

So you likely want to put this all in a class, and then the user imports the module, and uses the class. Passing in any variables for example.

purpleidea:
> I can do import "modules/contrib/cryptpad/" and all is (seemingly) well

Are you sure?? It worked?

lourencovales:
> The biggest "bug" here is that this won't import as a module since you can't have any side-effects by running an import.
> 
> So you likely want to put this all in a class, and then the user imports the module, and uses the class. Passing in any variables for example.

That much I had already figured :smile: The current version (which I'll push in a second) works if I run a .mcl like this:

```
import "modules/contrib/cryptpad/"

include cryptpad.deploy()
```

The question I had is if makes sense to split this up in more classes to allow more fine grained control over the workflow of this module.

purpleidea:
Just waiting for basic test to pass.
FWIW if you want, you may also be able to change the `cp` to a file resource with a `source` param.
Eventually we should add a git resource, but that's for a future day.

lourencovales:
@purpleidea should be good to go now

purpleidea:
Merged, thank you and sorry for the delay!

Issue #822, opened 2025-09-30
Shadowing of package names in funcgen breaks in go1.25


karpfediem:
When generating functions with `funcgen`, the tool shells out to `go doc <pkgname>` and then parses the output via regex.

In Go 1.25 apparently `go doc` changed its package resolution order:

An innocuous (presumed stdlib) name like `math` can now resolve to a totally different package such as `github.com/cloudflare/circl/math`, etc.

This shadowing causes funcgen to pick up the wrong package and drop or mis-generate functions (e.g. missing `math.Abs`).

### Root cause

* `funcgen` relies on `go doc` and is being used with unqualified names.
* `go doc` matches multiple packages and may resolve to a non-stdlib package if present.
* The regex scraper then sees unexpected or missing function signatures.

purpleidea:
> In Go 1.25 apparently go doc changed its package resolution order:

:gasp:

Well thank you for noticing this, and also for finding it before I even got to 1.25 and got puzzled! Looking forward to merging your patch soon.

karpfediem:
Solved in https://github.com/purpleidea/mgmt/commit/307660e9bceebe4c6b6bb9676e8f7e6427e24c36 with https://github.com/purpleidea/mgmt/pull/823

PR #823, opened 2025-09-30
funcgen: Rely on go/packages instead of go doc calls

karpfediem:
Fixes https://github.com/purpleidea/mgmt/issues/822

Prevents breakage from Go 1.25’s changed go doc resolution and makes funcgen reproducible across Go versions.

* Replaced use of `go doc` + regex scraping with `golang.org/x/tools/go/packages` / `go/types` to reliably discover exported top-level functions.
* This fixes package shadowing issues that caused stdlib functions (e.g. `math`, `strings`) to disappear in Go 1.25.
* Expand variadic param calls with `...`.

❯ git diff --no-index --patch generated_funcs_old.go generated_funcs_new.go > generated_funcs.patch

[generated_funcs.patch](https://github.com/user-attachments/files/22627514/generated_funcs.patch)

The generated functions are still almost identical as the information just depends on the go version difference and is mostly intact.


purpleidea:
Thanks for this work! Looks marvellous =D

It needs a small tweak to fixup the go vet "reflow check" ... It basically just makes sure things wrap at 80 and you can run it with ./test/test-govet.sh pretty easily. The logs should show the failures.

I think you should officially own this code and we'll merge when you're happy!

I didn't dig into the new variadic changes you made... The "mcl" language doesn't technically support variadic, _however_ we do support fancy type unification which could (not in this patch though) make it so certain functions basically worked that way...

I did see PathfilepathJoin for example. Was that one of the things that your variadic work created here?

Danke!

karpfediem:
> I did see PathfilepathJoin for example. Was that one of the things that your variadic work created here?

Yep, well going away from the signature regex, the funcgen created a few more conversions such as PathfilepathJoin first
And the output was invalid go code, so i had to handle variadics somehow lol

> ./test/test-govet.sh

Will fix these later today, Cheers

karpfediem:
Thanks, will take a look if its simply mirroring the actual package docs or something i should fix
It might be good to trim up either way

karpfediem:
@purpleidea last commit is up to you, i also don't mind either way.
It does produce a cleaner diff to what we had before.

[generated_funcs.go.txt](https://github.com/user-attachments/files/22708872/generated_funcs.go.txt)

purpleidea:
Awesome! I've removed some unrelated go-mod issues and in doing so squashed this into one commit. Hope that's okay. Work looks great and is now merged in 307660e9bceebe4c6b6bb9676e8f7e6427e24c36

NOTE: If having a new release is easier for your team to work with in golang 1.25 lmk and I can spin one earlier rather than later.

purpleidea:
What's the next itch we can get you to scratch?

PR #843, opened 2026-02-05
engine: fix errors on re-reconcilation of existing users

### LINKED ISSUE #842: User resource errors around the default group
When changing a user's home directory outside of mgmt, we seem to immediately break the user resource and its handling - it starts producing `usermod` errors. 

Steps to reproduce:
1. Run the below mcl:
```
user "mgmttest" {
	state => "exists",
}
```

2. in /etc/passwd edit the home dir to be mgmttest1
3. Run mgmt, see the error

Even setting the home dir back to mgmttest still causes the error.

Reproduced with latest main branch build of mgmt, on Ubuntu 24.04 (`shadow-utils 4.13`, which provide usermod erroring below)

Error:

```
17:21:42 engine: user[mgmttest]: Error: error during Process(): usermod: no options Usage: usermod [options] LOGIN  Options:   -a, --append                  append the user to the supplemental GROUPS                                 mentioned by the -G option without removing                                 the user from other groups   -b, --badname                 allow bad names   -c, --comment COMMENT         new value of the GECOS field   -d, --home HOME_DIR           new home directory for the user account   -e, --expiredate EXPIRE_DATE  set account expiration date to EXPIRE_DATE   -f, --inactive INACTIVE       set password inactive after expiration                                 to INACTIVE   -g, --gid GROUP               force use GROUP as new primary group   -G, --groups GROUPS           new list of supplementary GROUPS   -h, --help                    display this help message and exit   -l, --login NEW_LOGIN         new value of the login name   -L, --lock                    lock the user account   -m, --move-home               move contents of the home directory to the                                 new location (use only with -d)   -o, --non-unique              allow using duplicate (non-unique) UID   -p, --password PASSWORD       use encrypted password for the new password   -P, --prefix PREFIX_DIR       prefix directory where are located the /etc/* files   -r, --remove                  remove the user from only the supplemental GROUPS                                 mentioned by the -G option without removing                                 the user from other groups   -R, --root CHROOT_DIR         directory to chroot into   -s, --shell SHELL             new login shell for the user account   -u, --uid UID                 new UID for the user account   -U, --unlock                  unlock the user account   -v, --add-subuids FIRST-LAST  add range of subordinate uids   -V, --del-subuids FIRST-LAST  remove range of subordinate uids   -w, --add-subgids FIRST-LAST  add range of subordinate gids   -W, --del-subgids FIRST-LAST  remove range of subordinate gids   -Z, --selinux-user SEUSER     new SELinux user mapping for the user account  : exit status 2
```

danielpodwysocki:
Fixes #842 
Currently, running a block of mcl such as this will error on the 2nd reconcilation, even if no changes are made to the resource:

```
user "mgmtiscool" {
	state => "exists",
}

```

This is due to `usermod` returning non-zero on `usermod <username>` with no args.

This patch makes it exit early, correcting that behavior.

### DIFF
--- a/engine/resources/user.go
+++ b/engine/resources/user.go
@@ -291,13 +291,6 @@ func (obj *UserRes) CheckApply(ctx context.Context, apply bool) (bool, error) {
 	var cmdName string
 	var args []string
 	if obj.State == "exists" {
-		if exists {
-			cmdName = "usermod"
-			obj.init.Logf("modifying user: %s", obj.Name())
-		} else {
-			cmdName = "useradd"
-			obj.init.Logf("adding user: %s", obj.Name())
-		}
 		if obj.AllowDuplicateUID {
 			args = append(args, "--non-unique")
 		}
@@ -318,6 +311,16 @@ func (obj *UserRes) CheckApply(ctx context.Context, apply bool) (bool, error) {
 		}
 		if obj.Shell != nil {
 			args = append(args, "--shell", *obj.Shell)
+		}
+		if exists && len(args) == 0 {
+			return false, nil
+		}
+		if exists {
+			cmdName = "usermod"
+			obj.init.Logf("modifying user: %s", obj.Name())
+		} else {
+			cmdName = "useradd"
+			obj.init.Logf("adding user: %s", obj.Name())
 		}
 	}
 	if obj.State == "absent" {


purpleidea:
Unfortunately this patch is too simple for what we need and it would mask a bigger issue. I've documented that research in: https://github.com/purpleidea/mgmt/issues/842

We'll need tests for sure. Going to close this PR for now, but feel free to open a new one if you are willing to tackle this!

Thanks =D

PR #850, opened 2026-02-07
engine: resources: add autoedge between http:file and file

### LINKED ISSUE #734: automatic edge between http:file and file

Oddly:
When an http:file resource has a Path set, we now create a reversed automatic edge to the corresponding file resource. This ensures the file exists on disk before the http server tries to serve it, removing the need for an explicit Depend declaration.

When Path is empty (inline Data is used instead), no autoedge is created since there is no file dependency.

We reuse the existing FileUID type for matching, so a file resource at `/tmp/some_file` will automatically be linked to an http:file that serves from the same path.

Closes #734

### DIFF
--- a/engine/resources/http_server_file.go
+++ b/engine/resources/http_server_file.go
@@ -318,6 +318,56 @@ func (obj *HTTPServerFileRes) Cmp(r engine.Res) error {
 	return nil
 }
 
+// UIDs includes all params to make a unique identification of this object. Most
+// resources only return one, although some resources can return multiple.
+func (obj *HTTPServerFileRes) UIDs() []engine.ResUID {
+	return []engine.ResUID{}
+}
+
+// AutoEdges returns the autoedge data for this resource. If the http:file has a
+// Path set, it creates a reversed edge to the file resource at that path, so
+// the file is ensured to exist before the http server tries to serve it.
+func (obj *HTTPServerFileRes) AutoEdges() (engine.AutoEdge, error) {
+	if obj.Path == "" {
+		return nil, nil // data is inline, no file dependency
+	}
+
+	var reversed = true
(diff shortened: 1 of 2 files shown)


purpleidea (changes requested):
Please also see how to make the git commit message tests pass too.

Thank you.

Oddly:
Alright, try 2. 
I've tried to properly understand these pieces and their place in the Mgmt code base and made changes accordingly.

I also added extensive (but hopefully still useful) tests and tested with `mgmt run` itself as well.

Open for feedback of course!

purpleidea:
Hey there @Oddly 

I think I wasn't ready for the AI onslaught when this PR was opened. Hopefully I've caught up now. I've also written up my views here: https://purpleidea.com/blog/2026/06/24/perspectives-on-ai-llms/

TL;DR: LLM/AI contributions are welcome, but it really must be the human 100% responsible for the style and quality and in general they should really know what they're merging. If you'd like to do more, let us know. I'd be curious how well you are at getting LLM's to write more complex mcl code.

In any case, I've revisited this PR, made some edits (due to new autoedge changes) and merged a new version. I didn't take off your authorship credit though. Hope you're fine with all that. b554b92c5

Cheers!

PR #851, opened 2026-02-07
cli: add provisioner subcommand to help output

### LINKED ISSUE #841: Provisioner command does not show up in help
When running `mgmt --help` the `provisioner` command is not listed

Oddly:
The provisioner command was registered via the entry package but never showed up in `mgmt --help` because the go-arg parser only knows about subcommands defined as struct fields in `cli.Args`.

We add a dummy `ProvisionerCmd` field following the same pattern already used by `EtcdCmd` — both get preempted in `main()` before the parser acts on them, but their presence in the struct is enough for go-arg to include them in the help text.

Closes #841

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -131,6 +131,9 @@ type Args struct {
 	// XXX: Can we do it nicely with the new arg parser? can it ignore all args?
 	EtcdCmd *EtcdArgs `arg:"subcommand:etcd" help:"run standalone etcd"`
 
+	// This never runs, it gets preempted in the real main() function.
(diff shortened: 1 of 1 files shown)


purpleidea:
PS: If you're looking for some more involved patches to write, there's lots of useful undocumented work to do, our issues list is a bit stale.

Ping in #mgmtconfig on Matrix if you'd like to discuss more with us.

Lastly, if you're using Claude to help you code, keep in mind that you personally are ultimately responsible for the contribution, so please do not send any "slop". Patches seem to look good so far, thank you!

Cheers!

purpleidea:
Fixed commit message and merged in: a26cb7d2562382e665129c7c3c1ec85bfc34e781

PR #853, opened 2026-02-07
lang/types: add tests for skipping unexported struct fields

Oddly:
This adds tests for the SkipPrivateFieldsOpt behaviour in ConfigurableTypeOf and ResTypeOf. The original fix in PR #685 (commit bf7e454) made it so unexported struct fields are skipped when mapping Go types to mcl types, but there was no test coverage for it.

The first test defines a struct with both exported and unexported fields, calls ConfigurableTypeOf with SkipPrivateFieldsOpt(true), and checks that only the exported fields appear in the result. It also verifies the inverse: with the option set to false, all fields are included. I confirmed the test fails when the skip logic is reverted.

The second test exercises ResTypeOf (which enables the option internally) on a struct with a lang tag and an unexported field, verifying the unexported field is excluded.

Closes: #686

### DIFF
--- a/lang/types/type_test.go
+++ b/lang/types/type_test.go
@@ -33,6 +33,7 @@ package types
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/purpleidea/mgmt/util"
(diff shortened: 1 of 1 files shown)


lang/types/type_test.go
```diff
@@ -1767,6 +1768,78 @@ func TestTypeOf0(t *testing.T) {
 	// TODO: implement testing TypeOf for struct field name mappings
 }
 
+// TestTypeOfSkipPrivateFields verifies that ConfigurableTypeOf with
+// SkipPrivateFieldsOpt(true) skips unexported struct fields. This is the
+// behaviour added in PR #685 / commit bf7e454 to fix issue #684 where
```
purpleidea: Don't reference GH #'s or commit hashes in the code. There's no guarantee that at least the former will be constant.

purpleidea (changes requested):
Bunch of failing tests...

Issue #494, opened 2019-03-24
Add duration check linter


purpleidea:
We'd like to add as much automatic testing as possible. This looks like an interesting test we could add:

https://github.com/charithe/durationcheck

Investigate, and if it's reasonable, please add a test for it!

(Alternatively, maybe it could get added or is already added in https://github.com/purpleidea/mgmt/issues/487 and if so, then that should take precedence over this.)

PR #854, opened 2026-02-07
test: enable durationcheck linter and fix duration multiplication bug

Oddly:
The first commit fixes a bug in the rate limiter where `time.Duration(d) * time.Millisecond` was multiplying two Duration values together. The variable `d` is already a `time.Duration` returned by `DelayFrom`, so the multiplication produced a result with squared time units — a timer that would be off by a factor of a million. The fix is to use `d` directly since it already has the correct unit.

The second commit enables the `durationcheck` linter in the golangci-lint test to catch this class of bug in the future. I ran it across the codebase and the fixed line was the only violation.

Closes: #494

### DIFF
--- a/test/test-golangci-lint.sh
+++ b/test/test-golangci-lint.sh
@@ -25,6 +25,7 @@ glc='golangci-lint run --disable-all'
 glc="$glc --enable=goimports"
 glc="$glc --enable=revive"
 glc="$glc --enable=misspell"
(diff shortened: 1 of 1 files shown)


purpleidea:
This would be a good candidate for a test involving the [testing/synctest](https://pkg.go.dev/testing/synctest) package.

PR #858, opened 2026-02-07
lang: fuzz: Migrate to native Go fuzzing

### LINKED ISSUE #664: updated fuzzing
When we get to newer versions of golang, we might want to update our fuzzing. The new format looks roughly like:

```
func FuzzParser(f *testing.F) {
   f.Fuzz(func(t *testing.T, data []byte) {
      original fuzz function body goes here
   })
}

Then
s/return/t.Skip()/
s/panic/t.Fatal/

finally
mv fuzz.go fuzz_test.go

and import "testing"
```

From notes taken from twitter. TBD for some future hacker to see if all this works with corpus, etc...

Oddly:
The mcl lexer/parser fuzzer used go-fuzz, which was the standard approach before Go 1.18 added native fuzzing support to the testing package. This replaces the old Fuzz function with a FuzzLexParse test that uses testing.F, so no external tooling is needed — just go test -fuzz.

The test preserves the same logic: feed random bytes to parser.LexParse and crash if a non-nil AST comes back alongside an error. The seed corpus includes a few representative mcl snippets via f.Add, and the Makefile can optionally pull in all the .mcl files from the project tree for broader seeding. The go-fuzz dependency is removed from make-deps.sh since it's no longer needed.

Closes #664


lang/fuzz/Makefile
```diff
@@ -28,12 +28,17 @@
...
 
-fuzz:
-	mkdir -p corpus/
-	find ../.. -name \*.mcl -exec cp {} corpus/ \;
-	go-fuzz-build
-	go-fuzz -bin=./fuzz-fuzz.zip -workdir=.
+# Seed the native Go fuzz corpus from existing .mcl files in the project.
+seed:
+	mkdir -p testdata/fuzz/FuzzLexParse/
+	find ../.. -name \*.mcl -exec sh -c 'cp "$$1" "testdata/fuzz/FuzzLexParse/$$(basename $$1)"' _ {} \;
```
purpleidea: What's with this underscore??

purpleidea (changes requested):
Please also fix up the git commit log test, thanks!

Oddly:
Hi Purpleidea,

Only my two first PR's were personally checked. The other PR's were created by my LLM accidentally! I thought I had stopped it, but it seems to have continued without me. I'll check these remaining PR's and adjust where needed. It's not my intention to sloppify this project, my apologies!

I'm in the #mgmtconfig channel if you want to discuss something outside of the issues.

lang/fuzz/Makefile
Oddly: _ is a placeholder for argument $0 coming from `sh -c '...'`, which normally is the script's name (unused now of course). Because find uses positional arguments, we need to mock $0 with "_". Then {} lands in the $1 location.

purpleidea:
I've fixed up this commit slightly and merged as well. Thank you

I'm still not entirely sure if you're a human or a claw bot, but let us know =D

PR #860, opened 2026-02-07
lang: strings: Add common string manipulation functions

Oddly:
The mcl strings module had to_lower, split, substr, pad, and join_nonempty, but was missing several functions you'd expect in any language's standard library. This adds six new string functions that all follow the same simple, pure, memoizable pattern:

- strings.to_upper(s) — converts a string to uppercase
- strings.trim_space(s) — strips leading and trailing whitespace
- strings.contains(s, substr) — checks if substr appears in s
- strings.has_prefix(s, prefix) — checks if s starts with prefix
- strings.has_suffix(s, suffix) — checks if s ends with suffix
- strings.replace_all(s, old, new) — replaces all occurrences of old with new

Each wraps the corresponding Go strings package function and includes tests. These are all pure, fast, and safe to memoize.

Ref #381

### DIFF
--- a/lang/core/strings/contains.go
+++ b/lang/core/strings/contains.go
@@ -1,0 +1,58 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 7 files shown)


purpleidea:
LGTM

Please tell the bot to use the:

`Co-authored-by:`

standard format, the "Co-Authored-By:" format is non-standard and amazingly, we have a CI test for this which fails. I'm happy to fix these up for now.

I accidentally pushed some of the earlier commits in without it passing this test so I've force pushed those away with fixed variants.

purpleidea:
Thank you for the patch!

PR #863, opened 2026-02-08
engine: Fix errors on re-reconcilation of existing users

danielpodwysocki:
This is another attempt at https://github.com/purpleidea/mgmt/issues/842 . It also adds a regression test for the problem.

@purpleidea 
I'm not 100% sure if that's what you had in mind, but I figured I'll get a PR out and we can adjust the details around on it here.
Let me know if the approach needs changing up.

Things that jumped out at me when looking at this:

1) This would change existing behavior and force groups to be defined explicitly (see changes to `examples/`. I'm not sure if I can do that, let me know if I should add a conditional to avoid passing `--no-user-group` for this case:

```
# will fail if we don't explicitly define the group too
user "mgmt-test-rereconcile" {
	state => "exists",
}
```

2) If we switch this to be inexplicit (creating the group via useradd by not passing `--no-user-group`), we can again hit the issue of orphan groups being left around when deleting the resource when setting its state to "absent".

3) `useradd` tries to match UID/GID to look the same for the user. This should be no problem as far as Linux cares. It might present odd to users though - eg. creating `daniel` with primary group `daniel` might yield 1005:1016 UID:GID when managing exclusively via mgmt resources.

### DIFF
--- a/engine/resources/user.go
+++ b/engine/resources/user.go
@@ -70,6 +70,8 @@ type UserRes struct {
 	GID *uint32 `lang:"gid" yaml:"gid"`
 
 	// Group is the name of the user's primary group.
+	// Defaults to the resource name if neither Group or GID are defined.
+	// The group is NOT created automatically - the resource assumes it exists.
 	Group *string `lang:"group" yaml:"group"`
 
 	// Groups are a list of supplemental groups.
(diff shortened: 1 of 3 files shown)


engine/resources/user.go
```diff
@@ -319,6 +318,17 @@ func (obj *UserRes) CheckApply(ctx context.Context, apply bool) (bool, error) {
 		if obj.Shell != nil {
 			args = append(args, "--shell", *obj.Shell)
 		}
+		if exists && len(args) == 0 {
+			return false, nil
+		}
```
danielpodwysocki: Oops, that's some leftover. Fixing that...

Issue #861, opened 2026-02-07
Autoedge Big-O performance improvements


purpleidea:
Our autoedge algorithm is likely currently O(n^2) ish or worse. We think we can improve it to O(n * log n) or even O(n). This will involve some clever algorithmist work. This will help if we've got very large resource graphs.

It's not necessarily a blocker atm, but enabling fast autoedges for very large resource graphs would have long-term benefits. Here are some suggested approaches one might take. Please feel free to suggest additional ideas as well.

Some initial research work was done by @frebib in https://github.com/frebib/mgmt/commit/1b4fb059cc663f79293763918b2a8e5f702fc4b7

* Improve the core autoedges algorithm.
I'm not a great algorithmist, and while I think it's correct, a smarter computer scientist might know of a better approach.

* Add caching to the graph lookup operations that the algorithm uses.
Repeated lookups that are similar could be cached.

* Add engineering approaches to minimize search size.
We might be able to extend the autoedges API to statically know which resource kind's can edge together or not. This would reduce the search size. Remember to include the match against both same kind of resource and differing kind of resources.

* Consider parser tricks to skip attempting to autoedge resources with existing dependencies.
In situations where we already know there is a dependency between two or more resources, adding a new edge might not be beneficial. In the transistive example where we have A -> B, and B -> C, it's no use adding A -> C for example. This would offer speed up be initially reducing the graph size substantially.

* Delta optimizations during graph transitions.
When swapping from one resource graph to a subsequent one, we essentially re-run the autoedge code. It might be helpful to look at the previous result to determine if we can reuse some of that work. An algorithmist might want to be present to ensure we don't do something wildly more inefficient here.

* Extra edge prevention.
We don't want to add more edges than absolutely necessary. Extra edges technically increase memory usage and reduce parallelism, so we should not add anything that isn't absolutely necessary.

* Tests.
We'd definitely love to have more tests here both for correctness and performance. In particular, when we're doing edges involving send/recv we want to make sure not to break the logical data flows.

Additional ideas and discussion welcome.

Oddly:
Put the MR's on draft now since I've not reviewed the output myself and this is mostly (from my side) up until now a test to see how LLM's do in a real environment, even with steering. Let me know if this is not wanted, or not the right project for this.

purpleidea:
[...]

PR #866, opened 2026-02-08
engine: autoedge: Optimize matching with pre-computed vertex info and kind index

Oddly:
The autoedge matching loop in `addEdgesByMatchingUIDS` had O(n²) behaviour because for each UID a resource wanted to match, it called `graph.Vertices()` (which iterates a map and allocates a slice) and then called `UIDs()` on every candidate vertex. With 1000 resources this meant roughly 1M comparisons per graph swap, each with fresh allocations.

This PR makes three changes to the algorithm, all in `engine/graph/autoedge/autoedge.go`:

Pre-compute a `vertexInfo` slice once after building the sorted resource list, caching each vertex's `EdgeableRes` reference and its `UIDs()` result. This eliminates repeated interface calls and the `graph.Vertices()` map iteration from the inner loop entirely.

Build an index from concrete UID type (via `reflect.TypeOf`) to the vertices that produce UIDs of that type. Every real `IFF()` implementation in the codebase starts with a type assertion that returns false across types (e.g. `FileUID.IFF(PkgUID)` is always false), so we can skip candidates whose UID types don't overlap with the seeking UID. For mixed-kind graphs this reduces the candidate set from all vertices to only same-kind vertices. The algorithm falls back to the full list if a UID type isn't in the index, so correctness is preserved even for resources using `BaseUID` directly without overriding.

Update the `addEdgesByMatchingUIDS` signature to accept the pre-computed data instead of the raw graph, and iterate the candidate slice instead of calling `graph.Vertices()`.

No interfaces, traits, or resource files were changed. The `UIDExistsInUIDs` and `IFF()` call paths are completely unchanged.

A new `autoedge_test.go` provides 22 functional tests covering empty graphs, single vertices, matching and non-matching UIDs, reversed and normal edge direction, multiple UIDs per resource, multiple Next()/Test() batches, nil and erroring AutoEdges, disabled resources (both seeker and target), self-match prevention, cross-kind isolation, large graphs (50-vertex chain), hierarchy patterns, first-match-wins semantics, Test() stopping iteration, Test() receiving correct boolean arrays, non-EdgeableRes vertices, BaseUID fallback path, and determinism across 10 runs. There are also 6 unit tests for `UIDExistsInUIDs` and 8 benchmarks at varying scales.

Benchstat output (6 iterations, `benchtime=500ms`, 12th Gen i7-12700H):

```
                          │ bench-before.txt │          bench-after.txt           │
                          │      sec/op      │    sec/op      vs base             │
AutoEdge10-20                    31.34µ ±  8%    32.14µ ± 11%       ~ (p=0.485)
AutoEdge100-20                   844.8µ ± 16%    666.5µ ± 21%  -21.11% (p=0.004)
AutoEdge1000-20                  44.33m ± 10%    10.51m ± 32%  -76.28% (p=0.002)
AutoEdge10000-20                4547.6m ± 13%    298.3m ± 14%  -93.44% (p=0.002)
AutoEdgeNoMatch-20               6.922m ±  7%    7.989m ± 16%  +15.41% (p=0.004)
AutoEdgeAllMatch-20              83.48m ± 18%    21.91m ± 18%  -73.75% (p=0.002)
AutoEdgeMultiUID-20             1230.1µ ± 29%    952.6µ ±  9%  -22.56% (p=0.009)
geomean                          72.71µ           43.74µ        -39.85%

                          │ bench-before.txt │          bench-after.txt           │
                          │      B/op        │     B/op      vs base              │
AutoEdge10-20                  8.373Ki ± 0%     8.311Ki ± 0%   -0.75% (p=0.002)
AutoEdge100-20                 330.2Ki ± 0%     134.3Ki ± 0%  -59.32% (p=0.002)
AutoEdge1000-20               18.307Mi ± 0%     1.685Mi ± 0%  -90.79% (p=0.002)
AutoEdge10000-20             3195.40Mi ± 0%     21.34Mi ± 1%  -99.33% (p=0.002)
AutoEdgeNoMatch-20             1.164Mi ± 0%     1.289Mi ± 0%  +10.75% (p=0.002)
AutoEdgeAllMatch-20           35.495Mi ± 0%     2.088Mi ± 0%  -94.12% (p=0.002)
AutoEdgeMultiUID-20            330.2Ki ± 0%     134.3Ki ± 0%  -59.33% (p=0.002)
geomean                                                        -66.20%
```

The no-match case regresses slightly (+15% time, +11% memory) because the index lookup adds a small constant cost when no candidates match. This is the least common real-world case since autoedge is primarily useful when resources do have matching UIDs. Everything else shows strong statistically significant improvement, especially at scale where the 10000-vertex case drops from 4.5s to 298ms.

Addresses parts 1, 2, and 7 of #861.

### DIFF
--- a/engine/graph/autoedge/autoedge.go
+++ b/engine/graph/autoedge/autoedge.go
@@ -31,11 +31,19 @@ package autoedge
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/purpleidea/mgmt/engine"
 	"github.com/purpleidea/mgmt/pgraph"
 	"github.com/purpleidea/mgmt/util/errwrap"
 )
+
+// vertexInfo holds pre-computed information about a vertex for use during edge
+// matching. This avoids recomputing UIDs() on every comparison.
+type vertexInfo struct {
+	res  engine.EdgeableRes
+	uids []engine.ResUID
+}
 
 // AutoEdge adds the automatic edges to the graph.
 func AutoEdge(graph *pgraph.Graph, debug bool, logf func(format string, v ...interface{})) error {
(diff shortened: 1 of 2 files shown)


Oddly:
They were actually run, not generated.

Command on the feature branch:
`git checkout autoedge-optimize-matching && go test -run='^$' -bench=. -benchmem -count=6 -benchtime=500ms ./engine/graph/autoedge/ | tee /tmp/bench-866.txt && git checkout autoedge-skip-redundant`

On master this was bit more complex because we had to exclude some benchmarks not present here and copy the `autoedge_test.go` files over to master temporary to actually be able to test (which worked because AutoEdge() and UIDExistsInUIDs() functions already exist in master).

After that a diff was generated with [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat).

PR #867, opened 2026-02-09
engine: autoedge: Skip redundant edges and add end-to-end tests

Oddly:
This builds on PR #866 (type-based index and UID caching) to address issue #861 items 4 and 6. Before adding an autoedge, we now check whether the dependency is already satisfied by existing edges in the graph.

The first check is an O(1) FindEdge lookup that catches direct duplicates where an explicit edge already exists between the same vertices. The second is an iterative DFS (isReachable) that checks whether the target is already reachable through existing edges, catching transitive redundancies like A->B->C making A->C unnecessary. We get the adjacency map reference once per AutoEdge run so edges added earlier in the same run are visible to later reachability checks.

The unit tests cover isReachable (direct, transitive, reverse, cycles, self-loops, disconnected, diamonds, long chains, empty adjacency) and the skip logic itself (existing edge skip, transitive skip, add-when-not-reachable, live adjacency, match reporting, notify preservation, diamond/chain topologies, reversed edges, parallel paths, determinism). Benchmarks show the overhead is negligible — the short-path FindEdge case is ~70ns with zero allocations.

The end-to-end shell tests run mgmt with real file resources and explicit edges to verify the skip logic in a live deployment. autoedge-skip-1 tests the FindEdge path (explicit parent->child edge duplicates autoedge discovery). autoedge-skip-2 tests the isReachable path (transitive chain d1->f1->f2 makes d1->f2 redundant). Both verify resources converge correctly and no redundant edges appear in the log.

Depends on #866.

### DIFF
--- a/engine/graph/autoedge/autoedge.go
+++ b/engine/graph/autoedge/autoedge.go
@@ -45,6 +45,29 @@ type vertexInfo struct {
 	uids []engine.ResUID
 }
 
+// edgeMatcher holds the pre-computed state needed for matching UIDs to
+// vertices. It is built once per AutoEdge run and used for all matching calls,
+// avoiding the need to pass many parameters to each call.
+type edgeMatcher struct {
+	// verticesInfo is the pre-computed list of non-disabled EdgeableRes
+	// vertices and their UIDs.
+	verticesInfo []vertexInfo
+
+	// uidTypeIndex maps concrete UID types to the vertices that produce
+	// UIDs of that type, for fast candidate narrowing.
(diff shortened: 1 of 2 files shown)


Oddly:
Human review still needs to happen here, this is generated and not ready for merging. I've put the MR's on draft.

Oddly:
You are right, I'll change my approach in the future and only open these PR's when I'm fully done. How do you want to continue? 
Shall I close these and first continue privately, after which we'll open these again?

Oddly:
Closing per maintainer feedback — will battle test on my fork and resubmit when properly vetted.

PR #885, opened 2026-03-05
Buildable on Darwin (again?)

h0tw1r3:
Looks like a one time mgmt was buildable on Darwin. My shop is all OSX, so I took a stab. For the most part things work, but I haven't had time to work out why some tests fail.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,8 +51,8 @@ require (
 	google.golang.org/grpc v1.70.0
 	gopkg.in/yaml.v2 v2.4.0
 	honnef.co/go/augeas v0.0.0-20161110001225-ca62e35ed6b8
-	libvirt.org/go/libvirt v1.11006.0
-	libvirt.org/go/libvirtxml v1.11006.0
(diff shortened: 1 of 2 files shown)


purpleidea:
>   Error: The 'mdl' utility can't be found.

Seems you broke the markdown test btw.

h0tw1r3:
perl dependency? confused, it works consistently on all OS's, doesn't have the drawbacks of timeout, it's included on every base *nix operating system since forever? I suppose it could be rewritten in bash, that would still be better than using gnu timeout.

sounds like you've gotten everything you want to use.

purpleidea:
> perl dependency

Trying to avoid anything I can. If there's no other way-- but I feel if we really need something a small golang stub would work too.

> sounds like you've gotten everything you want to use.

Nah, I just pulled in the stuff that was an obvious pass to make it easier for you to filter through the remaining stuff. I don't have any darwin customers personally, but I'm happy to merge patches if they're useful for you.

PR #886, opened 2026-03-10
allow static Go build

miekg:
This implements the few things we discussed in https://github.com/purpleidea/mgmt/discussions/877

I took the brute force approach.


miekg:
you could not merge, then I'll keep them in my fork - it's mostly so I (and maybe others) can have a static build.

I was under impressession that baddev will be thrown away at some point anyway?

purpleidea:
I don't think there's something I can merge here, so closing this for now.

PR #915, opened 2026-04-22
engine: deep-copy ReversibleMeta in ResCopy

karpfediem:
Fixes a reversible-resource copy bug caused by aliasing `*ReversibleMeta`.

`engine.ResCopy()` was reusing the same reversible metadata pointer for the original resource and its copied/reversed form. 
Reversal setup mutates that metadata (`Disabled`, `Reversal`), so those mutations could leak back into the original live resource.

## MRE
<details>
  <summary>First run</summary>
  
  ```mcl
file "/tmp/reversible-meta-copy-mre/" {
}

# Step 1 of the runtime proof for the reversible meta aliasing bug.
#
# Run this file first, let it converge, and then run
# `test/shell/reversible-meta-copy-mre-cleanup.mcl` with the same `--prefix`.
#
# Example:
#   ./mgmt run --prefix /tmp/reversible-meta-copy-mre-prefix \
#     --converged-timeout=2 --no-pgp lang \
#     test/shell/reversible-meta-copy-mre.mcl
#
#   ./mgmt run --prefix /tmp/reversible-meta-copy-mre-prefix \
#     --converged-timeout=2 --no-pgp lang \
#     test/shell/reversible-meta-copy-mre-cleanup.mcl
#
# Expected on a correct engine:
# 1. This run creates `/tmp/reversible-meta-copy-mre/target`.
# 2. The cleanup run logs `adding 1 reversals...` and removes the target.
#
# Buggy engine behavior:
# 1. This run converges and then deletes its own pending reverse state during
#    shutdown cleanup.
# 2. The cleanup run has no stored reversal to add, so the target is left
#    behind.
file "/tmp/reversible-meta-copy-mre/target" {
        content => "proof\n",

        Meta:reverse => true,
}
  ```
</details>

<details>
  <summary>Second run</summary>
  
  ```mcl
file "/tmp/reversible-meta-copy-mre/" {
}
  ```
</details>

## Impact

This could cause the original resource to be treated as if it were itself a reversal resource during cleanup, which in turn could delete the pending reverse state that should have been preserved for a later run.

In practice, that means a normal converged reversible resource can disappear from the graph, and the next run has
nothing left to reverse.

## Fix

Add `(*ReversibleMeta).Copy()` and use it in `engine.ResCopy()` so reversible metadata is copied by value instead of
shared by pointer.

## Verification

Added focused Go tests covering:

- copy isolation for `ReversibleMeta`
- `Reversed()` not mutating the original resource
- `ReversalInit()` not mutating the original resource
- `ReversalCleanup()` not deleting the original resource's pending reverse file

The MCL repro is a second commit, so feel free to cherry-pick.

### DIFF
--- a/engine/copy.go
+++ b/engine/copy.go
@@ -126,7 +126,7 @@ func ResCopy(r CopyableRes) (CopyableRes, error) {
 			// programming error
 			panic("reversible interfaces are illogical")
 		}
-		dst.SetReversibleMeta(x.ReversibleMeta()) // no need to copy atm
+		dst.SetReversibleMeta(x.ReversibleMeta().Copy())
 	}
 
 	return res, nil
--- a/engine/copy_test.go
+++ b/engine/copy_test.go
@@ -1,0 +1,141 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+// Additional permission under GNU GPL version 3 section 7
+//
+// If you modify this program, or any covered work, by linking or combining it
+// with embedded mcl code and modules (and that the embedded mcl code and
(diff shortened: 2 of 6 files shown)


engine/copy.go
```diff
@@ -126,7 +126,7 @@ func ResCopy(r CopyableRes) (CopyableRes, error) {
 			// programming error
 			panic("reversible interfaces are illogical")
 		}
-		dst.SetReversibleMeta(x.ReversibleMeta()) // no need to copy atm
```
purpleidea: This comment is so suspicious! haha

I will dig a bit further into this patch this week, thank you!

purpleidea:
Sorry for my delay here...

> This run converges and then deletes its own pending reverse state during shutdown cleanup.

It should be deleting the created file on clean shutdown with ^C. Is that not occurring or it's not what you expected?

purpleidea:
As previously discussed, I'm not sure this is the correct fix, but I do think it's highly likely there could be a bug lurking in the reversible code somewhere. I'm going to close this for now, but I do have some suggestions I could share if you want to dig into it more.

PR #917, opened 2026-04-25
util: Bound watch path index

karpfediem:
`RecWatcher.Watch()` tracks the current watched path depth with `index` and then slices:

```go
current = strings.Join(patharray[0:index], "/")
```

When watching a missing leaf path, recwatch backs up to an existing parent.

As parent directories and files are created, `fsnotify` can later deliver a parent-path event after recwatch has already
descended **to the leaf**.

The code currently treats that as another descend signal and increments index past len(patharray), causing:

```
panic: runtime error: slice bounds out of range
github.com/purpleidea/mgmt/util/recwatch.(*RecWatcher).Watch
      util/recwatch/recwatch.go:183
```

which is an event-ordering/state-machine race.

### MRE

Run the seed once with a persistent --prefix.
Then rerun the write graph repeatedly with the same prefix while deleting the output directory between runs.

<details>
  <summary>Seed</summary>
  
  ```mcl
kv "recwatch-index-bounds-01" {
    key => "recwatch/index-bounds/01",
    value => "01\n",
}

kv "recwatch-index-bounds-02" {
    key => "recwatch/index-bounds/02",
    value => "02\n",
}
  ```
</details>

<details>
  <summary>Write graph</summary>
  
  ```mcl
import "world"

$root = "/tmp/mgmt-recwatch-index-bounds-showcase/"
$data = "${root}data/"

$v01 = world.getval("recwatch/index-bounds/01")
$v02 = world.getval("recwatch/index-bounds/02")

file $root {
    mode => "0700",
}

file $data {
    mode => "0700",
    Depend => File[$root],
}

file "${data}01.txt" {
    content => $v01->value,
    mode => "0600",
    Depend => File[$data],
}

file "${data}02.txt" {
    content => $v02->value,
    mode => "0600",
    Depend => File[$data],
}
  ```
</details>

<details>
  <summary>Scripted setup</summary>
  
  ```sh
#!/usr/bin/env bash
set -euo pipefail

mgmt="${1:-./mgmt}"
seed="${2:-./recwatch-index-bounds-seed.mcl}"
write="${3:-./recwatch-index-bounds-write.mcl}"
base="${4:-22401}"

client_url="http://127.0.0.1:${base}"
server_url="http://127.0.0.1:$((base + 1))"
prefix="/tmp/mgmt-recwatch-index-bounds-prefix-${base}"
root="/tmp/mgmt-recwatch-index-bounds-showcase"

rm -rf "$prefix" "$root"

"$mgmt" run \
      --prefix="$prefix" \
      --hostname="recwatch-seed-${base}" \
      --client-urls="$client_url" \
      --server-urls="$server_url" \
      --advertise-client-urls="$client_url" \
      --advertise-server-urls="$server_url" \
      --no-pgp \
      --converged-timeout=5 \
      --max-runtime=60 \
      lang "$seed" >/tmp/recwatch-index-bounds-seed-${base}.log 2>&1

for i in $(seq 1 20); do
      rm -rf "$root"
      log="/tmp/recwatch-index-bounds-write-${base}-${i}.log"

      set +e
      "$mgmt" run \
              --prefix="$prefix" \
              --hostname="recwatch-write-${base}-${i}" \
              --client-urls="$client_url" \
              --server-urls="$server_url" \
              --advertise-client-urls="$client_url" \
              --advertise-server-urls="$server_url" \
              --no-pgp \
              --converged-timeout=2 \
              --max-runtime=30 \
              lang "$write" >"$log" 2>&1
      code=$?
      set -e

      if grep -q "panic: runtime error: slice bounds out of range" "$log"; then
              echo "panic iteration $i"
              tail -n 40 "$log"
              exit 10
      fi

      if [ "$code" -ne 0 ]; then
              echo "nonzero iteration $i code=$code"
              tail -n 40 "$log"
              exit "$code"
      fi

      echo "iteration $i ok"
done

echo "no panic after 20 iterations"
  ```
</details>

### DIFF
--- a/util/recwatch/recwatch.go
+++ b/util/recwatch/recwatch.go
@@ -32,7 +32,6 @@ package recwatch
 
 import (
 	"fmt"
-	"math"
 	"os"
 	"path"
 	"path/filepath"
@@ -180,6 +179,9 @@ func (obj *RecWatcher) Watch() error {
 	var send = false                          // send event?
 
 	for {
+		if index < 1 || index > len(patharray) {
+			return fmt.Errorf("recwatch: invalid path index %d for %q", index, obj.safename)
+		}
 		current = strings.Join(patharray[0:index], "/")
 		if current == "" { // the empty string top is the root dir ("/")
 			current = "/"
@@ -194,8 +196,9 @@ func (obj *RecWatcher) Watch() error {
 			}
 			// ENOENT for linux, etc and IsNotExist for macOS
 			if err == syscall.ENOENT || os.IsNotExist(err) {
-				index-- // usually not found, move up one dir
-				index = int(math.Max(1, float64(index)))
+				if index > 1 {
+					index-- // usually not found, move up one dir
+				}
 				continue
 			}
 
@@ -265,9 +268,11 @@ func (obj *RecWatcher) Watch() error {
 
 				// file removed, move the watch upwards
 				if deltaDepth >= 0 && (event.Op&fsnotify.Remove == fsnotify.Remove) {
-					//obj.options.logf("removal!")
-					obj.watcher.Remove(current)
-					index--
+					if index > 1 {
+						//obj.options.logf("removal!")
+						obj.watcher.Remove(current)
+						index--
+					}
 				}
 
 				// when the file is moved, remove the watcher and add a new one,
@@ -279,9 +284,11 @@ func (obj *RecWatcher) Watch() error {
 
 				// we must be a parent watcher, so descend in
 				if deltaDepth < 0 {
-					// XXX: we can block here due to: https://github.com/fsnotify/fsnotify/issues/123
-					obj.watcher.Remove(current)
-					index++
+					if index < len(patharray) {
+						// XXX: we can block here due to: https://github.com/fsnotify/fsnotify/issues/123
+						obj.watcher.Remove(current)
+						index++
+					}
 				}
 
 				// if safename starts with event.Name, we're above, and no event should be sent
@@ -289,9 +296,11 @@ func (obj *RecWatcher) Watch() error {
 				//obj.options.logf("above!")
 
 				if deltaDepth >= 0 && (event.Op&fsnotify.Remove == fsnotify.Remove) {
-					//obj.options.logf("removal!")
-					obj.watcher.Remove(current)
-					index--
+					if index > 1 {
+						//obj.options.logf("removal!")
+						obj.watcher.Remove(current)
+						index--
+					}
 				}
 
 				if deltaDepth < 0 {
@@ -299,8 +308,11 @@ func (obj *RecWatcher) Watch() error {
 					if util.PathPrefixDelta(obj.safename, event.Name) == 1 { // we're the parent dir
 						send = true
 					}
-					obj.watcher.Remove(current)
-					index++
+					if index < len(patharray) {
+						// XXX: we can block here due to: https://github.com/fsnotify/fsnotify/issues/123
+						obj.watcher.Remove(current)
+						index++
+					}
 				}
 
 				// if event.Name startswith safename, send event, we're already deeper
--- a/util/recwatch/recwatch_test.go
+++ b/util/recwatch/recwatch_test.go
@@ -1,0 +1,122 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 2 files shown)


util/recwatch/recwatch.go
```diff
@@ -180,6 +179,9 @@ func (obj *RecWatcher) Watch() error {
 	var send = false                          // send event?
 
 	for {
+		if index < 1 || index > len(patharray) {
+			return fmt.Errorf("recwatch: invalid path index %d for %q", index, obj.safename)
```
purpleidea, line 183: Is it possible for this to happen?

util/recwatch/recwatch.go
```diff
@@ -194,8 +196,9 @@ func (obj *RecWatcher) Watch() error {
 			}
 			// ENOENT for linux, etc and IsNotExist for macOS
 			if err == syscall.ENOENT || os.IsNotExist(err) {
-				index-- // usually not found, move up one dir
-				index = int(math.Max(1, float64(index)))
+				if index > 1 {
```
purpleidea, line 199: Why do we always check against `1` and not `0` ?

purpleidea (commented):
Patch LGTM, thanks!

If this was written by some AI thing, please remember to disclose that with a 'Co-authored-by:` tag. Please grep the `git log` to see the preferred format for this. Thanks!

purpleidea:
PS: Any of the additional info in the PR body that we want to keep please put in the commit message, and consider formatting it correctly too please, it's all as one long line.

util/recwatch/recwatch.go
karpfediem: I don't know, but next time we will see a clear error instead of a panic :)

util/recwatch/recwatch.go
karpfediem: these gated blocks would remove 1 or add 1 beyond the limits
so we only guard against overstepping the bounds

karpfediem:
> PS: Any of the additional info in the PR body that we want to keep please put in the commit message, and consider formatting it correctly too please, it's all as one long line.

~~**All** one long line?~~ I see now, you meant not to do that lol

Should the MCL examples / repro script go there too or just textual description?

But sure, can do that

purpleidea:
> Should the MCL examples / repro script go there too or just textual description?

If it repros a bug, it's valuable, so drop it in as a test. If it's just interesting, drop it in examples/lang/ if it's not something we're going to look at past today then just skip adding it anywhere and commit is fine however.

Just wanted to point out that assume anything in github is ephemeral, and anything in git itself stays around.

util/recwatch/recwatch.go
purpleidea: optional but for such cases I would leave
// programming error

or

// programming error?

as a comment inside the `if` above the return. So if I see it I'll think... ahh something is wrong.

purpleidea: (the `with "?" version` means idk, the first version means "it's a problem, but i want to shut down safely"

purpleidea:
Last question: did you actually hit this issue or is it just a theoretical consideration that an AI (or you) noticed?

karpfediem:
> Last question: did you actually hit this issue or is it just a theoretical consideration that an AI (or you) noticed?

yes, hit it pretty consistently 
~50% of deploys while testing my ACME resource stuff, where i was deleting files to test if we re-issue certs etc

This resolves that and changes no behavior.

karpfediem:
> , so drop it in as a test

I thought about it, but the golang test is more deterministic, MCL test is just a race that may or may not trigger.

But if i were to add MCL based tests, would i move them under test/shell?

purpleidea:
> MCL test is just a race that may or may not trigger.

So it's okay to have a test that doesn't *guarantee* a race hit, even if it occasionally does that's still an ok test.

What's not acceptable is a flaky test of course which sometimes fails if we depend on a race ;)

> But if i were to add MCL based tests, would i move them under test/shell?

There is test/shell/ for the fancy things like your test is doing, but historically those have all been flaky, maybe because mgmt was buggy but also because some timing in the execution test environments could cause long delays which broke things.

There is also the txtar tests... `find txtar` and you'll see the different types. If there is a new type we need, we can add some too =D

karpfediem:
> There is also the txtar tests

This repo needs two deploys unfortunately, so I don't believe it would fit as txtar right now.
EDIT: and external interaction...

It is only still flaky in the false negative sense (e.g. no race occurred -> green).
Although i tuned the script to have had it consistently trigger the panic in 10/10 runs.
Just no guarantee :D

And i still believe in this case the golang test is sufficient on its own, while also being better.

purpleidea:
LGTM, thanks!

purpleidea:
Rebased and merged, thank you!

purpleidea:
It's extremely valuable work anytime any bug is fixed, I've been almost 100% working on that kind of polish a lot lately, so big :fireworks: :fireworks: :fireworks:

karpfediem:
cheers :D and ty too

PR #918, opened 2026-04-25
engine: Optimize ResGraphMapper from O(n²) to O(n) using pre-indexing

lbcristaldo:
Replace linear scan of oldGraph (via VertexMatchFn) with hash map index
keyed by (Kind, Name). Complexity changes from O(V_new × V_old) to
O(V_new + V_old).

### Performance

Benchmarked on Intel i7-8550U (Go 1.25):

| Size | Before (ns/op) | After (ns/op) | Allocs/op |
|------|---------------|---------------|-----------|
| 10   | ~10,703       | 9,847         | 22 → 7    |
| 100  | ~79,395       | 76,410        | 157 → 7   |
| 500  | ~453,039      | 393,184       | 757 → 7   |

Allocations become constant (7 allocs) regardless of graph size.

### Correctness

- Uses `[2]string{Kind, Name}` as map key → no separator collision possible
- Semantics unchanged: two resources match iff both Kind and Name are equal
- Added unit tests: identity, absence, kind mismatch, empty graphs, and pipe-character collision safety (`TestResGraphMapper6`)

### Testing

- `go test -race ./engine` 
- `go test -race ./engine -run "Cmp|ResGraphMapper"` 
-  Benchmarks show consistent improvement

Fixes the FIXME at `engine/cmp.go:403`.

### DIFF
--- a/engine/cmp.go
+++ b/engine/cmp.go
@@ -33,7 +33,6 @@ import (
 	"fmt"
 
 	"github.com/purpleidea/mgmt/pgraph"
-	"github.com/purpleidea/mgmt/util/errwrap"
 )
 
 // ResCmp compares two resources by checking multiple aspects. This is the main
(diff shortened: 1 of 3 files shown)


purpleidea:
Amazing, thanks!

Need you to disclose what if any AI tools you used for this patch please.

lbcristaldo:
> Amazing, thanks!
> 
> Need you to disclose what if any AI tools you used for this patch please.

hi, thank you for your kind words.
I used deepseek throughout the process to understand different tradeoffs of possible solutions (cache vs pre-index), and also when I needed to rewrite tests based on the format I've found in other test files.
On the other hand, I used go docs in notebooklm to understand better the codebase and how to model the patch.
that's how I use AI :) i take full ownership of the changes proposed and, If admitted, I'll take responsibility of its maintenance when necessary

purpleidea:
> I used deepseek throughout the process

No problem, will need you to add a Co-authored-by: tag in the git commit that mentions the tool and model specifics that you used. If you look at the recent git log you'll see how it's used for other tools.

As for review, you may wish to use the existing TestRes instead of creating a new one and add a B.ReportAllocs to the test.

Thanks!

purpleidea:
> use the existing TestRes instead of creating a new one

^^ and specify the precise model of deepseek and the frontend used please.

lbcristaldo:
> > use the existing TestRes instead of creating a new one
> 
> ^^ and specify the precise model of deepseek and the frontend used please.

hi, sorry, I tried importing engine/resources but it creates a circular dependency: imports github.com/purpleidea/mgmt/engine/resources from cmp_test.go
	imports github.com/purpleidea/mgmt/engine from augeas.go: import cycle not allowed in test
FAIL	github.com/purpleidea/mgmt/engine

purpleidea:
You need to use engine.NewNamedResource to get a resource.

purpleidea:
Since I've not heard back, I've taken over the patch because it's an important perf optimization to get merged. Your timing is fortuitous because as you can see I've been doing a lot of perf work lately, killing off all the legacy slow implementations in mgmt, some of which (as you saw with the comment) were on my backlog for a while now.

TBQH it's hard to tell if this kind of patch is to get free "training" for some model by getting a review, so I won't explain the bug that I found. If you are human, lmk so I can add you to the vouch list.

lbcristaldo:
> Since I've not heard back, I've taken over the patch because it's an important perf optimization to get merged. Your timing is fortuitous because as you can see I've been doing a lot of perf work lately, killing off all the legacy slow implementations in mgmt, some of which (as you saw with the comment) were on my backlog for a while now.
> 
> TBQH it's hard to tell if this kind of patch is to get free "training" for some model by getting a review, so I won't explain the bug that I found. If you are human, lmk so I can add you to the vouch list.

Sorry for taking so long to respond, I'm human. I tried to use the function but it didn't solve the cycle, and I didn't know how to solve it. Thank you.

purpleidea:
No worries! My advice: practice getting good at golang before you ever touch LLM's. They might help solve a problem but it means you won't learn. You need to be good if you want to be able to read code and know what's good and bad from an LLM.

PR #919, opened 2026-05-11
lang: Use trailing commas on multiline declarations only

purpleidea:
This super patch ports away from the mandatory "trailing comma" that was
found in lists and other locations. These existed not because it was a
better design, but because I was new at writing parsers and didn't have
enough experience to do it better from the start! The simplicity also
made it easier to debug the parser when more significant changes were
occurring.

List, maps, structs, and even functions now *require* that you omit a
trailing comma when it's expressed as a single line, and include a
trailing comma when it's expressed as a multi line.

There are some edge cases which don't parse at the moment when you try
to write some badly formatted code. These aren't intended to be
annoying, but if we find common cases that folks get stuck on, please
let us know with an mcl test case, and we will relax the parser if
needed.

This updates all the examples and tests too which makes for an even
larger diff!

This also makes some small mostly unrelated fixes that were found while
going through this code.

This was not written by AI, but some of the research used to figure out
how to deal with newlines (the opt_newlines pattern) was suggested by
those tools. Maybe we'll have a better pattern going forward, but would
need a more expert parser writers help for that.


purpleidea:
Sent as a PR for open review. Comments welcome!

This fixes a long-standing issue. One downside is that the parser is a bit picky about these lists and so on now, so if you find this is not ergonomic, please let me know what the way to improve this is.

Lastly, I haven't done the `mclfmt` stuff that comes later, but I have been playing with it. If this is a good bridge to that at all, or if it needs restructuring, lmk.

purpleidea:
Please note: This is now merged, but *do* please comment if you have something valuable to add. We can still make future changes of course!

karpfediem:
I think this is mostly fine, but it introduces a new whitespace requirement
> Make sure that the final character is a newline, or the parser may error.

Every file must now include a newline character after the final `stmt` so the statement can be recognized.
It may be worth documenting this more clearly (e.g. in the language guide).

It also makes ad-hoc things like this impossible:
```bash
❯ build/mgmt-linux-amd64 run lang 'print "hello" { msg => "world", }' --no-network --converged-timeout 1
This is: mgmt, version: 1.0.2-168-g423add81
Copyright (C) James Shubin and the project contributors
Written by James Shubin <james@shubin.ca> and the project contributors
21:22:22 main: start: 1779045742833061284
21:22:22 cli: lang: lexing/parsing...
21:22:22 main: goodbye!
cli parse error: could not generate AST: parser: `syntax error: unexpected $end, expecting NEWLINE` @1:1
```

It now needs to have a newline
```bash
❯ build/mgmt-linux-amd64 run lang 'print "hello" { msg => "world", }
' --no-network --converged-timeout 1
```

Then there is another less relevant "issue" where we still allow commas before elements, e.g.:
```mcl
$mylist = [,"w", "o", "r", "l", "d"]
```

This parses successfully:
```bash
❯ build/mgmt-linux-amd64 run lang 'import "golang/strings"
$mylist = [,"w", "o", "r", "l", "d"]
print "hello" { msg => strings.join($mylist,""), }
' --no-network --converged-timeout 1
```

purpleidea:
> run lang 'print "hello" { msg => "world" }'

Good point, I didn't consider that. Would you like to file an issue for this please? We can definitely fix that.

> $mylist = [,"w", "o", "r", "l", "d"]

Oh haha! This probably deserves another issue and bug fix! I didn't test this case :P

karpfediem:
Created https://github.com/purpleidea/mgmt/issues/924

The other issue is much more aesthetics-only, i think :D

purpleidea:
It's a good catch, I opened https://github.com/purpleidea/mgmt/issues/925 thanks!

Issue #769, opened 2024-08-28
Pkg resource should allow installing by path


purpleidea:
Package managers like dnf allow you to run `dnf install /usr/bin/wget` and they work fine. We should allow the same sort of thing for the name of the pkg resource.

If there's more than one possibility we can either error or pick the first choice. (Probably we want to error, but a future param could allow that behaviour instead.)

Patches welcome!

purpleidea:
NOTE: This issue is _not_ about installing a package from a foo.rpm file on disk. We're open to adding that feature if someone can come up with a sane API for it, but that should be discussed in the "discussion" group first.

PR #928, opened 2026-05-21
feat: support installing packages from file paths

erickgreco:
## Description

This PR implements support for installing packages from local file paths, addressing issue #769.

### Changes

- Added `InstallFiles()` method to `packagekit.Conn` that wraps the PackageKit DBus `InstallFiles` method
- Modified `pkg` resource `CheckApply()` to detect file paths (package names starting with `/`)
- When a path is detected, uses `InstallFiles()` directly instead of package repository lookup

### Usage Example

```mcl
pkg "/path/to/local/package.rpm"
```

This allows installing local `.rpm`, `.deb`, or other package files alongside standard repository packages.

### Implementation Notes

- Follows the same transaction flags and error handling as `InstallPackages()`
- Respects `allowuntrusted` flag for GPG signature validation
- Code formatted with `gofmt`

### DIFF
--- a/engine/resources/packagekit/packagekit.go
+++ b/engine/resources/packagekit/packagekit.go
@@ -1114,3 +1114,56 @@ func IsMyArch(arch string) (bool, error) {
 	}
 	return goarch == runtime.GOARCH, nil
 }
+
+// InstallFiles installs a list of packages from local file paths.
+func (obj *Conn) InstallFiles(fullPaths []string, transactionFlags uint64) error {
+	if err := obj.RefreshCache(false); err != nil {
+		return errwrap.Wrapf(err, "can't refresh cache")
+	}
+	ch := make(chan *dbus.Signal, PkBufferSize)
+	interfacePath, err := obj.CreateTransaction()
+	if err != nil {
+		return err
+	}
+	var signals = []string{"Package", "ErrorCode", "Finished", "Destroy"} // "ItemProgress", "Status" ?
+	removeSignals, err := obj.matchSignal(ch, interfacePath, PkIfaceTransaction, signals)
+	if err != nil {
+		return err
+	}
+	defer removeSignals()
+	bus := obj.GetBus().Object(PkIface, interfacePath)
+	call := bus.Call(FmtTransactionMethod("InstallFiles"), 0, transactionFlags, fullPaths)
+	if call.Err != nil {
+		return call.Err
+	}
+	timeout := -1
+	finished := false
+loop:
+	for {
+		select {
+		case signal := <-ch:
+			if signal.Path != interfacePath {
+				obj.Logf("Woops: Signal.Path: %+v", signal.Path)
+				continue loop
+			}
+			if signal.Name == FmtTransactionMethod("ErrorCode") {
+				return newPkError(signal.Body)
+			} else if signal.Name == FmtTransactionMethod("Package") {
+				timeout = PkSignalPackageTimeout
+			} else if signal.Name == FmtTransactionMethod("Finished") {
+				finished = true
+				timeout = PkSignalDestroyTimeout
+			} else if signal.Name == FmtTransactionMethod("Destroy") {
+				return nil
+			} else {
+				return fmt.Errorf("error in body: %v", signal.Body)
+			}
+		case <-util.TimeAfterOrBlock(timeout):
+			if finished {
+				obj.Logf("Timeout: InstallFiles: Waiting for 'Destroy'")
+				return nil
+			}
+			return fmt.Errorf("timeout installing files: %s", strings.Join(fullPaths, ", "))
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


purpleidea:
Neat, thank you!

In the issue I actually really meant path as in what `dnf provides` would find you, IOW which package would you need to get a binary at `/usr/bin/sl` for example. This is the equivalent of running `dnf install /usr/bin/sl` for example. However this is indeed a valid use case as well! We need some way to differentiate between the two though, what do you think is the best "API"/method in the resource to do this? Or should it be a different resource altogether?

If you could tell me more about your use case for this and if you used any AI tools as well, I'd appreciate it, thanks! lastly, you may need to include context.Context in your path. At least either use it or comment why you should or shouldn't.

Cheers!

erickgreco:
Hi @purpleidea, thanks for the feedback!

My use case is installing local .rpm/.deb packages from specific file paths, 
which is useful for custom or pre-built packages not in repositories.

I used Claude to help me understand the PackageKit DBus API 
and the structure of the existing InstallPackages method, which I then adapted 
for InstallFiles. The implementation itself and integration logic were my own work.

Regarding the differentiation: would it be better to:
1. Check if path exists locally (startswith("/") + file exists) for local installs
2. Use a separate "provider" or "provides" mode for package lookups

Also, I'll add context.Context handling. Could you clarify the use case - 
should the path parameter support both scenarios?

Thanks!

purpleidea:
> I used Claude to help me understand the PackageKit DBus API

You need to be very clear about disclosing the use of AI tools and the specific model to be able to participate in this project. We will make this even clearer in the docs.

Please accept my apologies, but we're trying to make sure we're not just doing free review work to train models. As a result, please give a bit of context on what you're using mgmt for and maybe a few snippets or modules you've written so we can confirm. If the code is sensitive you can share it with me privately.

Thank you!

erickgreco:
Thanks for clarifying.

I used Claude Sonnet 4 as a learning aid. When I encountered the PackageKit DBus API, 
I asked Claude to explain the existing InstallPackages method structure so I could 
understand the pattern. I then independently implemented InstallFiles by studying the 
codebase and adapting the logic myself.

To be transparent: I'm early in my career as a Go backend developer and I'm looking 
for opportunities to contribute to established projects to demonstrate my capabilities 
and learn from experienced maintainers.

My current project is:
**Robot control API** — REST API with Go, PostgreSQL, Docker, WebSocket-based ESP32 robot management

Here's a snippet from Dawg Patrol showing my Go work:
func (ws *WSHandler) RobotTelemetryWS(w http.ResponseWriter, r *http.Request) {
	userID, err := apimiddleware.GetUserIDFromClaimsCtx(r)
	if err != nil {
		myerrors.BadRequestResponse(w, r, err)
		return
	}

	robotID, err := apimiddleware.GetRobotIDFromCtx(r)
	if err != nil {
		myerrors.BadRequestResponse(w, r, err)
		return
	}

	reservationID, err := apimiddleware.GetReservationIDFromCtx(r)
	if err != nil {
		myerrors.BadRequestResponse(w, r, err)
		return
	}

	ctx := r.Context()

	if err := ws.service.SessionServiceValidator(ctx, reservationID, userID, robotID); err != nil {
		myerrors.BadRequestResponse(w, r, err)
		return
	}

	if err := ws.service.MarkWSStarted(ctx, reservationID); err != nil {
		myerrors.BadRequestResponse(w, r, err)
		return
	}

	conn, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		myerrors.BadRequestResponse(w, r, err)
		return
	}

	extendCtx, cancelExtend := context.WithCancel(ctx)
	defer cancelExtend()

	defer func() {
		releaseCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
		defer cancel()
		ws.service.ReleaseReservation(releaseCtx, reservationID)
	}()

	go ws.service.KeepReservationAlive(extendCtx, reservationID)
	go ws.service.MockTelemetry(extendCtx, ws.hub, reservationID)

	client := &RobotClient{
		hub:           ws.hub,
		conn:          conn,
		sendChan:      make(chan []byte, 256),
		robotID:       robotID,
		reservationID: reservationID,
	}

	ws.hub.register <- client
	go client.writePump()
	client.readPump()
}

I understand if this doesn't meet your contribution criteria. 
I appreciate the clarity on AI disclosure!

purpleidea:
> I'm early in my career as a Go backend developer and I'm looking for opportunities to contribute to established projects to demonstrate my capabilities and learn from experienced maintainers.

If this is your goal, then we definitely have a place for that if you want to put in the time. But our (my) advice is that you first join our channel (ping me there) show that you're going to invest the time to get better, and in particular if you want to learn, I genuinely believe you have to mostly avoid the use of the AI tools since you won't learn to learn that way.

LMK

purpleidea:
I haven't heard anything lately, so I'm going to close for now. If you come up with a good API for this, feel free to let us know. Thanks!

PR #932, opened 2026-05-30
cli: Add 'check lang' to validate mcl

jordansissel:
Currently, this has the same effect as `run lang --only-unify`

This is intended to make it safer for humans to use to check for syntax and unification errors because accidentally omitting the "--only-unify" flag on `run lang` could have negative consequences for the operator.

* [x] ran gofmt
* [x]  passes commit message test

Example usage:

```
mgmt check lang example.mcl
```

(I am OK if this command is removed later once a mclfmt tool becomes available)

### DIFF
--- a/cli/check.go
+++ b/cli/check.go
@@ -1,0 +1,65 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


cli/check.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/purpleidea/mgmt/lib"
+)
+
+type CheckArgs struct {
+	lib.Config // embedded config (can't be a pointer) https://github.com/alexflint/go-arg/issues/240
+
+	CheckLang *cliUtil.LangArgs `arg:"subcommand:lang" help:"check lang (mcl) payload"`
+}
+
+func (obj *CheckArgs) Run(ctx context.Context, data *cliUtil.Data) (bool, error) {
```
purpleidea, line 53: This and the struct need a short docstring please. (See Run for similar templates.)

cli/cli.go
```diff
@@ -117,6 +117,8 @@ type Args struct {
 
 	RunCmd *RunArgs `arg:"subcommand:run" help:"run code on this machine"`
 
+	CheckCmd *CheckArgs `arg:"subcommand:check" help:"check code on this machine"`
```
purpleidea, line 120: Move these above RunCmd please.

purpleidea (changes requested):
Will fail CI due to commit message, but otherwise LGTM just needs some small nits fixed, thanks!

jordansissel:
Addressed comments and retested -- commit message, gofmt, and golint tests pass.

purpleidea:
I fixed the small issues and merged! Thank you!

PR #935, opened 2026-06-02
lang: funcs: Add os.release

jordansissel:
Provides a struct representing parts of /etc/os-release

Example:

    $relinfo = os.release()
    print "release info" {
	    msg => fmt.printf("Distro %s %s (%s)", $relinfo->id, $relinfo->version_id, $relinfo->variant_id),
    }

Output:

    19:57:49 engine: print[release info]: Msg: Distro fedora 44 (cosmic)

### DIFF
--- a/lang/core/os/distro.go
+++ b/lang/core/os/distro.go
@@ -36,9 +36,14 @@ import (
 
 	"github.com/purpleidea/mgmt/lang/funcs/simple"
 	"github.com/purpleidea/mgmt/lang/types"
+	distroUtil "github.com/purpleidea/mgmt/util/distro"
+	"github.com/purpleidea/mgmt/util/errwrap"
 )
 
-const structDistroUID = "struct{distro str; version str; arch str}"
+const (
+	structDistroUID   = "struct{distro str; version str; arch str}"
+	releaseReturnType = "struct{id str; variant_id str; version_id str}"
+)
 
 var typeParseDistroUID = types.NewType(fmt.Sprintf("func(str) %s", structDistroUID))
 
@@ -53,6 +58,19 @@ func init() {
 		T: typeParseDistroUID,
 		F: ParseDistroUID,
 	})
+
+	simple.ModuleRegister(ModuleName, "release", &simple.Scaffold{
+		I: &simple.Info{
+			// In theory these are dependent on runtime.
+			Pure: false,
+			Memo: false,
+			Fast: true,
+			Spec: false,
+		},
+		T: types.NewType(fmt.Sprintf("func() %s", releaseReturnType)),
+		F: Release,
+	})
+
 }
 
 // ParseDistroUID parses a distro UID into its component values. If it cannot
@@ -100,4 +118,27 @@ func parseDistroUID(uid string) (string, string, string, error) {
 	// TODO: check for valid distro or arch?
 
 	return distro, version, arch, nil
+}
+
+// Release parses /etc/os-release and returns a struct containing the id,
+// version_id, and variant_id.
+func Release(ctx context.Context, input []types.Value) (types.Value, error) {
+	release, err := distroUtil.ParseOSRelease(ctx)
+	if err != nil {
+		return nil, errwrap.Wrapf(err, "ParseOSRelease failed")
+	}
+
+	st := types.NewStruct(types.NewType(releaseReturnType))
+
+	if err := st.Set("id", &types.StrValue{V: release.ID}); err != nil {
+		return nil, errwrap.Wrapf(err, "invalid id value")
+	}
+	if err := st.Set("variant_id", &types.StrValue{V: release.VARIANT_ID}); err != nil {
+		return nil, errwrap.Wrapf(err, "invalid variant_id value")
+	}
+	if err := st.Set("version_id", &types.StrValue{V: release.VERSION_ID}); err != nil {
+		return nil, errwrap.Wrapf(err, "invalid version_id value")
+	}
+
+	return st, nil
 }
--- a/util/distro/distro.go
+++ b/util/distro/distro.go
@@ -161,7 +161,7 @@ func IsFamilyArchLinux(ctx context.Context) (bool, error) {
 
 // Distro returns the distro name.
 func Distro(ctx context.Context) (string, error) {
-	output, err := parseOSRelease(ctx)
+	output, err := ParseOSRelease(ctx)
 	if err != nil {
 		return "", err
 	}
@@ -170,7 +170,7 @@ func Distro(ctx context.Context) (string, error) {
 
 // IsDistroDebian detects if the os distro is debian. (Not ubuntu!)
 func IsDistroDebian(ctx context.Context) (bool, error) {
-	output, err := parseOSRelease(ctx)
+	output, err := ParseOSRelease(ctx)
 	if err != nil {
 		return false, err
 	}
@@ -182,7 +182,7 @@ func IsDistroDebian(ctx context.Context) (bool, error) {
 
 // IsDistroFedora detects if the os distro is fedora.
 func IsDistroFedora(ctx context.Context) (bool, error) {
-	output, err := parseOSRelease(ctx)
+	output, err := ParseOSRelease(ctx)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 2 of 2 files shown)


jordansissel:
I ran some tests and got this output:

```
./lang/core/os/distro.go ...................................................V
Release parses /etc/os-release and returns a struct containing
the id, version_id, and variant_id.
failed with: block (distro.go:124) failed: line 2 is not reflowed properly
FAIL
The following tests have failed:
reflowed-comments ./lang/core/os/distro.go
```

I searched online for the 'is not reflowed properly' error message and got _zero_ results. I don't know what "reflowed" means in the context of a plain text comment? Can you advise?

---


This patch will allow me to remove this custom func and its tests! :)

```
$os_release = strings.split(os.readfile("/etc/os-release"), "\n")

# Parse /etc/os-release text format looking for a line starting with "{$field}="
# Returns the value assigned to that field.
$os_release_parse = func($text, $field) {
  golang_strings.replace(
    list_lookup_default(
      iter.filter($text, func($v) { strings.has_prefix($v, fmt.printf("%s=", $field)) }),
      0,
      "OSINFO $field MISSING",
    ),
    fmt.printf("%s=", $field),
    "",
    1,
  )
}

# Is this a reasonable way to assert a function is behaving as expected?
class test {
  $data = ["ID=foo", "VERSION=1.0"]

  $id = $os_release_parse($data, "ID")
  $version = $os_release_parse($data, "VERSION")
  $expect_id = "foo"
  $expect_version = "1.0"

  if $id != $expect_id {
    panic(
      fmt.printf("os_relase_parse failed. Expected ID field == '%v' but got '%v'", $expect_id, $id),
    )
  }

  if $version != $expect_version {
    panic(
      fmt.printf("os_relase_parse failed. Expected VERSION field == '%v' but got '%v'", $expect_version, $version),
    )
  }
}

include test
```

purpleidea:
> is not reflowed properly

Just means markdown comments need to be wrapped at 80 chars.

purpleidea:
> This patch will allow me to remove this custom func and its tests! :)

<3

Anything weird you see in your mcl means maybe it needs to be a function or resource =D

lang/core/os/distro.go
```diff
@@ -101,3 +120,31 @@ func parseDistroUID(uid string) (string, string, string, error) {
 
 	return distro, version, arch, nil
 }
+
+// Release parses /etc/os-release and returns a struct containing
+// the id, version_id, and variant_id.
+func Release(ctx context.Context, input []types.Value) (types.Value, error) {
+	fmt.Println("inside os.release()")
```
purpleidea: nuke all the printfs...

lang/core/os/distro.go
```diff
@@ -34,11 +34,17 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/purpleidea/mgmt/util/errwrap"
+
```
purpleidea: no nl here

lang/core/os/distro.go
```diff
@@ -34,11 +34,17 @@ import (
...
 	"github.com/purpleidea/mgmt/lang/funcs/simple"
 	"github.com/purpleidea/mgmt/lang/types"
+	distroUtil "github.com/purpleidea/mgmt/util/distro"
 )
 
-const structDistroUID = "struct{distro str; version str; arch str}"
+const (
+	structDistroUID   = "struct{distro str; version str; arch str}"
+	releaseReturnType = "struct{ id str; variant_id str; version_id str }"
```
purpleidea: fmt properly like above variant please

lang/core/os/distro.go
```diff
@@ -101,3 +120,31 @@ func parseDistroUID(uid string) (string, string, string, error) {
...
+// version_id, and variant_id.
+func Release(ctx context.Context, input []types.Value) (types.Value, error) {
+	fmt.Println("inside os.release()")
+	release, err := distroUtil.ParseOSRelease(ctx)
+	if err != nil {
+		return nil, errwrap.Wrapf(err, "ParseOSRelease failed")
+	}
+	fmt.Printf("is release nil? %#v", release)
+
+	st := types.NewStruct(types.NewType(releaseReturnType))
```
purpleidea, line 131: Main review question is why do we want these specific values and not others?
Do we want to return a struct or a map of them since they're all strings?

https://godocs.io/github.com/ashcrow/osrelease#OSRelease

purpleidea (changes requested):
Thanks for the patch, here are some review points!

lang/core/os/distro.go
jordansissel: Oops, thanks. I had these for debugging a crash caused by `types.NewType()` returning nil when an invalid type definition is given (perhaps something I'll send another PR for later)

lang/core/os/distro.go
jordansissel: Good questions!

> why do we want these specific values and not others?

I wanted these specific ones for helping mgmt make decisions about what resources are necessary on a given machine. The "ID" fields are most suitable for being processed by computers. I'm happy to add the other fields, but would not use them in mgmt for flow control, that is, I would not use the  "NAME" field in an if statement, I would use ID -- In practice, NAME is for display/humans, ID is for computers; same for VARIANT + VARIANT_ID. VERSION is an optional field, but VERSION_ID is not, etc.

Citations:
* Fields with 'ID' in the name often include this text in the [os-release documentation](https://www.freedesktop.org/software/systemd/man/247/os-release.html?__goaway_challenge=meta-refresh&__goaway_id=1af5d1051075bcf91c95dabbaa9646fd&__goaway_referer=https%3A%2F%2Fduckduckgo.com%2F): "suitable for processing by scripts or usage in generated filenames"
* "If you are reading this file from C code or a shell script to determine the OS or a specific version of it, use the ID and VERSION_ID fields"
* "The VARIANT_ID field should be used for making programmatic decisions."

> Do we want to return a struct or a map of them since they're all strings?

My choice for structure is that os-release has a specification and the contents are not an arbitrary string->string map. That is, there should never be a field "FANCYPANTS" because the spec doesn't describe it and thus we should not expect it.

In practice, the difference between `$foo->version_id` and `$foo["version_id"]` feels pretty small, and one (map keys) allows you to look something up with a variable, like `$foo[$field]`, but I do not expect such a use case.

I like a struct for this because the fields are a fixed, known set of names, and do not have a need to look up the "list of fields" (map keys) nor fetching a field by parameter `$foo[$field]`.

purpleidea: I do agree it should be a struct. I was hoping to avoid it being a huge struct. So at this point the question is if we want to include other fields in it or not.

I would include all the ones you think are valuable, and add a comment to the func definition that the returned struct fields may change in the future (we'll break API) if more are deemed useful.

jordansissel: My thought about avoiding the "display-only" fields is to avoid folks stumbling on things by accident. Choosing NAME ("Fedora") vs ID ("fedora") would certainly be something that I stumble on. so I'm -1 on including displayables in this structure.

That said, maybe an additional function for the display names? I'm open to that, and documentation could express that these values are for display/human reading, and to use the other function (that uses _ID fields) instead?

os.displayname() ? No particular thoughts on the naming.

Additionally:

Should we add ID_LIKE ? `id_like` 
> A space-separated list of operating system identifiers in the same syntax as the ID= setting.

In this case,  id_like would be a array of str. This could be used more casually for things that are unique to all derivations of a distro. The docs give an example of for ID=ubuntu, ID_LIKE=debian. For ID=centos, ID_LIKE="rhel fedora"

I can imagine using this, for example, I have exactly this in an mcl file today:

```
 if $osinfo["id"] == "debian" or $osinfo["id"] == "ubuntu" {
```

Perhaps a second function to add that would replace this:

```
if os.like("debian") {
```

os.like() would ask if `ID_LIKE` (string array) contains the string "debian" ?

jordansissel: For display name, maybe just including PRETTY_NAME ? The name of the field, to me, infers its use (pretty! for display!)

purpleidea: I agree with all your points. You just can't name the function `os.like`.

purpleidea:
OK to merge this first commit as-is?

jordansissel:
+1

purpleidea:
Merged, thank you!

PR #936, opened 2026-06-03
lang: funcs: Add sys.smbios.oem_strings()

jordansissel:
This reads the SMBIOS Type 11 data and returns a []str of the contents.

Example usage:

I booted qemu with the following flags:
* `-smbios type=11,path=config.json`  (This includes the contents of 'config.json' as a type 11 string, neat!)
* `-smbios type=11,value=this is the second one` (A second type 11 entry to make sure multiple entries works)

MCL:

```
import "sys"
import "fmt"
for $i, $v in sys.oem_strings() {
  print [fmt.printf("string %d", $i)] { msg => $v, Meta:autogroup => false, }
}
```

The mgmt output:

```
20:12:22 engine: print[string 0]: Msg: {
  "V1": {
    "distro": {
      "Debian": {
        "release": "trixie"
      }
    },
    "variant": "Default"
  }
}
20:12:22 engine: print[string 1]: Msg: this is the second one
```

----

I'd like to write tests for this just to keep it safe.

Implementation details:

There may already be existing golang smbios libraries, but the data structure is pretty simple (and extremely well documented), so I felt it was best to read and decode it directly.

Reference: https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.9.0.pdf

### DIFF
--- a/lang/core/sys/smbios.go
+++ b/lang/core/sys/smbios.go
@@ -1,0 +1,102 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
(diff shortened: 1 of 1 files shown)


purpleidea:
Quick draft comment: you named it sys.smbios... you probably meant to have a subpackage, like sys/smbios if there are planning to be a lot of smbios.whatever() functions. If not then you probably want sys.smbios_whatever()

jordansissel:
I figured it should move into a subpackage, but for the draft I was trying to save energy figuring out how to do the subpackage and get the prototype working instead 

I'll move it to a subpackage and then move this to ready-to-review

purpleidea:
You can look at the other lang/core/ dirs for sub package examples.

However we need a rough design (even if not implemented) about what else should go in there before we make it a subpackage. If we don't have more than say 3-4+ functions, let's just do sys.smbios_whatever()

purpleidea:
Also as part of this review: keep in mind (1) on some machines that file doesn't exist and (2) we must write safe code that can never panic. Yours is a big unsafe with looking into list indexes without length checking first for example.

PR #947, opened 2026-06-26
test: Add automatic link checker to markdown test suite

### LINKED ISSUE #440: [LOVE] Add automatic link checking to our docs
Someone pointed out that we have a dead link in our markdown docs. I thought, darn, we should have an automatic link checker. And turns out, it already exists.

Here's one:

https://github.com/raviqqe/liche/

Assuming there isn't a better alternative, we should integrate this into https://github.com/purpleidea/mgmt/blob/master/test/test-markdownlint.sh

Patches welcome!

0xp0tato:
test: Add automatic link checker to markdown test suite
This adds link checking to test/test-markdownlint.sh using lychee. Run with: ./test/test-markdownlint.sh --check-links

Fixes https://github.com/purpleidea/mgmt/issues/440

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/test/test-markdownlint.sh
+++ b/test/test-markdownlint.sh
@@ -92,8 +92,7 @@ bad_files=$(
 
 		# check links in docs
 		if $CHECK_LINKS; then
-			# if file is from the directory docs/ then check links
-			if [[ "$i" == docs/* ]] && ! "$LYCHEE" -n "$i" 1>&2; then
+			if ! "$LYCHEE" --root-dir . --accept 200..300,302,429 --timeout 120 --max-retries 2 -n -v "$i" 1>&2; then
 				echo "$i"
 			fi
 		fi


purpleidea:
Are these changes needed?  If you run it locally, does it find any bad links?

Perhaps we only need to add the --check-links are in the test.sh script when it's called? Although I suppose it doesn't make sense to run that every time.

0xp0tato:
> Are these changes needed? If you run it locally, does it find any bad links?

Yes, It found 3 links which were not working

https://jonathangold.ca/blog/aws-ec2-in-mgmt/ 
https://www.redhat.com/mailman/listinfo/mgmtconfig-listist 
http://log.winsos.net/2017/04/12/mgmt-aux-human-talks-grenoble.htmltml

> Perhaps we only need to add the --check-links are in the test.sh script when it's called? Although I suppose it doesn't make sense to run that every time.

Didn't touch ./test.sh but perhaps that can be done. (passing  --check-links to test.sh).
If that is needed, let me know and I will make the changes

purpleidea:
I found four links and updated them.

0xp0tato:
Will this PR be merged and the issue will be closed?

purpleidea:
Just need to know why those new args are needed or not first.

0xp0tato:
- "--root-dir . " is required since we wanted to scan all the .md files in the project and needed the provide a relative path
- "--accept 200..300,302,429" is for all the accepted status codes.  some valid links return 302 or 429 which lychee rejects by default but don't indicate broken links
- "--timeout 120" for pages which are slow. This could be only on my end since my internet speed fluctuates. This can be removed but i added this just as a preventive measure
- "--max-retries 2" for the same reason as above. ☝️
- "-v" for verbose output. just wanted to get some extra information out of the scan

purpleidea:
Thanks for the patch! I've rebased and merged it! \o/

0xp0tato:
thnx

PR #949, opened 2026-06-30
cli: Add Explicit key selection

karpfediem:
Adds a --ssh-id CLI flag to select an exact SSH private key path.
This sets the internal `SSHID` variable (`etcd/ssh/ssh.go`), which was previously not exposed directly.

With an unset/empty `SSHID` we collect SSH client signers only by scanning the default SSH directory for `id_*` private keys.

This change exposes the existing `SSHID` field as an explicit single-key alternative.

Why this was needed:
- My PK files are not neccessarily under ~/.ssh
- The current scan does not respect a custom HOME override
ExpandHome ends up using the actual user home setting via os.user
```go
	if strings.HasPrefix(p, "~/") {
		usr, err := user.Current()
		if err != nil {
			return p, fmt.Errorf("can't expand ~ into home directory")
		}
		return path.Join(usr.HomeDir, p[len("~/"):]) + suffix, nil
	}
```

### DIFF
--- a/cli/deploy.go
+++ b/cli/deploy.go
@@ -68,6 +68,10 @@ type DeployArgs struct {
 	// is specified, then it overrides looking for it in the URL.
 	SSHHostKey string `arg:"--ssh-hostkey" help:"use this ssh known hosts key when connecting over SSH"`
 
+	// SSHID is the private key path for SSH client auth with --ssh-url. If empty,
+	// mgmt scans the default SSH directory for id_* private keys.
+	SSHID string `arg:"--ssh-id" help:"private key for SSH client auth"`
+
 	Seeds []string `arg:"--seeds,separate,env:MGMT_SEEDS" help:"default etcd client endpoints"`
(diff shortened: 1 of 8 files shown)


purpleidea:
If LLM/AI was used in this commit, it must be disclosed. Please confirm, and thank you.

purpleidea:
Otherwise, patch LGTM, except you may wish to also plumb things through in lang/core/embedded/provisioner/main.mcl

Look for `--ssh-hostkey` and append the new arg everywhere that is seen... Should be an easy pattern match addition.

Thanks!

karpfediem:
> If LLM/AI was used in this commit, it must be disclosed. Please confirm, and thank you.

Both commits were written by me, without AI this time.

I've added the new flag to the embedded provisioner module code. Although they're really just going to be empty for now. Until the module code does anything fancy, like allowing to install a custom private key on the new host instead of auto-generating it.

purpleidea:
I've squashed and reflowed the commit message and merged! Thank you.

PR #950, opened 2026-06-30
engine: resources: packagekit: Use signal info for installed state

karpfediem:
Fix installed-state detection for the PackageKit package resource.

PackageKit resolve results expose the package state (installed/absent, or other classification) in the Package signal `info` argument:

```text
Package(info, package_id, summary)
```

The `package_id` is only the package identifier:

```text
name;version;arch;data
```

mgmt was using the package ID `data` field to decide whether a package was installed. That is backend-specific and breaks with the `apt` backend (Debian, ...).

Observed installed package data:

```text
Debian 13 / apt:
Package(1, "qemu-utils;1:10.0.8+ds-0+deb13u1+b2;amd64;auto:debian-stable-main", ...)

Fedora 42 / dnf:
Package(1, "bash;5.2.37-1.fc42;x86_64;installed:anaconda", ...)
```

In both cases `info=1` is `PK_INFO_ENUM_INSTALLED`.
For an available package that is not installed (e.g. `sl`) it shows 2:

```text
Package(2, "sl;5.02-1+b1;amd64;debian-stable-main", ...)
```

See the [enum source in PackageKit](https://github.com/PackageKit/PackageKit/blob/c6eb21d7f4e1005fd2296f50572420906d7d3ae4/lib/packagekit-glib2/pk-enum.h#L620-L660)
```h
typedef enum {
	PK_INFO_ENUM_UNKNOWN, // index 0
	PK_INFO_ENUM_INSTALLED, // index 1
	PK_INFO_ENUM_AVAILABLE, // index 2
// ...
```

The package ID `data` field still shows a difference between backends:

```text
apt: auto:debian-stable-main
dnf: installed:anaconda
```

The apt backend intentionally uses `manual:` / `auto:` package ID data for installed packages. So the portable state source is the PackageKit `info` enum, not the package ID string.

## Minimal reproduction

```mcl
pkg "qemu-utils" {
    state => "installed",
}
```

On Debian with `qemu-utils` already installed, upstream mgmt repeatedly applies the package because it classifies the apt package ID as not installed:

```text
engine: pkg[qemu-utils]: Check: pkg[qemu-utils]
engine: pkg[qemu-utils]: Apply: pkg[qemu-utils]
engine: pkg[qemu-utils]: Set(installed): pkg[qemu-utils]...
engine: pkg[qemu-utils]: Set(installed) success: pkg[qemu-utils]
engine: pkg[qemu-utils]: Check: pkg[qemu-utils]
engine: pkg[qemu-utils]: Apply: pkg[qemu-utils]
...
error: max runtime reached
```

## Fix

Preserve the PackageKit signal `info` value while resolving packages, and use that value for installed-state checks.

`ResolvePackages` still returns package IDs as before; the retained signal info is only used internally where mgmt needs package state.

There is no fallback to package ID `data` parsing for installed state.

## Tests

The unit test covers the new condition directly:

```text
info=1, package_id=...;auto:debian-stable-main => installed
info=2, package_id=...;debian-stable-main      => not installed
```

Focused test run:

```sh
nix shell nixpkgs#go --command env CGO_ENABLED=0 go test -tags 'noaugeas novirt' ./engine/resources/packagekit
```

Manual smoke runs with the patched binary:

```text
Debian 13, PackageKit 1.3.1, pkg "qemu-utils": converged.
Fedora 42, PackageKit 1.3.4, pkg "bash": converged.
```

### DIFF
--- a/engine/resources/packagekit/packagekit.go
+++ b/engine/resources/packagekit/packagekit.go
@@ -257,6 +257,15 @@ type PkPackageIDActionData struct {
 	Newest    bool
 }
 
+type resolvedPackage struct {
+	Info      uint32
+	PackageID string
+}
+
+// PackageKit's Package signal uses sequential PkInfoEnum values, not the
+// bitfield values used by the PkInfoEnum constants above.
+const packageInfoInstalledValue uint32 = 1
+
 // NewBus returns a new bus connection.
 func NewBus() (*Conn, error) {
 	// if we share the bus with others, we will get each others messages!!
@@ -371,11 +380,42 @@ func (obj *Conn) CreateTransaction() (dbus.ObjectPath, error) {
 
 // ResolvePackages runs the PackageKit Resolve method and returns the result.
 func (obj *Conn) ResolvePackages(ctx context.Context, packages []string, filter uint64) ([]string, error) {
-	packageIDs := []string{}
+	resolved, err := obj.resolvePackagesWithInfo(ctx, packages, filter)
+	if err != nil {
+		return nil, err
+	}
+	packageIDs := make([]string, 0, len(resolved))
+	for _, pkg := range resolved {
+		packageIDs = append(packageIDs, pkg.PackageID)
+	}
+	return packageIDs, nil
+}
+
+func newResolvedPackage(body []interface{}) (*resolvedPackage, bool) {
+	if len(body) < 2 {
+		return nil, false
+	}
+	info, _ := body[0].(uint32)
+	packageID, ok := body[1].(string)
+	if !ok {
+		return nil, false
+	}
+	return &resolvedPackage{
+		Info:      info,
+		PackageID: packageID,
+	}, true
+}
+
+func (obj resolvedPackage) installed() bool {
+	return obj.Info == packageInfoInstalledValue
+}
+
+func (obj *Conn) resolvePackagesWithInfo(ctx context.Context, packages []string, filter uint64) ([]resolvedPackage, error) {
+	resolved := []resolvedPackage{}
 	ch := make(chan *dbus.Signal, PkBufferSize)   // we need to buffer :(
 	interfacePath, err := obj.CreateTransaction() // emits Destroy on close
 	if err != nil {
-		return []string{}, err
+		return []resolvedPackage{}, err
 	}
 
 	// add signal matches for Package and Finished which will always be last
@@ -394,7 +434,7 @@ func (obj *Conn) ResolvePackages(ctx context.Context, packages []string, filter 
 		obj.Logf("ResolvePackages(): Call: Success!")
 	}
 	if call.Err != nil {
-		return []string{}, call.Err
+		return []resolvedPackage{}, call.Err
 	}
 loop:
 	for {
@@ -403,7 +443,7 @@ loop:
 		case signal, ok := <-ch:
 			if !ok {
 				// channel closed, e.g. the bus connection died
-				return []string{}, fmt.Errorf("signal channel closed unexpectedly")
+				return []resolvedPackage{}, fmt.Errorf("signal channel closed unexpectedly")
 			}
 			if obj.Debug {
 				obj.Logf("ResolvePackages(): Signal: %+v", signal)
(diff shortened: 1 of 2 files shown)


engine/resources/packagekit/packagekit.go
```diff
@@ -371,11 +380,42 @@ func (obj *Conn) CreateTransaction() (dbus.ObjectPath, error) {
...
+		Info:      info,
+		PackageID: packageID,
+	}, true
+}
+
+func (obj resolvedPackage) installed() bool {
+	return obj.Info == packageInfoInstalledValue
+}
+
+func (obj *Conn) resolvePackagesWithInfo(ctx context.Context, packages []string, filter uint64) ([]resolvedPackage, error) {
```
purpleidea, line 413: Should return ([]*resolvedPackage, error)

engine/resources/packagekit/packagekit.go
```diff
@@ -371,11 +380,42 @@ func (obj *Conn) CreateTransaction() (dbus.ObjectPath, error) {
...
+	if !ok {
+		return nil, false
+	}
+	return &resolvedPackage{
+		Info:      info,
+		PackageID: packageID,
+	}, true
+}
+
+func (obj resolvedPackage) installed() bool {
```
purpleidea, line 409: ptr for the method receiver please

engine/resources/packagekit/packagekit.go
```diff
@@ -371,11 +380,42 @@ func (obj *Conn) CreateTransaction() (dbus.ObjectPath, error) {
...
+		return nil, err
+	}
+	packageIDs := make([]string, 0, len(resolved))
+	for _, pkg := range resolved {
+		packageIDs = append(packageIDs, pkg.PackageID)
+	}
+	return packageIDs, nil
+}
+
+func newResolvedPackage(body []interface{}) (*resolvedPackage, bool) {
```
purpleidea, line 394: please add a short docstring here and below.

Let's also move this to the bottom right after `func (obj *Conn) PackagesToPackageIDs`

The rough convention I have for ordering is main struct with all it's methods, secondary structs with their methods, pub helper funcs, then private helper funcs.

I might have gotten this wrong or made exceptions in some places, it's not a hard rule, but just some attempt at organization.

engine/resources/packagekit/packagekit.go
```diff
@@ -371,11 +380,42 @@ func (obj *Conn) CreateTransaction() (dbus.ObjectPath, error) {
...
+	if len(body) < 2 {
+		return nil, false
+	}
+	info, _ := body[0].(uint32)
```
purpleidea, line 398: please pull out the other values from the comment below and the doc comment, eg:

> `// format is: name;version;arch;data`

and

> `//pkg_int, ok := signal.Body[0].(int)`

If you don't want to use them, at least write the code but leave commented.

We need to check "ok" in all cases too!

engine/resources/packagekit/packagekit.go
```diff
@@ -257,6 +257,15 @@ type PkPackageIDActionData struct {
 	Newest    bool
 }
 
+type resolvedPackage struct {
```
purpleidea, line 260: docstring plz here and for the fields.

engine/resources/packagekit/packagekit.go
```diff
@@ -371,11 +380,42 @@ func (obj *Conn) CreateTransaction() (dbus.ObjectPath, error) {
 
 // ResolvePackages runs the PackageKit Resolve method and returns the result.
 func (obj *Conn) ResolvePackages(ctx context.Context, packages []string, filter uint64) ([]string, error) {
-	packageIDs := []string{}
+	resolved, err := obj.resolvePackagesWithInfo(ctx, packages, filter)
+	if err != nil {
+		return nil, err
+	}
+	packageIDs := make([]string, 0, len(resolved))
```
purpleidea, line 387: nice!

engine/resources/packagekit/packagekit.go
```diff
@@ -257,6 +257,15 @@ type PkPackageIDActionData struct {
...
 }
 
+type resolvedPackage struct {
+	Info      uint32
+	PackageID string
+}
+
+// PackageKit's Package signal uses sequential PkInfoEnum values, not the
+// bitfield values used by the PkInfoEnum constants above.
+const packageInfoInstalledValue uint32 = 1
```
purpleidea, line 267: Are there other values here? Add them if they exist. Move up right above `type PkError struct {` just under the typedef consts. (But with the other consts.)

purpleidea:
PS: Without this patch ... how did this ever work on Debian??? Because I know it did.

karpfediem:
> PS: Without this patch ... how did this ever work on Debian??? Because I know it did.

Did some digging with help of AI:

PackageKit's aptcc backend used to emit installed packages as `installed:<repo>` since 2014:
https://github.com/PackageKit/PackageKit/commit/62ae5b7ed81ebcfa6d5dfc63390825270d06d2e1

Then in 2022 it changed apt package IDs to encode apt install mode instead:
manual:<repo> or auto:<repo>
https://github.com/PackageKit/PackageKit/commit/cfd297aeb51bde67269d74c5a762634314e918b6
(with the first release containing that commit being 1.2.5)

| Debian   | PackageKit (currently) | behavior                    |
|----------|------------------------|-----------------------------|
| bullseye | 1.2.2                  | installed:<repo>            |
| bookworm | 1.2.6                  | manual:<repo> / auto:<repo> |
| trixie   | 1.3.1                  | manual:<repo> / auto:<repo> |

So Debian did work, at least with older PackageKit.

I also believe it worked even on bookworm before.
But that could be explained by the fact that this change was (I would say hidden) inside a patch release.

In any case, the fix now is to just use the Package(info, package_id, summary) signal as it is intended in the current PackageKit.

purpleidea:
Nice! Just needs the fixups I mentioned in my review before merge.

purpleidea (changes requested):

purpleidea:
PS: I am stupid and forgot to click "submit review" I thought it happened with the comment. In any case, sorry about that. Patch LGTM, just needs a few fixups mentioned above.

PR #960, opened 2026-07-15
User resource: trim homedir trailing slash when invoking usermod

jordansissel:
The user resource already compares the current and desired homedir
without a trailing slash, but it also now sets the homedir without a
trailing slash.

This fixes a bug where a shell change is needed, and homedir is
specified but unchanged, and mgmt would include the trailing slash when
invoking usermod. If the user has any active processes at this time,
usermod will refuse, because it insists that the user be inactive while
a homedir change is applied -- even if the homedir is effectively
unchanged, usermod simply does a string comparison, which fails when
mgmt's provided directory includes a trailing slash and the existing
homedir does not have one.

https://github.com/purpleidea/mgmt/discussions/959

An example of this passing/failing:

Current values:
```
# Print homedir and shell
$ getent passwd dev | awk -F: '{print $6, $7}'
/home/dev /bin/zsh
```

mgmt 1.1.0:

```
# Try to set the shell to /bin/bash but also including homedir which effectively is the same path as the current homedir:
$ printf 'user "dev" { homedir => "/home/dev/", shell => "/bin/bash", state => "exists", }\n' | sudo mgmt run lang -
...

16:09:14 engine: user[dev]: modifying user: dev
16:09:14 engine: user[dev]: Error: usermod: user dev is currently used by process 1751 : exit status 8
```

With this patch:

```
16:11:20 engine: user[dev]: modifying user: dev
```

Verification after success:

```
# Bash is now the shell and homedir is unchanged
$ getent passwd dev | awk -F: '{print $6, $7}'
/home/dev /bin/bash
```


jordansissel:
I can't attest to the correctness of this in all situations. It is possible(?) that someone's homedir could be set to a path that has a trailing space, but I've never seen that. In practice, it appears that usermod will accept a `--home` with a trailing slash but those are removed before it is written to /etc/passwd:

```
[dev@localhost work]$ sudo usermod foo --home /some/path/here//////
[dev@localhost work]$ getent passwd foo
foo:x:1001:1002::/some/path/here:/bin/zsh
```

So, hopefully this doesn't cause any weird issues since usermod itself also strips the trailing slash(es) before writing the passwd file, and that gives me faith that we probably won't see a homedir field with a trailing slash.

purpleidea:
I merged a fixed version with a bigger comment and the tests fixed too.

Thank you!

PR #971, opened 2026-07-30
WIP: Perf replace bool slice with uint64 bit index in the lexer

miekg:
This replaces the []bool in lexer.nn.go with a bit index in a uint64, perfing shows the gains 3% perf.

However this file is generated from via nex from the lexer.nex. Hence WIP as this can't be merged as-is and needs upstream changes in nex (likely).

### DIFF
--- a/lang/lang_test.go
+++ b/lang/lang_test.go
@@ -34,6 +34,9 @@ package lang
 import (
 	"context"
 	"fmt"
+	"io"
+	"log"
+	"os"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


purpleidea:
Neat! I am a nex maintainer, although by accident, not by skill. We could definitely do something upstream if that's needed. I can't commit this here of course.

miekg:
diff should be fairly small, you can diff this file with the one that is generated by *nex* now.

*nex*'s master doesn't pass it's tests, so I'm not going to touch it any further. Also the few changes I tried to make didn't wholy do what I expected, but I've could have made a mistake there, hard to tell with a broken test suite.
As a guestimate I would say another 5% savings can found there, maybe more.

miekg:
it does depend on the max states the lexer might have... I've seen 10 in the code where a uint64 should be suffcient... but thus no more than 64...

Issue #978, opened 2026-08-29
panic with value res composite types


purpleidea:
@jordansissel reports in https://github.com/purpleidea/mgmt/discussions/977
I haven't tested this yet, but it's likely this is our bug. Either we need to catch composite types and disallow with panic, or handle them correctly.

```
$example = struct {
        foo => "Hello world",
}

value "example" {
        any => $example,

        Meta:export => [ "*" ],
}
```

This errors:

```
23:21:47 main: waiting...
23:21:47 gapi: lang: stream...
23:21:47 gapi: lang: function engine starting...
23:21:47 gapi: lang: interpret: interpreting...
panic: struct has unexported field: foo

goroutine 354 [running]:
github.com/purpleidea/mgmt/lang/types.(*Type).Reflect(0xc001283590)
        github.com/purpleidea/mgmt/lang/types/type.go:1076 +0xc85
github.com/purpleidea/mgmt/lang/types.(*StructValue).Value(0xc0012925d0)
        github.com/purpleidea/mgmt/lang/types/value.go:1413 +0x5b
github.com/purpleidea/mgmt/lang/types.Into({0x3d0c588, 0xc0012925d0}, {0x33967e0?, 0xc000373360?, 0x3?})
```

This also panics even without exporting (no Meta:export => ...)

arbelonson-source:
Dug into the mechanics here — not confident enough to just implement one of the two directions you're weighing in the discussion, so posting what I found instead of guessing. Disclosure up front: I run agent loops (Claude Code) that investigate issues like this; this is AI-produced analysis, not something I personally read the code for.

**Exact mechanism**: `Type.Reflect()` (`lang/types/type.go`, `KindStruct` case) builds a real Go struct via `reflect.StructOf` so `StructValue.Value()` (`lang/types/value.go`) can produce an `interface{}` for consumers like `Into()`/gob encoding. `reflect.StructOf` panics on any unexported field, and this codebase already knows it — there's a dead TODO sitting right there:

```go
if strings.Title(k) != k { // is exported?
    //k = strings.Title(k) // TODO: is this helpful?
    // reflect.StructOf would panic on anything unexported
    panic(fmt.Sprintf("struct has unexported field: %s", k))
}
```

[...]

PR #979, opened 2026-08-31
lang: types: Export struct field names in Reflect

arbelonson-source:
This allows composite types (struct, list, map, and nested combinations) in the value resource's `any` field, per your request on #978.

Our struct field names are lowercase, which `reflect.StructOf` panics on, since those are unexported in golang. Instead, `Reflect()` now exports each field name by capitalizing it and keeps the original in the `lang` struct tag — the same tag `TypeOf` and `Into` already read — so values round trip back out unchanged, including from `ValueOfGolang`. `StructValue.Value()` sets fields by index instead of by name, since the reflected names may now differ. Already-exported names keep their exact previous representation, with no tag.

Tests, as requested in TestAstFunc2/:
- `value-res-any-struct0.txtar` — the exact example from the issue
- `value-res-any-list0.txtar`, `value-res-any-map0.txtar`
- `value-res-any-nested0.txtar` — struct containing a str, an int, an inner struct, a list of structs, and a map with struct values

Verification notes, stated honestly:
- struct0 and nested0 panic on unmodified master with the exact reported panic (`struct has unexported field: foo`) and pass with this patch. list0/map0 pass on master too — scalar lists/maps never hit the struct path — so those two lock in coverage rather than guard a regression.
- `TestReflect0` expected the old panic; its own comment said the behaviour was provisional and to update it if this ever changed, so it now asserts the new name+tag mapping and the `TypeOf` round trip. Added `TestStruct3` covering the full value round trip (`Value()` → `ValueOfGolang` → `Cmp`) including a nested struct.
- Full `go test ./lang/...` passes except `lang/format`, which fails identically on unmodified master in my environment (missing generated golang func wrappers), so it's pre-existing and unrelated.

As you requested, this patch was generated by Claude Fable 5, via Claude Code (the earlier analysis comments on the issue were Claude Opus 5). AI-assisted-by: Claude Fable 5, via Claude Code

### DIFF
--- a/lang/interpret_test/TestAstFunc2/value-res-any-list0.txtar
+++ b/lang/interpret_test/TestAstFunc2/value-res-any-list0.txtar
@@ -1,0 +1,8 @@
+-- main.mcl --
+$example = ["a", "b", "c"]
+
+value "example" {
+	any => $example,
+}
+-- OUTPUT --
+Vertex: value[example]
--- a/lang/interpret_test/TestAstFunc2/value-res-any-map0.txtar
+++ b/lang/interpret_test/TestAstFunc2/value-res-any-map0.txtar
@@ -1,0 +1,8 @@
+-- main.mcl --
+$example = {"one" => 1, "two" => 2}
+
+value "example" {
+	any => $example,
+}
+-- OUTPUT --
+Vertex: value[example]
--- a/lang/interpret_test/TestAstFunc2/value-res-any-nested0.txtar
+++ b/lang/interpret_test/TestAstFunc2/value-res-any-nested0.txtar
@@ -1,0 +1,18 @@
+-- main.mcl --
+$inner = struct{
+	deep => ["x", "y"],
+}
+
+$example = struct{
+	foo => "Hello world",
+	bar => 42,
+	baz => $inner,
+	items => [struct{num => 1}, struct{num => 2}],
+	pairs => {"key" => struct{ok => true}},
+}
+
+value "example" {
+	any => $example,
+}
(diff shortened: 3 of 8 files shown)


arbelonson-source:
CI caught two things, both mine, and one of them means **a claim in my PR description above was wrong** — correcting that explicitly rather than quietly force-pushing over it:

I wrote that `lang/format` fails identically on unmodified master and is pre-existing. That was wrong. It did fail on my machine before my change, but for a *different, environmental* reason (I hadn't generated the golang func wrappers yet), and that masked the real failure. The actual CI failure was `file is not in canonical format: main.mcl` — my new `.txtar` used `struct {` with a space, copied from the issue text, where canonical mcl is `struct{`. Ran it through `txtarfmt`. `lang/format` now passes here too.

The other was the misspell linter: `capitalizing` → `capitalising`.

Both pushed. Full `go test ./lang/...` is green locally now, with no exclusions this time.

The environmental lesson, in case it's useful for the docs: on Arch, mgmt needs the `ragel6` package specifically — ragel 7 dropped the `-Z` Go output flag, so `make lang` fails on the interpolation step until you swap it.

AI-assisted-by: Claude Opus 5, via Claude Code (this reply; the patch itself is still Fable 5's, as requested)

purpleidea:
Sorry for the delay in replying, I have a mission to thoroughly review AI patches, and was backed up with work! All done, patch is quite good, thank you! I updated the commit message and squashed the CI fixup. Merged, thanks! If the human is there, give me a ping. Cheers

PR #980, opened 2026-09-01
util: atomicfile: Atomic file delivery

jordansissel:
This allows file writes to be delivered atomically by staging writes to a temporary file and using rename(2) to replace the target file only when we are ready. This means mgmt could update files (with the file resource, etc) without race conditions between `os.Create()` (which creates/truncates a file) and future writes to that File.

The gist is that we treat AtomicFile the same way we do a File, with two differences:
* atomicfile.New("destination") instead of os.Create("destination")
* call Commit() when you are ready to deliver the file to the destination path.

I have included implementations for Linux and macOS as they have different syscalls needed to achieve the same effects. Tests pass for me on Linux (fedora 44) and macOS (26.6.2).

A non-exhaustive list of places in mgmt where this pattern of atomic file delivery may be of value:
* file resource
* http:client when used with the 'file' param
* line resource
* ssh_authorized_key resource
* tar resource

Implementation thoughts: I tried to keep the temporary files invisible to the filesystem. On Linux, this is achieved with O_TMPFILE, and on macOS, I delete the filename as soon as I can while keeping the file open. The goal with this method is several:
1. I don't know yet how mgmt's use of inotify might make things _wierd_ with temporary files popping up.
2. If for any reason the program/computer crashes or mgmt decides not to complete the file writing, the temporary file is automatically freed by the filesystem.

Background: I have `http:server:file` serving files, and `file` writing contents to that file. There are occasions when the http server sends a zero-length file. My hypothesis is that this happens becasue the `file` resource writes to a file by calling os.Create on that path (which truncates it) and later writes data to that file. This gives us a small window of time where the file has zero length. It could also affect larger files where an external process might read  a file *while* it is being writtten by mgmt resulting in a partial read.

Out of scope (for this PR):
* Some filesystems on Linux do not support O_TMPFILE. Detecting this case would be a useful future improvement.
* Windows support
* Changes to resources to use this package should happen in a future PR, unless you feel otherwise.

### DIFF
--- a/util/atomicfile/atomicfile_linux.go
+++ b/util/atomicfile/atomicfile_linux.go
@@ -32,8 +32,12 @@
 package atomicfile
 
 import (
+	"io"
+	"io/fs"
+	"math/rand/v2"
 	"os"
 	"path"
+	"strconv"
 
 	"golang.org/x/sys/unix"
 )
@@ -47,39 +51,130 @@ func open(fspath string) (*os.File, error) {
 	// > Create an unnamed temporary regular file.  The path
 	// > argument specifies a directory; an unnamed inode will be
 	// > created in that directory's filesystem.
-	return os.OpenFile(path.Dir(fspath), unix.O_RDWR|unix.O_TMPFILE, 0600)
+	//
+	// Where a filesystem doesn't support O_TMPFILE, open() will return
+	// EOPNOTSUPP.
+	dir := path.Dir(fspath)
+	f, err := os.OpenFile(dir, unix.O_RDWR|unix.O_TMPFILE, 0600)
+	if err == nil {
+		return f, nil
+	}
+
+	// Check if OpenFile failed because O_TMPFILE isn't supported by the
+	// filesystem.
+	if err, ok := err.(*fs.PathError); ok && err.Err == unix.EOPNOTSUPP {
+		// O_TMPFILE doesn't work on this filesystem, try another method.
+		return openTemp(fspath)
+	}
+
+	return nil, err
+}
+
+func (obj *AtomicFile) onSameFilesystem() (bool, error) {
+	conn, err := obj.SyscallConn()
+	if err != nil {
+		return false, err
+	}
+
+	var srcstat, dststat unix.Stat_t
+	conn.Control(func(fd uintptr) {
+		err = unix.Fstat(int(fd), &srcstat)
+	})
+	if err != nil {
+		return false, err
+	}
+
+	if err = unix.Stat(obj.path, &dststat); err != nil {
+		if os.IsNotExist(err) {
+			// Full path doesn't exist. Check the parent directory.
+			if err = unix.Stat(path.Dir(obj.path), &dststat); err != nil {
+				return false, err
+			}
+
+			return srcstat.Dev == dststat.Dev, nil
+		}
+		return false, err
+	}
+
+	return srcstat.Dev == dststat.Dev, nil
 }
 
 func (obj *AtomicFile) commit() error {
-	temp, err := os.CreateTemp(path.Dir(obj.path), path.Base(obj.path)+".new*")
+	same, err := obj.onSameFilesystem()
 	if err != nil {
 		return err
 	}
+	if same {
+		err = obj.commitWithLinkat()
+		if err, ok := err.(*fs.PathError); ok && err.Err == unix.EACCES {
+			return obj.commitByCopy()
+		}
+		return err
+	} else {
+		return obj.commitByCopy()
+	}
+}
+
+func (obj *AtomicFile) commitByCopy() error {
+	temp, err := os.CreateTemp("", path.Base(obj.path))
+	if err != nil {
+		return err
+	}
+
+	defer os.Remove(temp.Name())
 	defer temp.Close()
 
-	err = os.Remove(temp.Name())
+	file, err := os.OpenFile(obj.path, os.O_WRONLY, 0600)
 	if err != nil {
 		return err
 	}
-	// Note: There is a race condition here between the Remove and Linkat calls.
-	// Above, CreateTemp is used to create a temporary, named file. It is removed
-	// in order to link our unnamed file to that pathname.
-	//
-	// If, between the Remove and Linkat, something else creates a file with the
-	// same name, the Linkat call will fail due to the filename already existing..
-	//
-	// A future improvement would be to try (possibly multiple times) different
-	// names to give to our temporary file.
+	defer file.Close()
 
-	// Previously, our open() func used O_TMPFILE to create an unnamed file, and
-	// in order to rename(2) to deliver our committed file, we must give our
-	// unnamed file a name!
-	// See linkat(2) and open(2)'s O_TMPFILE documentation on Linux for more information.
-	err = unix.Linkat(int(obj.Fd()), "", unix.AT_FDCWD, temp.Name(), unix.AT_EMPTY_PATH)
-	if err != nil {
+	if _, err = io.Copy(file, temp); err != nil {
 		return err
 	}
+
+	return nil
+}
+
+func (obj *AtomicFile) commitWithLinkat() error {
+	try := 0
+	base := path.Join(
+		path.Dir(obj.path),
+		path.Base(obj.path)+".new-",
+	)
+
+	// Select a filename that doesn't exist yet in order to give our
+	// currently unnamed file a real name so that we can rename() it.
+	//
+	// Note: This for loop was modeled after golang's CreateTemp implementation
+	var current string
+	for {
+		current = base + strconv.FormatUint(rand.Uint64(), 16)
+
+		// Previously, our open() func should have created an unnamed file. In
+		// order to rename(2) to deliver our committed file, we must give our
+		// unnamed file a name!
+		// See linkat(2) and open(2)'s O_TMPFILE documentation on Linux for more
+		// information.
+		err := unix.Linkat(int(obj.Fd()), "", unix.AT_FDCWD, current, unix.AT_EMPTY_PATH)
+		if err != nil {
+			if os.IsExist(err) {
+				if try++; try < 10000 {
+					continue
+				}
+
+				// Exhausted tries.
+				return &os.PathError{Op: "linkat", Path: current, Err: err}
+			} else {
+				return err
+			}
+		}
+
+		// Linkat succeeded!
+		break
+	}
+
 	defer obj.Close()
-
-	return os.Rename(temp.Name(), obj.path)
+	return os.Rename(current, obj.path)
 }
--- a/util/atomicfile/atomicfile_linux.go
+++ b/util/atomicfile/atomicfile_linux.go
@@ -1,0 +1,85 @@
+// Mgmt
+// Copyright (C) James Shubin and the project contributors
+// Written by James Shubin <james@shubin.ca> and the project contributors
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+// Additional permission under GNU GPL version 3 section 7
+//
+// If you modify this program, or any covered work, by linking or combining it
+// with embedded mcl code and modules (and that the embedded mcl code and
+// modules which link with this program, contain a copy of their source code in
+// the authoritative form) containing parts covered by the terms of any other
+// license, the licensors of this program grant you additional permission to
+// convey the resulting work. Furthermore, the licensors of this program grant
+// the original author, James Shubin, additional permission to update this
+// additional permission if he deems it necessary to achieve the goals of this
+// additional permission.
+
+//go:build linux
+
+package atomicfile
+
+import (
+	"os"
+	"path"
+
+	"golang.org/x/sys/unix"
+)
+
+func open(fspath string) (*os.File, error) {
+	// Note: this only works on certain filesystems on Linux. See the `O_TMPFILE`
+	// documentation on `open(2)` for more details.
+	//
+	// Relevant notes from Linux's open(2) are:
+	//
+	// > Create an unnamed temporary regular file.  The path
+	// > argument specifies a directory; an unnamed inode will be
+	// > created in that directory's filesystem.
+	return os.OpenFile(path.Dir(fspath), unix.O_RDWR|unix.O_TMPFILE, 0600)
+}
+
+func (obj *AtomicFile) commit() error {
+	temp, err := os.CreateTemp(path.Dir(obj.path), path.Base(obj.path)+".new*")
+	if err != nil {
+		return err
+	}
+	defer temp.Close()
+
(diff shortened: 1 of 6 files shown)


jordansissel:
>   failed with: block (./util/atomicfile/atomicfile.go:38) failed: line 11 wasn't trimmed properly:   - If the program crashes before a Commit(), the temporary file shall be

I made several attempts to fix this  but I don't know exactly what action will resolve "wasn't trimmed properly".

> Error: util/atomicfile/atomicfile_linux.go:50:9: G302: Expect file permissions to be 0600 or less (gosec)

I can make that change but this error I believe is misleading. The file cannot be opened by any other process due to the way O_TMPFILE works (the file has no name).

jordansissel:
> failed with: block (./util/atomicfile/atomicfile.go:38) failed: line 11 wasn't trimmed properly: - If the program crashes before a Commit(), the temporary file shall be

Guessing -- I think the test/tools/reflowed-comments tool doesn't understand(?) bullet points or left-indented text inside a comment? The check appears to compare the line with strings.TrimSpace(line) and that removes spaces on both sides. 'gofmt' appears to support them but changes the formatting a bit. It also doesn't understand that a bullet point can cross multiple lines of comment.

I dunno. I'm deep in the weeds here with tools that don't like how I typeset or write. plz advise ;P

purpleidea:
> `test/tools/reflowed-comments`

`make godocfmt` to fix.

purpleidea:
> Expect file permissions to be 0600 or less (gosec)

I'd set the perms to that anyways, but if it's not possible you can add a nolint tag. See the source for examples.

purpleidea:
> failed with: block (./util/atomicfile/atomicfile.go:38) failed

`Switch to * instead of - and it should work I think`

purpleidea:
> Changes to resources to use this package should happen in a future PR, unless you feel otherwise.

I would recommend patching one of the easiest or most important cases for you in the same PR just so we can feel out the ergonomics of the lib.

util/atomicfile/atomicfile_linux.go
```diff
@@ -0,0 +1,85 @@
...
+	// Relevant notes from Linux's open(2) are:
+	//
+	// > Create an unnamed temporary regular file.  The path
+	// > argument specifies a directory; an unnamed inode will be
+	// > created in that directory's filesystem.
+	return os.OpenFile(path.Dir(fspath), unix.O_RDWR|unix.O_TMPFILE, 0600)
+}
+
+func (obj *AtomicFile) commit() error {
+	temp, err := os.CreateTemp(path.Dir(obj.path), path.Base(obj.path)+".new*")
```
purpleidea: My one design suggestion is that everywhere in mgmt we have "vardir" API which is basically a place to get a "tmp" dir for various writes. This way there's only one place the user needs to look to find any temporary data or storage or anything beyond what we ask mgmt itself to do to the system... Can we plumb in the base dir of vardir to use that for any temp writing we need to do?

jordansissel: I was hoping you'd ask about this!

In order for this "atomic file replacement" to work, we can't really use mgmt's vardir. This is because the rename(2) syscall will faill with EXDEV (cross device). Because vardir may not necessarily live on the same filesystem as the target file, I can't see a way to use vardir for this purpose.

There may be alternative implementations, and I have some ideas for workarounds, but functionally, vardir is not suitable here.

I experienced this with `http:client`'s file feature which downloaded. As an example:

```
[dev@localhost ~]$ sudo mgmt run lang 'http:client "beep beep" { url => "https://github.com/robots.txt", file => "/home/dev/robots.txt", Meta:poll => -1, }
'

...

00:23:38 main: waiting...
00:23:38 engine: http:client[beep beep]: downloaded 2274 bytes
00:23:38 engine: http:client[beep beep]: Error: rename /var/lib/mgmt/engine/state/http:client-YmVlcCBiZWVw+beep beep/tmp /home/dev/robots.txt: invalid cross-device link
```

This happens because on my test system, /home/dev and /var/lib/mgmt are on different filesystem mounts, so rename(2) cannot succeed.

jordansissel: Oh and one other thing, the implementation on Linux should only have *extremely* short-lived files. This is because it uses O_TMPFILE which opens a file *without a name*, and closing the file causes the file to be reclaimed on the filesystem (unless you give it a name).

In practice, this should work as follows, and I'll use a large file download as an example.

If http:client uses AtomicFile to download a large file, then the following sequence occurs:

1. AtomicFile New() opens a new file using O_TMPDIR on the same filesystem as the target file. This file has no name and you cannot find it with `ls` or other file listing tools.
2. http:client writes to the file as the download makes progress.
3. The download completes successfully and all bytes are written to this temporary file.
4. chmod + chown, as needed
5. Call Commit()

Until step 5, the file is both invisible to humans *and* will be garbage collected and space reclaimed if:
* http:client calls Close[1] without calling Commit()
* mgmt crashes or is destroyed by the OS (oomkiler always targets your favorite process)
* the machine loses power or fails otherwise.

[1] https://github.com/purpleidea/mgmt/blob/master/engine/resources/http_client.go#L868

That is, in almost all cases, there will be no "place the user needs to look to find any temporary data". There's a tiny exception (if the machine/program crashes during Commit() between the steps where the file is given a name and the step where the file is renamed) though I am eager to find a way to combine that into a single syscall.

jordansissel:
> `make godocfmt` to fix.

```diff
-// Thus:
-//   - If the program crashes before a Commit(), the temporary file shall be
-//     lost (on purpose).
-//   - If the program calls Close() without a Commit(), the temporary file and
-//     its writes shall be lost (on purpose).
-//   - If the computer fails before a Commit(), the temporary file shall be
-//     lost (on purpose).
+// Thus: - If the program crashes before a Commit(), the temporary file shall be
+// lost (on purpose). - If the program calls Close() without a Commit(), the
+// temporary file and its writes shall be lost (on purpose). - If the computer
+// fails before a Commit(), the temporary file shall be lost (on purpose).
```

Hah, I uhh.. I dont' think that's better, though it is amusing! I understand if I am the first to write a doc comment that has bullet points. I can rephrase things, but, imo, this is most readable as a bulleted list. 

Thoughts?

purpleidea:
> this is most readable as a bulleted list.

No it definitely supports lists. They have to use * (star) before you run the `make godocfmt` to be seen as such. If you search for `* ` you'll see some comment examples.

util/atomicfile/atomicfile_linux.go
purpleidea: > syscall will faill with EXDEV (cross device)

I think this is the rare case most of the time, so I think it should work like this:

1) If it's same device, then use vardir
2) If it's not same device, fall back to this suffix game

Would that be acceptable?

I also considered extending the vardir API to let you have a separate vardir on each device, but I think this is too much complexity for now.

jordansissel: I agree we should try to handle it better (and your proposal goes in that direction)

> I think this is the rare case most of the time, so I think it should work like this:

A stock Fedora 44 cloud init image uses BTRFS and subvolumes(?) by default. This breaks rename(2) across mgmt's default vardir (as root) of /var/lib/mgmt because /var is a different mount point:

```
/dev/vda3       20866028 1893660  18566084  10% /
/dev/vda3       20866028 1893660  18566084  10% /boot
/dev/vda3       20866028 1893660  18566084  10% /home
/dev/vda3       20866028 1893660  18566084  10% /var
```

so I suspect this error will happen to any fedora 44 server/cloud (probably) server where http:client downloads a file to anywhere other than /var (say, downloading to /usr/local, or /home, etc) will trigger this, as I experienced (downloading a file to /usr/local/bin, I think?)

jordansissel:
> > this is most readable as a bulleted list.
> 
> No it definitely supports lists. They have to use * (star) before you run the `make godocfmt` to be seen as such. If you search for `* ` you'll see some comment examples.

I'll take a look and see what I can learn.

Using asterisks is rejected for me:

```
Thus:
* If the program crashes before a Commit(), the temporary file shall be
  lost (on purpose).
* If the program calls Close() without a Commit(), the temporary file and
  its writes shall be lost (on purpose).
* If the computer fails before a Commit(), the temporary file shall be
  lost (on purpose).
```

```
% ./test/reflowed-comments util/atomicfile/atomicfile.go
...
failed with: block (util/atomicfile/atomicfile.go:38) failed: line 12 wasn't trimmed properly:   lost (on purpose).
```

It bails on the wrapped line probably because I indent the second line to line up with the start of the line above?

Further, I believe I may be doing something wrong, because 'gofmt' rewrites my comments to use dashes instead of asterisks *and* it adds indentation which makes reflowed-comments' "trimming" check upset:

Below is before/after `make gofmt`

```diff
-// * If the program crashes before a Commit(), the temporary file shall be
-//   lost (on purpose).
-// * If the program calls Close() without a Commit(), the temporary file and
-//   its writes shall be lost (on purpose).
-// * If the computer fails before a Commit(), the temporary file shall be
-//   lost (on purpose).
+//   - If the program crashes before a Commit(), the temporary file shall be
+//     lost (on purpose).
+//   - If the program calls Close() without a Commit(), the temporary file and
+//     its writes shall be lost (on purpose).
+//   - If the computer fails before a Commit(), the temporary file shall be
+//     lost (on purpose).
```

jordansissel:
> I'll take a look and see what I can learn.

Ahh, your bulleted lists are not indented, but if I try that, it still fails `failed with: block (util/atomicfile/atomicfile.go:38) failed: line 12 is not reflowed properly`:

```
// AtomicFile is an os.File-like struct which allows you to write data and
// modify metadata before committing it to a specific filename.
//
// The intended use is to ensure complete files are only ever available at the
// given path.
//
// If, for any reason, a Commit() never occurs, this module attempts to ensure
// that no temporary files are left around.
//
// Thus:
// * If the program crashes before a Commit(), the temporary file shall be
// lost (on purpose).
// * If the program calls Close() without a Commit(), the temporary file and
// its writes shall be lost (on purpose).
// * If the computer fails before a Commit(), the temporary file shall be
// lost (on purpose).
//
// For any of the above scenarios, no disk space should be occupied by these
// lost files.
//
// This module focuses on delivering "whole files" and is not concerned with
// durability or serializability of operations.
```

`gofmt` does not modify this but I believe it is because it doesn't interpret this structure as a bulleted list (which explains why it doesn't add indentation and replace asterisks with dashes)

I hold no opinions on the final shape of this the AtomicFile doc comment, but I remain unable to please the formatting checker :(

util/atomicfile/atomicfile_linux.go
jordansissel: I did more research and didnt' find much in the way of avoiding temporarily naming the file something random (suffixed, in this case)...

* There's 'reflink' copying (using ioctl FICLONE)  but that requires a cooperating filesystem that supports COW.
* linkat(2) can't overwrite an existing file, and deleting the old file prior to calling linkat(2) would leave a race condition where the path doesn't exist.
* renameat(2) would be perfect except it doesn't support unnamed files (linkat's AT_EMPTY_PATH flag).
* I checked `ln` and `ln -f` 'force' hardlinking overwrites and existing path, but it does so with 3 syscalls (linkat to create a hardlink to a temporary name , renameat to rename that temporary file to the destination, then unlinkat to unlink the temporary name)
* I wondered if mmap could help us atomically write to a file (resize the file and link dirty pages  all at once?) but I couldn't find a mechanism to do this.

So far, everything I've found has more race conditions *or* works on fewer systems!

point of interest, from https://lwn.net/Articles/789600/

> in https://github.com/ostreedev/ostree we write all new objects to a "staging" directory that includes the value of /proc/sys/random/boot_id, and then syncfs() everything, then renameat() into place (I've been meaning to investigate doing parallel/asynchronous fsync instead). We assume that the files are garbage if the boot id doesn't match (i.e. system was rebooted before commit).

That sounds similar to what we're aiming at (a staging directory perhaps per-filesystem, where possible).

What about keeping a very small database *in* vardir to record information about temporary files that haven't been deleted yet? For a short implementation, it could be a directory full of symlinks to the temporary files. mgmt's job would be to check these records on startup and purge any likely-abandoned temporary files. The entries all being symlinks would make it easy for operators to identify orphaned files, too.

The sequence in Commit() might look like:

1. Create a 'temp file'  record in vardir (a symlink to the temporary file's name) 
2. Give the file a name
3. Rename the file
4. Delete the record

On startup, if any records leftover from step 1 exist, mgmt should delete them.

This might allow us to continue using the same directory as the to-be-replaced-file while still allowing mgmt to keep track of it.

(I'm also in favor, I think, of using vardir as the temporary location when it's on the same filesystem)

POSIX makes filesystems really hard! ;P

jordansissel:
I've updated this PR with a small change to the file resource that makes it use atomicfile.

In casual local testing, it see it working correctly on both Linux and macOS. I tested with the following:

First, run mgmt and use `test.fastcount()` for file contents. This causes a file to be rewritten extremely frequently, good!

```
 ./mgmt run lang --no-network --tmp-prefix 'import "fmt"
import "test"
file "/tmp/hello" { content => fmt.printf("ok: %d", test.fastcount()), state => "exists", }
'
```

Second, observe the readable byte count from the file as fast as we can:

```
while true ; do wc -c /tmp/hello; done | uniq
```

The `| uniq` cuts out repeat counts to make it easier to read the output.

On mgmt 1.1.0-69-gdf1c21a1 (baddev? I forget), this prints repeatedly, showing the race condition between `os.Create()` and `io.Copy` in the file resource.

```
10 /tmp/hello
0 /tmp/hello
10 /tmp/hello
0 /tmp/hello
10 /tmp/hello
0 /tmp/hello
10 /tmp/hello
```

### With this PR

With this pr and the file resource using atomicfile, the output is now stable as the path only ever contains a complete file:

```
11 /tmp/hello
```

(small note, because `test.fastcount()` is a number, the 'wc -c' count will change gradually as the number of digits in the counter increases)

jordansissel:
Also added to http:client, though i wasn't thorough in my review of the that resource and may have missed something.

jordansissel:
Added some extra strategies for cases where O_TMPFILE doesn't work, the filesystem rejects creation of new files, etc, with a final fallback of the following when it is detected that certain strategies aren't working (such as writes to /proc or /sys which reject new files, etc).

* For staging, use go's default temporary path (via os.CreateTemp("", ...)
* For committing, use io.Copy which isn't atomic,

purpleidea:
OK, the API fix came in my sleep, lmk what you think.

`atomic` becomes a *string. These are the options:

`atomic => nil (undefined)`
Attempt to make the edit atomically if the vardir is on the same filesystem. We detect this in advance, and if it's not on the same fs, then we don't do this atomically. (Current behaviour when arg omitted upgrades us to atomic where possible, and doesn't break if weird fs splits.)

`atomic => "" (empty string)`
Force atomic via vardir. If vardir is on a different filesystem then we error.

`atomic => "/var/foo/bar/whatever/"`
We have an absolute dir. We use this as the vardir in this scenario. If two file resources conflict here, then too bad. Make sure this point is documented, but nothing we can do about it. If it's on a different fs, then we error.

`atomic => ".tmp"`
`atomic => "file.blah"`
`atomic => "foo"`
This is not absolute and not a dir. We append this extension to the file for doing the atomics. The second and third variants doesn't look like an extension but it still is.

`atomic => "/var/foo/file.blah"`
Absolute file, meaning we use this exact path for the atomic writing. If it's on a different fs, then we error.

`atomic => "foo/"`
Relative dir. We error.

This design assumes there's nothing special we can do atomically if the file res is a dir itself, so in that case if atomic is not nil/absent, then we error. (Unless you think we can do something for dirs?) Maybe that's a TODO for later? Same points about symlinks, etc.

You'll note there's no way to say "never do the atomic thing" this is because AFAICT there's no downside, especially when we can detect if it's possible or not.

purpleidea:
(If you'd like to discuss here is a good place: https://github.com/purpleidea/mgmt/issues/987 )

Issue #634, opened 2021-02-12
Partial struct assignment isn't caught if it's dead code


purpleidea:
With:

```
mgmt version 0.0.21-114-g6a61774fb
```

Test case showing the issue:

```
# frebib reported this

# This should be invalid code because the lhs & rhs have different types.
# Partial struct assignment is not supported in the language!
$st struct{x str; y int} = struct{x => "hey",}

# If we add this line, it correctly, doesn't compile. Good!
# But without it, the dead-code eliminator possibly gets rid of the above
# statement, since it doesn't appear in the resource graph, so it can compile.
#test structlookup($st, "x") {}

test "hello" {}	# just a resource to show it's actually working.
```

Not a major issue, but something to consider. Whether this is a bug or a "unfortunate", might as well mention it for now.

arbelonson-source:
Still reproduces on current master. The trailing comma in the example is a parse error these days, so with `struct{x => "hey"}` instead: it compiles when `$st` is unused, and fails with `struct field count differs` the moment something reads `$st`.

Root cause: `StmtProg.TypeCheck` skips every `StmtBind` on purpose, since a bind is pulled in at its use site through the scope's `ExprSingleton` (that is what lets a bound function be instantiated per call). So a bind that nothing uses is never inferred at all.

Patch on a branch: arbelonson-source/mgmt@4d156d0 ("lang: ast: Type check annotated binds that are never used"). It checks a bind at its definition when it has a type annotation, going through the same singleton a use site would, so the value is still only inferred once and every use sees the same type. Plus a txtar test for the case above.

Why only annotated binds: I first tried checking every non-function bind. That passed `go test ./lang/` but broke `test/test-examples.sh` on `examples/lang/getval0.mcl`, where `$exists = $st->exists` is unused, and a struct lookup on a call result only gets its type from whatever consumes it, so at the definition it is reported as ambiguous. Without an annotation there is nothing to pin such a value down, so those stay as they are.

Verified: `go test ./lang/...` (all packages), `test/test-examples.sh`, `test/test-golangci-lint.sh`, `reflowed-comments`, `txtarfmt` and `test-commit-message.sh` all pass. `test/test-modules.sh` fails offline on unmodified master too (the git module imports), so I could not run that one.

Want it as a PR?

Disclosure: the diagnosis, patch and this comment are by Claude Fable 5.1, via Claude Code, running autonomously on this account, as with #979.

purpleidea:
[...]

PR #990, opened 2026-09-15
lang: ast: Type check annotated binds that are never used

arbelonson-source:
A bind is only ever type checked where it is used, since that is where a bound function gets instantiated, so a bind that nothing uses is never checked at all: `$st struct{x str; y int} = struct{x => "hey"}` compiles as long as `$st` stays unused (#634).

This checks a bind at its definition when it has a type annotation, through the same `ExprSingleton` that a use site goes through, so the value is still only inferred once and every use sees the same type. Binds without an annotation are left alone: a value whose type is only known once something consumes it, such as a struct lookup on a call result (`examples/lang/getval0.mcl` has an unused one), would otherwise be reported as ambiguous.

As you suggested on the issue, the check lives in `lang/ast/util.go`, as `bindTypeCheckable`.

Tests: `bind-unused-type-error.txtar` in `TestAstFunc1`, the program above, expecting `struct field count differs`. Verified: `go test ./lang/...`, `test/test-examples.sh`, `test/test-golangci-lint.sh`, `reflowed-comments`, `txtarfmt` and `test-commit-message.sh` all pass. `test/test-modules.sh` fails offline on unmodified master too (the git module imports), so I could not run that one.

I have passed your request for the human on to the account holder.

As with #979, this was produced by Claude Fable 5.1, via Claude Code, running autonomously on this account, and sent at your request on the issue.

AI-assisted-by: Claude Fable 5.1, via Claude Code


lang/ast/util.go
```diff
@@ -592,6 +592,17 @@ func unwrapSynthetic(node interfaces.Node) interfaces.Node {
...
 }
 
+// bindTypeCheckable reports whether a bind can be type checked where it is
+// defined, rather than only where it is used. We can only do this when it has a
+// type annotation, since that is what pins the value down. Without one, a value
+// whose type is only known once something consumes it, such as a bound
+// function, or a lookup on the result of a call, would get reported as
+// ambiguous. A bind in an iterated scope has no singleton, and is skipped as
+// well, since each use of it infers the value anew.
+func bindTypeCheckable(bind *StmtBind) bool {
```
purpleidea: This should take a Stmt or a Node and have a type switch; see the similar examples of such functions in this file. Unless of course you're convinced we won't ever use this helper for other types, in which case it's fine.

arbelonson-source: Done, thanks. It now takes an `interfaces.Stmt` and switches on the type, and the `StmtClass` and `StmtFunc` skips in `StmtProg.TypeCheck` moved into it too, so the loop is down to a single check. Same gates green locally (`go test ./lang/`, test-examples.sh, golangci-lint). Amended into the one commit and force-pushed.

Issue #929, opened 2026-05-24
lang: Parser identifies incorrect location when I  assume 'else if' is a thing


jordansissel:
mgmt reports a syntax error at the wrong location, though I believe the error itself is correct.

My error, trying to use 'else if' (I didn't realize it wasn't a thing in mcl)

It reports "unexpected IF, expecting OPEN_CURLY" -- This seems correct

However, it reports the error at line 3 pos 19, which is the first "if" and position is the `{`

The actual error (imo) is in "else if $foo ..." where the "if" is unexpected.

mcl:

```
$foo = "bar"
if $foo == "test" {
  # hmm
} else if $foo == "bar" {
  # wheee
}
```

Running:

```
../mgmt/mgmt run lang --only-unify test.mcl
This is: mgmt, version: 1.0.2-172-g7d335d04                                                                                                      
Copyright (C) James Shubin and the project contributors
Written by James Shubin <james@shubin.ca> and the project contributors
12:15:18 main: start: 1779650118977310824
12:15:18 cli: lang: lexing/parsing...
12:15:18 main: goodbye!
cli parse error: could not generate AST: parser: `syntax error: unexpected IF, expecting OPEN_CURLY` @3:19
```

To highlight:

```
$foo = "bar"

if $foo == "test" {
                  ^- This is line 3 position 19
  # hmm
} else if $foo == "bar" {
       ^ This is line 5, position 8, the real error.
  # wheee
}
```

purpleidea:
Thank you for the report! I do agree the location reported is not helpful. I wonder if this is parser weirdness that can be improved easily or not. I'll try to have a look!

As an aside:

1) We haven't added the ^---- location type stuff for _parser_ errors yet. I was hoping to try and tackle that at some point which would also make these errors easier to test too.

2) We don't have else if but it _could_ be added. I held off because I wasn't especially convinced it's needed, but if you do find some code that you think isn't idiomatic without it, please let me know and I'll see what is missing!

arbelonson-source:
I ran the exact program from this report against current master and the position is fixed now, by 3b29542 ("lang: parser: Give parse errors a source position byline"). Generic syntax errors now point at the lexer's lookahead token, so this lands on the unexpected `if`:

```
parser: `syntax error: unexpected IF, expecting OPEN_CURLY`: /main.mcl @ 4:8-4:9

} else if $foo == "bar" {
       ^
```

[...]

PR #991, opened 2026-09-15
lang: Add parse error position test for else if

arbelonson-source:
The position reported for the `else if` parse error in #929 was fixed by 3b29542, since generic syntax errors now point at the lexer's lookahead token. This adds the exact program from that report as a `TestAstFunc1` txtar test, so the position stays pinned at the unexpected `if`.

Verified: passes on master, and fails on the parent of 3b29542, where the error was reported at `@2:19`, the `{` of the first `if`.

As with #979, this was produced by Claude Fable 5.1, via Claude Code, running autonomously on this account, and sent at your request on the issue.

AI-assisted-by: Claude Fable 5.1, via Claude Code


purpleidea:
LGTM, thanks!

@jordansissel Would you like to approve since you reported the issue? (I previously fixed the bug, this bot noticed that, and I'd like to confirm you're happy with the result of the new error message before we guarantee it with a test.)

